PR #47, opened 2023-11-06
Update ci-gradle.yml runner to ubuntu-latest

nelsitoPuglisi:


bidetofevil (approved):
LGTM

fractalwrench:
CI needs to be green before we can merge - is KVM supported/installed by default on the latest version of ubuntu? The tests seem to be failing due to an issue around that

bidetofevil:
CI is failing because the our tests couldn't run, which is what is causing this issue. Can we override this and force a merege?

fractalwrench:
Ubuntu latest is 22.04, which is the same as what the failing github action reported running on: https://github.com/actions/runner-images#available-images

Is there an extra bit of context that I'm missing?

bidetofevil:
Specifying a specific version is a paid feature we only have in the private repo. For public ones, we need to use `latest`

PR #50, opened 2023-11-07
Additional EmbraceUrlConnectionDelegate unit tests and bug fixes

bidetofevil:
## Goal

Add additional unit tests for more cases for logging URLConnection requests and fix a few issues that were found. Specifically:

- Making it more consistent and efficient when the SDK is not on to bypass the logging code
- Add query string to the URL logged if there's a custom path override
- Record response body even when the network call data had been previously cached.

## Testing

Add unit tests of important cases where breakage may not be obvious

### DIFF
--- a/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/internal/network/http/EmbraceHttpUrlConnectionOverride.java
+++ b/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/internal/network/http/EmbraceHttpUrlConnectionOverride.java
@@ -28,7 +28,7 @@ class EmbraceHttpUrlConnectionOverride implements HttpPathOverrideRequest {
     public String getOverriddenURL(@NonNull String pathOverride) {
         try {
             return new URL(connection.getURL().getProtocol(), connection.getURL().getHost(),
-                connection.getURL().getPort(), pathOverride).toString();
+                connection.getURL().getPort(), pathOverride + "?" + connection.getURL().getQuery()).toString();
         } catch (MalformedURLException e) {
             InternalStaticEmbraceLogger.logError("Failed to override path of " +
                 connection.getURL() + " with " + pathOverride);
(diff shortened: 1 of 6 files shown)


bidetofevil:
* master
  * **PR #50** <a href="https://app.graphite.dev/github/pr/embrace-io/embrace-android-sdk/50" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/embrace-io/embrace-android-sdk/50?utm_source=stack-comment).

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/internal/network/http/EmbraceHttpUrlConnectionOverride.java
```diff
@@ -28,7 +28,7 @@ public String getHeaderByName(@NonNull String name) {
     public String getOverriddenURL(@NonNull String pathOverride) {
         try {
             return new URL(connection.getURL().getProtocol(), connection.getURL().getHost(),
-                connection.getURL().getPort(), pathOverride).toString();
+                connection.getURL().getPort(), pathOverride + "?" + connection.getURL().getQuery()).toString();
```
nelsitoPuglisi, line 31: What's the purpose of this change? Is it because we were discarding the query params?

nelsitoPuglisi (approved):

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/internal/network/http/EmbraceHttpUrlConnectionOverride.java
bidetofevil: Yeah. This is the bug I mention in the second bullet point in the PR description.  When you `toString()` a `URL`, it includes the query path as part of the returned string. When we return teh URL with an overridden path, we should do the same. (This is what the OkHttp implementation does)

PR #52, opened 2023-11-09
Extract component callbacks from activity service

fractalwrench:
## Goal

Extracts the component callbacks implementation from the activity service & places it into its own service. This has nothing to do with activities & therefore belongs in a separate file.

## Testing

Added unit test coverage.


bidetofevil (approved):
Very nice. LGTM.

PR #53, opened 2023-11-09
[EMBR-1230] Make rawPost private in ApiClient

lucaslabari:
## Goal

- The method postOnExecutor in EmbraceApiService was always being called with a true value for the compress parameter, so postRaw() method was never being called. 
- I removed the if and make rawPost private in ApiClient. 

## Testing

Updated unit tests.


bidetofevil (approved):
LGTM. We should probably revisit this implementation so we gzip the payload before we send it (since it's always compressed anyway) so we can save some room when we store it locally, but that's another PR.

PR #56, opened 2023-11-10
Avoid making http request on process termination

fractalwrench:
## Goal

Avoids making a HTTP request to send the crash on process termination. This is very unlikely to succeed as it's posted to a background thread & there isn't any synchronisation waiting for that thread to complete before the process terminates. This will likely lead to races where the same request gets made twice - this changes things so that we always just attempt to send the crash on the next launch.


bidetofevil (approved):
Yeah. Consistently saying that a crash won't be sent until the next time the app starts up seems reasonable. In fact, I had assumed that was what was happening...

PR #57, opened 2023-11-10
Write session to disk synchronously during crash

fractalwrench:
## Goal

Alters session persistence in the event of a crash. Previously we were calling `DeliveryService#sendSession()` which writes the session to disk then attempts a network request in a background thread. No blocking occurs before process termination so there isn't a guarantee this information will be persisted; additionally, the periodic caching could theoretically overwrite the file.

This changeset makes two changes: firstly, we suspend periodic caching of the session. Secondly, we write the session to disk synchronously to ensure it was persisted properly.

## Testing

Updated existing test coverage & manually tested


bidetofevil (approved):
Looks good. When going through the code, I noticed that we sometimes have a background deferring to another background thread to persist data to disk. I wonder if that's even necessary or just create extra overhead and weird race conditions. Anyway, we should look at that in a different task. This LGTM.

PR #58, opened 2023-11-10
Create property delegate function for ThreadLocal

fractalwrench:
## Goal

Neither `Gson` or `EmbraceSerializer` are thread-safe. This changeset updates both to use `ThreadLocal` to avoid any thread safety issues around JSON serialization, and adds some syntactic sugar that makes it easier to use thread locals.

## Testing

Added unit test coverage.


lucaslabari (approved):
LGTM

PR #59, opened 2023-11-11
Cleanup ApiClient and add more tests

bidetofevil:
## Goal

Make our delivery layer more testable by doing some non-code-path changing clean up to make `ApiClient` an interface, spin out the implementation to `ApiClientImpl`, and add a fake to enable better testing for components that depend on it (e.g. `EmbraceApiService`) as well as easier testing of that component as well.

Changes largely broken down by commits for easier review.

## Testing

Add additional tests for core workflows that weren't exercised. Left some edge cases as I plan to do more dramatic clean up of the interface later.


bidetofevil:
* master
  * **PR #59** <a href="https://app.graphite.dev/github/pr/embrace-io/embrace-android-sdk/59?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/embrace-io/embrace-android-sdk/59?utm_source=stack-comment).

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/comms/api/ApiClient.kt
```diff
@@ -1,179 +1,25 @@
...
- * testing is enabled when calling [Embrace.start()].
- */
-internal class ApiClient @JvmOverloads constructor(
-    private val logger: InternalEmbraceLogger = InternalStaticEmbraceLogger.logger
-) {
+    /**
+     * Executes [ApiRequest] as a POST with the given body defined by [payloadToCompress], returning the response as a [ApiResponse].
+     * The body will be gzip compressed.
+     */
+    fun executePost(request: ApiRequest, payloadToCompress: ByteArray): ApiResponse<String>
```
bidetofevil, line 13: The one slightly significant change is to return the full response abstraction rather than just the body. This enables consumers of this components to deal with the response better (e.g. non-200 responses), which will be done in a later PR.

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/comms/api/ApiService.kt
```diff
@@ -15,5 +15,5 @@ internal interface ApiService {
     fun sendEventAndWait(eventMessage: EventMessage)
     fun sendCrash(crash: EventMessage)
     fun sendAEIBlob(blobMessage: BlobMessage)
-    fun sendSession(sessionPayload: ByteArray, onFinish: (() -> Unit)?): Future<*>
+    fun sendSession(sessionPayload: ByteArray, onFinish: (() -> Unit)?): Future<*>?
```
bidetofevil, line 18: If the request can't be sent return a null Future and have the caller deal with it

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/comms/api/EmbraceApiService.kt
```diff
@@ -24,22 +23,23 @@ import java.util.concurrent.TimeUnit
...
     private val deliveryRetryManager: DeliveryRetryManager,
     private val lazyDeviceId: Lazy<String>,
-    private val appId: String
+    private val appId: String,
+    urlBuilder: ApiUrlBuilder,
+    networkConnectivityService: NetworkConnectivityService
 ) : ApiService, NetworkConnectivityListener {
 
+    private val configUrl = urlBuilder.getConfigUrl()
+    private val apiUrlBuilders = Endpoint.values().associateWith { urlBuilder.getEmbraceUrlWithSuffix(it.path) }
```
bidetofevil, line 39: We were fetching URLs from the builder each time we need it when they never change after this component's initialization. Simply get them al at the beginning and use them as needed.

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/comms/api/EmbraceUrl.kt
```diff
@@ -1,40 +1,38 @@
 package io.embrace.android.embracesdk.comms.api
 
 import java.io.IOException
+import java.net.HttpURLConnection
 import java.net.MalformedURLException
+import java.net.URL
 
-internal abstract class EmbraceUrl {
+internal class EmbraceUrl(val url: URL) {
```
bidetofevil, line 8: This class contains a lot of vestigial complexity like the `UrlFactory` that isn't needed. Remove them so we don't have untested dead code.

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/payload/Event.kt
```diff
@@ -14,15 +14,15 @@ internal data class Event constructor(
 
     @SerializedName("id")
     @JvmField
-    val eventId: String? = null,
+    val eventId: String,
```
bidetofevil, line 17: These are necessary fields that are null-checked - and runtime exceptions are thrown if they are. Instead of doing that, just make them required and non-null.

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/utils/exceptions/Unchecked.kt
```diff
@@ -1,64 +1,13 @@
...
-     *
-     * If a checked exception is thrown it is converted to a [RuntimeException].
-     *
-     * @param <T>   the type of the result
-     * @param block the code block to wrap
-     * @return the result of invoking the block
-     * @throws RuntimeException if an exception occurs
-     </T> */
-    @JvmStatic
-    fun <T> wrap(block: CheckedSupplier<T>): T {
```
bidetofevil, line 44: We don't need these anymore, and arguably, we probably shouldn't have them.

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/comms/delivery/EmbraceDeliveryService.kt
```diff
@@ -63,7 +63,7 @@ internal class EmbraceDeliveryService(
                         apiService.sendSession(
                             session,
                             onFinish
-                        )[SEND_SESSION_TIMEOUT, TimeUnit.SECONDS]
+                        )?.get(SEND_SESSION_TIMEOUT, TimeUnit.SECONDS)
```
fractalwrench, line 66: Is this correct? If `sendSession` fails to schedule a future then `onFinish` will result in data loss of the session

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/utils/exceptions/Unchecked.kt
fractalwrench: These are a hangover from the Java days & we're nearly at the point where they can be fully removed..

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/comms/api/ApiClientImpl.kt
```diff
@@ -0,0 +1,158 @@
...
+        logger.logDeveloper("ApiClient", request.httpMethod.toString() + " " + request.url)
+        logger.logDeveloper("ApiClient", "Request details: $request")
+
+        var connection: EmbraceConnection? = null
+        return try {
+            connection = request.toConnection()
+            setTimeouts(connection)
+            if (payload != null) {
+                logger.logDeveloper("ApiClient", "Payload size: " + payload.size)
+                connection.outputStream?.write(payload)
```
fractalwrench, line 60: future: thoughts on serializing JSON (or cached file) directly to the request outputstream?

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/comms/api/EmbraceApiService.kt
```diff
@@ -165,20 +154,12 @@ internal class EmbraceApiService(
      * @param networkEvent the event containing the network call information
      */
     override fun sendNetworkCall(networkEvent: NetworkEvent) {
-        logger.logDeveloper(TAG, "sendNetworkCall")
-
-        val url = Unchecked.wrap {
-            EmbraceUrl.getUrl(
-                urlBuilder.getEmbraceUrlWithSuffix("network")
-            )
+        apiUrlBuilders[Endpoint.NETWORK]?.let { url ->
```
fractalwrench, line 157: Is it worth wrapping these types of lookups in checkNotNull? It _should_ never be the case that an endpoint will be null

lucaslabari (approved):
LGTM

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/comms/delivery/EmbraceDeliveryService.kt
lucaslabari: It seems that onFinish is only executed on the finally blocked inside the executor's task. 
The only case where the future can be null is when the endpoint URL is null and that's not expected in any scenario. 
Anyway I think we already have a scenario of data loss here, because in EmbraceApiService we are calling onFinish() even when the request fails or is not executed because there is no network connection. 
We should review this path.

fractalwrench:
Also, heads up that I've made some changes in a similar area. If you try to get this merged first then I'll rebase away the conflicts: https://github.com/embrace-io/embrace-android-sdk/pull/61

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/comms/delivery/EmbraceDeliveryService.kt
bidetofevil: This is an annoying leftover of the current API as we've validated that the URL isn't null. I should've taken care of this as part of the refactor and I'll do that later.

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/comms/api/ApiClientImpl.kt
bidetofevil: We are caching the payload as bytes to enable retries. @lucaslabari is actually going to do the gzipping of those bytes earlier, but I wonder if caching a zipped (or unzipped, depending on size) payload and doing the serialization here would make more sense.

Serializing is so expensive and if we can defer that and still don't blow up disk usage, that might be worth trying!

Let me profile this later.

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/comms/api/EmbraceApiService.kt
bidetofevil: Yeah definitely. Will be doing that later for sure. The sober light of caffeine and y'alls comments makes this obvious lol

bidetofevil:
> Also, heads up that I've made some changes in a similar area. If you try to get this merged first then I'll rebase away the conflicts: #61

Haha, yup sorry! Hope this should be at least straightforward

PR #61, opened 2023-11-13
Extract logic for constructing ApiRequest from ApiService

fractalwrench:
## Goal

Extracts the logic for constructing `ApiRequest` objects into a separate file. This makes it easier to unit test the business logic for mapping events/sessions to `ApiRequest`.

This is necessary for future work that will give failed request types different priorities on the retry queue. This is mostly just a case of lifting functions into a separate source file, although I did remove/alter some redundant code in a couple of functions.

## Testing

Added unit test coverage.


bidetofevil (approved):
Oh yeah this is nice. It'll work well with the changes I had and replace the clunky map I have!

PR #62, opened 2023-11-13
Allow session retries to continue

fractalwrench:
## Goal

Allows session retries to continue past the limit of 200 requests. There are a couple of ways this could be done - I chose the fairly naive approach of just allowing session requests to go past the arbitrary limit. This approach is fairly simple to implement & it's unlikely that this will add more than a few extra retries to the queue.

## Testing

Added unit test coverage.


bidetofevil (approved):
I think this is fine for a quick fix for 6.1.0, but @lucaslabari will need to apply the retries per end point, so his solution may subsume this - which is fine, as now we have tests to verify that it'll work as it does now

PR #63, opened 2023-11-13
Convert ApiResponse into a seal class.

lucaslabari:
## Goal

- Convert ApiResponse into a seal class.
- This way ApiClient will always return an ApiResponse of type Success, Failure or Error. 

## Testing

Updated unit tests.

### DIFF
--- a/embrace-android-sdk/config/detekt/baseline.xml
+++ b/embrace-android-sdk/config/detekt/baseline.xml
@@ -1,7 +1,6 @@
-<?xml version="1.0" ?>
+<?xml version='1.0' encoding='UTF-8'?>
 <SmellBaseline>
-  <ManuallySuppressedIssues></ManuallySuppressedIssues>
+  <ManuallySuppressedIssues/>
   <CurrentIssues>
-    <ID>UseCheckOrError:EmbraceApiService.kt$EmbraceApiService$throw IllegalStateException("Failed to retrieve from Embrace server.")</ID>
   </CurrentIssues>
 </SmellBaseline>
--- a/embrace-android-sdk/config/detekt/baseline.xml
+++ b/embrace-android-sdk/config/detekt/baseline.xml
@@ -1,6 +1,7 @@
-<?xml version='1.0' encoding='UTF-8'?>
+<?xml version="1.0" ?>
 <SmellBaseline>
-  <ManuallySuppressedIssues/>
+  <ManuallySuppressedIssues></ManuallySuppressedIssues>
   <CurrentIssues>
+    <ID>UseCheckOrError:EmbraceApiService.kt$EmbraceApiService$throw IllegalStateException("Failed to retrieve from Embrace server.")</ID>
   </CurrentIssues>
 </SmellBaseline>
--- a/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/comms/api/ApiClientImpl.kt
+++ b/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/comms/api/ApiClientImpl.kt
@@ -1,13 +1,16 @@
 package io.embrace.android.embracesdk.comms.api
 
 import io.embrace.android.embracesdk.comms.api.ApiClient.Companion.NO_HTTP_RESPONSE
+import io.embrace.android.embracesdk.comms.api.ApiClient.Companion.TOO_MANY_REQUESTS
 import io.embrace.android.embracesdk.comms.api.ApiClient.Companion.defaultTimeoutMs
 import io.embrace.android.embracesdk.logging.InternalEmbraceLogger
 import java.io.ByteArrayOutputStream
 import java.io.IOException
 import java.io.InputStream
 import java.io.InputStreamReader
-import java.net.HttpURLConnection
+import java.net.HttpURLConnection.HTTP_ENTITY_TOO_LARGE
+import java.net.HttpURLConnection.HTTP_NOT_MODIFIED
+import java.net.HttpURLConnection.HTTP_OK
 import java.util.zip.GZIPOutputStream
 
 /**
@@ -84,16 +87,29 @@ internal class ApiClientImpl(
     private fun executeHttpRequest(connection: EmbraceConnection): ApiResponse {
         return try {
             val responseCode = readHttpResponseCode(connection)
-
             val responseHeaders = readHttpResponseHeaders(connection)
-            return if (responseCode == HttpURLConnection.HTTP_OK) {
-                val responseBody = readResponseBodyAsString(connection.inputStream)
-                ApiResponse.Success(responseBody, responseHeaders)
-            } else {
-                val errorMessage = connection.errorStream?.let {
-                    readResponseBodyAsString(it)
+
+            return when (responseCode) {
+                HTTP_OK -> {
+                    val responseBody = readResponseBodyAsString(connection.inputStream)
+                    ApiResponse.Success(responseBody, responseHeaders)
                 }
-                ApiResponse.Failure(errorMessage, responseCode, responseHeaders)
+                HTTP_NOT_MODIFIED -> {
+                    ApiResponse.NotModified
+                }
+                HTTP_ENTITY_TOO_LARGE -> {
+                    ApiResponse.PayloadTooLarge
+                }
+                TOO_MANY_REQUESTS -> {
+                    val retryAfter = responseHeaders["Retry-After"]?.toLongOrNull()
+                    ApiResponse.TooManyRequests(retryAfter)
+                }
+                NO_HTTP_RESPONSE -> {
+                    ApiResponse.Incomplete(IllegalStateException("Connection failed or unexpected response code"))
+                }
+                else -> {
+                    ApiResponse.Failure(responseCode, responseHeaders)
+                }
             }
         } catch (exc: Throwable) {
             ApiResponse.Incomplete(IllegalStateException("Error occurred during HTTP request execution", exc))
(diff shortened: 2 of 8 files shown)


embrace-android-sdk/config/detekt/baseline.xml
```diff
@@ -1,6 +1,7 @@
-<?xml version='1.0' encoding='UTF-8'?>
+<?xml version="1.0" ?>
 <SmellBaseline>
-  <ManuallySuppressedIssues/>
+  <ManuallySuppressedIssues></ManuallySuppressedIssues>
   <CurrentIssues>
+    <ID>UseCheckOrError:EmbraceApiService.kt$EmbraceApiService$throw IllegalStateException("Failed to retrieve from Embrace server.")</ID>
```
bidetofevil, line 5: Lets not use the baseline file since they get hidden. Can you suppress with the @Suppress annotation instead?

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/comms/api/ApiClientImpl.kt
```diff
@@ -27,13 +27,14 @@
...
+        return try {
             connection = request.toConnection()
             setTimeouts(connection)
             connection.connect()
-            return executeHttpRequest(connection)
+            val response = executeHttpRequest(connection)
+            response
         } catch (ex: Throwable) {
-            throw IllegalStateException(ex.localizedMessage ?: "", ex)
+            ApiResponse.Error(IllegalStateException(ex.localizedMessage ?: "", ex))
```
bidetofevil, line 37: Just add the @Suppress to this method, as this is something we'll want to change. I don't think it's an IllegalStateException at all, and I'm not sure why we're even wrapping it. So we can count them together? I think we lose some data fidelity by munging them all together.

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/comms/api/ApiResponse.kt
```diff
@@ -1,7 +1,21 @@
...
+
+    /**
+     * Represents a failed API call. (status code 400-499 or 500-599)
+     */
+    data class Failure(val errorMessage: String?, val code: Int, val headers: Map<String, String>?) : ApiResponse<Nothing>()
+
+    /**
+     * Represents an exception thrown while making the API call.
+     */
+    data class Error(val exception: Throwable) : ApiResponse<Nothing>()
```
bidetofevil, line 20: I would call this `Incomplete` or something to indicate that a response was not received. It's too close to `Failure`

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/comms/api/ApiResponse.kt
```diff
@@ -1,7 +1,21 @@
...
+internal sealed class ApiResponse<out T> {
+    /**
+     * Represents a successful API call (status code 200-299)
+     */
+    data class Success<T>(val body: T?, val headers: Map<String, String>?) : ApiResponse<T>()
+
+    /**
+     * Represents a failed API call. (status code 400-499 or 500-599)
+     */
+    data class Failure(val errorMessage: String?, val code: Int, val headers: Map<String, String>?) : ApiResponse<Nothing>()
```
bidetofevil, line 15: You can have a body with a non-2xx response, so we should also use `String` as a type.

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/comms/api/ApiResponse.kt
```diff
@@ -1,7 +1,21 @@
...
 
-internal data class ApiResponse<T>(
-    val statusCode: Int?,
-    val headers: Map<String, String>,
-    val body: T?
-)
+/**
+ * ApiResponse is a sealed class that represents the result of an API call.
+ */
+internal sealed class ApiResponse<out T> {
```
bidetofevil, line 6: Do we really need a type here? The body is either a String or it doesn't exist. This seems superfluous?

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/comms/api/ApiResponse.kt
bidetofevil, line 15: Also, any reason why you aren't sub-dividing the different groups of failure so we can handle them without looking at the status code? It would be nice to just have a `when` statement that exhaustively handles all the different classes without us have to look at the status code.

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/comms/api/EmbraceApiService.kt
```diff
@@ -60,27 +60,32 @@
...
-                logger.logInfo("Confirmed config has not been modified.")
-                cachedResponse.remoteConfig
-            }
-
-            ApiClient.NO_HTTP_RESPONSE -> {
-                logger.logInfo("Failed to fetch config (no response).")
-                null
+            is ApiResponse.Failure -> {
+                when (response.code) {
+                    HttpURLConnection.HTTP_NOT_MODIFIED -> {
```
bidetofevil, line 71: `NOT_MODIFIED` is 304 and shouldn't be considered a failure. This is another reason why expanding the number of sealed classes would be useful.

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/comms/api/ApiClientImpl.kt
```diff
@@ -84,16 +82,21 @@
...
-                headers,
-                readResponseBodyAsString(connection.inputStream)
-            )
+
+            val responseHeaders = readHttpResponseHeaders(connection)
+            return if (responseCode == HttpURLConnection.HTTP_OK) {
+                val responseBody = readResponseBodyAsString(connection.inputStream)
+                ApiResponse.Success(responseBody, responseHeaders)
+            } else {
+                val errorMessage = connection.errorStream?.let {
```
bidetofevil, line 93: I'm not sure if our server returns anything useful in the body. Is there a specific reason why you want to access this now when we didn't use it before?

bidetofevil (commented):
A question overall: we do a lot of logging all over this class. Is it really necessary? Given how many API requests we make, and how standard they all are, we would get a lot of noise that isn't particularly useful

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/comms/api/ApiResponse.kt
lucaslabari: It was already a type in the previous version of ApiResponse, I left it that way in case the body changes at any time to be a stream or something else, so we don't need to make changes in a lot of places. 
But it adds an unnecessary complexity, I'll remove it.

lucaslabari: Updated

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/comms/api/ApiClientImpl.kt
lucaslabari: We don't need a Suppress here, as the warning to not throw an IllegalStateException is in EmbraceApiService because it was not in the context of a try catch. 
I agree with you in that we are unnecessary wrapping exceptions into a IllegalStateException, not sure why we were doing this, that's why we better investigate if there is a reason and address it in another PR.

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/comms/api/EmbraceApiService.kt
lucaslabari: Updated.

lucaslabari:
> A question overall: we do a lot of logging all over this class. Is it really necessary? Given how many API requests we make, and how standard they all are, we would get a lot of noise that isn't particularly useful

There was already a lot of logs, I thnk we can review which of them makes sense and remove the others on separate PR.

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/comms/api/ApiResponse.kt
```diff
@@ -1,7 +1,36 @@
...
+     */
+    object PayloadTooLarge : ApiResponse()
+
+    /**
+     * Represents an API call that returned a 429 Too Many Requests status code.
+     */
+    data class TooManyRequests(val retryAfter: Long?) : ApiResponse()
+
+    /**
+     * Represents a failed API call. (status code 400-499 or 500-599)
```
bidetofevil, line 28: I would leave out this comment since 413 and 429s are call out explicitly

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/comms/api/EmbraceApiService.kt
```diff
@@ -53,33 +52,39 @@ internal class EmbraceApiService(
...
         var request = prepareConfigRequest(configUrl)
         val cachedResponse = cachedConfigProvider(configUrl, request)
         if (cachedResponse.isValid()) { // only bother if we have a useful response.
             request = request.copy(eTag = cachedResponse.eTag)
         }
-        val response = apiClient.executeGet(request)
 
-        return when (response.statusCode) {
-            HttpURLConnection.HTTP_OK -> {
+        return when (val response = apiClient.executeGet(request)) {
```
bidetofevil, line 63: Does it make sense just to preserve the existing behaviour when dealing with how to handle a reponses and change this for real when you add the code the handle 429?

bidetofevil (approved):
LGTM

PR #67, opened 2023-11-14
Perform session gating asynchronously

fractalwrench:
## Goal

Performs session gating asynchronously. #55 pointed out that this is something we can feasibly do when saving sessions.

## Testing

Updated unit test coverage.


bidetofevil (approved):
Nice

PR #68, opened 2023-11-14
Add lock for creating session ordinal

fractalwrench:
## Goal

Adds synchronisation so that there is no possibility of a race condition when creating the session ordinal. Admittedly this function is only called in one place when a session is started, but there isn't currently any synchronisation for the session starting so a double-call could lead to misleading dropped session %.

This lock might be something we can remove when looking at how the session boundary control works more generally.


bidetofevil (commented):
Can we just use an `AtomicInteger` instead?

fractalwrench:
Closing as #76 should make this redundant, since the function will be behind a lock.

PR #69, opened 2023-11-14
Separate lifecycle callbacks for activity + process state

fractalwrench:
## Goal

Separates the lifecycle callbacks for the activity + process state. Currently these are implemented together in the same class   (`ActivityService`) but these are distinct responsibilities so we should separate them.

I have refactored the implementation so that `ProcessStateService` is responsible for tracking the process foreground/background state, and that `ActivityTracker` is responsible for tracking activity lifecycle callbacks.

## Testing


embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/EmbraceImpl.java
```diff
@@ -1518,8 +1526,13 @@
...
 
     @Nullable
-    ActivityService getActivityService() {
-        return activityService;
+    ProcessStateService getActivityService() {
+        return processStateService;
+    }
+
+    @Nullable
+    ActivityTracker getActivityLifecycleTracker() {
```
bidetofevil, line 1534: Who calls this? If we don't need this I'd rather not have it

fractalwrench: `EmbraceAutomaticVerification` calls it (and other services) unfortunately

bidetofevil: Ack....

bidetofevil (approved):
Looks good to me

PR #71, opened 2023-11-15
Extract session message collator

fractalwrench:
## Goal

Extracts the logic used to construct a session message from the business logic that decides when a session should start/end.

## Testing


lucaslabari (approved):
LGTM

PR #72, opened 2023-11-15
Extract session properties into separate service

fractalwrench:
## Goal

Extracts the session properties into a separate service from `SessionService`

## Testing

Updated unit test coverage.

### DIFF
--- a/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/session/properties/SessionPropertiesServiceImpl.kt
+++ b/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/session/properties/SessionPropertiesServiceImpl.kt
@@ -1,0 +1,42 @@
+package io.embrace.android.embracesdk.session.properties
+
+import io.embrace.android.embracesdk.logging.InternalEmbraceLogger
+import io.embrace.android.embracesdk.logging.InternalStaticEmbraceLogger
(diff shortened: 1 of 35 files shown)


embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/session/properties/SessionPropertiesServiceImpl.kt
```diff
@@ -0,0 +1,42 @@
+package io.embrace.android.embracesdk.session.properties
+
+import io.embrace.android.embracesdk.logging.InternalEmbraceLogger
+import io.embrace.android.embracesdk.logging.InternalStaticEmbraceLogger
+import io.embrace.android.embracesdk.ndk.NdkService
+
+internal class SessionPropertiesServiceImpl(
```
lucaslabari, line 7: If we want to be consistent with other services this should be called `EmbraceSessionProperties`.  Do we want to start naming them with Impl?

lucaslabari (approved):
LGTM

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/session/properties/SessionPropertiesServiceImpl.kt
fractalwrench: Good point - I've renamed it for consistency.

PR #74, opened 2023-11-15
Reduce duplication in session ending logic

fractalwrench:
## Goal

The existing implementation of `SessionHandler` duplicates business logic on how sessions should be ended. This changeset reduces that duplication & creates an explicit enum type that models the different way a session can end.

@bidetofevil would appreciate an in-depth review on this one, as it feels like there is a lot of complexity here.

## Testing

Relied on existing unit test coverage in `SessionHandlerTest`.

### DIFF
--- a/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/session/SessionHandler.kt
+++ b/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/session/SessionHandler.kt
@@ -47,6 +47,32 @@ internal class SessionHandler(
     private val sessionPeriodicCacheExecutorService: ScheduledExecutorService
 ) : Closeable {
 
+    /**
+     * Defines the states in which a session can end.
+     */
+    private enum class SessionEndType(
+        val endedCleanly: Boolean,
+        val forceQuit: Boolean,
+        val shouldStopCaching: Boolean
+    ) {
+
+        /**
+         * The end session happened in the normal way (i.e. process state changes or manual/timed end).
+         */
+        NORMAL_END(true, false, true),
+
+        /**
+         * The end session is being constructed so that it can be periodically cached. This avoids
+         * the scenario of data loss in the event of NDK crashes.
+         */
+        CACHED_END(false, true, false),
+
+        /**
+         * The end session is being constructed because of a JVM crash.
+         */
+        JVM_CRASH_END(false, false, true);
+    }
+
     var scheduledFuture: ScheduledFuture<*>? = null
 
     /**
@@ -109,14 +135,26 @@ internal class SessionHandler(
         completedSpans: List<EmbraceSpanData>? = null
     ) {
         logger.logDebug("Will try to run end session full.")
-        runEndSessionFull(
-            endType,
+        if (originSession == null) {
+            return
+        }
+        val fullEndSessionMessage = runEndSessionImpl(
+            SessionEndType.NORMAL_END,
             originSession,
             sessionProperties,
             sdkStartupDuration,
-            endTime,
-            completedSpans
-        )
+            completedSpans,
+            endType,
+            endTime
+        ) ?: return
+
+        // Clean every collection of those services which have collections in memory.
+        memoryCleanerService.cleanServicesCollections(exceptionService)
+        metadataService.removeActiveSessionId(originSession.sessionId)
+        logger.logDebug("Services collections successfully cleaned.")
+        sessionProperties.clearTemporary()
+        logger.logDebug("Session properties successfully temporary cleared.")
+        deliveryService.sendSession(fullEndSessionMessage, SessionMessageState.END)
     }
 
     /**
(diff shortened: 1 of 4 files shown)


embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/session/SessionHandler.kt
```diff
@@ -48,6 +48,32 @@ internal class SessionHandler(
     private val sessionPeriodicCacheExecutorService: ScheduledExecutorService
 ) : Closeable {
 
+    /**
+     * Defines the states in which a session can end.
+     */
+    private enum class SessionEndType(
+        val endedCleanly: Boolean,
```
bidetofevil, line 55: Do we need to add a comment about what each of these properties mean or will it be obvious at their usage points?

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/session/SessionHandler.kt
```diff
@@ -111,14 +137,26 @@ internal class SessionHandler(
         completedSpans: List<EmbraceSpanData>? = null
```
bidetofevil, line 137: <i class='graphite__hidden'>[Re: lines 133 to 133]<br/></i>

Why not make this non-null and force the caller to only invoke it if there's an original session? Calling a method that immediately returns makes it seem like the method should't be called at all, and it should be obvious at the call site that this is a no-op<span class='graphite__hidden'><br/><br/>See this comment inline on <a href="https://app.graphite.dev/github/pr/embrace-io/embrace-android-sdk/74?utm_source=unchanged-line-comment">Graphite</a>.</span>

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/session/SessionHandler.kt
```diff
@@ -233,126 +279,44 @@ internal class SessionHandler(
...
-     * it updates cache and it sends it to our servers.
-     * It also stops periodic caching and automatic session stopper.
+     * 'Ends' the active session. Note that this logic is also used for caching the session
+     * periodically so the session won't always end. The behavior is controlled by the
+     * [SessionEndType] passed to this function.
      */
-    private fun runEndSessionFull(
-        endType: SessionLifeEventType,
-        originSession: Session?,
+    private fun runEndSessionImpl(
```
bidetofevil, line 286: Since we aren't always really ending the session, what do you think about renaming this to indicate that it's just taking a current snapshot of the session?

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/session/SessionHandler.kt
```diff
@@ -233,126 +279,44 @@ internal class SessionHandler(
...
+        endType: SessionEndType,
+        activeSession: Session,
         sessionProperties: EmbraceSessionProperties,
         sdkStartupDuration: Long,
+        completedSpans: List<EmbraceSpanData>?,
+        lifeEventType: SessionLifeEventType,
```
bidetofevil, line 292: `SessionLifeEventType` is kind of confusing. Is there a way to refactor this out or somehow combine it with another concept? Have this and `SessionEndType` creates a number of permutations, many of which may not even be valid, which makes the code more complex than i needs to be (I hope)

bidetofevil (approved):
Logic looks good. Most of my comments is about naming and flow and would make this code more readable. The name used and the number of variations depending on the combination of parameters can probably be simplified to reduce complexity, which there is a lot of here already....

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/session/SessionHandler.kt
fractalwrench: Renamed this function

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/session/SessionHandler.kt
fractalwrench: I have renamed `SessionEndType` to `SessionSnapshotType`. I can't think of a clean way to combine the two concepts as they are separate ideas: one is how the session was started; the other is how the session was written to disk.

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/session/SessionHandler.kt
fractalwrench: Added comments

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/session/SessionHandler.kt
fractalwrench: I've moved the null-check higher up.

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/session/SessionHandler.kt
```diff
@@ -233,126 +279,44 @@ internal class SessionHandler(
...
-            return
+        crashId: String? = null,
+    ): SessionMessage? {
+        if (endType.shouldStopCaching) {
+            stopPeriodicSessionCaching()
         }
-
-        stopPeriodicSessionCaching()
-
         if (!configService.dataCaptureEventBehavior.isMessageTypeEnabled(MessageType.SESSION)) {
```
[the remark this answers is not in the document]
fractalwrench: This does feel a bit weird but the caller would have to check ~5 places, whereas the current code only checks 2 places. I'd propose changing this in a future PR

bidetofevil (approved):
Awesome. Looks good!

PR #75, opened 2023-11-15
Fix incorrect usage of @Synchronized in SessionHandler

fractalwrench:
## Goal

Ending the session uses `@Synchronized` but every other function synchronises on the `lock` property. This changeset updates it to use the same lock.

Further changes to the synchronisation will be made in future PRs.


bidetofevil (approved):
Yikes and LGTM

PR #76, opened 2023-11-15
Improve session locking mechanism

fractalwrench:
## Goal

This changeset pushes locking logic down into `SessionHandler` so that we don't need to worry about it in the `SessionService`. A lock must be acquired for starting/ending a session, and when a crash or periodic cache happens. I have opted to use `synchronized` as that is what was there before.

@bidetofevil would appreciate an in-depth review on this one, as it's a complex area...

## Testing

Relied on existing unit test cases.

### DIFF
--- a/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/session/SessionHandler.kt
+++ b/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/session/SessionHandler.kt
@@ -78,6 +78,11 @@ internal class SessionHandler(
     var scheduledFuture: ScheduledFuture<*>? = null
 
     /**
+     * Guards session state changes.
+     */
+    private val lock = Any()
+
+    /**
      * It performs all corresponding operations in order to start a session.
      */
     fun onSessionStarted(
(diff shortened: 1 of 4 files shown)


embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/session/SessionHandler.kt
```diff
@@ -77,6 +77,11 @@ internal class SessionHandler(
     @VisibleForTesting
     var scheduledFuture: ScheduledFuture<*>? = null
 
+    /**
+     * Guards session state changes.
+     */
+    private val lock = Any()
```
bidetofevil, line 83: If we're not ensuring that start/stop are done atomically, we are not preventing the case where data can be recorded while there's no active session? It was probably a problem before, but is there a plan to solve that later?

bidetofevil (approved):
Moving the lock down makes sense. But to improve this further, I wonder if we can have more guarantee that session starts and ends are paired more consistently rather than relying on the foreground/background callbacks to fire and be processed in order. For instance, is there a case where the background/foreground is done so quickly, you get two foreground callback firings before you get the background one? I know the Activity handlers have a built in delay before firing certain lifecycle methods. I wonder if we need to protect ourselves from race conditions that arise as result?

PR #77, opened 2023-11-15
New workflow to update baseline profile file

nelsitoPuglisi:
## Goal

Adding a workflow that will update the `baseline-prof.txt` file. Execute that workflow before creating the snapshot.


lucaslabari (approved):
Nice !!!

.github/workflows/generate_baseline_profile.yml
```diff
@@ -0,0 +1,95 @@
...
+        required: true
+
+jobs:
+  test:
+    runs-on: macos-latest
+    strategy:
+      matrix:
+        jdk-version: [ "11" ]
+        ndk-version: [ "21.4.7075529" ]
+        api-level: [ 29 ]
```
bidetofevil, line 17: Should we target a more current API version? Seems like the recommended right now is 33

.github/workflows/generate_baseline_profile.yml
```diff
@@ -0,0 +1,95 @@
...
+        uses: actions/cache@v3
+        id: avd-cache
+        with:
+          path: |
+            ~/.android/avd/*
+            ~/.android/adb*
+          key: avd-${{ matrix.api-level }}
+
+      - name: Clean Managed Devices
+        # TODO add cd ./android-sdk-benchmark
```
bidetofevil, line 61: What are these TODOs for?

PR #78, opened 2023-11-16
Use remote limit as ceiling for network call limits and apply consistently

bidetofevil:
## Goal

There were two problems that were addressed in this PR: 

1. The initial issue was that if a remote config is not defined for network behaviour, the local settings on network call limits will be used as is, which means customers can set the limit to whatever they want, which sort of defeats the purpose of having a limit to prevent runaway payload size.
2. The second that I found while working on this is that we are clearing the limits prematurely, every time we cache a session. This means the window in which the limit will be applied before resetting was impossibly short.

This PR addresses both issues by using the default limits in the remote config (or defaults implied by the absence of a config) to act as the ceiling to whatever is set locally. This make the local setting a way of further limit what is set on the remote, and won't allow the overriding of it. I also removed the erroneous limit clearing, which ironically is probably the source of the original customer issue anyway, just obscured by what I found with the local limits

## Testing

Added more test cases to verify the fixed behaviour as well as corner cases not covered previously

## Release Notes

**WHAT**:<br>
Properly enforce network call per session limit and use the remote limit as a ceiling for any local limits set.
**WHY**:<br>
Limiting the number of session calls recorded prevents session payloads from growing to an unhealthy size.

### DIFF
--- a/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/config/behavior/NetworkBehavior.kt
+++ b/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/config/behavior/NetworkBehavior.kt
@@ -4,6 +4,7 @@ import io.embrace.android.embracesdk.config.local.SdkLocalConfig
 import io.embrace.android.embracesdk.config.remote.NetworkCaptureRuleRemoteConfig
 import io.embrace.android.embracesdk.config.remote.RemoteConfig
 import java.util.regex.Pattern
+import kotlin.math.min
 
 /**
  * Provides the behavior that functionality relating to network call capture should follow.
@@ -68,28 +69,27 @@ internal class NetworkBehavior(
      * List of domains to be limited for tracking.
      */
     fun getNetworkCallLimitsPerDomain(): Map<String, Int> {
-        return remote?.networkConfig?.domainLimits
-            ?: transformLocalDomainCfg()
-            ?: HashMap()
-    }
+        val limitCeiling = getLimitCeiling()
+        val domainLimits: MutableMap<String, Int> = remote?.networkConfig?.domainLimits?.toMutableMap() ?: mutableMapOf()
 
-    private fun transformLocalDomainCfg(): Map<String, Int>? {
-        val mergedLimits: MutableMap<String, Int> = HashMap()
-        for (domain in local?.networking?.domains ?: return null) {
-            if (domain.domain != null && domain.limit != null) {
-                mergedLimits[domain.domain] = domain.limit
+        local?.networking?.domains?.forEach { localDomainLimit ->
+            if (localDomainLimit.domain != null && localDomainLimit.limit != null) {
+                domainLimits[localDomainLimit.domain] =
+                    domainLimits[localDomainLimit.domain]?.let { remoteLimit ->
+                        min(remoteLimit, localDomainLimit.limit)
+                    } ?: min(limitCeiling, localDomainLimit.limit)
             }
         }
-        return mergedLimits
+
+        return domainLimits
     }
 
     /**
-     * Gets the capture limit for network calls.
+     * Gets the default limit for network calls for all domains where the limit is not specified.
      */
     fun getNetworkCaptureLimit(): Int {
-        return remote?.networkConfig?.defaultCaptureLimit
-            ?: local?.networking?.defaultCaptureLimit
-            ?: DEFAULT_NETWORK_CALL_LIMIT
+        val remoteDefault = getLimitCeiling()
+        return min(remoteDefault, local?.networking?.defaultCaptureLimit ?: remoteDefault)
     }
 
     /**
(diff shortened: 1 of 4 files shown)


bidetofevil:
* master
  * **PR #78** <a href="https://app.graphite.dev/github/pr/embrace-io/embrace-android-sdk/78?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/embrace-io/embrace-android-sdk/78?utm_source=stack-comment).

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/config/behavior/NetworkBehavior.kt
```diff
@@ -68,28 +69,27 @@ internal class NetworkBehavior(
...
-            ?: transformLocalDomainCfg()
-            ?: HashMap()
-    }
-
-    private fun transformLocalDomainCfg(): Map<String, Int>? {
-        val mergedLimits: MutableMap<String, Int> = HashMap()
-        for (domain in local?.networking?.domains ?: return null) {
-            if (domain.domain != null && domain.limit != null) {
-                mergedLimits[domain.domain] = domain.limit
+        val limitCeiling = getLimitCeiling()
```
bidetofevil, line 72: The algo is:

- Use the remote settings as a base
- For domains where there is both a local and remote entry, apply the local setting per domain if the local limit is smaller than the remote one
- For domains with only a local entry, apply the local setting or the ceiling as defined by the default limit set on the remote, which ever is smaller.

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/network/logging/EmbraceNetworkLoggingService.kt
```diff
@@ -166,88 +168,66 @@
...
-                // only capture if the ipAddressCount has not exceeded defaultLimit
-                logger.logDeveloper("EmbraceNetworkLoggingService", "capturing network call")
+            if (ipAddressNetworkCallCount.getAndIncrement() < defaultPerDomainCallLimit) {
                 storeNetworkCall(callId, networkCall)
-            } else {
-                logger.logDeveloper("EmbraceNetworkLoggingService", "capture limit exceeded")
             }
             return
+        } else if (!domainSetting.containsKey(domain)) {
+            createLimitForDomain(domain)
```
bidetofevil, line 182: Added the fetching of the domain limits here to run before we record a network request. It was doing it after and I have no idea why.

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/network/logging/EmbraceNetworkLoggingService.kt
```diff
@@ -166,88 +168,66 @@
...
                     "EmbraceNetworkLoggingService",
-                    "Call per domain $domain ${count.requestCount + 1}"
+                    "Call per domain $domain ${countPerSuffix.requestCount + 1}"
                 )
             }
         }
     }
 
-    private fun storeSettings(url: String) {
+    private fun createLimitForDomain(domain: String) {
```
bidetofevil, line 218: The old implementation is really wasteful in that it rebuild the domain limits map at every call, often unnecessarily. So we build and cache it for reuse at init time and redo it when the session ends instead.

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/network/logging/EmbraceNetworkLoggingService.kt
```diff
@@ -68,8 +74,6 @@
             logger.logError(msg, IllegalStateException(msg), true)
         }
 
-        // clear calls per domain and session network calls lists before be used by the next session
```
bidetofevil, line 71: Lol is one line is probably the most important change in the PR. We were resetting this every time a session payload is built, which effectively nerfs the limit.

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/config/behavior/NetworkBehavior.kt
fractalwrench: Consider leaving as a comment in the code?

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/network/logging/EmbraceNetworkLoggingService.kt
fractalwrench: Periodic session caching strikes again 💥

fractalwrench (approved):
LGTM

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/config/behavior/NetworkBehavior.kt
bidetofevil: Good idea

PR #80, opened 2023-11-16
Move session enum and session properties

fractalwrench:
## Goal

Moves an enum out of the `SessionHandler` class & also avoids unnecessary passing of the `sessionProperties` as a function parameter.

## Testing


bidetofevil (approved):
Hell yeah. Love to see unused dependancies ripped out!

PR #81, opened 2023-11-16
Extract session ordinal logic to preference service

fractalwrench:
## Goal

Extracts the business logic for incrementing the session number to the preference service. This means `SessionHandler` no longer needs knowledge about how the ordinal is calculated.

As far as the session number calculation goes, I think the implementation should be thread-safe. Starting a session is always synchronised with a lock, and that's the only place the function is called. Flushing the change to shared preferences is done using `apply()` which is async, but internally within the OS implementation the change is kept in an in-memory hashmap. The only two scenarios where the ordinal won't be incremented is if I/O fails before the process can terminate (e.g. due to a race), or if storage is wiped by a user (in which case a new device ID/ordinal is created).

## Testing

Updated unit test coverage.


bidetofevil (approved):
LGTM.

Unrelated to this change, we should probably a similar counting for background activities (or account for background activities in this scheme... some how). I suspect we lose more of those than regular sessions given how much of an second-class citizen it seems to be.

PR #82, opened 2023-11-16
Rollback behavior for 4xx and 5xx status code.

lucaslabari:
## Goal

- In [this PR](https://github.com/embrace-io/embrace-android-sdk/pull/63) we introduced some changes on the behavior when 4xx and 5xx status code is received.
- The change on that PR was not to retry when receiving those status codes because they are not expected to succeed on a retry. 
- We are now rolling back that change because 2 functional tests failed and we plan to reintroduce these changes soon. 

## Testing

Covered by existing tests.

### DIFF
--- a/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/comms/api/EmbraceApiService.kt
+++ b/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/comms/api/EmbraceApiService.kt
@@ -216,29 +216,9 @@ internal class EmbraceApiService(
 
     @Suppress("UseCheckOrError")
     private fun executePost(request: ApiRequest, payload: ByteArray) {
-        when (val response = apiClient.executePost(request, payload)) {
-            is ApiResponse.Success -> {}
-            is ApiResponse.TooManyRequests -> {
-                // Temporarily, we just throw an exception. In the future,
-                // we will use the retryAfter to schedule a retry.
-                val retryAfter = response.retryAfter ?: 3
-                throw IllegalStateException("Too many requests. Will retry after $retryAfter seconds.")
-            }
-            is ApiResponse.PayloadTooLarge -> {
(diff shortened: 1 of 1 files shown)


embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/comms/api/EmbraceApiService.kt
```diff
@@ -216,29 +216,9 @@ internal class EmbraceApiService(
...
-            is ApiResponse.Incomplete -> {
-                // Retry on incomplete response.
-                throw IllegalStateException("Failed to retrieve from Embrace server.")
-            }
-            is ApiResponse.NotModified -> {
-                // Not expected to receive a 304 response for a POST request.
-            }
+        val response = apiClient.executePost(request, payload)
+        if (response !is ApiResponse.Success) {
+            throw IllegalStateException("Failed to retrieve from Embrace server.")
```
bidetofevil, line 221: post will probably not ever get a NOT_MODIFIED right now, but what do you think about just adding a `throw` in the `is` blocks for `Failure` and `PayloadTooLarge`? That way we keep as much of the existing implementation as possible, but still preserve the same behaviour

bidetofevil (approved):
I'd consider just adding throws to the failure and payloadTooLarge blocks instead as the current behaviour doesn't throw if we get a 304 (even though the server doesn't return that right now I don't think for posts)

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/comms/api/EmbraceApiService.kt
lucaslabari: I prefer to keep this working as it was before the other PR so then we can analyze all cases together and do the corresponding changes.

PR #83, opened 2023-11-16
Enforce call order for session/background activity state transition

fractalwrench:
## Goal

Enforces a specific call order for the session/background activity state transition. Previously we prioritised the session service but were not prioritising the background activity service.

The existing behavior almost certainly led to subtle bugs. As an example, the background activity service sets a session ID on the `MetadataService`. If this was called in the wrong order then events/logs could have the wrong session ID towards the end of a session.

## Testing

Added unit test coverage.

### DIFF
--- a/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/EmbraceImpl.java
+++ b/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/EmbraceImpl.java
@@ -762,6 +762,7 @@ final class EmbraceImpl {
                 serviceRegistry.close();
                 internalEmbraceLogger.logDeveloper("Embrace", "Services closed");
                 workerThreadModule.close();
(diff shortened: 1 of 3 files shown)


bidetofevil (approved):
Very nice. I wonder if we can merge the session and background activity services in the future, as it seems like they're serving very similar functions, but slightly differently.

fractalwrench:
Converting to a draft - there are some functional test failures here around rejected executions that I need to dig into further.

fractalwrench:
Updated changeset to unregister the lifecycle callbacks, as this was failing the tests in a subtle way (by throwing `RejectedExecutionException` when the previous `EmbraceImpl` in `BaseTest` was shutdown).

bidetofevil:
Functional tests doing their job!

bidetofevil (approved):
LGTM

PR #84, opened 2023-11-16
Get rid of unnecessary @VisibleForTest annotations

bidetofevil:
## Goal

Get rid of unnecessary annotations to expose internal member variables to tests. It's good that we don't need any of them because we really shouldn't expose internal state for tests if we don't have to, which for the most part we really shouldn't have to.


bidetofevil:
* master
  * **PR #84** <a href="https://app.graphite.dev/github/pr/embrace-io/embrace-android-sdk/84?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/embrace-io/embrace-android-sdk/84?utm_source=stack-comment).

priettt (approved):

PR #85, opened 2023-11-17
Use the over-the-wire response size when logging URLConnection network calls

bidetofevil:
## Goal

When we are wrapping the input bytestream so that we can log the response body when that feature is turned on, don't use the byte count from the wrapped, ungzipped byte stream for the size of the response. When this feature is off and we don't wrap the input stream, we just use the bytes over the wire (as defined by the content-length header). Reading the uncompressed bytes would be wrong anyway since it's not what's gone through the wire, so if we don't get a content-header, tough beans.

## Testing

Changed the existing test that was asserting the value of the compressed bytes.


bidetofevil:
* master
  * **PR #85** <a href="https://app.graphite.dev/github/pr/embrace-io/embrace-android-sdk/85?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/embrace-io/embrace-android-sdk/85?utm_source=stack-comment).

fractalwrench (approved):
LGTM

PR #87, opened 2023-11-17
Add ordinal number to track background activity requests

fractalwrench:
## Goal

Each session has an incrementing integer included in its payload that helps us diagnose what % of sessions don't make it to the backend. Background activities don't have this in their payload so it's impossible to check the loss rate. This was noted in #81 

I've added this using the same key that is used for sessions (`s.sn`). Capturing this data will allow us to run similar analysis as we have done for sessions.

## Testing

Added unit test coverage.


lucaslabari (approved):
LGTM

pablomatiasgomez (approved):

PR #88, opened 2023-11-17
Add tests for state transition

fractalwrench:
## Goal

Adds tests for the existing session/background activity state transition behavior & formalises some of the interfaces by making them extend `ProcessStateListener`. I've separated this out from #83 to make it easier to debug test failures on that branch.

## Testing

Added unit test coverage.


bidetofevil (approved):
LGTM

PR #89, opened 2023-11-17
Remove process listening from user service

fractalwrench:
## Goal

Removes an unnecessary listener - the user service does not implement the onForeground/onBackground callbacks so doesn't need to implement this interface.


bidetofevil (approved):
LGTM

PR #90, opened 2023-11-17
Remove unused method from DeliveryService.

lucaslabari:
## Goal

- Removed sendEvent() method from DeliveryService because it was not being called. 
- Renamed sendLogs() to sendLog()

## Testing

Updated unit tests.


bidetofevil (approved):
LGTM. Digging deeper into the code, I wonder if we can remove some of these methods in DeliveryService whose implementation is just calling ApiService - and just have the the caller consume the ApiService directly?

lucaslabari:
> LGTM. Digging deeper into the code, I wonder if we can remove some of these methods in DeliveryService whose implementation is just calling ApiService - and just have the the caller consume the ApiService directly?

Yeap, that would be a good think to do so we remove the deliveryService from being the middleman.

PR #94, opened 2023-11-20
Move active session down to SessionHandler

fractalwrench:
## Goal

Moves the `activeSession` property down to `SessionHandler` & makes it private. This avoids having to pass state between the two classes unnecessarily.

## Testing


bidetofevil (approved):
LGTM. 

One question though: is there a reason (other than inertia) that we are using so many volatile member variables instead of using atomic references? Any time we have mutable state that can be accessed by multiple threads, we should think carefully about synchronization, and while making something atomic doesn't automatically guarantee thread safety, at least it synchronizes access and forces us (in theory) to think about how that is changed.

fractalwrench:
I think it's mainly inertia & there isn't a good reason why better synchronisation mechanisms aren't being used.

PR #95, opened 2023-11-20
Move periodic caching to SessionHandler

fractalwrench:
## Goal

Moves the periodic caching functionality to the `SessionHandler` class.

## Testing


bidetofevil (approved):
Very nice!

PR #96, opened 2023-11-20
Move disabling of session creation

fractalwrench:
## Goal

Moves up the disabling of sessions to `SessionService`.

## Testing


bidetofevil (approved):
LGTM. We may want too look at why we're checking whether we can start a session when we are handling a crash though. Seems odd.

PR #99, opened 2023-11-20
Synchronize network call logging

bidetofevil:
## Goal

Properly synchronize the update and fetching of network calls. The previous strategy only accounted for the modification of the network call storage, but left limits tracking and limit overage reporting open to race conditions. This change makes sure that while a network call is being logged, we can't fetch a current snapshot for the session payload. 

Also, we were trying to be clever about when when to lock when we take the snapshot, but it just didn't work (i.e. the things can change once the calls were fetched but before the payload is returned). As much as we can, we put work outside the synchronization block if we could (i.e. when we validate the staleness of the cache, which should not be a problem now), but in most cases, the locking isn't too bad (i.e. cache read and limits overage data construction), and all we're blocking are background threads. So this should be OK? *sweat_smile

## Testing

Existing tests pass. The stateless internal error was happening once every ~650 session or so, and this change should bring it down to 0

### DIFF
--- a/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/network/logging/EmbraceNetworkLoggingService.kt
+++ b/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/network/logging/EmbraceNetworkLoggingService.kt
@@ -53,28 +53,33 @@ internal class EmbraceNetworkLoggingService(
     private var domainSuffixCallLimits = configService.networkBehavior.getNetworkCallLimitsPerDomainSuffix()
 
     override fun getNetworkCallsForSession(): NetworkSessionV2 {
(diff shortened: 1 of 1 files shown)


bidetofevil:
* master
  * **PR #99** <a href="https://app.graphite.dev/github/pr/embrace-io/embrace-android-sdk/99?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/embrace-io/embrace-android-sdk/99?utm_source=stack-comment).

fractalwrench (approved):
LGTM

PR #102, opened 2023-11-21
Add remote config for disabling the session start message

fractalwrench:
## Goal

Adds a remote config field that prevents the session start message from being sent. We don't really use the start message & it adds bandwidth when making requests + complexity to cache failed requests on disks. Therefore we will add a remote config that disables sending it so that we can experiment with disabling it by default.

## Testing

Added unit test coverage. Confirmed in an example app that sessions still show on the dashboard.

### DIFF
--- a/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/config/behavior/SessionBehavior.kt
+++ b/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/config/behavior/SessionBehavior.kt
@@ -143,6 +143,14 @@ internal class SessionBehavior(
         getFullSessionEvents().contains(SessionGatingKeys.FULL_SESSION_ERROR_LOGS)
 
     /**
+     * Checks whether the session start message should be sent to the API or not.
+     */
+    fun isStartMessageEnabled(): Boolean {
+        return thresholdCheck.isBehaviorEnabled(remote?.sessionConfig?.pctStartMessageEnabled)
+            ?: true
+    }
+
+    /**
      * Checks whether a feature should be gated.
      * If [getSessionComponents] is null, this will return false.
      * If [getSessionComponents] is empty, this will return true.
--- a/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/config/remote/SessionRemoteConfig.kt
+++ b/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/config/remote/SessionRemoteConfig.kt
@@ -11,6 +11,12 @@ internal data class SessionRemoteConfig(
     @SerializedName("enable")
     val isEnabled: Boolean? = null,
 
+    /**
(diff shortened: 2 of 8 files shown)


embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/config/behavior/SessionBehavior.kt
```diff
@@ -142,6 +142,14 @@ internal class SessionBehavior(
     fun shouldSendFullForErrorLog() =
         getFullSessionEvents().contains(SessionGatingKeys.FULL_SESSION_ERROR_LOGS)
 
+    /**
+     * Checks whether the session start message should be sent to the API or not.
+     */
+    fun isStartMessageEnabled(): Boolean {
+        return thresholdCheck.isBehaviorEnabled(remote?.sessionConfig?.pctStartMessageEnabled)
+            ?: true
```
bidetofevil, line 150: What do you think about getting wild and defaulting to `false` 😈

bidetofevil (approved):
LGTM. But what do you think about defaulting to false? Or do you want to enable this for a couple of apps first, monitor, then changing it in the next SDK version?

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/config/behavior/SessionBehavior.kt
fractalwrench: I'd be happy with that - probably something to run past @fnewberg when he's back.

bidetofevil: Lets just do it this way and we can turn this on by default in 6.2. Given that I want to start the release today, I'm just going to merge this for you so we can get it as part of 6.1. Since it's default off right now, it should be fine. :-)

fnewberg: @fractalwrench did you talk with pablo about just turning this off? i'm pretty sure we can just remove it entirely rather than adding a config for it

fnewberg:
> @fnewberg tagging you for review as this adds a new remote config field

if we decide to keep the start message (see earlier comment) i'll add a config for this when i'm back

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/config/behavior/SessionBehavior.kt
fractalwrench: IIRC our previous conversations suggested adding a config flag to disable this in case anything unforeseen did go wrong. We can switch the default or remove the flag entirely once we've validated that is true

Pablo did not have any concerns about not sending the start message

PR #104, opened 2023-11-21
Detect unbalanced process lifecycle calls

fractalwrench:
## Goal

Detects when a process lifecycle state change is unbalanced. Theoretically the `androidx` library should not make unbalanced calls, but I'd like to confirm this hypothesis by logging an internal error that will show up in Grafana if the scenario does occur.

If an unbalanced call does occur for an `onForeground` event it may cause data loss as a new session will be created & the session ordinal incremented. Therefore I have only logged an internal error if this happens for the foreground lifecycle state change. 

## Testing

Added unit test coverage.


bidetofevil (approved):
LGTM

PR #105, opened 2023-11-21
Harden session message creation

fractalwrench:
## Goal

Catches any exceptions that are thrown from individual services when building the session message. Exceptions could result in the loss of a start/end message so if they happen we should catch it, set the field to null, and log an internal error so we can fix the error in future releases.

## Testing

### DIFF
--- a/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/capture/EmbracePerformanceInfoService.kt
+++ b/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/capture/EmbracePerformanceInfoService.kt
@@ -11,8 +11,10 @@ import io.embrace.android.embracesdk.capture.powersave.PowerSaveModeService
 import io.embrace.android.embracesdk.capture.strictmode.StrictModeService
 import io.embrace.android.embracesdk.logging.InternalStaticEmbraceLogger.Companion.logDeveloper
 import io.embrace.android.embracesdk.network.logging.NetworkLoggingService
+import io.embrace.android.embracesdk.payload.AppExitInfoData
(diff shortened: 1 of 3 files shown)


embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/capture/EmbracePerformanceInfoService.kt
```diff
@@ -69,10 +88,23 @@
...
+                powerSaveModeService.getCapturedData()?.toList()
+            },
         )
     }
+
+    private inline fun <R> captureSafely(result: () -> R): R? {
+        return try {
+            result()
+        } catch (exc: Throwable) {
+            InternalStaticEmbraceLogger.logger.logError("Exception thrown capturing session data", exc)
```
bidetofevil, line 106: Can we pass in the logger instance we initialize in the service so this would bubble up while we run integration tests?

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/session/SessionMessageCollator.kt
```diff
@@ -147,10 +148,22 @@
         )
     }
 
+    private inline fun <R> captureSafely(result: () -> R): R? {
```
bidetofevil, line 151: I wonder if we can stick this in a location where it can be used in the various places we have to do this. And it might be good if we have have an indication of what failed in the internal error so we know what exactly is going wrong.

This could be done in a separate PR BTW

bidetofevil (approved):
Couple of suggestions that can be done in a separate PR, but LGTM in general.

PR #107, opened 2023-11-21
[EMBR-1196] Create a retry queue for each endpoint.

lucaslabari:
## Goal

- Created a retry queue for each endpoint. 
- Now it loads and saves from cache the new map that holds the retry queues for each endpoint. 
- When loads return null, we still need to verify if it's returning null because of an exception caused by an old file version. (we can discuss doing this differently)
- Now, the limits are established for each endpoint.

## Testing

Added new unit tests and updated others.

### DIFF
--- a/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/comms/api/EmbraceApiService.kt
+++ b/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/comms/api/EmbraceApiService.kt
@@ -228,20 +228,7 @@ internal class EmbraceApiService(
             BLOBS("blobs"),
             LOGGING("logging"),
             NETWORK("network"),
-            SESSIONS("sessions");
-
-            companion object {
-                fun fromEmbraceUrl(embraceUrl: EmbraceUrl): Endpoint? {
-                    return when (embraceUrl.url.path.substringAfterLast("/")) {
-                        EVENTS.path -> EVENTS
-                        BLOBS.path -> BLOBS
-                        LOGGING.path -> LOGGING
-                        NETWORK.path -> NETWORK
-                        SESSIONS.path -> SESSIONS
-                        else -> null
-                    }
-                }
-            }
+            SESSIONS("sessions")
         }
     }
 }
--- a/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/comms/api/EmbraceApiService.kt
+++ b/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/comms/api/EmbraceApiService.kt
@@ -228,7 +228,8 @@ internal class EmbraceApiService(
             BLOBS("blobs"),
             LOGGING("logging"),
             NETWORK("network"),
-            SESSIONS("sessions")
+            SESSIONS("sessions"),
+            UNKNOWN("unknown")
         }
     }
 }
--- a/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/comms/api/EmbraceApiService.kt
+++ b/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/comms/api/EmbraceApiService.kt
@@ -228,7 +228,20 @@ internal class EmbraceApiService(
             BLOBS("blobs"),
             LOGGING("logging"),
             NETWORK("network"),
-            SESSIONS("sessions")
+            SESSIONS("sessions");
+
+            companion object {
+                fun fromEmbraceUrl(embraceUrl: EmbraceUrl): Endpoint? {
+                    return when (embraceUrl.url.path.substringAfterLast("/")) {
+                        EVENTS.path -> EVENTS
+                        BLOBS.path -> BLOBS
+                        LOGGING.path -> LOGGING
+                        NETWORK.path -> NETWORK
+                        SESSIONS.path -> SESSIONS
+                        else -> null
+                    }
+                }
+            }
         }
     }
 }
--- a/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/comms/api/EmbraceUrl.kt
+++ b/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/comms/api/EmbraceUrl.kt
@@ -12,7 +12,7 @@ internal class EmbraceUrl(val url: URL) {
         return EmbraceConnectionImpl(url.openConnection() as HttpURLConnection, this)
     }
 
-    fun toEndpoint(): Endpoint? {
+    fun endpoint(): Endpoint? {
         return when (url.path.substringAfterLast("/")) {
             Endpoint.EVENTS.path -> Endpoint.EVENTS
             Endpoint.BLOBS.path -> Endpoint.BLOBS
--- a/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/comms/api/EmbraceUrl.kt
+++ b/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/comms/api/EmbraceUrl.kt
@@ -1,5 +1,6 @@
 package io.embrace.android.embracesdk.comms.api
 
+import io.embrace.android.embracesdk.comms.api.EmbraceApiService.Companion.Endpoint
 import java.io.IOException
 import java.net.HttpURLConnection
 import java.net.MalformedURLException
@@ -9,6 +10,17 @@ internal class EmbraceUrl(val url: URL) {
     @Throws(IOException::class)
     fun openConnection(): EmbraceConnection {
         return EmbraceConnectionImpl(url.openConnection() as HttpURLConnection, this)
+    }
+
+    fun toEndpoint(): Endpoint? {
+        return when (url.path.substringAfterLast("/")) {
+            Endpoint.EVENTS.path -> Endpoint.EVENTS
+            Endpoint.BLOBS.path -> Endpoint.BLOBS
+            Endpoint.LOGGING.path -> Endpoint.LOGGING
+            Endpoint.NETWORK.path -> Endpoint.NETWORK
+            Endpoint.SESSIONS.path -> Endpoint.SESSIONS
+            else -> null
+        }
     }
 
     override fun toString(): String {
--- a/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/comms/api/EmbraceUrl.kt
+++ b/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/comms/api/EmbraceUrl.kt
@@ -12,14 +12,14 @@ internal class EmbraceUrl(val url: URL) {
         return EmbraceConnectionImpl(url.openConnection() as HttpURLConnection, this)
     }
 
-    fun endpoint(): Endpoint? {
+    fun endpoint(): Endpoint {
         return when (url.path.substringAfterLast("/")) {
             Endpoint.EVENTS.path -> Endpoint.EVENTS
             Endpoint.BLOBS.path -> Endpoint.BLOBS
             Endpoint.LOGGING.path -> Endpoint.LOGGING
             Endpoint.NETWORK.path -> Endpoint.NETWORK
             Endpoint.SESSIONS.path -> Endpoint.SESSIONS
-            else -> null
+            else -> Endpoint.UNKNOWN
         }
     }
 
--- a/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/comms/delivery/DeliveryFailedApiCalls.kt
+++ b/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/comms/delivery/DeliveryFailedApiCalls.kt
@@ -1,8 +1,78 @@
 package io.embrace.android.embracesdk.comms.delivery
 
 import io.embrace.android.embracesdk.comms.api.ApiRequest
+import io.embrace.android.embracesdk.comms.api.EmbraceApiService.Companion.Endpoint
+import java.util.concurrent.ConcurrentHashMap
 import java.util.concurrent.ConcurrentLinkedQueue
 
+/**
+ * A queue containing failed API calls.
+ */
 internal class DeliveryFailedApiCalls : ConcurrentLinkedQueue<DeliveryFailedApiCall>()
 
-internal data class DeliveryFailedApiCall(val apiRequest: ApiRequest, val cachedPayload: String)
+/**
+ * A failed API call.
+ */
+internal data class DeliveryFailedApiCall(
+    val apiRequest: ApiRequest,
+    val cachedPayload: String,
+    val queueTime: Long? = null
+)
+
+/**
+ * A map containing a list of failed API calls per endpoint.
+ */
+internal class FailedApiCallsPerEndpoint : ConcurrentHashMap<Endpoint, DeliveryFailedApiCalls>() {
+
+    /**
+     * Adds a failed API call in the corresponding endpoint's list.
+     */
+    fun add(endpoint: Endpoint, failedApiCall: DeliveryFailedApiCall) {
+        if (this.containsKey(endpoint)) {
+            this[endpoint]?.add(failedApiCall)
+        } else {
+            val failedApiCalls = DeliveryFailedApiCalls()
+            failedApiCalls.add(failedApiCall)
+            this[endpoint] = failedApiCalls
+        }
+    }
+
+    /**
+     * Returns the number of failed API calls in the corresponding endpoint's list.
+     */
+    fun failedApiCallsCount(endpoint: Endpoint) = this[endpoint]?.size ?: 0
+
+    /**
+     * Returns true if there are any failed API calls in any endpoint's list.
+     */
+    fun hasAnyFailedApiCalls(): Boolean {
+        return this.values.any { it.isNotEmpty() }
+    }
+
+    /**
+     * Returns true if there are no failed API calls in any endpoint's list.
+     */
+    fun hasNoFailedApiCalls(): Boolean {
+        return !hasAnyFailedApiCalls()
+    }
+
+    /**
+     * Returns the next failed API call to be sent and removes it from the corresponding
+     * endpoint's list.
+     */
+    fun pollNextFailedApiCall(): DeliveryFailedApiCall? {
+        this[Endpoint.SESSIONS]?.let { sessions ->
+            if (sessions.isNotEmpty()) {
+                return sessions.poll()
+            }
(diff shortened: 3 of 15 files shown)


embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/comms/delivery/DeliveryFailedApiCalls.kt
```diff
@@ -1,8 +1,78 @@
...
+        this[Endpoint.SESSIONS]?.let { sessions ->
+            if (sessions.isNotEmpty()) {
+                return sessions.poll()
+            }
+        }
+
+        val entryToPollFrom = this
+            .entries
+            .filter { it.value.isNotEmpty() }
+            .minByOrNull { it.value.peek()?.queueTime ?: Long.MAX_VALUE }
```
fractalwrench, line 73: This behavior is inconsistent with returning from the `sessions` collection higher up. Having said that, I'm unsure on whether we should bother changing the behavior to get deterministic request order as part of this PR. @bidetofevil thoughts?

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/comms/delivery/EmbraceDeliveryRetryManager.kt
```diff
@@ -42,16 +47,23 @@ internal class EmbraceDeliveryRetryManager(
...
-            retryQueue.add(failedApiCall)
-            cacheManager.saveFailedApiCalls(retryQueue)
-
-            // By default there are no scheduled retry jobs pending. If the retry queue was initially empty, try to schedule a retry.
-            if (scheduleJob) {
-                scheduleFailedApiCallsRetry(RETRY_PERIOD)
+
+        val endpoint = Endpoint.fromEmbraceUrl(request.url)
+        endpoint?.let { e ->
+            if (isBelowRetryLimit(e)) {
```
fractalwrench, line 53: This works fine, ideally I'd like to change the strategy so the oldest retry gets dropped, rather than everything new. Also, we should ideally record somewhere that a retry was dropped. Not something to do in this PR

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/comms/api/EmbraceApiService.kt
```diff
@@ -228,7 +228,20 @@ internal class EmbraceApiService(
             BLOBS("blobs"),
             LOGGING("logging"),
             NETWORK("network"),
-            SESSIONS("sessions")
+            SESSIONS("sessions");
+
+            companion object {
+                fun fromEmbraceUrl(embraceUrl: EmbraceUrl): Endpoint? {
```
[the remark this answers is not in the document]
lucaslabari: Agree, that's a better way. Done.

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/comms/delivery/DeliveryFailedApiCalls.kt
```diff
@@ -1,8 +1,78 @@
...
+ */
 internal class DeliveryFailedApiCalls : ConcurrentLinkedQueue<DeliveryFailedApiCall>()
 
-internal data class DeliveryFailedApiCall(val apiRequest: ApiRequest, val cachedPayload: String)
+/**
+ * A failed API call.
+ */
+internal data class DeliveryFailedApiCall(
+    val apiRequest: ApiRequest,
+    val cachedPayload: String,
```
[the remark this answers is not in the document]
lucaslabari: Renamed, but added @SerializedName("cachedPayload") so it doesn't break compatibility with already existing files.

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/comms/delivery/DeliveryFailedApiCalls.kt
```diff
@@ -1,8 +1,78 @@
...
+    val queueTime: Long? = null
+)
+
+/**
+ * A map containing a list of failed API calls per endpoint.
+ */
+internal class FailedApiCallsPerEndpoint : ConcurrentHashMap<Endpoint, DeliveryFailedApiCalls>() {
```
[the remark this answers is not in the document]
lucaslabari: Yeap, totally. Updated.

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/comms/delivery/DeliveryFailedApiCalls.kt
```diff
@@ -1,8 +1,78 @@
...
+
+    /**
+     * Adds a failed API call in the corresponding endpoint's list.
+     */
+    fun add(endpoint: Endpoint, failedApiCall: DeliveryFailedApiCall) {
+        if (this.containsKey(endpoint)) {
+            this[endpoint]?.add(failedApiCall)
```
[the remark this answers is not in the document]
lucaslabari: I didn't use putIfAbsent because it's not what I needed for this case, but ended up using `getOrPut()` that is also a thread safe method.

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/comms/delivery/EmbraceDeliveryCacheManager.kt
```diff
@@ -184,31 +185,56 @@ internal class EmbraceDeliveryCacheManager(
...
+     * Loads the old version of the [FAILED_API_CALLS_FILE_NAME] file where
+     * it was storing [DeliveryFailedApiCalls] instead of [FailedApiCallsPerEndpoint]
+     */
+    private fun loadFailedApiCallsOldVersion(): FailedApiCallsPerEndpoint? {
+        var cachedApiCallsPerEndpoint: FailedApiCallsPerEndpoint? = null
+        val cachedApiCalls = cacheService.loadObject(FAILED_API_CALLS_FILE_NAME, DeliveryFailedApiCalls::class.java)
+
+        if (cachedApiCalls != null) {
+            cachedApiCallsPerEndpoint = FailedApiCallsPerEndpoint()
+            cachedApiCalls.forEach { cachedApiCall ->
```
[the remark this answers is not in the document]
lucaslabari: As you mentioned in another comment I wrapped the map inside `FailedApiCallsPerEndpoint` and improved the implementation of the add method, so we the `associateBy` wouldn't apply.

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/comms/delivery/DeliveryFailedApiCalls.kt
```diff
@@ -1,8 +1,78 @@
...
+    fun hasNoFailedApiCalls(): Boolean {
+        return !hasAnyFailedApiCalls()
+    }
+
+    /**
+     * Returns the next failed API call to be sent and removes it from the corresponding
+     * endpoint's list.
+     */
+    fun pollNextFailedApiCall(): DeliveryFailedApiCall? {
+        this[Endpoint.SESSIONS]?.let { sessions ->
```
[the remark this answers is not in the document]
lucaslabari: I don't think so, what we want here is to poll the next item we want to retry, taking into account the priorities between types/endpoints.

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/comms/delivery/DeliveryFailedApiCalls.kt
lucaslabari: So, the idea we discussed with @bidetofevil was to prioritize sessions and then send payloads in the order they were enqued. 
Basically, we are adding the prioritization of sessions, because the order by queued time was already there as the previous implementation was a FIFO queue.

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/comms/delivery/EmbraceDeliveryRetryManager.kt
lucaslabari: Why are new attempts more valuable than the oldest ones?

fractalwrench: Good question - I don't think it's a clear cut distinction. I'd say they're slightly more valuable as folks care about things that happened recently. Also, there's less chance we're dealing with a bad scenario, like data corruption meaning an old session was stuck on a device & is unsendable.

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/comms/api/EmbraceUrl.kt
```diff
@@ -11,6 +12,17 @@ internal class EmbraceUrl(val url: URL) {
         return EmbraceConnectionImpl(url.openConnection() as HttpURLConnection, this)
     }
 
+    fun toEndpoint(): Endpoint? {
```
bidetofevil, line 15: what about just calling it `endpoint()` since we're not transforming the object?

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/comms/api/EmbraceUrl.kt
```diff
@@ -11,6 +12,17 @@ internal class EmbraceUrl(val url: URL) {
...
+        return when (url.path.substringAfterLast("/")) {
+            Endpoint.EVENTS.path -> Endpoint.EVENTS
+            Endpoint.BLOBS.path -> Endpoint.BLOBS
+            Endpoint.LOGGING.path -> Endpoint.LOGGING
+            Endpoint.NETWORK.path -> Endpoint.NETWORK
+            Endpoint.SESSIONS.path -> Endpoint.SESSIONS
+            else -> null
```
bidetofevil, line 22: Do we want to have a default value that isn't null so if we have a ApiRequest that doesn't fit this scheme later, it'll at least fall through to a generic queue? Something like `Endpoint.Unknown`?

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/comms/delivery/EmbraceDeliveryRetryManager.kt
```diff
@@ -84,25 +96,25 @@ internal class EmbraceDeliveryRetryManager(
     }
 
     /**
-     * Returns true if there is an active pending retry task
+     * Returns true if the number of retries for the endpoint is below the limit.
      */
+    private fun isBelowRetryLimit(endpoint: Endpoint): Boolean {
```
bidetofevil, line 101: Can this just be returned by `FailedApiCallsPerEndpoint` instead of exposing the count?

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/comms/delivery/FailedApiCallsPerEndpoint.kt
```diff
@@ -0,0 +1,80 @@
...
+     */
+    fun add(endpoint: Endpoint, failedApiCall: DeliveryFailedApiCall) {
+        val failedApiCallsForEndpoint = failedApiCallsMap.getOrPut(endpoint) { DeliveryFailedApiCalls() }
+        failedApiCallsForEndpoint.add(failedApiCall)
+    }
+
+    /**
+     * Returns the list of failed API calls for the corresponding endpoint.
+     */
+    fun get(endpoint: Endpoint): DeliveryFailedApiCalls? {
```
bidetofevil, line 24: Do we need to expose this? Is it possible for the interface of the object to just be `add()` `getNext()`, `clear()`, and `hasAnyFailedApiCalls()`, leaving the limits, counts, etc. be internal to the object?

Basically, the retry manager should ask: "do you have a request for me to retry?" and this will just return if it has one. Not sure if it needs to expose more than that right now before we add the 429 stuff...

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/comms/delivery/FailedApiCallsPerEndpoint.kt
```diff
@@ -0,0 +1,80 @@
...
+     */
+    fun hasNoFailedApiCalls(): Boolean {
+        return !hasAnyFailedApiCalls()
+    }
+
+    /**
+     * Returns the next failed API call to be sent and removes it from the corresponding
+     * endpoint's list.
+     */
+    fun pollNextFailedApiCall(): DeliveryFailedApiCall? {
```
bidetofevil, line 63: probably need to synchronize this call with any methods that can modify the internal map? Given the amount of traffic we have, I don't think we need to be more clever than that the the locking

bidetofevil (commented):
Looks good in general! Just some comments about naming, whether we need to expose all the methods that are currently public on `FailedApiCallsPerEndpoint`, and a couple of simplification/cleanup suggestions

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/comms/delivery/EmbraceDeliveryRetryManager.kt
lucaslabari: @eliabsisay I created [this Story](https://www.notion.so/embraceio/DCI-1678d2876bc247a8951b7ee79f25afe2?p=c24c5a8772ec4089890531bd7a693d72&pm=s) to work on it.

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/comms/api/EmbraceUrl.kt
lucaslabari: yeap, that's better, updated.

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/comms/delivery/EmbraceDeliveryCacheManager.kt
```diff
@@ -184,29 +184,68 @@ internal class EmbraceDeliveryCacheManager(
...
+     */
+    override fun loadFailedApiCalls(): FailedApiCallsPerEndpoint {
         logger.logDeveloper(TAG, "Loading failed api calls")
-        val cached = executorService.submit<DeliveryFailedApiCalls> {
-            cacheService.loadObject(FAILED_API_CALLS_FILE_NAME, DeliveryFailedApiCalls::class.java)
+        val cached = executorService.submit<FailedApiCallsPerEndpoint> {
+            val loadFailedApiCalls = runCatching {
+                cacheService.loadObject(FAILED_API_CALLS_FILE_NAME, FailedApiCallsPerEndpoint::class.java)
+            }
+            if (loadFailedApiCalls.isSuccess) {
```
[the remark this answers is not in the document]
lucaslabari: We don't need to delete it because it will be overwritten the next time we save it. And deleting would be bad because we could be missing data if the requests retry don't succeed.

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/comms/delivery/FailedApiCallsPerEndpoint.kt
```diff
@@ -0,0 +1,80 @@
...
+ * A map containing a list of failed API calls per endpoint.
+ */
+internal class FailedApiCallsPerEndpoint {
+
+    private val failedApiCallsMap = ConcurrentHashMap<Endpoint, DeliveryFailedApiCalls>()
+
+    /**
+     * Adds a failed API call in the corresponding endpoint's list.
+     */
+    fun add(endpoint: Endpoint, failedApiCall: DeliveryFailedApiCall) {
```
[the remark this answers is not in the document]
lucaslabari: Good catch, updated.

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/comms/api/EmbraceUrl.kt
lucaslabari: Yes, that's a good idea, I added `Endpoint.Unknown`

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/comms/delivery/EmbraceDeliveryRetryManager.kt
lucaslabari: Yes, I moved this to FailedApiCallsPerEndpoint.

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/comms/delivery/FailedApiCallsPerEndpoint.kt
lucaslabari: Done. I had to update a bit the code in the retryManager to avoid using the count. I leaved just the methods you are mentioning.

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/comms/delivery/FailedApiCallsPerEndpoint.kt
```diff
@@ -0,0 +1,80 @@
+package io.embrace.android.embracesdk.comms.delivery
+
+import io.embrace.android.embracesdk.comms.api.EmbraceApiService.Companion.Endpoint
+import java.util.concurrent.ConcurrentHashMap
+
+/**
+ * A map containing a list of failed API calls per endpoint.
+ */
+internal class FailedApiCallsPerEndpoint {
```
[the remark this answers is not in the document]
lucaslabari: I will do the rename in another PR so I don't complicate the review of this PR and also because we need to modify other components to be consistent on the naming we give to these Api Calls.

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/comms/delivery/FailedApiCallsPerEndpoint.kt
lucaslabari: As we talked on Slack, this class is accessed by a single thread, so I just added a comment in case we start accessing it from multiple threads.

bidetofevil (approved):
Looks good! Lets wait until the release is done before merging this?

PR #108, opened 2023-11-22
More EmbraceImpl cleanup

bidetofevil:
## Goal

Cleaning up `EmbraceImpl` so it's more organized and easier to tease apart. Grouping variables and methods together and giving them the appropriate access modifiers will make it easier to extract features into their own classes to reduce this size of this and make it easier to convert to Kotlin.

The only functional change is to consolidate the `isStarted()` check for many of these calls. The common pattern is to run the main body of the method if the SDK is started, and then log an error if it's not. Instead of having each method implement the if-else-log logic, just used a helper method to do that.

## Testing

This was mostly about moving stuff around, so existing tests should cover this.

### DIFF
--- a/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/Embrace.java
+++ b/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/Embrace.java
@@ -654,7 +654,7 @@ public final class Embrace implements EmbraceAndroidApi {
 
     @InternalApi
     void installUnityThreadSampler() {
-        getImpl().installUnityThreadSampler();
+        impl.installUnityThreadSampler();
     }
 
     /**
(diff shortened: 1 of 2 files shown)


bidetofevil:
* master
  * **PR #108** <a href="https://app.graphite.dev/github/pr/embrace-io/embrace-android-sdk/108?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/embrace-io/embrace-android-sdk/108?utm_source=stack-comment).

lucaslabari (approved):

fractalwrench (approved):
LGTM

PR #109, opened 2023-11-22
Move session implementation to session handler

fractalwrench:
## Goal

I spotted a bit of session handling logic in `EmbraceImpl` so moved it to the `SessionHandler` class, which will aid any future refactoring to make the logic easier to understand.

## Testing

Added unit test coverage.


bidetofevil (approved):
LGTM

PR #111, opened 2023-11-22
Extract creation of background activity message

fractalwrench:
## Goal

Extracts the code that creates background activity objects in the start/stop state out of the service. This will make it easier to refactor in future (and possibly align with the `SessionMessage` as there's a lot of duplication).

## Testing


bidetofevil (approved):
LGTM. No tests that need moving as well now that we have a the Collator object?

PR #112, opened 2023-11-22
Avoid setting process summary state for background activities

fractalwrench:
## Goal

Avoid setting the process summary state for background activities. Currently we only use the AEI to link ANRs to a given session. Overwriting this with an ID for a background activity could plausibly reduce our match rate of ANRs -> Session.


bidetofevil (approved):
LGTM

PR #114, opened 2023-11-23
Harden creation of background activity message

fractalwrench:
## Goal

Catches any exceptions that might be thrown when creating the background activity message & logs them as internal errors. This should improve the reliability of delivery & let us know about any exceptions that we are missing.

## Testing


bidetofevil (approved):
Nice

PR #115, opened 2023-11-23
[POC] combine envelope classes into one

fractalwrench:
## Goal

@bidetofevil could use your opinion on whether it's wise to consolidate some of the payload classes that duplicated fields, and the approach we should take if so. This PR alters the `BackgroundActivityMessage` to make it capable of sending both a `BackgroundActivity` and a `Session`. The only difference between these two classes is whether a background activity or session is serialized under the 's' key, so therefore I've used generics to specify that type.

That ties us to these types always having the same schema. I'm reasonably sure this will always be true (we can check before committing to this approach). Another disadvantage is that it slightly complicates Gson (de)serialization.

A second option I considered is having some form of inheritance hierarchy, either via an interface or an abstract class. This would be highly beneficial for related classes such as `BlobMessage` + `EventMessage`, which share most but not all of the same fields. One downside with this approach is that it negates the possibility of using data classes (we currently use copy & equals functions, but could implement our own).

IMO it makes sense to combine some of these types into combined types, because it will stop future divergence in the payload fields. It should also mean we can share more code (particularly between the session/background activity services, which is the main driver for why I looked at this in the first place).

### DIFF
--- a/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/comms/delivery/DeliveryCacheManager.kt
+++ b/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/comms/delivery/DeliveryCacheManager.kt
@@ -1,5 +1,6 @@
 package io.embrace.android.embracesdk.comms.delivery
 
+import io.embrace.android.embracesdk.payload.BackgroundActivity
 import io.embrace.android.embracesdk.payload.BackgroundActivityMessage
 import io.embrace.android.embracesdk.payload.EventMessage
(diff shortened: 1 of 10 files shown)


bidetofevil:
Combining the envelope for session and background activity seems like a no-brainer. The serialized form is going to be the same, so we wouldn't even have to worry about compatibility, right? The extra complexity to deserialization is a just that we have to get a type, which I don't think is that bad and won't require additional changes if the envelope changes.

Regarding Blob/Event messages, what is the overlap like? I typically balk at created inheritance hierarchies, but if the overlap is a lot, this might be worth it.

PR #116, opened 2023-11-23
Improve logging in session service

fractalwrench:
## Goal

Improves the log messages & moves them to the `SessionHandler` class where appropriate. I also took the chance to inline some functions that are no longer necessary in `SessionService`.

## Testing

### DIFF
--- a/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/payload/Session.kt
+++ b/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/payload/Session.kt
@@ -1,7 +1,6 @@
 package io.embrace.android.embracesdk.payload
 
 import com.google.gson.annotations.SerializedName
-import io.embrace.android.embracesdk.session.EmbraceSessionService
 import io.embrace.android.embracesdk.session.MESSAGE_TYPE_START
 
 /**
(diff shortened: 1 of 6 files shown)


embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/payload/Session.kt
```diff
@@ -142,6 +141,11 @@ internal data class Session @JvmOverloads internal constructor(
 
     companion object {
 
+        /**
+         * Signals to the API that the application was in the foreground.
+         */
+        internal const val APPLICATION_STATE_FOREGROUND = "foreground"
+
         @JvmStatic
         fun buildStartSession(
```
bidetofevil, line 150: Do we even need this given that we're getting rid of the startSession message, or is this doing something else as well?

bidetofevil (approved):
LGTM

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/payload/Session.kt
fractalwrench: We still need this right now as the start message is sent by default. After it's removed we will still need it to populate initial session state, but can rename it to something more appropriate

PR #126, opened 2023-11-24
Add limits for data capture

fractalwrench:
## Goal

Adds some extra limits for data capture, so that the session payload isn't unbounded.

## Testing

Added unit test coverage.


bidetofevil (approved):
Good call. 

Should we let the user know when a limit has been reached in these cases with like a breadcrumb the first time it has been hit in the session, or should we devise a more comprehensive scheme later for all such limit-hitting things?

PR #129, opened 2023-11-27
Extract location of storage directory to one place

fractalwrench:
## Goal

Extracts the location where the storage directory is set to one place in the `EssentialServiceModule`. This will make it much simpler to alter the location of the storage directory (for example, if we want to store in the filesDir rather than cacheDir).

## Testing

Updated unit test coverage.


bidetofevil (approved):
Nice! LGTM

PR #131, opened 2023-11-27
Use static imports in test

fractalwrench:
## Goal

Uses static imports in tests so our style remains consistent.


bidetofevil (approved):
LGTM

PR #132, opened 2023-11-27
Suppress deprecation warnings in tests

fractalwrench:
## Goal

Suppresses some deprecation warnings in the tests that were cluttering up the build log.


bidetofevil (approved):
LGTM. 

Can we do this at a higher level so we don't have to do it on a per-file/test basis? Like everything under `test` or `androidTest` we ignore deprecation warnings.

PR #134, opened 2023-11-27
Fix warning about Java/Kotlin compiler version mismatch

fractalwrench:
## Goal

Fixes a compiler warning about our build's Java/Kotlin versions mismatching. Note that this isn't a problem with the code we were shipping to our customers, rather it's just an annoying message that shows up each build.


bidetofevil (approved):
LGTM

PR #135, opened 2023-11-27
Bump gradle version to 7.6.3

fractalwrench:
## Goal

Bumps the gradle version from 7.4 to 7.6.3.

## Testing

Ran the build, and also did a dry-run of publishing to the snapshot repo. I didn't encounter any obvious build errors


bidetofevil (approved):
Seems reasonable. Why not something even more recent? Are we using deprecated APIs that were removed?

fractalwrench:
@bidetofevil Bumping to Gradle 8.4 felt like it was more likely to break stuff, so I just picked the latest 7.x release. There's no known technical blocker to bumping to 8.4 in future

PR #140, opened 2023-11-29
Add integration test for manually ending session

fractalwrench:
## Goal

Adds an integration test that manually ends a session. This verifies the current behavior (which seems to be buggy & a potential cause of dropped sessions).


bidetofevil (approved):
LGTM

PR #142, opened 2023-11-29
Avoid exponential increase in timed sessions

fractalwrench:
## Goal

Sessions can be automatically closed & a new one started when a customer has opted in to the use of a TIMED session. This is helpful for kiosk mode, where a session may last for multiple hours & capture a large amount of data.

A bug in the `SessionHandler` implementation meant that the automatic closing was not cancelled after each session was ended, meaning there would be an exponential increase in session messages each time the automatic closing happened.

As this is not the default behavior it's assumed this will not affect a large number of customers.

## Testing

Wrote an integration test that discovered this bug.

### DIFF
--- a/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/session/SessionHandler.kt
+++ b/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/session/SessionHandler.kt
@@ -72,6 +72,7 @@ internal class SessionHandler(
     internal fun getSessionId(): String? = activeSession?.sessionId
 
     var scheduledFuture: ScheduledFuture<*>? = null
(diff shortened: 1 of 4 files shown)


embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/session/SessionHandler.kt
```diff
@@ -347,7 +348,8 @@ internal class SessionHandler(
         }
 
         try {
-            this.automaticSessionStopper.scheduleWithFixedDelay(
+            closerFuture?.cancel(true)
```
bidetofevil, line 351: oh no.....

PR #143, opened 2023-11-29
Add integration test for session crash

fractalwrench:
## Goal

Adds an integration test that confirms on a crash, the session is written to disk. I've made a small tweak to make the `activeSession` null, as if the lifecycle callback ends the session, technically we could end up with duplicate session messages racing with each other. This seems reasonably unlikely in practice.


bidetofevil (approved):
LGTM

PR #145, opened 2023-11-29
[EMBR-1534] Rename failed API call classes to Pending API calls.

lucaslabari:
## Goal

- Renamed DeliveryFailedApiCall to PendingApiCall
- Renamed DeliveryFailedApiCalls to PendingApiCallsQueue
- Renamed FailedApiCallsPerEndpoint to PendingApiCalls
- Renamed DeliveryRetryManager to PendingApiCallsSender

## Testing

Updated reference in tests.


bidetofevil (approved):
Looks good. 

I'm wondering if it makes sense in a later PR to funnel all API calls through this instead of sending it directly in ApiService. We'll want to persist the data right before we send the API request because it could fail in the middle, so why not have it all done here instead of only queuing this up when the request fails?

lucaslabari:
> Looks good.
> 
> I'm wondering if it makes sense in a later PR to funnel all API calls through this instead of sending it directly in ApiService. We'll want to persist the data right before we send the API request because it could fail in the middle, so why not have it all done here instead of only queuing this up when the request fails?

That's a good idea; let's discuss it on Slack and create a ticket to address it.

PR #147, opened 2023-11-30
[EMBR-1489] Prioritize most recent api calls when limit is reached.

lucaslabari:
## Goal

- When adding a new pending api call on a full queue (reached the limit) we were not doing the insertion. 
- That means that we were prioritizing old API calls over new ones. 
- This PR adds the new value when the queue is full and removes the oldest one. 
- That means that we are now prioritizing the most recent API calls. 

## Testing

Updated tests accordingly.

### DIFF
--- a/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/comms/delivery/PendingApiCalls.kt
+++ b/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/comms/delivery/PendingApiCalls.kt
@@ -4,15 +4,15 @@ import io.embrace.android.embracesdk.comms.api.EmbraceApiService.Companion.Endpo
 import java.util.concurrent.ConcurrentHashMap
 
 /**
- * A map containing a list of pending API calls per endpoint.
+ * A map containing a queue of pending API calls for each endpoint.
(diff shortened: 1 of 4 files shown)


embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/comms/delivery/PendingApiCalls.kt
```diff
@@ -4,61 +4,64 @@ import io.embrace.android.embracesdk.comms.api.EmbraceApiService.Companion.Endpo
...
+        pendingApiCallsMap[Endpoint.SESSIONS]?.let { sessionsQueue ->
+            synchronized(sessionsQueue) {
+                if (sessionsQueue.isNotEmpty()) {
+                    return sessionsQueue.poll()
+                }
             }
         }
 
-        val entryToPollFrom = pendingApiCallsMap
+        val queueWithMinTime = pendingApiCallsMap
```
bidetofevil, line 43: Can we make sure that sessions aren't pruned? Maybe have a method that only returns the non-session queues and go through that rather than looking at all the entries in the map?

PR #148, opened 2023-12-01
Avoid dropping session when endSession called

fractalwrench:
## Goal

Avoids dropping a session when `endSession` is called and manual session is not enabled. The previous logic did not check whether a session had ended before starting a new one, which resulted in the previous session message getting dropped. This changeset returns the session message if one was ended, or returns null if not.

## Testing

Covered by updating existing integration tests.


bidetofevil (approved):
LGTM

PR #149, opened 2023-12-04
Log internal errors when session writing fails

fractalwrench:
## Goal

Alters logging to ensure an internal error is sent when writing or reading a session fails. This should get us more insight when using the Grafana internal errors dashboard.


bidetofevil (approved):
LGTM

PR #150, opened 2023-12-04
Serialize session message to stream

fractalwrench:
## Goal

Updates the session message serializer so that it's capable of writing information to a stream. This changeset doesn't update the SDK to write directly to streams but lays the groundwork in future. Additionally restructured + renamed a few functions in the class.

## Testing


bidetofevil (approved):
Nice. LGTM.

PR #151, opened 2023-12-04
Reduce memory usage by serializing session via stream

fractalwrench:
## Goal

Reduces memory usage by serializing a session via a `Writer` under certain circumstances. Specifically if the process has crashed or the session is being periodically cached, the bytes from the session payload are not required because no network request is made. Rather than retaining these in memory, we should stream the results directly to disk to reduce memory pressure.

## Testing

Added unit test coverage & manually verified that sessions are still written to disk for each method


bidetofevil (approved):
Maybe this will increase our crash delivery rate if we don't OOM while we're saving a session for a crash. This should help tremendously for the session caching work too.

PR #152, opened 2023-12-05
Gzip encode HTTP requests via stream

fractalwrench:
## Goal

Performs GZIP encoding on the HTTP request `OutputStream` rather than copying the results to a temporary string buffer. This should reduce memory usage as we no longer need to retain the buffer that holds the gzipped contents.

I investigated the possibility of streaming JSON serialization to further reduce memory usage, but Gson doesn't support that without migrating to the manual `JsonReader/JsonWriter` implementations (which we probably don't want to do). Therefore migrating to a different JSON library (probably kotlinx) would be a prerequisite if we wanted to optimize this area of the codebase.

## Testing

Updated unit tests. Confirmed that requests still show up in dashboard.

### DIFF
--- a/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/comms/api/ApiClient.kt
+++ b/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/comms/api/ApiClient.kt
@@ -1,5 +1,7 @@
 package io.embrace.android.embracesdk.comms.api
 
+import java.io.ByteArrayInputStream
+
 /**
  * A simple interface to make internal HTTP requests to the Embrace API
  */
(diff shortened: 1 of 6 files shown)


fractalwrench:
@lucaslabari tagging you for review as I think you did the original gzip implementation, so wanted to check this approach hasn't already been discounted

bidetofevil (commented):
Nice. Eventually, I think we should start persisting gzipped payloads in order to save disk space, so we can move this higher on up the chain. 

I'll leave this for Lucas to approve though given Jamie's question about the original implementation

lucaslabari (approved):
LGTM.

PR #153, opened 2023-12-06
Reduce direct usage of gson

fractalwrench:
## Goal

Alters test code further to avoid directly using Gson. This will make it easier to swap out for another json library.


bidetofevil (approved):
LGTM!

PR #155, opened 2023-12-06
Remove unnecessary serialization functions

fractalwrench:
## Goal

Removes some unnecessary function overloads on `EmbraceSerializer`. This also renames a couple of signatures & takes an `InputStream` & `OutputStream` rather than `Reader/Writer`.

## Testing


bidetofevil (approved):
LGTM

PR #159, opened 2023-12-07
Move embrace serializer to separate package

fractalwrench:
## Goal

Moves the serializer & custom JSON adapters to a separate package so that it's easy to identify where JSON adapters live.


bidetofevil (approved):
LGTM. The test failure will be dealt with after the other PR merges eh?

PR #161, opened 2023-12-07
Fix flaky behavior in CrashSessionTest

fractalwrench:
## Goal

I noticed that `CrashSessionTest` [can flake](https://github.com/embrace-io/embrace-android-sdk/actions/runs/7129063657?pr=159) - I think this is due to a timing issue that causes `AnrService#handleCrash()` to sometimes get invoked. `FakeAnrService` currently throws a `NotImplementedError`. Therefore I've updated all the fake services to do nothing in the case of this function being invoked.


bidetofevil (approved):
LGTM

PR #162, opened 2023-12-07
Enforce static analysis against business logic in data classes

fractalwrench:
## Goal

This is somewhat opinionated but the changeset adds [Detekt rules](https://github.com/embrace-io/embrace-android-sdk/pull/new/data-class-business-logic) that flag any data class containing business logic. This is defined as any mutable variables or the presence of functions that are not on an allowlist. I would like the config/payload classes to be as dumb as possible to make creating fakes etc easy.

I have enabled this for the `io.embrace.android.embracesdk.payload` package only at this point & have suppressed the existing warnings (this prevents the problem getting worse). I haven't enabled it for the config classes as those are mostly not data classes.

### DIFF
--- a/config/detekt/detekt.yml
+++ b/config/detekt/detekt.yml
@@ -64,3 +64,11 @@ style:
     active: true
   UnusedImports:
     active: true
+
+  # Avoid business logic in payload/config classes
+  DataClassShouldBeImmutable:
+    active: true
+    includes: [ '**/io/embrace/android/embracesdk/payload/**' ]
+  DataClassContainsFunctions:
+    active: true
+    includes: [ '**/io/embrace/android/embracesdk/payload/**' ]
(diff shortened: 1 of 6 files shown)


bidetofevil (approved):
I agree with this. Lets bring it up tomorrow in the guild meeting so everyone is onboard and we can make this a more general thing? I want to fix these rather than suppress them going forward so lets make a plan to clean this up after merge this,.

config/detekt/detekt.yml
```diff
@@ -63,4 +63,12 @@ style:
...
     active: true
   UnusedImports:
-    active: true
\ No newline at end of file
+    active: true
+
+  # Avoid business logic in payload/config classes
+  DataClassShouldBeImmutable:
+    active: true
+    includes: [ '**/io/embrace/android/embracesdk/payload/**' ]
```
bidetofevil, line 71: Do we want to remove the restrictions or are there too many data classes we are mutating now?

bidetofevil (approved):
LGTM

config/detekt/detekt.yml
fractalwrench: I decided to confine this to payload/config as that's where most of the data classes should be where we care about business logic not being present

PR #163, opened 2023-12-07
Added EmbraceTelemetryService

priettt:
## Goal

- Count public API calls
- Group other general telemetry, such as disk usage (to be done in the future) or app attributes.

The information is sent privately on the current session span, so we pass EmbraceTelemetryService as a parameter to the SpansService, and fetch and clear the telemetry info every time we flush the spans (when the sessions end).

### DIFF
--- a/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/EmbraceImpl.java
+++ b/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/EmbraceImpl.java
@@ -92,6 +92,7 @@ import io.embrace.android.embracesdk.session.lifecycle.ActivityTracker;
 import io.embrace.android.embracesdk.session.lifecycle.ProcessStateService;
 import io.embrace.android.embracesdk.session.properties.EmbraceSessionProperties;
 import io.embrace.android.embracesdk.session.properties.SessionPropertiesService;
+import io.embrace.android.embracesdk.telemetry.EmbraceTelemetryService;
 import io.embrace.android.embracesdk.utils.PropertyUtils;
 import io.embrace.android.embracesdk.worker.ExecutorName;
 import io.embrace.android.embracesdk.worker.WorkerThreadModule;
@@ -237,6 +238,9 @@ final class EmbraceImpl {
 
     @Nullable
     private volatile WebViewService webViewService;
+
+    @Nullable
+    private volatile EmbraceTelemetryService telemetryService;
 
     @Nullable
     private NativeThreadSamplerService nativeThreadSampler;
@@ -458,6 +462,9 @@ final class EmbraceImpl {
         exceptionsService = nonNullExceptionsService;
         serviceRegistry.registerService(exceptionsService);
         internalEmbraceLogger.addLoggerAction(sdkObservabilityModule.getInternalErrorLogger());
+
+        telemetryService = sdkObservabilityModule.getEmbraceTelemetryService();
+        serviceRegistry.registerService(telemetryService);
 
         serviceRegistry.registerService(essentialServiceModule.getNetworkConnectivityService());
 
@@ -768,7 +775,7 @@ final class EmbraceImpl {
      * @param userId the unique identifier for the user
      */
     void setUserIdentifier(@Nullable String userId) {
-        if (checkSdkStarted("set user identifier")) {
+        if (checkSdkStartedAndLogPublicApiUsage("set user identifier")) {
             if (!configService.getDataCaptureEventBehavior().isMessageTypeEnabled(MessageType.USER)) {
                 internalEmbraceLogger.logWarning("User updates are disabled, ignoring identifier update.");
                 return;
@@ -783,7 +790,7 @@ final class EmbraceImpl {
      * Clears the currently set user ID. For example, if the user logs out.
      */
     void clearUserIdentifier() {
-        if (checkSdkStarted("clear user identifier")) {
+        if (checkSdkStartedAndLogPublicApiUsage("clear user identifier")) {
             if (!configService.getDataCaptureEventBehavior().isMessageTypeEnabled(MessageType.USER)) {
                 internalEmbraceLogger.logWarning("User updates are disabled, ignoring identifier update.");
                 return;
@@ -798,7 +805,7 @@ final class EmbraceImpl {
      * @param email the email address of the current user
      */
     void setUserEmail(@Nullable String email) {
-        if (checkSdkStarted("set user email")) {
+        if (checkSdkStartedAndLogPublicApiUsage("set user email")) {
             if (!configService.getDataCaptureEventBehavior().isMessageTypeEnabled(MessageType.USER)) {
                 internalEmbraceLogger.logWarning("User updates are disabled, ignoring email update.");
                 return;
@@ -813,7 +820,7 @@ final class EmbraceImpl {
      * Clears the currently set user's email address.
      */
     void clearUserEmail() {
-        if (checkSdkStarted("clear user email")) {
+        if (checkSdkStartedAndLogPublicApiUsage("clear user email")) {
             if (!configService.getDataCaptureEventBehavior().isMessageTypeEnabled(MessageType.USER)) {
                 internalEmbraceLogger.logWarning("User updates are disabled, ignoring email update.");
                 return;
@@ -828,7 +835,7 @@ final class EmbraceImpl {
      * Sets this user as a paying user. This adds a persona to the user's identity.
      */
     void setUserAsPayer() {
-        if (checkSdkStarted("set user as payer")) {
+        if (checkSdkStartedAndLogPublicApiUsage("set user as payer")) {
             if (!configService.getDataCaptureEventBehavior().isMessageTypeEnabled(MessageType.USER)) {
                 internalEmbraceLogger.logWarning("User updates are disabled, ignoring payer user update.");
                 return;
@@ -844,7 +851,7 @@ final class EmbraceImpl {
      * paying for the service and has reverted back to a basic user.
      */
     void clearUserAsPayer() {
-        if (checkSdkStarted("clear user as payer")) {
+        if (checkSdkStartedAndLogPublicApiUsage("clear user as payer")) {
             if (!configService.getDataCaptureEventBehavior().isMessageTypeEnabled(MessageType.USER)) {
                 internalEmbraceLogger.logWarning("User updates are disabled, ignoring payer user update.");
                 return;
@@ -861,7 +868,7 @@ final class EmbraceImpl {
      * @param persona the persona to set
      */
     void addUserPersona(@NonNull String persona) {
-        if (checkSdkStarted("set user persona")) {
+        if (checkSdkStartedAndLogPublicApiUsage("set user persona")) {
             if (!configService.getDataCaptureEventBehavior().isMessageTypeEnabled(MessageType.USER)) {
                 internalEmbraceLogger.logWarning(ERROR_USER_UPDATES_DISABLED);
                 return;
@@ -878,7 +885,7 @@ final class EmbraceImpl {
      * @param persona the persona to clear
      */
     void clearUserPersona(@NonNull String persona) {
-        if (checkSdkStarted("clear user persona")) {
+        if (checkSdkStartedAndLogPublicApiUsage("clear user persona")) {
             if (!configService.getDataCaptureEventBehavior().isMessageTypeEnabled(MessageType.USER)) {
                 internalEmbraceLogger.logWarning(ERROR_USER_UPDATES_DISABLED);
                 return;
@@ -893,7 +900,7 @@ final class EmbraceImpl {
      * Clears all custom user personas from the user.
      */
     void clearAllUserPersonas() {
-        if (checkSdkStarted("clear user personas")) {
+        if (checkSdkStartedAndLogPublicApiUsage("clear user personas")) {
             if (!configService.getDataCaptureEventBehavior().isMessageTypeEnabled(MessageType.USER)) {
                 internalEmbraceLogger.logWarning(ERROR_USER_UPDATES_DISABLED);
                 return;
@@ -908,7 +915,7 @@ final class EmbraceImpl {
      * Adds a property to the current session.
      */
     boolean addSessionProperty(@NonNull String key, @NonNull String value, boolean permanent) {
-        if (checkSdkStarted("add session property")) {
+        if (checkSdkStartedAndLogPublicApiUsage("add session property")) {
             return sessionPropertiesService.addProperty(key, value, permanent);
         }
         return false;
@@ -918,7 +925,7 @@ final class EmbraceImpl {
      * Removes a property from the current session.
      */
     boolean removeSessionProperty(@NonNull String key) {
-        if (checkSdkStarted("remove session property")) {
+        if (checkSdkStartedAndLogPublicApiUsage("remove session property")) {
             return sessionPropertiesService.removeProperty(key);
         }
         return false;
@@ -929,7 +936,7 @@ final class EmbraceImpl {
      */
     @Nullable
     Map<String, String> getSessionProperties() {
-        if (checkSdkStarted("get session properties")) {
+        if (checkSdkStartedAndLogPublicApiUsage("get session properties")) {
             return sessionPropertiesService.getProperties();
         }
         return null;
@@ -941,7 +948,7 @@ final class EmbraceImpl {
      * @param username the username to set
      */
     void setUsername(@Nullable String username) {
-        if (checkSdkStarted("set username")) {
+        if (checkSdkStartedAndLogPublicApiUsage("set username")) {
             if (!configService.getDataCaptureEventBehavior().isMessageTypeEnabled(MessageType.USER)) {
                 internalEmbraceLogger.logWarning("User updates are disabled, ignoring username update.");
                 return;
@@ -956,7 +963,7 @@ final class EmbraceImpl {
      * Clears the username of the currently logged in user, for example if the user has logged out.
      */
     void clearUsername() {
-        if (checkSdkStarted("clear username")) {
+        if (checkSdkStartedAndLogPublicApiUsage("clear username")) {
             if (!configService.getDataCaptureEventBehavior().isMessageTypeEnabled(MessageType.USER)) {
                 internalEmbraceLogger.logWarning("User updates are disabled, ignoring username update.");
                 return;
@@ -980,7 +987,7 @@ final class EmbraceImpl {
     void startMoment(@NonNull String name,
                      @Nullable String identifier,
                      @Nullable Map<String, Object> properties) {
-        if (checkSdkStarted("start moment")) {
+        if (checkSdkStartedAndLogPublicApiUsage("start moment")) {
             eventService.startEvent(name, identifier, normalizeProperties(properties));
             onActivityReported();
         }
@@ -996,7 +1003,7 @@ final class EmbraceImpl {
      * @param properties custom key-value pairs to provide with the moment
      */
     void endMoment(@NonNull String name, @Nullable String identifier, @Nullable Map<String, Object> properties) {
-        if (checkSdkStarted("end moment")) {
+        if (checkSdkStartedAndLogPublicApiUsage("end moment")) {
             eventService.endEvent(name, identifier, normalizeProperties(properties));
             onActivityReported();
         }
@@ -1018,7 +1025,7 @@ final class EmbraceImpl {
      */
     @NonNull
     String getTraceIdHeader() {
-        if (checkSdkStarted("get traceIdHeader") && configService != null) {
+        if (checkSdkStartedAndLogPublicApiUsage("get traceIdHeader") && configService != null) {
             return configService.getNetworkBehavior().getTraceIdHeader();
         }
         return NetworkBehavior.CONFIG_TRACE_ID_HEADER_DEFAULT_VALUE;
@@ -1030,13 +1037,13 @@ final class EmbraceImpl {
     }
 
     void recordNetworkRequest(@NonNull EmbraceNetworkRequest request) {
-        if (checkSdkStarted("record network request") && embraceInternalInterface != null) {
+        if (checkSdkStartedAndLogPublicApiUsage("record network request") && embraceInternalInterface != null) {
             embraceInternalInterface.recordAndDeduplicateNetworkRequest(UUID.randomUUID().toString(), request);
         }
     }
 
     void recordAndDeduplicateNetworkRequest(@NonNull String callId, @NonNull EmbraceNetworkRequest request) {
-        if (checkSdkStarted("record network request")) {
+        if (checkSdkStartedAndLogPublicApiUsage("record network request")) {
             logNetworkRequestImpl(
                 callId,
                 request.getNetworkCaptureData(),
@@ -1184,7 +1191,7 @@ final class EmbraceImpl {
         @Nullable String library,
         @Nullable String exceptionName,
         @Nullable String exceptionMessage) {
-        if (checkSdkStarted("log message")) {
+        if (checkSdkStartedAndLogPublicApiUsage("log message")) {
             try {
                 remoteLogger.log(
                     message,
@@ -1213,7 +1220,7 @@ final class EmbraceImpl {
      * @param message the name of the breadcrumb to log
      */
     void addBreadcrumb(@NonNull String message) {
-        if (checkSdkStarted("log breadcrumb")) {
+        if (checkSdkStartedAndLogPublicApiUsage("log breadcrumb")) {
             breadcrumbService.logCustom(message, sdkClock.now());
             onActivityReported();
         }
@@ -1223,7 +1230,7 @@ final class EmbraceImpl {
      * Logs an internal error to the Embrace SDK - this is not intended for public use.
      */
     void logInternalError(@Nullable String message, @Nullable String details) {
-        if (checkSdkStarted("log internal error")) {
+        if (checkSdkStartedAndLogPublicApiUsage("log internal error")) {
             if (message == null) {
                 return;
             }
@@ -1242,7 +1249,7 @@ final class EmbraceImpl {
      * Logs an internal error to the Embrace SDK - this is not intended for public use.
      */
     void logInternalError(@NonNull Throwable error) {
-        if (checkSdkStarted("log internal error")) {
+        if (checkSdkStartedAndLogPublicApiUsage("log internal error")) {
             exceptionsService.handleInternalError(error);
         }
     }
@@ -1253,7 +1260,7 @@ final class EmbraceImpl {
      * Cleans all the user info on the device.
      */
     void endSession(boolean clearUserInfo) {
-        if (checkSdkStarted("end session")) {
+        if (checkSdkStartedAndLogPublicApiUsage("end session")) {
             sessionService.endSessionManually(clearUserInfo);
         }
     }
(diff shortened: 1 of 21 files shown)


embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/injection/InitModule.kt
```diff
@@ -16,6 +17,11 @@ internal interface InitModule {
      */
     val clock: Clock
 
+    /**
+     * Service to track usage of public APIs and other internal metrics
+     */
+    val embraceTelemetryService: EmbraceTelemetryService
```
fractalwrench, line 23: Consider creating a `TelemetryService` interface that defines the functions that form this service's internal API. This will make testing much easier as we can create fake/no-ops & lets folks know what is the internal API & what isn't

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/EmbraceImpl.java
```diff
@@ -1627,11 +1632,14 @@ private Map<String, Object> normalizeProperties(@Nullable Map<String, Object> pr
...
 
-    private boolean checkSdkStarted(@NonNull String action) {
+    private boolean checkSdkStartedAndLogPublicApiUsage(@NonNull String action) {
         boolean isStarted = isStarted();
         if (!isStarted) {
             internalEmbraceLogger.logSDKNotInitialized(action);
         }
+        if (telemetryService != null) {
+            telemetryService.onPublicApiCalled(action);
+        }
```
fractalwrench, line 1642: This brings up an interesting question: do we count a public API being called if the SDK hasn't been initialized? Whichever way we go on this it would be good to document it in the `onPublicApiCalled` function kdoc

embrace-android-sdk/src/test/java/io/embrace/android/embracesdk/telemetry/EmbraceTelemetryServiceTest.kt
```diff
@@ -0,0 +1,51 @@
+package io.embrace.android.embracesdk.telemetry
+
+import org.junit.Assert.assertEquals
+import org.junit.Before
+import org.junit.Test
+
+internal class EmbraceTelemetryServiceTest {
```
fractalwrench, line 7: An integration test would be helpful too, as it'd verify all the separate components are working together as expected. It might be possible to adapt an existing test case to do this.

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/telemetry/EmbraceTelemetryService.kt
```diff
@@ -0,0 +1,77 @@
+package io.embrace.android.embracesdk.telemetry
+
+import io.embrace.android.embracesdk.capture.metadata.EmbraceMetadataService
+import io.embrace.android.embracesdk.internal.spans.toEmbraceAttributeName
+import java.util.Locale
+
+/*
+    Service for tracking usage of public APIs, and different internal metrics about the app.
```
fractalwrench, line 8:
```suggestion
/**
 * Service for tracking usage of public APIs, and different internal metrics about the app.
```

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/telemetry/EmbraceTelemetryService.kt
```diff
@@ -0,0 +1,77 @@
...
+
+    /*
+        Tracks the usage of a public API by name. Adds a suffix for easier identification. Replaces whitespaces with underscores.
+    */
+    fun onPublicApiCalled(name: String) {
+        val suffixedName = "emb.usage.${
+        name.toLowerCase(Locale.ENGLISH)
+            .trim()
+            .replace(" ", "_")
+        }"
```
fractalwrench, line 25: Because this function is on a hot execution path it would be ideal to avoid all these operations (changing case, creating string subsequences, etc).

Could we either alter the code so that `onPublicApiCalled` always receives a sanitized string, or defer the sanitisation to the point where spans are flushed?

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/telemetry/EmbraceTelemetryService.kt
```diff
@@ -0,0 +1,77 @@
...
+        Returns a map with every telemetry value. For now, it's just usage counts of public APIs.
+     */
+    fun getTelemetryAttributes(): Map<String, String> {
+        val telemetryMap = mutableMapOf<String, String>()
+
+        telemetryMap.putAll(getUsageCountTelemetry())
+
+        telemetryMap.putAll(getAppAttributes())
+
+        return telemetryMap
```
fractalwrench, line 39: `return getUsageCountTelemetry().plus(getAppAttributes())`

Would avoid returning an instance of a mutable map

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/telemetry/EmbraceTelemetryService.kt
```diff
@@ -0,0 +1,77 @@
...
+    }
+
+    private fun getUsageCountTelemetry() = usageCountMap.mapValues {
+        it.value.toString()
+    }.also {
+        usageCountMap.clear()
+    }
+
+    /**
+     * Interesting attributes about the running app environment. These should be the same for every session, so we only compute them once.
```
fractalwrench, line 49: IMO it'd be better to compute the OkHttp/app attributes as a `lazy` property. That way the functions won't mutate any state so will be easier to test & we can be sure it will behave ok if this is called concurrently.

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/telemetry/EmbraceTelemetryService.kt
```diff
@@ -0,0 +1,77 @@
...
+
+        telemetryMap.putAll(getAppAttributes())
+
+        return telemetryMap
+    }
+
+    private fun getUsageCountTelemetry() = usageCountMap.mapValues {
+        it.value.toString()
+    }.also {
+        usageCountMap.clear()
```
fractalwrench, line 45: Is it necessary to reset this each time the function is called? Might we want to keep the counts for each span?

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/telemetry/EmbraceTelemetryService.kt
```diff
@@ -0,0 +1,77 @@
...
+        val telemetryMap = mutableMapOf<String, String>()
+
+        telemetryMap.putAll(getUsageCountTelemetry())
+
+        telemetryMap.putAll(getAppAttributes())
+
+        return telemetryMap
+    }
+
+    private fun getUsageCountTelemetry() = usageCountMap.mapValues {
```
fractalwrench, line 42: I think `mapValues` ultimately creates an iterator which means this is likely to throw `ConcurrentModificationException` if other threads insert values while this is iterating. This seems likely given the scope of places this can be called. Using `ConcurrentHashMap` or similar would help avoid this

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/telemetry/EmbraceTelemetryService.kt
```diff
@@ -0,0 +1,77 @@
+package io.embrace.android.embracesdk.telemetry
+
+import io.embrace.android.embracesdk.capture.metadata.EmbraceMetadataService
+import io.embrace.android.embracesdk.internal.spans.toEmbraceAttributeName
+import java.util.Locale
+
+/*
+    Service for tracking usage of public APIs, and different internal metrics about the app.
+ */
+internal class EmbraceTelemetryService(
```
fractalwrench, line 10: Worth chatting to @bidetofevil about the best approach for cleaning data as he's more familiar with the span session model, but I think it'd be worth making this implement `DataCaptureService`. That interface has functions that handle getting capturing data & also cleaning old data when the session boundary ends.

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/injection/InitModule.kt
priettt: Do you always create the interfaces by default? I usually wait until I actually need a fake or other implementation, but I can add it, no problem!

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/EmbraceImpl.java
priettt: Not yet, but if we wanted to, I'd like it to be in a different metric. It would be easy to add to the telemetry service, and it would let us differentiate between calls. Do you think we should also add this new metric?

I'll clarify the current behavior in the function kdoc!

embrace-android-sdk/src/test/java/io/embrace/android/embracesdk/telemetry/EmbraceTelemetryServiceTest.kt
priettt: I'll add one :)

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/telemetry/EmbraceTelemetryService.kt
priettt: ugh I never get these right haha

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/telemetry/EmbraceTelemetryService.kt
priettt: I thought it was going to be better to do these operations gradually rather than all at the same time 😅

Either way, I'll modify calls to `checkSdkStartedAndLogPublicApiUsage("set user identifier")` to sanitize them.

It will only affect `logSDKNotInitialized()`, so it wouldn't be too inconvenient.

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/telemetry/EmbraceTelemetryService.kt
priettt: nice catch 🙌

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/injection/InitModule.kt
fractalwrench: I usually try to create one by default - the benefit looks a bit more marginal here as the service doesn't take any params etc

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/telemetry/EmbraceTelemetryService.kt
fractalwrench: It's an interesting point. Something like `logBreadcrumb` might get spammed many times a second, but session serialization also happens every ~2s. I think it'd be preferable to avoid sanitisation completely & just pass in something that we can use as-is. e.g. `setUserIdentifier` is good enough for both uses

priettt: We don't do this on session serialization! `flushSpans` only happens when the session actually ends, so the frequency is lower. 

I think your way is better though: if we sanitize when we add it to the map, we may sanitize the same string multiple times (multiple calls to logBreadcrumb for instance). If we do it at the end, we are going to do it just once for every String! 

I still have to append the `emb.usage.` prefix, so I'll do it on getUsageTelemetry

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/injection/InitModule.kt
bidetofevil: Agreed with creating the interface for anything we are calling a Service. We almost always want to create a fake for tests given the way our modules and instantiation works, and doing so also forces us to think about what logic needs to expose externally vs things that should be kept internal, especially when the implementation evolves.

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/EmbraceImpl.java
bidetofevil: We have `SdkNotInintialized` error tracking somewhere else, but yeah this seems like it would be reasonable to track in this manner so we can remove the error we log, which is too heavyhanded for we need anyway

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/telemetry/EmbraceTelemetryService.kt
bidetofevil: The issue I have with the existing `DataCaptureService` model is the lack of atomicity for flushing and clearing. It maybe less important for general telemetry (although I think this is something we should improve on as well), but I'd like to see an iteration on that where we can enforce atomicity with like a `flush()` function.

If this works right now, I would take a look at that later, perhaps as an extension of the interface or another that builds on it, and make this adhere to that.

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/internal/spans/SpansServiceImpl.kt
```diff
@@ -160,11 +160,12 @@ internal class SpansServiceImpl(
 
     override fun flushSpans(appTerminationCause: EmbraceAttributes.AppTerminationCause?): List<EmbraceSpanData> {
         synchronized(completedSpans) {
+
+            val telemetryAttributes = embraceTelemetryService.getTelemetryAttributes()
```
bidetofevil, line 164: Right not, session spans don't survive native crashes and sudden process terminations, so this would not be recorded in those cases. And even when we address that, it'll likely go through a different code path that won't access to this.

I think this is OK for now, but I would add a comment here about this limitation and how we need to address this in the future when we address ending in-progress spans left dangling after a crash

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/telemetry/EmbraceTelemetryService.kt
bidetofevil: Another option you can consider is creating a static map of the method name to the sanitized name and have it defined there. Having a map like that is nice because we can use it in a test to verify that each method has been called in `NullParametersTests` and others like that, which we have no good way to verify right now. We can also make it a failing condition in another test that all public methods go through this helper so new methods are handled appropriately.

If you go with this suggestion, all that other stuff I mention will of course be another PR - probably a KTLO task we'll prioritize separately. Totally OK not to go with that as well as what y'all are talking about seem fine too

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/telemetry/EmbraceTelemetryService.kt
bidetofevil: +1

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/telemetry/EmbraceTelemetryService.kt
```diff
@@ -0,0 +1,77 @@
...
+        val telemetryMap = mutableMapOf<String, String>()
+
+        telemetryMap.putAll(getUsageCountTelemetry())
+
+        telemetryMap.putAll(getAppAttributes())
+
+        return telemetryMap
+    }
+
+    private fun getUsageCountTelemetry() = usageCountMap.mapValues {
```
bidetofevil, line 42: It's not clear from the method name that this also clears the map - usually getters shouldn't be destructive, so I would change the name or separate out the clearing logic.

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/telemetry/EmbraceTelemetryService.kt
bidetofevil: Yeah this whole thing can be a lazy/singleton as it is immutable. Now that are are calling this multiple times for an app launch, we should optimize this.

priettt: In this case, lazy would work exactly like checking if it's initialized before computing, but it gives us concurrency safety, right?

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/telemetry/EmbraceTelemetryService.kt
priettt: We only add telemetry to the "current session" span. As getUsageCountTelemetry will only be called when the session ends, we need to flush the counts so we don't send the same ones for the next session

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/telemetry/EmbraceTelemetryService.kt
priettt: Converting the usageCountMap to ConcurrentHashMap will be enough? 

I am now doing this for getting the attributes. Should I change anything else?

```
        usageCountMap
            .entries.associate {
                it.key.toEmbraceUsageAttributeName() to it.value.toString()
            }
```

Also, should this be a ConcurrentHashMap? Lazy already handles concurrency right?

```
private val appAttributes: Map<String, String> by lazy { getAppAttributes() }
```

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/internal/spans/SpansServiceImpl.kt
priettt: We could cache telemetry on its own? Either way, yeah, let's think about it later.

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/EmbraceImpl.java
priettt: But we wouldn't show the developers that they are trying to use something with the SDK not initialized yet. I think we should keep doing that!

If we wanted telemetry for this, I could add it on every call of `logSDKNotInitialized`, similar to what we did on checkSdkStartedAndLogPublicApiUsage

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/telemetry/EmbraceTelemetryService.kt
priettt: I already added this: 
```
      "attributes": {
        "emb.sequence_id": "1",
        "emb.private": "true",
        "emb.type": "SESSION",
        "emb.okhttp3": "true",
        "emb.okhttp3_on_classpath": "4.9.3",
        "emb.kotlin_on_classpath": "1.4.32",
        "emb.is_emulator": "__EMBRACE_TEST_IGNORE__",
        "emb.usage.log_breadcrumb": "1",
        "emb.usage.set_user_email": "1",
        "emb.usage.set_user_identifier": "1",
        "emb.usage.set_user_persona": "1",
        "emb.usage.set_username": "1"
      },
```

To the SessionMessageTest session-end.json golden file. Do you think we should add some other test? Do we need to call every public API, or would this be enough?

fractalwrench: That seems reasonable for coverage

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/telemetry/EmbraceTelemetryService.kt
fractalwrench: Yep

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/telemetry/EmbraceTelemetryService.kt
fractalwrench: `ConcurrentHashMap` takes a collection as a constructor parameter so I'd advise creating one from the result of whatever you've got now.

`lazy` handles concurrency around initialising a variable multiple times. It doesn't alter how the underlying collection works - and by default maps aren't thread-safe in Kotlin

priettt: usageCountMap is now a ConcurrentHashMap!

I'm still unsure about appAttributes: if lazy ensures that it will only be initialized by one thread, and after that the value will be cached, wouldn't it be already thread safe?

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/internal/spans/SpansServiceImpl.kt
```diff
@@ -161,7 +161,9 @@ internal class SpansServiceImpl(
     override fun flushSpans(appTerminationCause: EmbraceAttributes.AppTerminationCause?): List<EmbraceSpanData> {
         synchronized(completedSpans) {
 
-            val telemetryAttributes = embraceTelemetryService.getTelemetryAttributes()
+            /* Right now, session spans don't survive native crashes and sudden process terminations,
```
bidetofevil, line 164: We should use `//` to mark comments inside the body of a method?

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/telemetry/EmbraceTelemetryService.kt
```diff
@@ -2,76 +2,59 @@ package io.embrace.android.embracesdk.telemetry
...
-     */
-    fun getTelemetryAttributes(): Map<String, String> {
-        val telemetryMap = mutableMapOf<String, String>()
-
-        telemetryMap.putAll(getUsageCountTelemetry())
-
-        telemetryMap.putAll(getAppAttributes())
-
-        return telemetryMap
+    override fun getTelemetryAttributes(): Map<String, String> {
```
bidetofevil, line 22: Actually, it's the public method that should make it clear that the getter is destructive. We don't want consumers of this API to call this multiple times because they are unaware that i'll clear the underlying counts.

embrace-android-sdk/src/test/java/io/embrace/android/embracesdk/fakes/injection/FakeInitModule.kt
```diff
@@ -1,14 +1,17 @@
...
 import io.embrace.android.embracesdk.internal.clock.NormalizedIntervalClock
 import io.embrace.android.embracesdk.internal.clock.SystemClock
 import io.embrace.android.embracesdk.internal.spans.EmbraceSpansService
 import io.embrace.android.embracesdk.internal.spans.SpansService
+import io.embrace.android.embracesdk.telemetry.TelemetryService
 
 internal class FakeInitModule(
     override val clock: Clock = NormalizedIntervalClock(systemClock = SystemClock()),
-    override val spansService: SpansService = EmbraceSpansService(clock = FakeOpenTelemetryClock(embraceClock = clock))
+    override val telemetryService: TelemetryService = FakeTelemetryService(),
```
bidetofevil, line 15: Can we use a regular `EmbraceTelemetryService` here? We use the `FakeInitModule` in the `IntegrationTestRule` and it would be good to able to inspect the logged telemetry in those tests.

embrace-android-sdk/src/test/java/io/embrace/android/embracesdk/internal/spans/SpansServiceImplTest.kt
```diff
@@ -683,7 +684,8 @@ internal class SpansServiceImplTest {
         spansService = SpansServiceImpl(
             sdkInitStartTimeNanos = TimeUnit.MILLISECONDS.toNanos(sdkInitStartTimeMillis),
             sdkInitEndTimeNanos = TimeUnit.MILLISECONDS.toNanos(sdkInitEndTimeMillis),
-            clock = FakeOpenTelemetryClock(embraceClock = clock)
+            clock = FakeOpenTelemetryClock(embraceClock = clock),
+            EmbraceTelemetryService()
```
bidetofevil, line 688: Use named parameter if the other parameters are too?

bidetofevil (commented):
A few minor things. Are you planning to add to the integration tests to verify this?

priettt:
> A few minor things. Are you planning to add to the integration tests to verify this?

@bidetofevil I added the attributes we expect to the session-end integration test. Do you think we should add more tests?

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/internal/spans/SpansServiceImpl.kt
priettt: But it's a multi-line comment! Wouldn't it be weird to have it like this? 

```
// Right now, session spans don't survive native crashes and sudden process terminations,
// so telemetry will not be recorded in those cases, for now.
```

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/telemetry/TelemetryService.kt
```diff
@@ -9,6 +9,7 @@ internal interface TelemetryService {
 
     /**
      * Returns a map with every telemetry value. This is called when the session ends.
+     * We clear the usage count map so we don't count the same usages in the next session.
      */
     fun getTelemetryAttributes(): Map<String, String>
```
bidetofevil, line 14: Sorry - I should have been more specific. I think the NAME of the method should reflect that it will be destructive and not just a regular getting.

bidetofevil (approved):
Looks good!

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/telemetry/EmbraceTelemetryService.kt
bidetofevil: Regarding `appAttributes`, since we return an immutable map, concurrent access should be fine. And since `lazy` blocks on the invocation of the supplier method, we should be good there too?

bidetofevil: Regarding `usageCountMap`, is the concern that access to `usageCountMap` isn't synchronized, so even if the underlying data structure is thread safe (i.e. cool to access on multiple threads), you can be in the middle of `getAndClearTelemetryAttributes()`  and have someone call `onPublicApiCalled()` and cause a `ConcurrentModificationException`?

bidetofevil: I think if we can make the snapshotting and clearing of `usageCountMap` atomic (and minimize the creation of new map instances), this should be fine? Do don't want to simply synchronize the two public methods because we want to preserve the concurrent invocations of `onPublicApiCalled()`, so we'll have to be a bit more thoughtful around this.

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/telemetry/TelemetryService.kt
```diff
@@ -0,0 +1,15 @@
+package io.embrace.android.embracesdk.telemetry
+
+internal interface TelemetryService {
+
+    /**
+     * Tracks the usage of a public API by name. We only track public APIs that are called when the SDK is initialized.
```
fractalwrench, line 6: nit: it's worth documenting the convention that strings should use underscores & no spaces

fractalwrench (approved):
LGTM. This could use a rebase as I'm not sure what the % code coverage is on the change

bidetofevil (approved):
Looks good!

PR #164, opened 2023-12-08
Extract business logic from payload classes

fractalwrench:
## Goal

Attempts to extract some business logic that was present in the payload classes to separate classes/extension functions.

## Testing


bidetofevil (approved):
LGTM

PR #166, opened 2023-12-08
Remove unused uptime from device payload

fractalwrench:
## Goal

Removes the unused `uptime` field from the `DeviceInfo` payload. This is not needed & was breaking the cache used for JSON serialization.


bidetofevil (approved):
Nice

PR #167, opened 2023-12-08
Remove capture of unused ANR process errors

fractalwrench:
## Goal

Removes the capture of unused ANR process error info captured from the ActivityManager. We have not used this & are unlikely to do so, so should remove it from the SDK.


bidetofevil (approved):
LGTM

fnewberg (approved):

PR #168, opened 2023-12-08
Remove strictmode violation capture

fractalwrench:
## Goal

Removes the capture of unused strictmode violations. We have not used this & are unlikely to do so, so should remove it from the SDK.


bidetofevil (approved):
LGTM. 

That said, this seems useful on the surface. Any reason why we haven't decided to show this on the dashboard?

fnewberg (approved):

PR #169, opened 2023-12-08
Remove unused activity lifecycle breadcrumbs

fractalwrench:
## Goal

Removes the capture of unused beta features. We have not used this & are unlikely to do so, so should remove it from the SDK.


bidetofevil (approved):
LGTM.

Tracking thermal state might be something useful though. While the current implementation might not be that useful, I wonder if we can leverage this somehow, like during an ANR to see if the current slow down might be caused by thermal throttling?

fractalwrench:
Updated the changeset to just remove activity lifecycle breadcrumbs, as discussed

PR #170, opened 2023-12-09
Monitor the responsiveness of the heartbeat thread

bidetofevil:
## Goal

Track statistics for the time lapse interval between when the ANR heartbeat thread checks to see if it needs to ping the main thread, as well as the interval between when the main thread response to a heartbeat message.

## Testing

Add appropriate unit tests


bidetofevil:
* master
  * **PR #170** <a href="https://app.graphite.dev/github/pr/embrace-io/embrace-android-sdk/170?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/embrace-io/embrace-android-sdk/170?utm_source=stack-comment).

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/anr/detection/ResponsivenessMonitor.kt
```diff
@@ -0,0 +1,106 @@
...
+ * durational buckets separated by the given boundaries. The counts of the buckets are provided along with start and end times of the
+ * first N outliers, defined by any interval greater than [outlierThreshold], where N specified by [outliersLimit].
+ *
+ * The bucket name is the boundary value (inclusive) down to the next boundary value or 0. The last bucket is anything greater than the
+ * last boundary and less than [Long.MAX_VALUE].
+ *
+ * Note that this does not handle backpressure so don't use it if the rate of pings is expected to exceed the rate at which the pings can
+ * be processed sequentially.
+ */
+internal class ResponsivenessMonitor(
```
bidetofevil, line 18: Instances of this is currently only used on a single (background) thread, but I synchronized access out of habit. In the old days, I remember there being a small perf penalty on the JVM to locking even if the code runs on the same thread. Is it the same for ART?

If we're not sure, I can write a mircrobenchmark test to check it out if there's a material difference. I feel like keeping it for now in case this gets used on multiple threads later.

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/payload/PerformanceInfo.kt
```diff
@@ -74,5 +75,11 @@ internal data class PerformanceInfo(
      * StrictMode violations captured during the session
      */
     @SerializedName("v")
-    val strictmodeViolations: List<StrictModeViolation>? = null
+    val strictmodeViolations: List<StrictModeViolation>? = null,
+
+    /**
+     * Stats about the responsiveness of the ANR monitoring components
+     */
+    @SerializedName("rms")
```
bidetofevil, line 83: Not sure if this is the best place for this since it'll always update, essentially busting the cache all the time. Is there something else that updates that frequently in this payload that makes this irrelevant? If not, I should probably make it parallel to `p` and just add the plumbing to do that?

bidetofevil (commented):
I will be adding some additional serialization tests and some component tests for the components that use the `ResponsivenessMonitor`, but I don't want to block the review as I think the unit tests and manual tests are probably good enough to validate that it (mostly) works

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/anr/detection/LivenessCheckScheduler.kt
```diff
@@ -59,7 +60,8 @@
     fun startMonitoringThread() {
         enforceThread(anrMonitorThread)
         if (!state.started.getAndSet(true)) {
-            logger.logInfo("Started heartbeats to capture ANRs.")
+            logger.logInfo("Start ANR detection...")
+            resetResponsivenessMonitors()
```
bidetofevil, line 64: This line will be removed in the next PR as it's no longer necessary to do this manual clean up when the bug where we were still pinging the thread in the background is fixed.

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/payload/PerformanceInfo.kt
fractalwrench: `p` has its cache busted by `networkInterfaceIntervals` already so this seems a good place to put it

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/anr/detection/ResponsivenessMonitor.kt
fractalwrench: I'm reasonably sure there is still a small perf penalty, but I think it makes sense to keep this as-is for now.

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/anr/detection/ResponsivenessMonitor.kt
```diff
@@ -0,0 +1,106 @@
...
+    private val latestPing = AtomicLong(unsetPingTime)
+    private val errors = AtomicLong(0)
+
+    @Synchronized
+    fun ping() {
+        val previousPing = latestPing.get()
+        latestPing.set(clock.now())
+        if (previousPing != unsetPingTime) {
+            (latestPing.get() - previousPing).let { gap ->
+                if (gap >= 0) {
```
fractalwrench, line 38: might be good to break this into a separate function - I found it hard to hold all the information in my brain at once

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/anr/detection/ResponsivenessMonitor.kt
```diff
@@ -0,0 +1,106 @@
...
+    @Synchronized
+    fun reset() {
+        firstPing.set(unsetPingTime)
+        latestPing.set(unsetPingTime)
+        gaps.putAll(gaps.keys.associateWith { 0L })
+        outliers.clear()
+        errors.set(0)
+    }
+
+    data class Snapshot(
```
fractalwrench, line 75: Can this go in `io.embrace.android.embracesdk.payload`? By convention all of the payload classes live there now. Ultimately I would like to move them to a separate module

fractalwrench (approved):
LGTM

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/anr/detection/ResponsivenessMonitor.kt
bidetofevil: no free lunch, etc. this is on a the background thread so I care less, but I do want to profile this eventually

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/anr/detection/ResponsivenessMonitor.kt
bidetofevil: Yup. I can do that. I tried using actual names instead of `it` to make it more readible, but this is still pretty terse. I'm sure I'll thank myself in 3 months when I look at this again.

PR #171, opened 2023-12-10
Prevent duplicate heartbeats

bidetofevil:
## Goal

The callback in EmbraceAnrService that is called when the app foregrounds resets the state of the ANR tracking object. This is because we can't do this when the app backgrounds, in case the method to retrieve ANR data happens after the onBackground callback is called, as it is called on another thread. However, the reset is too aggressive, as it needlessly resets the ANR tracking state to false as well as resetting the stats, which is what we really need. As this is called right before we check the state to create the heartbeat recurring task, this will always be called, regardless of what the state of the service prior to the reset was.

When we background the app, we set this to false, so doing this is just a no-op. But upon app start, this set to true, which means there should already be a heartbeat job that's scheduled. This means the foregrounding will create an another one, rendering the first one "lost" as the reference to it gets overridden, so this task will always keep running.

This PR fixes this issue by not not reseting the state of the service, but just the stats. In additional, the logic to stop and reschedule the heartbeat has been consolidated and streamlined so we schedule another job to do the reschedule, thus ensuring the recurring task won't be running when it gets canceled and replaced.

## Testing

Add component test to verify the case being fixed, as well as additional changes in the supporting infra to ensuring that double job creations won't happen.

### DIFF
--- a/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/anr/detection/ThreadMonitoringState.kt
+++ b/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/anr/detection/ThreadMonitoringState.kt
@@ -46,7 +46,6 @@ internal class ThreadMonitoringState(
      */
     fun resetState() {
         anrInProgress = false
-        started.set(false)
         lastTargetThreadResponseMs = clock.now()
         lastMonitorThreadResponseMs = clock.now()
         lastSampleAttemptMs = 0
--- a/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/anr/detection/LivenessCheckScheduler.kt
+++ b/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/anr/detection/LivenessCheckScheduler.kt
@@ -72,19 +72,10 @@ internal class LivenessCheckScheduler internal constructor(
      */
     fun stopMonitoringThread() {
         enforceThread(anrMonitorThread)
-
         if (state.started.get()) {
-            monitorFuture?.let { monitorTask ->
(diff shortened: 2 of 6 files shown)


bidetofevil:
* master
  * **PR #170** <a href="https://app.graphite.dev/github/pr/embrace-io/embrace-android-sdk/170?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
    * **PR #171** <a href="https://app.graphite.dev/github/pr/embrace-io/embrace-android-sdk/171?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/embrace-io/embrace-android-sdk/171?utm_source=stack-comment).

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/anr/detection/ThreadMonitoringState.kt
```diff
@@ -46,7 +46,6 @@ internal class ThreadMonitoringState(
      */
     fun resetState() {
         anrInProgress = false
-        started.set(false)
```
bidetofevil, line 49: This is the bug fix

fractalwrench (approved):
LGTM

bidetofevil:
### Merge activity

* **Dec 12, 11:18 AM**: @@bidetofevil started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/embrace-io/embrace-android-sdk/171).

PR #172, opened 2023-12-11
Consolidate save session function overloads

fractalwrench:
## Goal

Consolidates 3 separate functions for saving a session on the `DeliveryService` into one location instead, that behaves differently according to an existing enum.

## Testing


bidetofevil (approved):
LGTM

PR #173, opened 2023-12-11
Remove redundant SessionMessageState enum

fractalwrench:
## Goal

Removes the redundant `SessionMessageState` enum. This was not used in the session payload and was conflating two separate concepts: the start/end message, and the circumstances in which the payload was written to disk. The `SessionSnapshotType` enum has superseded this.

## Testing


bidetofevil (approved):
Oh that was confusing. LGTM

PR #174, opened 2023-12-11
Consolidate send + save functions for sessions

fractalwrench:
## Goal

Combines the send/save functions for delivering sessions. This achieves a few goals:

1. Simplifies the internal API of `DeliveryService`
2. Ensures that a HTTP request is made for a session on a crash, which follows the same behavior as crashes. Data analysis has showed there is a surprisingly high success rate for deliverability, so we should enable it for sessions
3. Simplified the threading model somewhat - I/O is performed on a background thread, but other activities such as serialization happen on the calling thread. In practice this was always a background thread so shouldn't present an issue

## Testing

### DIFF
--- a/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/comms/delivery/EmbraceDeliveryCacheManager.kt
+++ b/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/comms/delivery/EmbraceDeliveryCacheManager.kt
@@ -61,18 +61,18 @@ internal class EmbraceDeliveryCacheManager(
         snapshotType: SessionSnapshotType
     ): ByteArray? {
         return when (snapshotType) {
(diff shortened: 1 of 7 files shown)


embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/comms/delivery/EmbraceDeliveryCacheManager.kt
```diff
@@ -61,18 +61,18 @@ internal class EmbraceDeliveryCacheManager(
         snapshotType: SessionSnapshotType
     ): ByteArray? {
         return when (snapshotType) {
```
bidetofevil, line 63: There probably isn't enough decisions being made that checks `SnapshotType` to formalize them as properties of the enum?

bidetofevil (approved):
LGTM

PR #175, opened 2023-12-11
Combine save/send operations for crashes in DeliveryService

fractalwrench:
## Goal

Combines the save/send operation for crashes in the `DeliveryService` as they are both only called from one place.

## Testing


bidetofevil (approved):
LGTM

PR #176, opened 2023-12-11
Rename function to sendMoment

fractalwrench:
## Goal

Renames a function on `DeliveryService` to indicate better that it sends moments, not any events.


bidetofevil (approved):
NICE. Not having call it as such was really annoying when I was working at this layer and wonder what else is being sent "async" like that

PR #179, opened 2023-12-12
Prioritise session message delivery

fractalwrench:
## Goal

Prioritises the execution of HTTP requests that are session messages. While we recently refactored the pending API call logic to prioritise session messages, we did not do so with the `ApiService`.

It's plausible that this could affect the deliverability of sessions, although the session should still get saved on disk. In a future changeset I'll look at adding similar logic to calls to the `CacheService` as this area is more likely to cause data loss if a `Runnable` does not execute.

### Changeset

This alters the SDK by:

- Submitting HTTP requests in a `NetworkRequestRunnable` wrapper class that contains a reference to the `ApiRequest` & an auto-incrementing ordinal
- Creating an executor for sending HTTP requests that uses a `PriorityBlockingQueue`
- Prioritising items on the queue that are a session HTTP request, then by request order (FIFO)
- Used a distinct executor for retrying HTTP requests

## Testing

Added test coverage.

### DIFF
--- a/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/injection/EssentialServiceModule.kt
+++ b/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/injection/EssentialServiceModule.kt
@@ -131,8 +131,8 @@ internal class EssentialServiceModuleImpl(
     private val backgroundExecutorService =
         workerThreadModule.backgroundExecutor(ExecutorName.BACKGROUND_REGISTRATION)
 
-    private val apiRetryExecutor =
-        workerThreadModule.scheduledExecutor(ExecutorName.API_RETRY)
+    private val networkRequestExecutor =
+        workerThreadModule.backgroundExecutor(ExecutorName.NETWORK_REQUEST)
 
     private val deliveryCacheExecutorService =
         workerThreadModule.backgroundExecutor(ExecutorName.DELIVERY_CACHE)
@@ -277,7 +277,7 @@ internal class EssentialServiceModuleImpl(
     override val pendingApiCallsSender: PendingApiCallsSender by singleton {
         EmbracePendingApiCallsSender(
             networkConnectivityService,
-            apiRetryExecutor,
+            workerThreadModule.scheduledExecutor(ExecutorName.BACKGROUND_REGISTRATION),
             deliveryCacheManager,
             initModule.clock
         )
@@ -289,7 +289,7 @@ internal class EssentialServiceModuleImpl(
             serializer = coreModule.jsonSerializer,
             cachedConfigProvider = { url: String, request: ApiRequest -> cache.retrieveCachedConfig(url, request) },
             logger = coreModule.logger,
-            scheduledExecutorService = apiRetryExecutor,
+            executorService = networkRequestExecutor,
             cacheManager = deliveryCacheManager,
             pendingApiCallsSender = pendingApiCallsSender,
             lazyDeviceId = lazyDeviceId,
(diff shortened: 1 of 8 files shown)


embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/worker/WorkerThreadModule.kt
```diff
@@ -35,7 +35,7 @@ internal enum class ExecutorName(internal val threadName: String) {
     CACHED_SESSIONS("cached-sessions"),
     SEND_SESSIONS("send-sessions"),
     DELIVERY_CACHE("delivery-cache"),
-    API_RETRY("api-retry"),
+    NETWORK_REQUEST("api-retry"),
```
lucaslabari, line 38: is there any reason for not renaming "api-retry" to "network_request"?

lucaslabari:
LGTM. 
With @bidetofevil we were talking about saving all requests as PendingApiCalls and sending them through there; that way, we would also prioritize sessions. 
Anyway, I think the benefit of this approach is that it runs in a separate thread, and this would make sure that incoming requests don't get blocked by the existing ones, especially at the moment the device gets connected again.

lucaslabari (approved):

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/worker/NetworkRequestRunnable.kt
```diff
@@ -0,0 +1,40 @@
+package io.embrace.android.embracesdk.worker
+
+import io.embrace.android.embracesdk.comms.api.ApiRequest
+import java.util.concurrent.PriorityBlockingQueue
+import java.util.concurrent.atomic.AtomicLong
+
+private val counter = AtomicLong(0)
```
bidetofevil, line 7: Any reason this isn't an `AtomicInteger` if you're just converting it to an Int anyway?

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/injection/EssentialServiceModule.kt
```diff
@@ -277,7 +277,7 @@ internal class EssentialServiceModuleImpl(
     override val pendingApiCallsSender: PendingApiCallsSender by singleton {
         EmbracePendingApiCallsSender(
             networkConnectivityService,
-            apiRetryExecutor,
+            workerThreadModule.scheduledExecutor(ExecutorName.BACKGROUND_REGISTRATION),
```
bidetofevil, line 280: There's a KTLO task to look at our background thread usage to see if they're being used appropriately so we can consolidate threads to minimize unnecessary concurrent execution and maximize utilization. It's great to see this done incrementally - the retry checking job shouldn't be the same thread that requests are actually run on

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/worker/WorkerThreadModule.kt
bidetofevil: Yeah I'd definitely change this as the thread name shows up in traces

bidetofevil (approved):
Couple of comments but generally looks good. 

This had me look at `EmbraceDeliveryService` and how it's pretty inconsistent in terms of whether the public `send` methods execute on the calling thread or not. It feels like this should be streamlined especially since we're going to have one thread send requests anyway, so I don't see why we need so many different threads doing stuff at the same time just to be blocked by the network request sending thread anyway.

Any hoo, that's for another time

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/worker/WorkerThreadModule.kt
fractalwrench: No reason. I've renamed it

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/worker/NetworkRequestRunnable.kt
fractalwrench: Force of habit - I forgot that `Comparator` returns an `Int`. Switched to using `AtomicInteger` instead

PR #184, opened 2023-12-13
Miscellaneous tweaks to serialization in preparation for Moshi

fractalwrench:
## Goal

Makes some miscellaneous tweaks to serialization that will make it possible to serialize JSON using Moshi. This makes the following changes:

- Avoids initializing fields & converts properties to ctor params, as Moshi can't serialize them otherwise
- Makes private fields internal where necessary
- Add default initializers for properties where necessary

## Testing


bidetofevil (approved):
LGTM

PR #185, opened 2023-12-15
Remove session message serializer

fractalwrench:
## Goal

Removes the `SessionMessageSerializer`. This was intended as a stopgap to reduce the CPU impact of JSON serialization. In practice its cache is regularly bust on the largest part of the payload meaning it doesn't give enough of a benefit given the extra complexity it introduces. Introducing Moshi is another factor that IMO allows us to get rid of this.

## Testing


bidetofevil (approved):
Nice. I'm hoping we can get Moshi in before we ship 6.3?

PR #186, opened 2023-12-15
Avoid access exceeds rights error message

fractalwrench:
## Goal

Avoids a warning message in the gradle build script that access exceeds its rights when accessing version fields from Groovy. Making the properties JVM fields is enough to stop the warning.


bidetofevil (approved):
LGTM

BTW, are we able to use the version catalog instead of this now or is there also no good way to access these in the plugin code?

fractalwrench:
There isn't a technical blocker to migrating to version catalog - probably not a big priority yet imo

PR #187, opened 2023-12-15
Move functional tests

fractalwrench:
## Goal

Moves the functional tests to `androidTest` of the main SDK module. This allows them to access internally visible code etc that previously required abusing Java's package-private visiblity.


bidetofevil (approved):
NICE. Since all the deps are declared as `testImplementation`, this shouldn't affect the binary we ship at all right?

BTW, we should probably extract all the other versions out of here so we can keep them consistent and in the same place Another patch of course.

fractalwrench:
Yep - this shouldn't affect the binary that is shipped at all.

PR #189, opened 2023-12-15
EMBR-632 Handle 429 response 

lucaslabari:
## Goal

- Added logic to the Endpoint class to manage the rate limited endpoints and the number of retries. 

## Testing

- Added unit tests.

## Release Notes

**WHAT**: Handle 429 responses from backend. 
**WHY**: When an app exceeds an allowed number of requests, our backend will send a 429 response code that notifies the client that it should stop sending requests until a retryAfter expires.
**WHO**: All customers.


embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/comms/api/EmbraceApiService.kt
```diff
@@ -170,38 +172,45 @@ internal class EmbraceApiService(
...
         payload: ByteArray,
         request: ApiRequest,
         onComplete: (() -> Any)?
     ): Future<*> {
         return executorService.submit(
             NetworkRequestRunnable(request) {
-                try {
-                    if (lastNetworkStatus != NetworkStatus.NOT_REACHABLE) {
-                        executePost(request, payload)
+                val pendingApiCall = pendingApiCallsSender.savePendingApiCall(request, payload)
```
bidetofevil, line 187: The reason you are saving the pending call always is not directly related to the 429 work, but just to take care of the case when we crash during the a network call and we haven't persisted it, so we lose it?

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/comms/api/EmbraceApiService.kt
```diff
@@ -170,38 +172,45 @@ internal class EmbraceApiService(
...
+
+                    if (response is ApiResponse.Success) {
+                        // Remove the API call from cache if it was sent successfully.
+                        pendingApiCallsSender.removePendingApiCall(pendingApiCall)
+                        onComplete?.invoke()
                     } else {
-                        pendingApiCallsSender.scheduleApiCall(request, payload)
-                        logger.logWarning("No connection available. Request was queued to retry later.")
+                        // Schedule the API call to be sent later if it should be retried.
+                        if (pendingApiCallsSender.shouldRetry(response)) {
```
bidetofevil, line 201: What if there are no retry jobs outstanding to check for requests to retry? If `pendingApiCallsSender.scheduleApiCall()` isn't called, it won't check for that case and schedule a job.

bidetofevil (commented):
Not finished. Will continue tomorrow, but here's some initial feed back.

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/comms/delivery/EmbracePendingApiCallsSender.kt
```diff
@@ -32,25 +36,62 @@ internal class EmbracePendingApiCallsSender(
...
 
-    override fun setSendMethod(sendMethod: (request: ApiRequest, payload: ByteArray) -> Unit) {
+    override fun setSendMethod(sendMethod: (request: ApiRequest, payload: ByteArray) -> ApiResponse) {
         this.sendMethod = sendMethod
     }
 
-    override fun scheduleApiCall(request: ApiRequest, payload: ByteArray) {
-        logger.logDeveloper(TAG, "Scheduling api call for retry")
+    override fun shouldRetry(response: ApiResponse): Boolean {
+        return when (response) {
```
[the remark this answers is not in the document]
lucaslabari: Sounds good, I was already thinking of refactoring that and moving it to ApiResponse. Updated.

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/comms/api/EmbraceApiService.kt
lucaslabari: Yeap, but as we talked on slack we will address that on a separate task.

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/comms/api/EmbraceApiService.kt
```diff
@@ -170,38 +172,45 @@ internal class EmbraceApiService(
...
+                            pendingApiCallsSender.scheduleApiCall(response)
+                        }
+                        onComplete?.invoke()
```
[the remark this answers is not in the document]
lucaslabari: The reason of removing the try-catch-finally is that this piece of code doesn't throw exceptions any more, try-catch is used in ApiClient and when an exception is cought we return to the ApiService a ApiResponse.Incomplete object. 
With other methods called in this function is the same, they don't throw anything, they try-catch inside.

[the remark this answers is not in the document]
bidetofevil: So `onComplete()` is invoked before regardless of what the status was because it was in the `finally` block. If you kept the same pattern (but got rid of the catch block), you wouldn't have to call that in two places.

Ah, and good to know `executePost()` already contains try-catch logic. Good to know.

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/comms/delivery/EmbracePendingApiCallsSender.kt
```diff
@@ -32,25 +36,38 @@ internal class EmbracePendingApiCallsSender(
...
-        val scheduleJob = pendingApiCalls.hasAnyPendingApiCall().not()
 
+        // Save the pending api call to disk.
+        val pendingApiCall = PendingApiCall(request, cachedPayloadName, clock.now())
         pendingApiCalls.add(pendingApiCall)
         cacheManager.savePendingApiCalls(pendingApiCalls)
+        return pendingApiCall
+    }
+
+    override fun scheduleApiCall(response: ApiResponse) {
```
bidetofevil, line 54: This should probably renamed now that it doesn't schedule an API call? Maybe `scheduleRetry()`?

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/comms/delivery/EmbracePendingApiCallsSender.kt
```diff
@@ -59,22 +76,16 @@ internal class EmbracePendingApiCallsSender(
...
-            NetworkStatus.NOT_REACHABLE -> {
-                synchronized(this) {
-                    lastDeliveryTask?.let { task ->
-                        if (task.cancel(false)) {
-                            logger.logDebug("Api Calls Delivery Action was stopped because there is no connection. ")
-                            lastDeliveryTask = null
-                        } else {
-                            logger.logError("Api Calls Delivery Action could not be stopped.")
-                        }
+        if (status.isReachable) {
```
bidetofevil, line 79: Nice change. this is much more readable and exhaustive!

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/comms/delivery/EmbracePendingApiCallsSender.kt
```diff
@@ -163,49 +130,104 @@ internal class EmbracePendingApiCallsSender(
             logger.logError("Cannot schedule delivery of pending api calls.", e)
         }
     }
+    private fun executeDelivery(delayInSeconds: Long) {
```
bidetofevil, line 133: nit: new line

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/comms/delivery/EmbracePendingApiCallsSender.kt
```diff
@@ -163,49 +130,104 @@
...
-            // We don't need to retry sending in the future as we'd get the same result.
-            // That's the reason for returning true.
+
+        if (payload == null) {
+            // If payload is null, the file could have been removed. We don't have to retry this call.
+            logger.logDeveloper(TAG, "Could not retrieve cached api payload")
+            return null
         }
-        return true
+        return sendMethod(call.apiRequest, payload)
```
bidetofevil, line 204: Do we need to check the response and delete it from the cache if we're not going to retry?

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/comms/delivery/PendingApiCalls.kt
```diff
@@ -59,17 +72,26 @@ internal class PendingApiCalls(
...
+    fun hasPendingApiCallsToSend(): Boolean {
+        return pendingApiCallsMap.entries.any {
+            it.value.hasPendingApiCallsToSend(it.key)
+        }
+    }
+
+    /**
+     * Returns true if the queue has at least one pending API call and the endpoint is not rate limited.
+     */
+    private fun MutableList<PendingApiCall>.hasPendingApiCallsToSend(endpoint: Endpoint): Boolean {
```
bidetofevil, line 86: You can probably attach this to `List` since you are not calling any of the MutableList interfaces?

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/comms/delivery/PendingApiCalls.kt
```diff
@@ -59,17 +72,26 @@ internal class PendingApiCalls(
...
+        return this.isNotEmpty() && rateLimitHandler?.isRateLimited(endpoint) != true
     }
 
     /**
-     * Returns true if there is al least one pending API call in any endpoint's queue.
+     * Returns true if the endpoint's queue has reached its limit.
      */
-    fun hasAnyPendingApiCall(): Boolean {
-        return pendingApiCallsMap.values.any { it.isNotEmpty() }
+    private fun MutableList<PendingApiCall>.hasReachedLimit(endpoint: Endpoint): Boolean {
```
bidetofevil, line 93: Same here too and attaching to `List`

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/comms/delivery/RateLimitHandler.kt
```diff
@@ -0,0 +1,26 @@
+package io.embrace.android.embracesdk.comms.delivery
+
+import io.embrace.android.embracesdk.comms.api.EmbraceApiService.Companion.Endpoint
+
+internal interface RateLimitHandler {
```
bidetofevil, line 5: I need to think about this some more, but the `RateLimitHandler` is depended on by 3 different components, and 1 of them is set after initialization. I wonder if there's a better way to shape this responsibilities of this component so that it needs to be invoked directly by fewer components?

bidetofevil (commented):
Just a few comments. Overall, this looks good, though `RateLimitHandler` is required by a number of components, and I'm wondering if we can reduce how many components have to know about this. 

Let me mull this over.

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/comms/delivery/EmbracePendingApiCallsSender.kt
```diff
@@ -84,16 +95,15 @@
...
+    fun isDeliveryTaskActive(): Boolean = lastDeliveryTask?.let { task ->
+        !task.isCancelled && !task.isDone
+    } ?: false
 
     /**
      * Return true if the conditions are met for a delivery task to be scheduled
      */
     private fun shouldScheduleDelivery(): Boolean {
-        return !isDeliveryTaskActive() && pendingApiCalls.hasAnyPendingApiCall()
+        return !isDeliveryTaskActive() && pendingApiCalls.hasPendingApiCallsToSend()
```
bidetofevil, line 106: I'm wondering if you can check to see if this endpoint to see if it's rate limited instead of asking `pendingApiCalls` to do so in `hasPendingApiCallsToSend()` so that it won't have to know about rate limiting?

bidetofevil:
Do you think we can build the rate limiting logic, both to track that an `Endpoint` is rate limited and to do a retry scheduling, in the actual `Endpoint` object?

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/comms/delivery/EmbracePendingApiCallsSender.kt
lucaslabari: Yeap, totally !!

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/comms/delivery/PendingApiCalls.kt
lucaslabari: Good catch !!!

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/comms/api/EmbraceApiService.kt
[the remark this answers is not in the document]
lucaslabari: The idea behind that was to avoid so many nested blocks, but you are right that calling the `onComplete()` in just one place is much better. 
I added the `try-finally` block but I created a separate function to simplify and make more readable that part of the code.

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/comms/delivery/EmbracePendingApiCallsSender.kt
lucaslabari: Absolutely... I lost the deletion of the payload on the refactor. Updated.

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/comms/delivery/RateLimitHandler.kt
lucaslabari: I moved the logic to the Endpoint class and now it looks much better. Maybe the scheduleRetry() method in the Endpoint class is not the best option, but in the future we can do a refactor to unify it in the PendingApiCallsSender without overloading that class too much.

bidetofevil (commented):
This is looking really good! Just a couple of comments about the changes to `Endpoint`

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/comms/api/Endpoint.kt
```diff
@@ -0,0 +1,71 @@
...
+    SESSIONS("sessions"),
+    UNKNOWN("unknown");
+
+    var isRateLimited = false
+        private set
+
+    var rateLimitRetryCount = 0
+        private set
+
+    fun setRateLimited() {
```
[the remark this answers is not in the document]
lucaslabari: Good idea... updated.

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/comms/api/Endpoint.kt
```diff
@@ -0,0 +1,71 @@
...
+
+internal enum class Endpoint(val path: String) {
+    EVENTS("events"),
+    BLOBS("blobs"),
+    LOGGING("logging"),
+    NETWORK("network"),
+    SESSIONS("sessions"),
+    UNKNOWN("unknown");
+
+    var isRateLimited = false
```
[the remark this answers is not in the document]
lucaslabari: yeap... added volatile to `isRateLimited` and converted rateLimitRetryCount to AtomicInteger.

bidetofevil (approved):
Good to go! I see some duplication in the two set so retry logic that we can probably unify, but no need to do that now.

PR #194, opened 2023-12-18
Refactor pending api call queue

fractalwrench:
## Goal

Refactors the pending API call queue implementation. Moshi does not handle serializing queues out of the box, so the easiest solution is to use a List instead.

## Testing

### DIFF
--- a/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/comms/delivery/CacheService.kt
+++ b/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/comms/delivery/CacheService.kt
@@ -87,4 +87,9 @@ internal interface CacheService {
      * @return list of file names
      */
     fun listFilenamesByPrefix(prefix: String): List<String>?
+
+    /**
+     * Loads pending API calls from their old location on disk.
+     */
+    fun loadOldPendingApiCalls(name: String): List<PendingApiCall>?
 }
--- a/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/comms/delivery/EmbraceCacheService.kt
+++ b/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/comms/delivery/EmbraceCacheService.kt
@@ -160,6 +160,20 @@ internal class EmbraceCacheService(
         }
     }
 
+    @Suppress("UNCHECKED_CAST")
+    override fun loadOldPendingApiCalls(name: String): List<PendingApiCall>? {
+        val file = File(storageDir, EMBRACE_PREFIX + name)
+        try {
+            val results = serializer.fromJson(file.inputStream(), ArrayList::class.java)
+            return results as List<PendingApiCall>? ?: return emptyList()
+        } catch (ex: FileNotFoundException) {
+            logger.logDebug("Cache file cannot be found " + file.path)
+        } catch (ex: Exception) {
+            logger.logDebug("Failed to read cache object " + file.path, ex)
+        }
+        return null
+    }
+
     companion object {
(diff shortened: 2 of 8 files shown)


embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/comms/delivery/EmbraceCacheService.kt
```diff
@@ -160,6 +160,20 @@
         }
     }
 
+    @Suppress("UNCHECKED_CAST")
+    override fun loadOldPendingApiCalls(name: String): List<PendingApiCall>? {
+        val file = File(storageDir, EMBRACE_PREFIX + name)
+        try {
+            val results = serializer.fromJson(file.inputStream(), ArrayList::class.java)
+            return results as List<PendingApiCall>? ?: return null
```
bidetofevil, line 168: Is there utility to return an empty list here so we can differentiate between the failure and nothing-found cases? If not, this is fine

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/comms/delivery/PendingApiCalls.kt
```diff
@@ -17,11 +20,11 @@ internal class PendingApiCalls {
      */
     fun add(pendingApiCall: PendingApiCall) {
         val endpoint = pendingApiCall.apiRequest.url.endpoint()
-        val pendingApiCallsForEndpoint = pendingApiCallsMap.getOrPut(endpoint) { PendingApiCallsQueue() }
+        val pendingApiCallsForEndpoint = pendingApiCallsMap.getOrPut(endpoint, ::mutableListOf)
```
bidetofevil, line 23: Is there a concern that `ArrayList` isn't thread safe? Or are access into a particular queue run serially so there's never multiple threads access it?

bidetofevil (commented):
This looks good. We're not doing anything fancy with the queue so replacing it with a list should be fine. 

I'm going to get Lucas to have a look at the access of each queue to see if we need threadsafety either in the object or controlled through synchronizing access. He can approve it after we've determined that it's all good.

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/comms/delivery/EmbraceCacheService.kt
fractalwrench: Updated to return an empty list.

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/comms/delivery/CacheService.kt
```diff
@@ -87,4 +87,9 @@ internal interface CacheService {
      * @return list of file names
      */
     fun listFilenamesByPrefix(prefix: String): List<String>?
+
+    /**
+     * Loads pending API calls from their old location on disk.
```
lucaslabari, line 92: This method loads the old version of the serialized structure for pending API calls, but the location on disk is the same.

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/comms/delivery/PendingApiCalls.kt
lucaslabari: I don't think that the fact of `ArrayList` not being thread-safe could be a problem here, as we are synchronizing manually on each access to the lists. 
The lock is held on each list so different lists can be accessed at the same time.

lucaslabari (approved):
LGTM.

PR #195, opened 2023-12-18
Switch JSON library from Gson to Moshi

fractalwrench:
## Goal

This changeset replaces Gson with [Moshi](https://github.com/square/moshi) & uses it for all JSON marshalling in our SDK. Our benchmarking using the [microbenchmark library](https://developer.android.com/topic/performance/benchmarking/microbenchmark-overview) indicated that on a Pixel 6 running Android 14, Moshi was ~2x faster at serializing session payloads & had comparable performance to Gson at deserialization. It's probable that Moshi performs even better than Gson on older devices, because Gson uses reflection whereas Moshi [uses codegen](https://github.com/square/moshi?tab=readme-ov-file#codegen) to create adapters.

We want to switch JSON library for these main reasons:

1. Gson is widely considered a legacy library in the Android community
2. Moshi offers superior performance
3. Gson was designed for Java & therefore allows undesirable behaviors (such as populating non-null Kotlin properties with null)
4. Customers are more likely to be using Moshi/[Okio](https://github.com/square/okio) in their apps than Gson, so the size impact of our SDK will probably be smaller
5. The API of Moshi is conceptually very similar to Gson
6. Stream support! This will allow us to reduce memory footprint considerably

## Changeset

- Replaced Gson annotations with Moshi annotations
- Altered payload classes where necessary. This involved making properties non-private, and moving properties to either be declared in the constructor or changing `val` to `var`. This is because Moshi populates the values via ctors/setters behind the scenes
- Added custom adapter for `EmbraceUrl`
- Altered `EmbraceSerializer` to use a thread local Moshi instance
- Added build dependency on Moshi + kapt (later versions of Moshi use KSP, but this generates Kotlin bytecode that does not support 1.4)
- Removed Gson dependency + proguard rules

## Testing

Testing relied on the following:

- There is fairly extensive unit test coverage that checks (de)serialization against golden files
- Functional tests also assert that key JSON objects are present in payloads
- Confirmed sessions/crashes/ANRs/logs are displayed in the dashboard as before
- Confirmed when no network is available cached requests are sent on next launch
- Compared the JSON files generated by both Gson + Moshi when opening the android-test-suite app & performing no action, and confirming that the payloads were the same

## Release Notes

**WHAT**:<br>Switched JSON library from Gson to Moshi
**WHY**:<br>Performance benchmarks indicate this will reduce the CPU impact of the SDK
**WHO**:<br>Everyone


embrace-android-sdk/embrace-proguard.cfg
```diff
@@ -8,17 +8,6 @@
...
--keepattributes *Annotation*
--dontwarn sun.misc.**
--keep class com.google.gson.examples.android.model.** { *; }
--keep class * implements com.google.gson.TypeAdapterFactory
--keep class * implements com.google.gson.JsonSerializer
--keep class * implements com.google.gson.JsonDeserializer
--keep class com.google.gson.reflect.TypeToken { *; }
--keep class * extends com.google.gson.reflect.TypeToken
-
 ## Proguard configuration for Arrow
```
bidetofevil, line 11: Nothing needed for Moshi huh? That's nice!

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/payload/Session.kt
```diff
@@ -132,9 +134,9 @@ internal data class Session @JvmOverloads internal constructor(
      * Enum to discriminate the different ways a session can start / end
      */
     enum class SessionLifeEventType {
-        @SerializedName("s")
-        STATE, @SerializedName("m")
-        MANUAL, @SerializedName("t")
+        @Json(name = "s")
+        STATE, @Json(name = "m")
```
bidetofevil, line 138: Nit: I know it's already there, but do you mind fixing the wrapping of the annotation so it appears in a separate line?

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/payload/TapBreadcrumb.kt
```diff
@@ -50,8 +52,8 @@ internal class TapBreadcrumb(
     override fun getStartTime(): Long = timestamp
 
     internal enum class TapBreadcrumbType {
-        @SerializedName("s")
-        TAP, @SerializedName("l")
+        @Json(name = "s")
```
bidetofevil, line 55: Same nit about annotation

embrace-android-sdk/src/test/java/io/embrace/android/embracesdk/AnrSampleTest.kt
```diff
@@ -33,9 +33,8 @@ internal class AnrSampleTest {
         assertEquals(listOf(threadInfo), obj.threads)
     }
 
-    @Test
+    @Test(expected = JsonDataException::class)
```
bidetofevil, line 36: Do we catch/handle these new runtime exceptions so we don't introduce app crashes if something bad happens?

bidetofevil (commented):
Mostly looks good. Just need to confirm that the runtime exceptions from failed deserialization will be handled as the tests previously was checking for a thrown runtime exception but will now be.

bidetofevil (approved):
Going to preemptively approve this if the runtime exceptions are handled if deserialization fails. If it's not, I assume you'll fix it before merging

embrace-android-sdk/embrace-proguard.cfg
fractalwrench: Moshi bundles its own proguard rules, and as we use codegen for serialization our adapters are safe.

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/payload/Session.kt
fractalwrench: Fixed

embrace-android-sdk/src/test/java/io/embrace/android/embracesdk/AnrSampleTest.kt
fractalwrench: I checked that all `fromJson` function invocations are surrounded by a try-catch or equivalent.

nelsitoPuglisi (approved):

priettt (approved):
Looks good! If you haven't yet, I'd give it a go on a proguard & minify enabled test app, so we are sure we are getting everything correctly

lucaslabari (approved):
COOL !!!

PR #196, opened 2023-12-19
Stream pending api calls to disk

fractalwrench:
## Goal

Streams pending API calls to disk on a background thread. This avoids serialization on the main thread that could happen often if a network request failed, and also reduces the memory pressure where the number of requests is high.

## Testing

Manually tested & relied on existing test coverage.


lucaslabari (approved):
LGTM

PR #197, opened 2023-12-19
Make requests using streams

fractalwrench:
## Goal

Instead of serializing to a `ByteArray` this alters requests so that they write directly to the HTTP request's outputstream. This reduces the memory impact of the SDK and should have negligible impact on CPU usage (the only negative case is where a network connection is poor, the request fails & it's necessary to reserialize the payload to disk).

This is implemented by taking a lambda parameter that writes to an `OutputStream`. I have created a typealias for convenience. This PR makes the change for all requests other than sessions; subsequent requests will attempt to remove the remaining uses of in-memory buffers.

## Testing

Added to existing unit test coverage.


bidetofevil (approved):
Nice. Are we able to covert the remaining cases of `cacheBytes()` to do something similar?

fractalwrench:
Yep, next on my stack. I want to break up the tasks to make review (and implementation) easier.

PR #198, opened 2023-12-19
Stream session HTTP requests

fractalwrench:
## Goal

Uses streams when sending session HTTP requests. This should reduce memory usage.

## Testing

### DIFF
--- a/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/comms/delivery/EmbraceDeliveryService.kt
+++ b/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/comms/delivery/EmbraceDeliveryService.kt
@@ -2,6 +2,7 @@ package io.embrace.android.embracesdk.comms.delivery
 
 import io.embrace.android.embracesdk.comms.api.ApiService
 import io.embrace.android.embracesdk.gating.GatingService
+import io.embrace.android.embracesdk.internal.serialization.EmbraceSerializer
 import io.embrace.android.embracesdk.logging.InternalEmbraceLogger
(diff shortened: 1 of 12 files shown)


embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/comms/delivery/EmbraceDeliveryService.kt
```diff
@@ -37,14 +39,20 @@
...
+        cacheManager.saveSession(sanitizedSessionMessage, snapshotType)
+        if (snapshotType == SessionSnapshotType.PERIODIC_CACHE) {
             return
         }
 
         try {
-            val future = apiService.sendSession(sessionBytes) {
-                cacheManager.deleteSession(sessionMessage.session.sessionId)
+            val sessionId = sessionMessage.session.sessionId
+            val action = cacheManager.loadSessionAsAction(sessionId) ?: { stream ->
```
bidetofevil, line 49: This would only apply if the app had a session that was previously cached in the old format?

fractalwrench: This codepath is taken by all sessions. Sessions are written to disk in the cache before a HTTP request is sent. If the request fails that cache file is deleted after a 'retry' file is created in a separate directory.

The system is a little confusing but I would not want to change it unless we make wider changes to the persistence layer

bidetofevil (approved):
Looks good - just a question about fallback logic and whether we need a test to verify that. If it's all good from your standpoint, merge away

PR #201, opened 2023-12-20
Add integration test for background activity

fractalwrench:
## Goal

Adds a basic integration test that asserts background activities are recorded.


bidetofevil (approved):
LGTM

PR #202, opened 2023-12-20
Stream background activity requests

fractalwrench:
## Goal

Streams http requests that send background activities.

## Testing

Relied on existing test coverage + manual testing.


bidetofevil (approved):
Looks good. Man, we really need to merge this background activity stuff in with the session.

PR #206, opened 2023-12-21
Avoid using mocks in delivery service test

fractalwrench:
## Goal

Refactors the delivery service test to use fakes rather than mocks, as the mocks were becoming an obstacle when refactoring the delivery code.

### DIFF
--- a/embrace-android-sdk/src/test/java/io/embrace/android/embracesdk/comms/delivery/EmbraceDeliveryServiceTest.kt
+++ b/embrace-android-sdk/src/test/java/io/embrace/android/embracesdk/comms/delivery/EmbraceDeliveryServiceTest.kt
@@ -2,67 +2,61 @@ package io.embrace.android.embracesdk.comms.delivery
 
 import com.google.common.util.concurrent.MoreExecutors
 import io.embrace.android.embracesdk.EmbraceEvent
-import io.embrace.android.embracesdk.comms.api.ApiService
+import io.embrace.android.embracesdk.FakeNdkService
 import io.embrace.android.embracesdk.fakeBackgroundActivity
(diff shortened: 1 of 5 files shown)


embrace-android-sdk/src/test/java/io/embrace/android/embracesdk/comms/delivery/EmbraceDeliveryServiceTest.kt
```diff
@@ -2,67 +2,61 @@ package io.embrace.android.embracesdk.comms.delivery
...
-            mockDeliveryCacheManager = mockk(relaxed = true)
-            every { mockDeliveryCacheManager.loadCrash() } returns null
-            every { mockDeliveryCacheManager.getAllCachedSessionIds() } returns emptyList()
-            apiService = mockk(relaxed = true)
-            gatingService = FakeGatingService()
-            logger = InternalEmbraceLogger()
-        }
+    @Before
+    fun setUp() {
+        executor = MoreExecutors.newDirectExecutorService()
```
lucaslabari, line 43: Isn't it better to use BlockingScheduledExecutorService?

lucaslabari (approved):

embrace-android-sdk/src/test/java/io/embrace/android/embracesdk/comms/delivery/EmbraceDeliveryServiceTest.kt
fractalwrench: I'll leave this as-is since I was just moving the point at which it's initialized, but in future perhaps we can do a roundup of anything using `MoreExecutors`

PR #210, opened 2023-12-21
Consolidate mock creation for system apis

fractalwrench:
## Goal

Consolidates the creation of mocks for Android framework APIs. This should restrict the point at which mocks are created to one place & allows us to perform useful setup in one place in the test code, rather than multiple places.


lucaslabari (approved):
LGTM

PR #212, opened 2023-12-22
Create interface for internal error service

fractalwrench:
## Goal

Created an interface to formalise `EmbraceInternalErrorService`, and to make it easier to create fakes for testing. I also renamed variable names scattered through the codebase for consistency.

## Testing


bidetofevil (approved):
Nice. We need to improve the implementation and perhaps even the interface, but this is a good start.

PR #215, opened 2023-12-22
Reduce SDK usage of threads

fractalwrench:
## Goal

Removes some underutilised threads that were generally only used to submit one Runnable in each class they were used. The `BACKGROUND_REGISTRATION` is used instead, so we can get rid of unnecessary threads.

Thread count can be reduced further but this will require a bit of extra refactoring of the caching + delivery layers.

## Testing

Relied on existing test coverage & manually checked sessions + NDK crashes were sent from the NDK test app.

### DIFF
--- a/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/worker/WorkerThreadModule.kt
+++ b/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/worker/WorkerThreadModule.kt
@@ -31,15 +31,10 @@ internal interface WorkerThreadModule : Closeable {
  */
 internal enum class ExecutorName(internal val threadName: String) {
     BACKGROUND_REGISTRATION("background-reg"),
-    SCHEDULED_REGISTRATION("scheduled-reg"),
     CACHED_SESSIONS("cached-sessions"),
     SEND_SESSIONS("send-sessions"),
     DELIVERY_CACHE("delivery-cache"),
     NETWORK_REQUEST("network-request"),
-    NATIVE_CRASH_CLEANER("native-crash-cleaner"),
-    NATIVE_STARTUP("native-startup"),
-    SESSION_CACHE_EXECUTOR("session-cache"),
+    PERIODIC_CACHE("periodic-cache"),
     REMOTE_LOGGING("remote-logging"),
-    SESSION_CLOSER("session-closer"),
-    SESSION_CACHING("session-caching"),
 }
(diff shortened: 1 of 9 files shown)


embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/worker/WorkerThreadModule.kt
```diff
@@ -31,15 +31,10 @@ internal interface WorkerThreadModule : Closeable {
  */
 internal enum class ExecutorName(internal val threadName: String) {
     BACKGROUND_REGISTRATION("background-reg"),
-    SCHEDULED_REGISTRATION("scheduled-reg"),
     CACHED_SESSIONS("cached-sessions"),
     SEND_SESSIONS("send-sessions"),
```
bidetofevil, line 35: Is this the thread that preps the session payload before sending it out to the `DeliveryService` to send?

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/worker/WorkerThreadModule.kt
```diff
@@ -31,15 +31,10 @@ internal interface WorkerThreadModule : Closeable {
...
     DELIVERY_CACHE("delivery-cache"),
     NETWORK_REQUEST("network-request"),
-    NATIVE_CRASH_CLEANER("native-crash-cleaner"),
-    NATIVE_STARTUP("native-startup"),
-    SESSION_CACHE_EXECUTOR("session-cache"),
+    PERIODIC_CACHE("periodic-cache"),
     REMOTE_LOGGING("remote-logging"),
```
bidetofevil, line 39: This is the analog for `SEND_SESSION` but for logging?

bidetofevil (approved):
LGTM. 

How did you verify that combining multiple jobs and running them on the same executor won't create some weird deadlock where the currently executing job queues up another job that it blocks on in order to complete? Or were the use case pretty straight forward such that none of the jobs on these removed executors actually block on anything?

fractalwrench:
@bidetofevil IMO the use case was fairly straightforward. I eyeballed the result then manually verified nothing was obviously broken. My changes were:

- SCHEDULED_REGISTRATION - only used to record a span, so switched to BACKGROUND_REGISTRATION
- SESSION_CLOSER - only used to close a session (when timed sessions are disabled this shouldn't have a large effect)
- Power save/thermal status workers - BACKGROUND_REGISTRATION is not generally used after SDK startup so sharing the executor for scheduled execution shouldn't make much of a difference

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/worker/WorkerThreadModule.kt
fractalwrench: Yep - it processes log messages

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/worker/WorkerThreadModule.kt
fractalwrench: This is probably next on my target list - it's actually used to load background activities + moments & initiate HTTP requests via the API service. I'm not convinced it is necessary

PR #223, opened 2024-01-02
Add docs for what each executor does

fractalwrench:
## Goal

Adds some docs that explain what each executor is currently responsible for.


nelsitoPuglisi:
Thank you!

PR #225, opened 2024-01-02
Remove unnecessary executor for background activity + moments

fractalwrench:
## Goal

Removes an unnecessary executor that was being used in the delivery service for background activities + moments.

Previously this was required because the delivery service loaded the payload before calling the ApiService. This now happens in a lambda function that is executed within the ApiService on its executor, therefore the `SEND_SESSIONS` executor isn't necessary.

## Testing

Updated existing test coverage, and manually verified with debugger that the calls are still made on the updated executors.


bidetofevil (approved):
LGTM

PR #226, opened 2024-01-03
Avoid rejected execution exceptions

fractalwrench:
## Goal

Adds a RejectedExecutionHandler that logs a warning when instantiating any executors. This avoids throwing `RejectedExecutionException` when the executor is shutdown (i.e. the process is terminating) or the task is rejected for whatever reason.

Previously catching the exception was left as a choice for the individual caller - the codebase is inconsistent in whether the exception is properly handled or not. Adopting a consistent policy will avoid any chance of crashes in customer apps.

This causes a difference in behavior in that a `Future` will always be returned now when submitting a task. Waiting for the `Future` using `get()` will still block but will obviously never return a value if the executor is shutdown.

## Testing

Updated existing test coverage. I also visually verified that every point in the codebase that holds a reference to a Future handles it appropriately with a timeout on the `get()`. This resulted in adding arbitrary timeouts to a couple of places in the codebase that did not have any.


bidetofevil (approved):
Very nice!

PR #228, opened 2024-01-03
Avoid using callable when submitting to executor

fractalwrench:
## Goal

Some old code was using `Callable` rather than `Runnable` when submitting to executors. This wasn't necessary & made the indent longer than it had to be, and complicated in-progress refactoring of our threading model. Therefore I've removed the usages where unnecessary.

## Testing


bidetofevil (approved):
Nice

PR #229, opened 2024-01-03
[EMBR-1624] Store sessions in the files directory.

lucaslabari:
## Goal

- All the Embrace files were stored in the `cache` directory. 
- To avoid loosing data when cache is cleared, we now use the `files` directory. 
- Moved from `cache` to `files/embrace`:
   - last Embrace session
   - failed_api_calls
   - payloads (sessions, logs, etc)
   - crash marker file
   - ndk folder with crashes 
- Cached config is the only think we kept in the `cache` folder. 
- Added a way to serialize and deserialize a list of Objects. 
- Fixed a bug in the method that loads the old pending API calls. 

## Testing

Added unit tests. 

## Release Notes

**WHAT**: Moved files from `cache` folder to `files/embrace` folder.


embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/injection/StorageModule.kt
```diff
@@ -0,0 +1,54 @@
...
+import io.embrace.android.embracesdk.comms.delivery.DeliveryCacheManager
+import io.embrace.android.embracesdk.comms.delivery.EmbraceCacheService
+import io.embrace.android.embracesdk.comms.delivery.EmbraceDeliveryCacheManager
+import io.embrace.android.embracesdk.storage.EmbraceStorageManager
+import io.embrace.android.embracesdk.storage.StorageManager
+import io.embrace.android.embracesdk.worker.ExecutorName
+import io.embrace.android.embracesdk.worker.WorkerThreadModule
+import java.io.File
+
+internal interface StorageModule {
```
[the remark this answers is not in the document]
lucaslabari: Added.

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/injection/StorageModule.kt
```diff
@@ -0,0 +1,54 @@
...
+    val storageManager: StorageManager
```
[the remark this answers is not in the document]
lucaslabari: Done. I renamed it to StorageService.

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/storage/StorageManager.kt
```diff
@@ -0,0 +1,14 @@
+package io.embrace.android.embracesdk.storage
+
+import java.io.File
+import java.io.FilenameFilter
+
+internal interface StorageManager {
```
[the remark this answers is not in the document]
lucaslabari: Added.

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/storage/EmbraceStorageManager.kt
```diff
@@ -0,0 +1,57 @@
...
+
+internal class EmbraceStorageManager(
+    private val coreModule: CoreModule
+) : StorageManager {
+
+    override val cacheDirectory: Lazy<File> = lazy {
+        coreModule.context.cacheDir
+    }
+
+    override val filesDirectory: Lazy<File> = lazy {
```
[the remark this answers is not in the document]
lucaslabari: The problem with doing a custom getter that tries to create the directory each time is that we could end with some files in the cache dir and some on the files/embrace dir. I think it will probably work even in a situation like that, but I think it would be more consistent if we use the same dir. Now, if the `files/embrace` directory can't be created it will use the cache folder as fallback for all files.

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/storage/EmbraceStorageService.kt
```diff
@@ -0,0 +1,55 @@
...
+
+import io.embrace.android.embracesdk.injection.CoreModule
+import java.io.File
+import java.io.FilenameFilter
+
+internal class EmbraceStorageService(
+    private val coreModule: CoreModule
+) : StorageService {
+
+    override val cacheDirectory: Lazy<File> = lazy {
```
fractalwrench, line 11:
```suggestion
    override val cacheDirectory: File by lazy {
```

Using [property delegation](https://kotlinlang.org/docs/delegated-properties.html) for both of these properties would avoid having to use `cacheDirectory.value` everywhere

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/internal/serialization/EmbraceSerializer.kt
```diff
@@ -44,6 +44,13 @@ internal class EmbraceSerializer {
         }
     }
 
+    fun <T> toJsonFromList(src: List<T>, clazz: Class<T>): String {
+        val adapter = impl.adapter<List<T>>(
+            Types.newParameterizedType(List::class.java, clazz)
```
fractalwrench, line 49: I'm fairly sure that you can use `fun <T> toJson(src: T, type: Type)` to achieve this without adding a new function. That should also be the case for the new `fromJson` overload that was added. Ping me if you need assistance as I spent a bit of time tackling this

fractalwrench (approved):
LGTM pending resolution of the inline comments

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/comms/delivery/EmbraceCacheService.kt
```diff
@@ -149,6 +148,24 @@ internal class EmbraceCacheService(
         return null
     }
 
+    /**
+     * Gets a file from the files directory.
+     * Generally used to write files, so we store them in the files directory.
+     * The files directory is the default directory.
+     */
+    private fun getFileFromFilesDir(name: String): File {
```
bidetofevil, line 156: Do you think we can hide the logic of where the files are stored within the `StorageService`? If that part is abstracted away fully, `EmbraceCacheService` should haven't to know where the files are stored. Is there a reason it needs to? In other words, can we defer all the "where the file is stores" logic to the storage service and have this completely not know about it?

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/injection/StorageModule.kt
```diff
@@ -0,0 +1,57 @@
...
+    val deliveryCacheManager: DeliveryCacheManager
+}
+
+internal class StorageModuleImpl(
+    workerThreadModule: WorkerThreadModule,
+    initModule: InitModule,
+    coreModule: CoreModule,
+) : StorageModule {
+
+    private val deliveryCacheWorker =
```
bidetofevil, line 30: This could probably be in lined since there's only one user of it?

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/internal/serialization/EmbraceSerializer.kt
lucaslabari: Updated. I had to add a new fromJson that receives the InputStream and the Type, as it was there for toJson but not for fromJson.

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/injection/StorageModule.kt
lucaslabari: Done

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/storage/EmbraceStorageService.kt
lucaslabari: Done.

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/storage/StorageService.kt
```diff
@@ -0,0 +1,27 @@
...
+import java.io.FilenameFilter
+
+/**
+ * Provides File instances for files in the cache and files directories.
+ * Previous versions of the SDK used to store files in the cache directory, but now we use the
+ * files directory for everything except the config cache.
+ */
+internal interface StorageService {
+
+    val cacheDirectory: File
```
bidetofevil, line 13: Is it possible to remove this from interface and just have it as part of the implementation?

lucaslabari: Updated.

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/storage/EmbraceStorageService.kt
```diff
@@ -0,0 +1,55 @@
...
+
+import io.embrace.android.embracesdk.injection.CoreModule
+import java.io.File
+import java.io.FilenameFilter
+
+internal class EmbraceStorageService(
+    private val coreModule: CoreModule
+) : StorageService {
+
+    val cacheDirectory: File by lazy {
```
bidetofevil, line 11: private?

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/comms/delivery/EmbraceCacheService.kt
lucaslabari: As we discussed, I ended up dividing the implementation of getFile into two methods:
- getFileForRead
- getFileForWrite
This way we hide completely the implementation details related to the storage directories.

PR #230, opened 2024-01-04
Encapsulate executors in BackgroundWorker/ScheduledWorker classes

fractalwrench:
## Goal

Encapsulates uses of `ExecutorService` in a `BackgroundWorker/ScheduledWorker` class. There are a couple of reasons I've chosen this approach:

1. It hides implementation details of executors that don't need to be exposed as part of our internal API (such as whether it's shutdown)
2. It hides functions such as `scheduleAtFixedRate` that we definitely do not want to use within the SDK
3. This is a pre requisite for refactoring how task prioritisation works. It's much easier to use a custom queue that ranks tasks by priority if the logic is encapsulated in one place

## Testing

Relied on existing test coverage as these classes just call into the underlying executor.


bidetofevil (approved):
We previously had similar implementations that abstract away the underlying executors but effectively exposed the same API but with a lot of duplicate code, so we just removed it as we were needlessly maintaining abstractions without getting any benefits.

I like your current approach better because it has a different purpose (restricting the interface) and removes a lot of the duplication. That said, I wonder if we can simplify `BackgroundWorker` even further so `eagerLazyLoad()` isn't exposed, and it'll be similar to `ScheduledWorker` and just be deferring to the underlying executor? That way, we'll have our custom abstractions that limits the interface exposed, but super thin ones that just defers.

If you think that's worth doing, feel free to merge this and do that after. Otherwise, this is fine as it is.

fractalwrench:
I've updated the implementation so that `eagerLazyLoad` is defined as an extension for now.

PR #232, opened 2024-01-04
Implement priority queueing for background tasks

fractalwrench:
## Goal

This changeset introduces priority to tasks that are submitted to a `BackgroundWorker`. This is desirable as there are often scenarios where certain tasks should be prioritised before others. As an example, customers care more about sessions/crashes than other request types, so it makes sense to prioritise these HTTP requests before attempting logs/moments.

I've added a `TaskPriority` enum which has an arbitrary 4 values - we could add more if this is seen as desirable. I aimed to strike a balance here - an alternate approach could define specific orders for every task that is posted to a given executor. However, I think it's best to keep things simple & avoid introducing business logic - most of the time we only need to prioritise a few essential tasks above everything else.

When submitting a `Runnable` or `Callable` a `TaskPriority` must also now be specified. This is held in a wrapper class & propagated to a custom `RunnableFuture` that holds the priority information, along with a timestamp of the submission time. The `Executor` ultimately uses a custom queue with a comparator that sorts by a reduction of priority + submission time.

### Discussion

I have left the task prioritisation as-is in this changeset. The exception is for session HTTP requests which already have a higher priority compared to other requests. Future changesets will alter task prioritisation.

I've mitigated resource starvation by using an aging approach for scheduling. A score is calculated by summing the submission time and a delay that is smaller for more important `TaskPriority` values. Tasks with the lowest score are executed first. This helps ensure important tasks are given priority but don't saturate the queue & prevent lower priority tasks from ever executing.

## Testing

Added new test coverage where applicable.


bidetofevil (commented):
Looks good!. A couple of thoughts:

- This doesn't handle starvation. For our existing use case, it should fine, but we should explicitly state that or do something where "time in queue" is a component to calculate priority so that normal priority tasks are dealt with if they're waiting long enough. I probably wouldn't do that until we need to, so leaving a comment ought to be good enough right now.

- If we are going to limit the threadpool to having only one thread, we should probably put that fact in the name, e.g. `SerialPriorityExecutor`

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/comms/delivery/EmbraceDeliveryCacheManager.kt
```diff
@@ -207,7 +208,7 @@ internal class EmbraceDeliveryCacheManager(
             }
             loadApiCallsResult.getOrNull() ?: loadPendingApiCallsOldVersion()
         }
-        val cached = backgroundWorker.submit(callable).get()
+        val cached = backgroundWorker.submit(TaskPriority.NORMAL, callable).get()
```
bidetofevil, line 211: Should we just rely on the default parameter when the task doesn't care about priority explicity?

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/worker/PriorityThreadPoolExecutor.kt
```diff
@@ -0,0 +1,71 @@
...
+ * 4. Tasks are executed in order of lowest score to highest score
+ *
+ * This helps mitigate against resource starvation scenarios where a large number of high priority
+ * tasks prevent lower priority tasks from ever completing execution.
+ */
+internal class PriorityThreadPoolExecutor(
+    private val clock: Clock,
+    threadFactory: ThreadFactory,
+    handler: RejectedExecutionHandler,
+    corePoolSize: Int,
```
bidetofevil, line 28: Ah, nice. Good to just make the threadpool configurable instead of narrowing down the threadpool attributes

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/worker/PriorityThreadPoolExecutor.kt
```diff
@@ -0,0 +1,71 @@
...
+        return PriorityRunnableFuture(runnableFuture, runnable.priority, clock.now())
+    }
+
+    companion object {
+        internal fun createPriorityQueue(): PriorityBlockingQueue<Runnable> =
+            PriorityBlockingQueue(
+                100,
+                compareBy<Runnable> { runnable ->
+                    require(runnable is PriorityRunnableFuture<*>) {
+                        "Runnable must be PriorityRunnableFuture"
```
bidetofevil, line 65: Since the executor controls the creation of the runnable, this is literally impossible right now right?

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/worker/PriorityThreadPoolExecutor.kt
```diff
@@ -0,0 +1,71 @@
...
+                    }
+                    runnable.submitTime + runnable.priority.delayThresholdMs
```
bidetofevil, line 67: oh nice solution

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/worker/PriorityRunnable.kt
```diff
@@ -0,0 +1,55 @@
...
+     *
+     * The higher the priority, the lower the delay threshold.
+     *
+     * See [PriorityThreadPoolExecutor] for further detail.
+     */
+    val delayThresholdMs: Long
+) {
+    CRITICAL(0),
+    HIGH(5_000),
+    NORMAL(15_000),
```
bidetofevil, line 53: Maybe something to think about - 15s isn't a tonne of delay when we are talking about network requests running serially, especially in network conditions that will cause a backlog of queued requests. Bumping this to a higher number might make this work better in practice.

bidetofevil (approved):
Looks good. A couple of comments about upping the delay for NORMAL priority and maybe simplifying the calling case if you don't need a special priority. Feel free to merge with or without those changes as you see fit!

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/worker/PriorityRunnable.kt
fractalwrench: What sort of delay do you think would be appropriate? Perhaps `30` for NORMAL and `60` for LOW?

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/worker/PriorityThreadPoolExecutor.kt
fractalwrench: Yep - this should not be possible but the typecast is still required

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/comms/delivery/EmbraceDeliveryCacheManager.kt
fractalwrench: Yep this feels a reasonable thing to change, so that `TaskPriority` only needs to be specified when overriding the default.

PR #233, opened 2024-01-05
Use same class for session/background activity payload

fractalwrench:
## Goal

The session and background activity payloads use the same schema but for legacy reasons is implemented as two separate classes. This could lead to unnecessary divergence in behavior & also makes it harder to share code between the session/background activity services. Therefore this changeset combines them to use one type.

## Testing


bidetofevil (approved):
Yesssssss!

PR #235, opened 2024-01-08
Refactor ANR detection to use ScheduledWorker

fractalwrench:
## Goal

Refactored the ANR detection to use `ScheduledWorker`, so that thread creation in the core SDK module remains in one place.

## Testing

### DIFF
--- a/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/injection/AnrModule.kt
+++ b/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/injection/AnrModule.kt
@@ -20,10 +20,8 @@ import io.embrace.android.embracesdk.capture.monitor.NoOpResponsivenessMonitorSe
 import io.embrace.android.embracesdk.capture.monitor.ResponsivenessMonitorService
 import io.embrace.android.embracesdk.internal.ApkToolsConfig
 import io.embrace.android.embracesdk.internal.SharedObjectLoader
-import java.util.concurrent.Executors
-import java.util.concurrent.ScheduledExecutorService
-import java.util.concurrent.ThreadFactory
-import java.util.concurrent.atomic.AtomicReference
+import io.embrace.android.embracesdk.worker.WorkerName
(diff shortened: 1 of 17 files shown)


embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/worker/ScheduledWorker.kt
```diff
@@ -43,4 +44,19 @@ internal class ScheduledWorker(
...
+
+    /**
+     * Submits a task for execution and returns a [Future].
+     */
+    fun <T> submit(callable: Callable<T>): Future<T> = impl.submit(callable)
+
+    @Deprecated(
+        "Use scheduleWithFixedDelay instead.",
+    )
+    fun scheduleAtFixedRate(
```
bidetofevil, line 56: Any reason we don't swap this out now?

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/injection/AnrModule.kt
```diff
@@ -125,20 +125,4 @@ internal class AnrModuleImpl(
             logger = coreModule.logger
         )
     }
-
-    private val anrMonitorThreadFactory = ThreadFactory { runnable: Runnable ->
```
bidetofevil, line 129: Ah very nice

bidetofevil (approved):
Looks good. We might want to replace the existing use of `scheduleAtFixedRate` soon so we can remove the "deprecated" method. if we do that, we can probably simplify the heartbeat checking logic too.

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/worker/ScheduledWorker.kt
fractalwrench: I don't want to touch the ANR detection logic at this time as it's fairly complex & the main purpose of this changeset was to consolidate thread creation in one place.

PR #236, opened 2024-01-08
Remove unnecessary thread used for loading sessions

fractalwrench:
## Goal

Removes the `CACHED_SESSIONS` worker that was only used for transforming the cached session from the previous process into a request that can be sent to the Embrace API. I have altered this to use the `DELIVERY_CACHE` worker instead as it does perform some I/O, but it will be a one-off during SDK initialization that should be fairly short-lived. I also removed an unnecessary use of a worker in the `DeliveryCacheManager` as it's always called from the thread that a job was being submitted to.

## Testing

Relied on existing test coverage & confirmed sessions still show in dashboard


bidetofevil (approved):
LGTM

PR #237, opened 2024-01-08
Enhance background activity service test coverage

fractalwrench:
## Goal

Increases test coverage for the BackgroundActivityService class & moves some constants to a consistent location.

### DIFF
--- a/embrace-android-sdk/src/test/resources/breadcrumb_empty.json
+++ b/embrace-android-sdk/src/test/resources/breadcrumb_empty.json
@@ -3,7 +3,7 @@
     "id": "fakeSessionId",
     "st": 160000000000,
     "sn": 1,
-    "ty": "st",
+    "ty": "en",
     "as": "foreground",
     "cs": true,
     "sm": "s",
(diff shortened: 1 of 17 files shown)


bidetofevil (commented):
Looks good. Will review when the tests pass

embrace-android-sdk/src/test/resources/breadcrumb_empty.json
```diff
@@ -3,7 +3,7 @@
     "id": "fakeSessionId",
     "st": 160000000000,
     "sn": 1,
-    "ty": "st",
```
nelsitoPuglisi, line 6: Wasn't the type validated before?

nelsitoPuglisi (approved):

embrace-android-sdk/src/test/resources/breadcrumb_empty.json
fractalwrench: I have altered the default type to be `en` rather than `st` because we no longer send a session start message. `st` will never actually get sent in a payload to the backend as we only send messages with `en` now

PR #238, opened 2024-01-09
Remove useless session message creation

fractalwrench:
## Goal

Removes a useless creation of a session message object that wasn't actually used within the `SessionHandler`.

## Testing


lucaslabari (approved):
LGTM

PR #239, opened 2024-01-09
Combine payload creation into one class

fractalwrench:
## Goal

Combines the creation of session/background activity payloads into one class. This changeset also unifies the creation of the initial session message.

## Testing


embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/session/PayloadMessageCollator.kt
```diff
@@ -172,4 +190,97 @@
             spans = spans
         )
     }
+
+    /**
+     * Creates a background activity stop message.
+     */
+    fun createStopMessage(
```
bidetofevil, line 197: Is there still a "message" that's not "stop" now?

bidetofevil (approved):
LGTM

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/session/PayloadMessageCollator.kt
fractalwrench: Renamed function

PR #242, opened 2024-01-11
Ignored some methods that are called too often in logPublicApiUsage

priettt:
- Renamed some logging to reflect the correct method names
- Flipped some conditionals to avoid logging public API usages that don't actually happen.


bidetofevil (approved):
LGTM. Would've been nice to have a default parameter so you don't have to pass in `false` when using `checkSdkStarted()` directly, but Java....

priettt:
> LGTM. Would've been nice to have a default parameter so you don't have to pass in `false` when using `checkSdkStarted()` directly, but Java....

That was my first thought lol

PR #247, opened 2024-01-15
Alter priority for session-related tasks

fractalwrench:
## Goal

This changeset alters the priority of several session-related tasks in the SDK. Specifically it prioritises anything that writes a session or background activity to disk. I achieved this by reviewing the delivery/cache classes & manually checking what jobs were submitted to individual executors.

I decided to prioritise the following:

- Writing sessions to disk (Critical as if this doesn't happen, it will lead to data loss)
- Loading/sending cached sessions (High, as this is important, but is less likely to result in data loss if it doesn't get scheduled)
- Prioritising session requests over other HTTP requests (already present)

Session snapshots are deprioritised as they are less important than these other tasks.

Additionally I switched out a `BackgroundWorker` for a `ScheduledWorker` in the `BackgroundActivityService` to avoid creating an unnecessary thread.

## Testing


bidetofevil (approved):
LGTM

PR #248, opened 2024-01-15
Added crashNumber to the Crash payload

priettt:
## Goal

The idea is to number every crash so we know if there are crashes that do not arrive at the backend, similar to `session.session_number`

## Testing

Verified that crash_number is added to the crash payload on the android-test-suite, using proxyman

### DIFF
--- a/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/capture/crash/EmbraceCrashService.kt
+++ b/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/capture/crash/EmbraceCrashService.kt
@@ -19,6 +19,7 @@ import io.embrace.android.embracesdk.payload.Event
 import io.embrace.android.embracesdk.payload.EventMessage
 import io.embrace.android.embracesdk.payload.JsException
 import io.embrace.android.embracesdk.payload.extensions.CrashFactory
+import io.embrace.android.embracesdk.prefs.PreferencesService
 import io.embrace.android.embracesdk.session.BackgroundActivityService
 import io.embrace.android.embracesdk.session.SessionService
 import io.embrace.android.embracesdk.session.properties.SessionPropertiesService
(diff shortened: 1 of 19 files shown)


embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/capture/crash/EmbraceCrashService.kt
```diff
@@ -76,14 +78,15 @@ internal class EmbraceCrashService(
             // is enabled for an Unity build. When a native crash occurs and the NDK sends an
             // uncaught exception the SDK assign the unity crash id as the java crash id.
             val unityCrashId = ndkService.getUnityCrashId()
+            val crashNumber = preferencesService.incrementAndGetCrashNumber()
```
bidetofevil, line 81: Should we do a best-effort only here in case the preferences service is unavailable or throws an exception? The fallback for the `Future.get()` seems to be doing it on the calling thread which could fail too.

priettt: Maybe we should add a try catch here, as it's used by every "incrementAndGet" function: 

```
private fun incrementAndGetOrdinal(key: String): Int {
        val ordinal = (prefs.getIntegerPreference(key) ?: 0) + 1
        prefs.setIntegerPreference(key, ordinal)
        return ordinal
}
```

Right now, the session number doesn't have any error handling, so we could solve some issues if we add it here

bidetofevil: Try catch and return `-1` or some other value that well tell us an error has occurred while incrementing seems like a good idea

bidetofevil (approved):
LGTM. I'd wait unti you and @fractalwrench figure out the one-ordinal-or-two issues - I'm happy either way.

PR #249, opened 2024-01-16
Remove timed sessions

fractalwrench:
## Goal

This changeset removes the concept of a 'timed' session - a session that automatically times out after X seconds and creates a new one. This legacy functionality is not used by customers & complicates our implementation, so we're removing it.

## Testing

Removed obsolete test coverage.


bidetofevil (approved):
LGTM

PR #250, opened 2024-01-16
Simplify how background activities are saved

fractalwrench:
## Goal

This changeset simplifies how the background activities are cached, while not making any changes to the underlying persistence mechanism. Rather than posting a runnable to the main thread to schedule a worker, now the SDK will just schedule a task with an appropriate delay.

## Testing

Added unit test coverage.


bidetofevil (approved):
Nice. LGTM

PR #251, opened 2024-01-16
Remove deprecated async end flag

fractalwrench:
## Goal

Removes a deprecated flag `asyncEnd` that is no longer respected, as it was increasing complexity in session handling code. Users who have specified this in their `embrace-config.json` file won't be affected by this change - the warning simply won't show up anymore


bidetofevil (approved):
Love me some cleanups

PR #252, opened 2024-01-16
Remove unnecessary code in session control

fractalwrench:
## Goal

Removes `triggerStatelessSessionEnd`, `startSession` and `close` from the internal `SessionService` interface. `triggerStatelessSessionEnd/startSession` are not called apart from within the service itself, and `close()` is not necessary to call - the executor is shutdown via the `WorkerThreadModule`.

There are also some style tweaks I made during these changes.

## Testing


lucaslabari (approved):
LGTM

PR #254, opened 2024-01-17
Create session orchestrator class

fractalwrench:
## Goal

This changeset creates a `SessionOrchestrator` class that is meant to be the sole class that is responsible for controlling when sessions + background activities start/end. This does not fully achieve the intended architecture changes that have been discussed internally but is a good first step towards combining some of this logic.

The `SessionOrchestrator` registers for a callback on the `ProcessStateService`, which notifies its listeners when the app enters/exits the foreground/background. Additionally, the `SessionOrchestrator` is notified when a developer requests to manually end a session. Finally, `SessionOrchestrator` also contains logic for starting an initial background activity/session depending on the process state.

I took this opportunity to rename functions on the `SessionService` and `BackgroundActivityService` so that they are a bit more consistent. This should help us eventually unify the two.

## Testing

Updated & added to existing test coverage.

### DIFF
--- a/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/session/EmbraceBackgroundActivityService.kt
+++ b/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/session/EmbraceBackgroundActivityService.kt
@@ -12,7 +12,6 @@ import io.embrace.android.embracesdk.payload.Session
 import io.embrace.android.embracesdk.payload.Session.LifeEventType
 import io.embrace.android.embracesdk.payload.SessionMessage
 import io.embrace.android.embracesdk.session.PayloadMessageCollator.PayloadType
-import io.embrace.android.embracesdk.session.lifecycle.ProcessStateService
 import io.embrace.android.embracesdk.worker.ScheduledWorker
 import java.util.concurrent.TimeUnit
 import java.util.concurrent.atomic.AtomicInteger
@@ -20,7 +19,6 @@ import kotlin.math.max
 
 internal class EmbraceBackgroundActivityService(
     private val metadataService: MetadataService,
-    processStateService: ProcessStateService,
     private val deliveryService: DeliveryService,
     private val configService: ConfigService,
     private val ndkService: NdkService,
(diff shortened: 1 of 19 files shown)


embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/session/EmbraceBackgroundActivityService.kt
```diff
@@ -42,17 +40,11 @@ internal class EmbraceBackgroundActivityService(
     var backgroundActivity: Session? = null
     private val manualBkgSessionsSent = AtomicInteger(0)
 
-    private var lastSendAttempt: Long
+    private var lastSendAttempt: Long = clock.now()
```
bidetofevil, line 43: Should this be set to 0 or something to indicate that we didn't know if there was even a last attempt?

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/session/orchestrator/SessionOrchestratorImpl.kt
```diff
@@ -0,0 +1,45 @@
...
+            // If the app goes to foreground before the SDK finishes its startup,
+            // the session service will not be registered to the activity listener and will not
+            // start the cold session.
+            // If so, force a cold session start.
+            sessionService.startSessionWithState(true, clock.now())
+        }
+    }
+
+    override fun onForeground(coldStart: Boolean, startupTime: Long, timestamp: Long) {
+        backgroundActivityService?.endBackgroundActivityWithState(timestamp)
```
bidetofevil, line 30: Being able to synchronize Session and BA start/stop is such a win

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/session/EmbraceBackgroundActivityService.kt
bidetofevil: Ah, I see this was just what the code used to do. Never mind - feel free to not make this change!

bidetofevil (approved):
LGTM. Feel free to ignore the comment about the lastSent time in BA since what you have is just what was there before.

PR #255, opened 2024-01-17
Refactor how session/background services are gated

fractalwrench:
## Goal

Refactors how the session/background activity services are gated by creating a `ConfigGate` class. This is responsible for checking whether a feature is enabled & holding state so that either a service is returned, or `null` is returned if not. The aim is to free up the session & background activity service from having to hold state about whether it's enabled or not & gate every single function call.

This is a little problematic in that the service can be disabled at any time in the process lifecycle, so it'd be great to get a detailed review to see whether I've missed anywhere this might cause us state management pain. Having said that, the existing services don't do a great job of this either. Ultimately I'd like to change this so that there's a register/deregister call on every service.

## Testing

Added to existing test coverage.

### DIFF
--- a/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/session/orchestrator/SessionOrchestratorImpl.kt
+++ b/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/session/orchestrator/SessionOrchestratorImpl.kt
@@ -1,19 +1,36 @@
 package io.embrace.android.embracesdk.session.orchestrator
 
+import io.embrace.android.embracesdk.config.ConfigService
+import io.embrace.android.embracesdk.internal.MessageType
 import io.embrace.android.embracesdk.internal.clock.Clock
 import io.embrace.android.embracesdk.session.BackgroundActivityService
+import io.embrace.android.embracesdk.session.ConfigGate
 import io.embrace.android.embracesdk.session.SessionService
 import io.embrace.android.embracesdk.session.lifecycle.ProcessStateService
 
 internal class SessionOrchestratorImpl(
     private val processStateService: ProcessStateService,
-    private val sessionService: SessionService,
-    private val backgroundActivityService: BackgroundActivityService?,
-    clock: Clock
+    sessionServiceImpl: SessionService,
+    backgroundActivityServiceImpl: BackgroundActivityService?,
+    clock: Clock,
+    configService: ConfigService
 ) : SessionOrchestrator {
+
+    private val sessionGate = ConfigGate(sessionServiceImpl) {
+        configService.dataCaptureEventBehavior.isMessageTypeEnabled(MessageType.SESSION)
+    }
+    private val backgroundActivityGate = ConfigGate(backgroundActivityServiceImpl) {
+        configService.isBackgroundActivityCaptureEnabled()
+    }
+    private val sessionService: SessionService?
+        get() = sessionGate.getService()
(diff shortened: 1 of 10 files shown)


embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/session/orchestrator/SessionOrchestratorImpl.kt
```diff
@@ -1,19 +1,36 @@
...
-    private val backgroundActivityService: BackgroundActivityService?,
-    clock: Clock
+    sessionServiceImpl: SessionService,
+    backgroundActivityServiceImpl: BackgroundActivityService?,
+    clock: Clock,
+    configService: ConfigService
 ) : SessionOrchestrator {
 
+    private val sessionGate = ConfigGate(sessionServiceImpl) {
+        configService.dataCaptureEventBehavior.isMessageTypeEnabled(MessageType.SESSION)
```
bidetofevil, line 20: The SDK can be started with sessions turned off? Is this something we actually support?

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/session/orchestrator/SessionOrchestratorImpl.kt
```diff
@@ -22,24 +39,24 @@ internal class SessionOrchestratorImpl(
             // the session service will not be registered to the activity listener and will not
             // start the cold session.
             // If so, force a cold session start.
-            sessionService.startSessionWithState(true, clock.now())
+            sessionService?.startSessionWithState(true, clock.now())
         }
     }
 
     override fun onForeground(coldStart: Boolean, startupTime: Long, timestamp: Long) {
         backgroundActivityService?.endBackgroundActivityWithState(timestamp)
```
bidetofevil, line 47: I assume that starting and ending a BA will just work even if the instantiation is done outside the init block? We may want to add a test to verify this

bidetofevil (approved):
This is likely already a problem but I'm highlight in anyway:

In `EmbraceImpl.java`, we initialize and register the `EmbraceBackgroundActivityService` at SDK start up time. Through this reference, we call `save()` in `onActivityReported()`. If the feature is enabled after startup, this reference is missing so `onActivityReported()` doesn't do anything. I think we should fix this, but not as part of this change.

In general, I like this pattern for services are mostly only consumed in Kotlin where null handling is much nicer than Java. I might want to simply the `SpansService` to use this pattern as well. I didn't find anything from looking at the code that would make this worse than what existed before, so I'm good with this change - with maybe an additional test added.

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/session/orchestrator/SessionOrchestratorImpl.kt
fractalwrench: The pre-existing behavior for both sessions/BA is if something was started & then disabled, it gets sent. If it was disabled and then an end attempt is made, nothing is recorded. I added some unit tests to simulate this.

fractalwrench:
`EmbraceImpl` should never have a null reference to the service after the changes to `SessionModuleImpl`. Tagging for re-review as I made a few changes to the unit tests + EmbraceImpl to reflect this.

bidetofevil (approved):
LGTM

PR #259, opened 2024-01-18
Remove disabled message type

fractalwrench:
## Goal

Removes the `disabledMessagesType` remote config field as Fredric ran queries that showed this was only used for one (deleted) appId. Getting rid of this simplifies our session & event capture logic.


bidetofevil (approved):
oooooh so great!

PR #260, opened 2024-01-18
Move session handler into session service

fractalwrench:
## Goal

Moves the `SessionHandler` implementation to be part of `SessionService`. `SessionService` was effectively just decorating the implementation previously.

## Testing


bidetofevil (approved):
Looks great! Normally, I prefer grouping logical functionality together into components even if there's only one user, but in this case, there's some dependencies like `setSdkStartupInfo()` that make the abstractions blurry, so you have the complexity of two components but having some undesirable coupling, so they aren't really as independent that you'd like.

Smashing these together makes sense. `SessionHandler` just require too many dependencies to be passed into it for it to function well and be tested independently from `SessionService`

PR #261, opened 2024-01-18
Remove unused startupTime parameter

fractalwrench:
## Goal

Removes an unused parameter that is tracked & inserted into the session payload via other means.

## Testing


bidetofevil (approved):
Another leaky abstraction removed. LGTM

PR #262, opened 2024-01-18
Extract initial session/BA params into separate class

fractalwrench:
## Goal

Extracts the parameters required to create an initial session/background activity object into a sealed class hierarchy. This should make it clearer what the differences are between session/background activity & improve testability. A future changeset will attempt the same for the end session message.

## Testing


bidetofevil (approved):
LGTM

PR #264, opened 2024-01-18
Rename & hide payload creation functions

fractalwrench:
## Goal

Renames a few functions for greater consistency with how the session building functions work. Additionally this simplifies the code so that the function that creates the `BackgroundActivity` is now hidden, similarly to how construction of sessions behave.

## Testing


bidetofevil (approved):
LGTM

PR #265, opened 2024-01-18
EMBR-1246 Compress payloads for Pending Api Calls.

lucaslabari:
## Goal

- Compress all payloads stored as pending API calls. 
- When serializing we now compress the data too, so when going through the ApiService and ApiClient it will be sent compressed. Then if it's stored as a pending API call, it's already compressed. 
- To avoid losing uncompressed payloads stored in the device by previous versions, the `loadPayload()` method in `EmbraceCacheService` does the compression in the SerializationAction. This way we don't loose those payloads and it will also compress stored last sessions until we do the changes to store them compressed. 

## Testing

Added unit tests and updated others. 

## Release Notes

Compress payloads before storing them in the device.


embrace-android-sdk/src/test/java/io/embrace/android/embracesdk/internal/compression/GzipCompressorTest.kt
```diff
@@ -0,0 +1,30 @@
...
+
+    private lateinit var compressor: GzipCompressor
+
+    @Before
+    fun setUp() {
+        compressor = GzipCompressor(logger = InternalEmbraceLogger())
+    }
+
+    @Test
+    fun testCompression() {
```
fractalwrench, line 20: Would be good to have a test case for when the payload is not compressed too

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/comms/delivery/EmbraceCacheService.kt
```diff
@@ -57,13 +59,27 @@ internal class EmbraceCacheService(
...
+                if (compressor.isCompressed(file)) {
+                    file.inputStream().buffered().use { input ->
+                        input.copyTo(stream)
+                    }
+                } else {
+                    compressor.compress(stream) {
+                        file.inputStream().buffered().use { input ->
+                            input.copyTo(it)
+                        }
+                    }
```
fractalwrench, line 82: Could `Compressor` extend `OutputStream` and delegate most of its work (apart from checking whether info is compressed) to `GzipOutputStream`? IMO this would avoid having to introduce logic like this throughout the codebase

fractalwrench (commented):
Left a comment about extending `OutputStream` as I think that'll reduce the amount of code we need to write in future. This is great - I'd always thought this would take more code to achieve. Out of curiosity, do you have an idea of what the payload size was before + after this change?

embrace-android-sdk/src/test/java/io/embrace/android/embracesdk/internal/compression/GzipCompressorTest.kt
lucaslabari: Here the compressor class just compresses what appears in the outputStream of the SerializationAction provided. The tests for covering the behavior of when the stored payload is compressed or not are in these tests:
- `test cachePayload stores uncompressed data and loadPayload returns compressed data`
- `test cachePayload stores compressed data and loadPayload returns compressed data`
in the `EmbraceCacheServiceTest` class.

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/comms/delivery/EmbraceCacheService.kt
```diff
@@ -57,13 +59,27 @@ internal class EmbraceCacheService(
...
+     * payloads from previous versions, we need to check if the file is compressed or not.
+     */
     override fun loadPayload(name: String): SerializationAction {
         logger.logDeveloper(TAG, "Attempting to read bytes from $name")
         return { stream ->
             val file = storageService.getFileForRead(EMBRACE_PREFIX + name)
             try {
-                file.inputStream().buffered().use { input ->
-                    input.copyTo(stream)
+                if (compressor.isCompressed(file)) {
```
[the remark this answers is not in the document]
fractalwrench: I'm not a fan of catching a thrown exception and doing a fallback. It should be possible to achieve what we want by reading the first N bytes into a temporary buffer, checking whether they indicate compression, then writing to the OutputStream as necessary.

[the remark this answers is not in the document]
bidetofevil: That works too and is much preferred. Whatever that avoids the double read in the default non fallback case.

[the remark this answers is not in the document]
lucaslabari: Updated with the OutputStream containing a buffer to check for the gzip magic number in order to determine if the data is already compressed or not.

fractalwrench (approved):
LGTM

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/internal/compression/CompressionOutputStream.kt
```diff
@@ -0,0 +1,70 @@
+package io.embrace.android.embracesdk.internal.compression
+
+import java.io.OutputStream
+import java.util.zip.GZIPOutputStream
+
+/**
+ * This Output Stream is used to compress the data that is being written to it.
+ * It will only compress the data if it's not already compressed.
+ * Given we started compressing payloads to disk from version 6.3.0 onwards,
```
bidetofevil, line 9: This is a standalone component, so a description of why we are using it doesn't seem appropriate. Maybe adding this a the place we are using it is better?

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/internal/compression/CompressionOutputStream.kt
```diff
@@ -0,0 +1,70 @@
...
+        if (!drainedBuffer) {
+            impl = generateOutputStream()
+            magicNumberBuffer.forEach {
+                impl.write(it.toInt())
+            }
+            drainedBuffer = true
+        }
+    }
+
+    private fun generateOutputStream(): OutputStream {
```
bidetofevil, line 54: If 0 or 1 byte is written, this will always create a GZipOutputStream. This can happen if `close()` is called too before 2 or more bytes are written. 

Does it make sense to modify the code so the stream is only set if two bytes are buffered, and if the drain happens before that, you just use the stream passed in?

One way you can do that is separating the stream creation method from the drain method, and only call the former in the `write()` method, and have the latter defer to stream passed in if the buffered stream isn't set. There are probably other ways to do this so feel free to go some other route.

embrace-android-sdk/src/test/java/io/embrace/android/embracesdk/EmbraceCacheServiceTest.kt
```diff
@@ -110,11 +133,13 @@ internal class EmbraceCacheServiceTest {
     }
 
     @Test
-    fun `test loadPayload with non-existent file returns null`() {
+    fun `test loadPayload with non-existent file returns empty string in the output stream`() {
         val action = service.loadPayload("some_file.jpeg")
         val stream = ByteArrayOutputStream()
         action(stream)
-        assertEquals("", String(stream.toByteArray()))
+        println(stream.toByteArray())
```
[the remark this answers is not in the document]
lucaslabari: Ups, I was checking a value for the test. I removed it now.

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/internal/compression/CompressionOutputStream.kt
lucaslabari: I made a modification to return the passed stream in the scenario when the buffer is empty because no bytes were written. But for the case were, the output stream is closed and one byte was written, we still want to compress that byte. That's why for that case, we still use the GzipOutputStream.

bidetofevil: Got it. Thanks!

bidetofevil (approved):
Looks good!

PR #266, opened 2024-01-19
Add SpansRepository to track started spans

bidetofevil:
## Goal

Add a component to keep track of spans that have been started so their reference can be obtained later using just a spanId

## Testing

Added unit tests

### DIFF
--- a/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/internal/spans/SpansRepository.kt
+++ b/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/internal/spans/SpansRepository.kt
@@ -1,0 +1,62 @@
+package io.embrace.android.embracesdk.internal.spans
+
+import io.embrace.android.embracesdk.spans.EmbraceSpan
+import java.util.concurrent.ConcurrentHashMap
+
+/**
+ * Allows the tracking of [EmbraceSpan] instances so that their references can be retrieved with its associated spanId
+ */
+internal class SpansRepository {
+    private val activeSpans: MutableMap<String, EmbraceSpan> = ConcurrentHashMap()
+    private val completedSpans: MutableMap<String, EmbraceSpan> = ConcurrentHashMap()
+
+    /**
+     * Track the [EmbraceSpan] if it has been started and it's not already tracked.
+     */
+    fun started(embraceSpan: EmbraceSpan) {
+        embraceSpan.spanId?.let { spanId ->
+            if (activeSpans[spanId] == null && completedSpans[spanId] == null) {
+                synchronized(activeSpans) {
+                    if (activeSpans[spanId] == null && completedSpans[spanId] == null) {
+                        if (embraceSpan.isRecording) {
+                            activeSpans[spanId] = embraceSpan
+                        } else {
+                            completedSpans[spanId] = embraceSpan
+                        }
+                    }
+                }
+            }
+        }
+    }
+
+    /**
+     * Transition active span to completed span if the span is tracked and the span is actually stopped.
+     */
+    fun stopped(spanId: String) {
+        synchronized(activeSpans) {
+            activeSpans[spanId]?.takeIf { !it.isRecording }?.let { activeSpans.remove(spanId) }?.let { span ->
+                completedSpans[spanId] = span
+            }
+        }
+    }
+
+    /**
+     * Return the [EmbraceSpan] with the corresponding [spanId] if it's tracked. Return null otherwise.
+     */
+    fun getSpan(spanId: String): EmbraceSpan? {
+        synchronized(activeSpans) {
+            activeSpans[spanId]?.let { return it }
+        }
+        return completedSpans[spanId]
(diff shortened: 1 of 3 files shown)


bidetofevil:
* `master`
  * **PR #266** <a href="https://app.graphite.dev/github/pr/embrace-io/embrace-android-sdk/266?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/embrace-io/embrace-android-sdk/266?utm_source=stack-comment).

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/internal/spans/SpansRepository.kt
```diff
@@ -0,0 +1,62 @@
+package io.embrace.android.embracesdk.internal.spans
+
+import io.embrace.android.embracesdk.spans.EmbraceSpan
+import java.util.concurrent.ConcurrentHashMap
+
+/**
+ * Allows the tracking of [EmbraceSpan] instances so that their references can be retrieved with its associated spanId
+ */
+internal class SpansRepository {
```
bidetofevil, line 9: I'm still missing a strategy to clear the tracked spans because you can theoretically use as parents spans that are already sent off device. I'm going to do some profiling to see how much memory keeping track of 1000+ spans take, but I'm likely going to add a facility to clear the tracked spans at some point.

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/internal/spans/SpansRepository.kt
```diff
@@ -0,0 +1,62 @@
...
+    private val activeSpans: MutableMap<String, EmbraceSpan> = ConcurrentHashMap()
```
fractalwrench, line 10: Strictly speaking I don't think this needs to be `ConcurrentHashMap` as access is always synchronised, it's fine to leave it as-is though

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/internal/spans/SpansRepository.kt
```diff
@@ -0,0 +1,62 @@
...
+    private val completedSpans: MutableMap<String, EmbraceSpan> = ConcurrentHashMap()
+
+    /**
+     * Track the [EmbraceSpan] if it has been started and it's not already tracked.
+     */
+    fun started(embraceSpan: EmbraceSpan) {
+        embraceSpan.spanId?.let { spanId ->
+            if (activeSpans[spanId] == null && completedSpans[spanId] == null) {
+                synchronized(activeSpans) {
+                    if (activeSpans[spanId] == null && completedSpans[spanId] == null) {
```
fractalwrench, line 20: nit: this is nested pretty deep. I'd suggest `val spanId = embraceSpan.spanId ?: return`, or possibly creating an abstraction for checking the same condition before & after the synchronized block

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/internal/spans/SpansRepository.kt
```diff
@@ -0,0 +1,62 @@
...
+            }
+        }
+    }
+
+    /**
+     * Transition active span to completed span if the span is tracked and the span is actually stopped.
+     */
+    fun stopped(spanId: String) {
+        synchronized(activeSpans) {
+            activeSpans[spanId]?.takeIf { !it.isRecording }?.let { activeSpans.remove(spanId) }?.let { span ->
```
fractalwrench, line 37: Is the span guaranteed to always be `false` for `isRecording` at this point?

fractalwrench (approved):
LGTM

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/internal/spans/SpansRepository.kt
bidetofevil: I'll change that. No need to pay the penalty if we don't have to.

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/internal/spans/SpansRepository.kt
bidetofevil: Yeah I tacked this on pretty late. Let me think about this - early return probably solves this nicely.

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/internal/spans/SpansRepository.kt
bidetofevil: It's not if someone make a mistake and calls this on a span that isn't actually stopped. This provides another level of validation.

PR #267, opened 2024-01-19
Add tests for constructing final payload

fractalwrench:
## Goal

Adds some tests for constructing the final session/background activity payloads. This allowed for us to remove some repetition in how the session message was built, by using the background activity message as a base.


bidetofevil (approved):
LGTM

PR #268, opened 2024-01-19
Create final envelope params class

fractalwrench:
## Goal

Creates a class that holds the parameters used to create the final session/BA objects. I also consolidated some null checks & data population within `PayloadMessageCollator`.

## Testing


bidetofevil (approved):
Very nice. We may want to call this something other than "envelope" in the future when these things are collapsed since it may be confused with the payload envelope, but this is looking great.

PR #269, opened 2024-01-19
Added telemetry to track how much storage we use

priettt:
## Goal

After a minute, we will track how much space the files on the StorageService are taking, and how much available space is left

image

### DIFF
--- a/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/injection/StorageModule.kt
+++ b/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/injection/StorageModule.kt
@@ -27,7 +27,11 @@ internal class StorageModuleImpl(
 ) : StorageModule {
 
     override val storageService: StorageService by singleton {
-        EmbraceStorageService(coreModule.context)
+        EmbraceStorageService(
+            coreModule.context,
+            initModule.telemetryService,
+            workerThreadModule.scheduledWorker(WorkerName.PERIODIC_CACHE)
+        )
     }
 
     override val cache by singleton {
--- a/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/injection/StorageModule.kt
+++ b/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/injection/StorageModule.kt
@@ -9,6 +9,7 @@ import io.embrace.android.embracesdk.storage.EmbraceStorageService
 import io.embrace.android.embracesdk.storage.StorageService
 import io.embrace.android.embracesdk.worker.WorkerName
 import io.embrace.android.embracesdk.worker.WorkerThreadModule
+import java.util.concurrent.TimeUnit
 
 /**
  * Contains dependencies that are used to store data in the device's storage.
@@ -29,8 +30,7 @@ internal class StorageModuleImpl(
     override val storageService: StorageService by singleton {
         EmbraceStorageService(
             coreModule.context,
-            initModule.telemetryService,
-            workerThreadModule.scheduledWorker(WorkerName.PERIODIC_CACHE)
+            initModule.telemetryService
         )
     }
 
@@ -53,5 +53,11 @@ internal class StorageModuleImpl(
             initModule.clock,
             coreModule.jsonSerializer
         )
+    }
+
+    init {
+        workerThreadModule
+            .scheduledWorker(WorkerName.PERIODIC_CACHE)
+            .schedule<Unit>({ storageService.logStorageTelemetry() }, 1, TimeUnit.MINUTES)
     }
 }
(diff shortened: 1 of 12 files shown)


embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/telemetry/EmbraceTelemetryService.kt
```diff
@@ -21,8 +23,13 @@ internal class EmbraceTelemetryService(
         }
     }
 
+    override fun logStorageTelemetry(fileToSizeMap: Map<String, Int>) {
```
bidetofevil, line 26: I don't think we to store all the file names and the associated size - just the size ALL the files are taking on disk. Instead of passing back a map, we can just add the bytes all up and return a number, which will be a lot easier to consume and make dashboards from.

If we want to get fancy in the the future, we can add different categories (e.g. sessions), but I think one number representing the total amount of cached data we have would be sufficient.

fractalwrench:
While this information is useful, I wonder whether if it would also be worth tracking the % of available storage. This would probably be more useful than absolute numbers by itself, as we could calculate metrics such as "In 99% of devices Embrace uses <2% of available storage". It should be possible to calculate these values using `StatFs`.

Absolute numbers are definitely helpful to capture - but there's a big difference in using 1Mb of storage on a device from 2015, versus using 1Mb on the latest Google Pixel

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/storage/EmbraceStorageService.kt
```diff
@@ -52,6 +55,21 @@ internal class EmbraceStorageService(
...
     }
 
+    /**
+     * Logs storage telemetry, including the current size utilized by Embrace.
+     */
+    override fun logStorageTelemetry() {
+        val availableStorage = storageAvailabilityChecker.getAvailableBytes()
+        val storageUsed = listFiles { _, _ -> true }
+            .filter { it.isFile }
+            .sumOf { it.length() }
```
bidetofevil, line 65: will this `sumOf` be a `Long`? We have to make sure it won't overflow, andI think an 32-bit integer could

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/injection/StorageModule.kt
```diff
@@ -50,4 +56,10 @@ internal class StorageModuleImpl(
             coreModule.jsonSerializer
         )
     }
+
+    init {
+        workerThreadModule
+            .scheduledWorker(WorkerName.PERIODIC_CACHE)
```
bidetofevil, line 62: Looking at the description, `BACKGROUND_REGISTRATION` might be a better fit for this?

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/telemetry/EmbraceTelemetryService.kt
```diff
@@ -21,8 +22,13 @@ internal class EmbraceTelemetryService(
         }
     }
 
+    override fun logStorageTelemetry(storageTelemetry: Map<String, String>) {
+        this.storageTelemetryMap.putAll(storageTelemetry)
```
bidetofevil, line 26:
```suggestion
        storageTelemetryMap.putAll(storageTelemetry)
```

bidetofevil: `this` seems extraneous?

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/telemetry/TelemetryService.kt
```diff
@@ -8,6 +8,12 @@ internal interface TelemetryService {
      */
     fun onPublicApiCalled(name: String)
 
+    /**
+     * Tracks the storage being used. storageTelemetry is a map of storage telemetry names and their values, such as:
```
bidetofevil, line 12: I assume this is in bytes? Might want to clarify that in the comment

bidetofevil (approved):
Some basic suggestions, but looks good overall.

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/storage/EmbraceStorageService.kt
priettt: Yup, it's a Long! And then we convert it to a String so we don't have any overflow risk unless we find a phone with 4 Exbibytes of available storage 😅

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/injection/StorageModule.kt
priettt: I wanted to use that one but it says: 

> recurring tasks that don't use I/O or block for long periods of time.

and this reads from the disk. 

I guess it does fall into

> Used primarily to perform short-lived tasks that need to execute only once

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/telemetry/EmbraceTelemetryService.kt
priettt: leftover of a previous name 😬

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/injection/StorageModule.kt
bidetofevil: Serializing disk access probably isn't a bad idea. We may want to give it a different name so it's not so specific about session/BA payload saving, but yeah, this seems fine.

PR #273, opened 2024-01-22
Consolidate payload creation in one place

fractalwrench:
## Goal

Further builds on #268 to consolidate the creation of session & background activity payloads from one place. This extracts logic about what should/shouldn't be collected to `FinalEnvelopeParams`

## Testing


bidetofevil (approved):
LGTM

PR #274, opened 2024-01-22
Extract SDK startup duration capture to separate service

fractalwrench:
## Goal

Extracts the capture of SDK startup duration to a separate service, as it doesn't really belong in the `SessionService`.

## Testing


bidetofevil (approved):
Very nice. I bolted this onto the startup moment code initially. Could that be put into to this as well?

fractalwrench:
@bidetofevil I can extract that in a future PR. Where does it currently live in the code?

bidetofevil:
> @bidetofevil I can extract that in a future PR. Where does it currently live in the code?

It's mostly under the Events stuff - `EmbraceEventService` and `EventHandler` (but you can also end it manually, which we ignore for the span).  Search for references to `STARTUP_EVENT_NAME` and you'll find them.

PR #275, opened 2024-01-22
Alter how session boundary is cleared

fractalwrench:
## Goal

Extracts the logic for clearing the session boundary to the `SessionOrchestrator` so that it is responsible for resetting `MemoryCleanerListener` services.

## Testing

### DIFF
--- a/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/session/orchestrator/SessionOrchestratorImpl.kt
+++ b/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/session/orchestrator/SessionOrchestratorImpl.kt
@@ -2,8 +2,10 @@ package io.embrace.android.embracesdk.session.orchestrator
 
 import io.embrace.android.embracesdk.config.ConfigService
 import io.embrace.android.embracesdk.internal.clock.Clock
+import io.embrace.android.embracesdk.logging.InternalErrorService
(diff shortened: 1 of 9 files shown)


embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/session/orchestrator/SessionOrchestratorImpl.kt
```diff
@@ -38,11 +42,13 @@ internal class SessionOrchestratorImpl(
 
     override fun onForeground(coldStart: Boolean, timestamp: Long) {
         backgroundActivityService?.endBackgroundActivityWithState(timestamp)
```
bidetofevil, line 44: All these methods to start/stop sessions are running serially because they are always done on a particular thread?

bidetofevil (approved):
LGTM

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/session/orchestrator/SessionOrchestratorImpl.kt
fractalwrench: Just keeping existing behavior for now, which is to run on whatever the calling thread is. `SessionService` does use a lock that I want to move up to this class eventually.

PR #276, opened 2024-01-22
Make session orchestrator responsible for clearing session props

fractalwrench:
## Goal

Makes the session orchestrator responsible for clearing temporary session properties.

## Testing

Added to existing test coverage.

### DIFF
--- a/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/session/orchestrator/SessionOrchestratorImpl.kt
+++ b/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/session/orchestrator/SessionOrchestratorImpl.kt
@@ -8,6 +8,7 @@ import io.embrace.android.embracesdk.session.ConfigGate
 import io.embrace.android.embracesdk.session.MemoryCleanerService
 import io.embrace.android.embracesdk.session.SessionService
 import io.embrace.android.embracesdk.session.lifecycle.ProcessStateService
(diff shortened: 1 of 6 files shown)


embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/session/orchestrator/SessionOrchestratorImpl.kt
```diff
@@ -67,5 +69,6 @@ internal class SessionOrchestratorImpl(
      */
     private fun prepareForNewEnvelope() {
         memoryCleanerService.cleanServicesCollections(internalErrorService)
+        sessionProperties.clearTemporary()
```
bidetofevil, line 72: What happens before this change if a session property is set while the user is in the background? It looks like it would persist in the next session, but with this change, it'll be cleared, which tbh seems like the correct thing to do, but it'll be a behavioural change.

If Eliab is OK with this subtle change/fix, I'm good too.

bidetofevil (approved):
Ready to go if the behaviour change is OK with Eliab.

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/session/orchestrator/SessionOrchestratorImpl.kt
fractalwrench: Started discussion in Slack

PR #278, opened 2024-01-23
Extract session id tracking

fractalwrench:
## Goal

Extracts the tracking of session ID from the `MetadataService` to a separate class. This helps breakdown the monolithic `MetadataService` and is the first step in moving the session ID tracking to be controlled via `SessionOrchestrator`.

## Testing

### DIFF
--- a/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/session/id/SessionIdTrackerImpl.kt
+++ b/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/session/id/SessionIdTrackerImpl.kt
@@ -1,0 +1,41 @@
+package io.embrace.android.embracesdk.session.id
+
+import android.app.ActivityManager
+import android.os.Build
+import io.embrace.android.embracesdk.logging.InternalStaticEmbraceLogger
+
+internal class SessionIdTrackerImpl(
+    private val activityManager: ActivityManager? = null
+) : SessionIdTracker {
+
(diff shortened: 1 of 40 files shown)


embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/session/id/SessionIdTrackerImpl.kt
```diff
@@ -0,0 +1,41 @@
...
+    private val activityManager: ActivityManager? = null
+) : SessionIdTracker {
+
+    @Volatile
+    private var sessionId: String? = null
+
+    override fun getActiveSessionId(): String? = sessionId
+
+    override fun setActiveSessionId(sessionId: String?, isSession: Boolean) {
+        this.sessionId = sessionId
```
bidetofevil, line 17: do we need to synchronize access to this?

bidetofevil (approved):
LGTM. Added a question about synchronization but if this is just an intermediate step don't worry about it as it's just what is happening now.

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/session/id/SessionIdTrackerImpl.kt
fractalwrench: Ultimately yes - I want to handle that in `SessionOrchestrator` eventually.

PR #280, opened 2024-01-23
Extract how user info is cleared

fractalwrench:
## Goal

Extracts user info clearing to the `SessionOrchestrator`.

## Testing

Updated unit tests.


bidetofevil (approved):
Looks good. We may want to abstract away the userInfo into a service that we pass into the two consuming services rather than making copies, which seems more error prone

PR #282, opened 2024-01-23
EMB-627 Added support for publishing, closing and releasing sonatype repositories

cesarmax22:
## Goal

Added the ability to:
- publish to sonatype staging repository
- close staging repository
- release from staging repository to maven central

Modified workflow so we now have:
- Pre-Release: this let us publish and close to staging sonatype repo. We can now smoke test
- Release: this let us release to maven central

## Release Notes

**WHAT**:<br>
**WHY**:<br>
**WHO**:<br>


.github/workflows/pre-release-workflow.yml
```diff
@@ -0,0 +1,124 @@
...
+#    with:
+#      sdk_version: ${{ github.event.inputs.current_version }}
+#      token: ${{ secrets.CD_GITHUB_TOKEN }}
+#
+#  measure-startup-time:
+#    uses: embrace-io/android-sdk-benchmark/.github/workflows/macrobenchmark.yml@main
+#    needs: release
+#    with:
+#      sdk_version: ${{ github.event.inputs.current_version }}
+#      token: ${{ secrets.CD_GITHUB_TOKEN }}
```
fractalwrench, line 124: If these are unused they can be deleted

fractalwrench (commented):
Nice, this should make our lives easier. For releasing, I like your idea of a pre-release workflow (that creates a snapshot we can test against) and a release workflow (that closes the snapshot).

Before we merge this I think it'd also be good to compare the POM + other artefacts this generates and check they are the same as what the old solution generates. This could be achieved by getting the files from maven local (`~/.m2/repository/io`) and eyeballing if there are any differences. I've been bitten in the past by POM files changing or artefacts going missing on upgrades like this.

cesarmax22:
I did a quick testing on android-test-suite with the version generated by nexus-plugin and it seems to work fine

.github/workflows/pre-release-workflow.yml
```diff
@@ -0,0 +1,124 @@
...
+        uses: actions/upload-artifact@v3
+        with:
+          name: android-sdk-test-results
+          path: embrace-android-sdk/build/reports/tests/
+
+      - name: Configure git
+        run: |
+          git config --global user.name 'embrace-ci[bot]'
+          git config --global user.email 'embrace-ci@users.noreply.github.com'
+          git config --global url."https://${{ secrets.CD_GITHUB_USER }}:${{ secrets.CD_GITHUB_TOKEN }}@github.com".insteadOf "https://github.com"
```
bidetofevil, line 81: is this the accepted way of passing a token around for git to consume?

lucaslabari (approved):

bidetofevil (approved):
Looks good to me as long as the way we are passing the token is a acceptable way of doing so. I know we typically want to avoid putting credentials in plain text in the URL but  if this is a one off use thing or one of those that can't be reused in other contexts, then it's fine.

.github/workflows/pre-release-workflow.yml
cesarmax22: I will create a ticket so we can revisit this

PR #283, opened 2024-01-23
Extract sending of cached sessions

fractalwrench:
## Goal

Extracts the sending of cached sessions from `SessionService` into `EmbraceImpl` to avoid side-effects when initializing the session service. Longer-term it would be good to consolidate this in `DeliveryService` but for now there are circular dependencies (`DeliveryService` initialises before `NdkService` but needs `NdkService` to check files exist).

## Testing

### DIFF
--- a/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/EmbraceImpl.java
+++ b/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/EmbraceImpl.java
@@ -623,6 +623,12 @@ final class EmbraceImpl {
         backgroundActivityService = sessionModule.getBackgroundActivityService();
         serviceRegistry.registerServices(sessionService, backgroundActivityService);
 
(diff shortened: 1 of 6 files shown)


embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/EmbraceImpl.java
```diff
@@ -623,6 +623,12 @@ private void startImpl(@NonNull Context context,
         backgroundActivityService = sessionModule.getBackgroundActivityService();
         serviceRegistry.registerServices(sessionService, backgroundActivityService);
 
+        // Send any sessions that were cached and not yet sent.
```
bidetofevil, line 626: This makes a lot of sense. The triggerer of the sending of cached sessions should be done by separate component or maybe even my the delivery service itself, as you said.

bidetofevil (approved):
LGTM

PR #284, opened 2024-01-23
Parameterise session delivery in SessionService

fractalwrench:
## Goal

Parameterises session delivery in SessionService, avoiding unnecessary duplication.

## Testing


bidetofevil (approved):
LGTM

PR #287, opened 2024-01-24
Extract logic for restricting manual ending of sessions

fractalwrench:
## Goal

Extracts the logic for restricting manual ending of sessions to the `SessionOrchestrator`. We restrict this so that sessions must be >5s to avoid poor instrumentation accidentally sending too many sessions.

## Testing

Added unit test coverage.


bidetofevil (approved):
LGTM

PR #288, opened 2024-01-24
Move NDK service to be updated within session id tracker

fractalwrench:
## Goal

Alters the session ID tracker so that the NDK service (if enabled) is updated about session ID changes.

## Testing

Updated unit test coverage.


bidetofevil (approved):
LGTM

PR #289, opened 2024-01-24
Remove dead code from background activity service

fractalwrench:
## Goal

Removes some dead code from the `BackgroundActivityService` that is only called from tests.


bidetofevil (approved):
Cool. Was this always the case that these were only called in tests or did the recent refactoring remove the use case in production code for this?

fractalwrench:
Searching 5.x, it seems this was only ever called from tests.

bidetofevil:
> Searching 5.x, it seems this was only ever called from tests.

Adding specific public functions just for testing is the absolute worse......

PR #290, opened 2024-01-24
Hide internal functions in SessionService

fractalwrench:
## Goal

Makes some internal functions private by altering the unit tests so they check the end result, rather than the internal implementation.


bidetofevil (approved):
Very nice. One side effect of ripping apart a big component into multiple is that if the boundaries and responsibilities are clean, you seen bad patterns like this where one component is invoking effectively private methods in another just to do a test. Love seeing us get rid of more of these!

PR #291, opened 2024-01-24
Avoid creating useless background activity

fractalwrench:
## Goal

Avoids creating a new background activity _after_ a crash has happened, as no useful information will be captured or persisted.


bidetofevil (approved):
LGTM

PR #292, opened 2024-01-24
Extract periodic caching to separate class

fractalwrench:
## Goal

Extracts the periodic caching of sessions to a separate class. This will make its eventual removal easier.

## Testing


bidetofevil (approved):
LGTM

PR #295, opened 2024-01-24
Make session orchestrator responsible for crashes

fractalwrench:
## Goal

Alters the `CrashService` so it calls into the `SessionOrchestrator` rather than individual services.

## Testing

Updated unit tests.


bidetofevil (approved):
Awesome!

PR #296, opened 2024-01-24
Extract state synchronisation to session orchestrator

fractalwrench:
## Goal

Moves most of the existing state synchronisation to the `SessionOrchestrator`. This adds synchronisation for the background activity capture & moves it for the session capture. There are two instances that can't be migrated out of the session/BA services yet (both to do with periodic caching) so I've chosen to pass a lock as a parameter instead. As periodic session caching will eventually be removed this should be possible to simplify in future.

## Testing

Relied on existing testing.

### DIFF
--- a/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/injection/SessionModule.kt
+++ b/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/injection/SessionModule.kt
@@ -82,6 +82,8 @@ internal class SessionModuleImpl(
         )
     }
 
+    private val orchestrationLock = Any()
+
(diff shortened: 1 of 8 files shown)


bidetofevil (approved):
Looks good. This will look so much nicer once we get rid of the session and BA services.

Also, I wonder if we can do these all on a background thread. If so, we can get rid of the synchronization and run them serially on a background thread instead of potentially locking on or blocking the main thread. I guess that makes the transitions a little less predictable, but it would be really nice if we can do that, as I've seen ANRs blamed on us being we were doing stuff during the transition.

Anyway, just a thought. I think this is good to go,

PR #299, opened 2024-01-25
Reduce default CI timeout

fractalwrench:
## Goal

The default timeout on CI jobs in [Github actions is 6 hours](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idtimeout-minutes). I've reduced this to 1 hour or 30 minutes depending on how long the job is reasonably expected to take. This should make it quicker to detect when a CI job has hanged indefinitely.


leandro-godon (approved):
lgtm

PR #300, opened 2024-01-25
Move responsibility for clearing session to delegate

fractalwrench:
## Goal

Fully moves the responsibility for clearing the session to a delegate. This continues what previous refactors have done by moving the breadcrumb service/network service state reset into `SessionOrchestratorDelegate`. I also ensured that the very first session resets the boundary & that the timestamp is predominantly calculated in the `SessionOrchestrator`.

## Testing

Updated unit tests.


embrace-android-sdk/src/test/java/io/embrace/android/embracesdk/session/SessionHandlerTest.kt
```diff
@@ -70,8 +69,6 @@ import org.junit.Test
 internal class SessionHandlerTest {
```
bidetofevil, line 69: At some point, we should move these tests to the appropriate test classes that are testing those objects now

bidetofevil (approved):
LGTM

PR #307, opened 2024-01-26
Inline functions + parameterise timestamp

fractalwrench:
## Goal

Inlines some functions & parameterises the timestamp calculation.

## Testing


lucaslabari (approved):
LGTM

PR #308, opened 2024-01-26
Extract background activity caching to separate class

fractalwrench:
## Goal

Extracts the caching of the background activity to a separate class. Additionally this now cancels any scheduled snapshot that happens after a background activity was ended.

### DIFF
--- a/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/EmbraceImpl.java
+++ b/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/EmbraceImpl.java
@@ -1538,7 +1538,7 @@ final class EmbraceImpl {
 
     private void onActivityReported() {
         if (backgroundActivityService != null) {
-            backgroundActivityService.save();
+            backgroundActivityService.saveBackgroundActivitySnapshot();
         }
     }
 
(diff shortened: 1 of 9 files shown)


embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/EmbraceImpl.java
```diff
@@ -1538,7 +1538,7 @@ FlutterInternalInterface getFlutterInternalInterface() {
 
     private void onActivityReported() {
```
bidetofevil, line 1539: I would LOVE to get rid of this. I'm sure we have calls that are missing this right now.

That said, I like the idea of someone being able to notify the caching mechanism that something has changed and the payload needs to be regenerated again. Perhaps this is something we can work on when the orchestration and new payload population work is done so we do a better job for background activities and also be more efficient rather than always saving every 2s for foreground sessions.

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/session/EmbraceBackgroundActivityService.kt
```diff
@@ -77,30 +68,17 @@ internal class EmbraceBackgroundActivityService(
...
-    /**
-     * Cache the activity, with performance information generated up to the current point.
-     */
-    private fun cacheBackgroundActivity(timestamp: Long) {
-        synchronized(orchestrationLock) {
-            try {
-                lastSaved = timestamp
-                val initial = backgroundActivity ?: return
+    override fun saveBackgroundActivitySnapshot() {
+        periodicCacher.scheduleSave {
```
bidetofevil, line 78: Nice

bidetofevil (approved):
LGTM. I have a long comment about redoing the session/BA caching scheme which may already be on your radar already :-)

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/EmbraceImpl.java
fractalwrench: Yep, it would be good to get rid of this. There's a task to align the session/BA persistence with each other, so we'll see whether it's possible as part of that.

PR #309, opened 2024-01-26
Prepare for hoisting params into session orchestrator

fractalwrench:
## Goal

Preparatory refactor for hoisting the `session/backgroundActivity` properties into the `SessionOrchestrator` class. This changeset involves returning `Session` rather than string when creating a session, and miscellaneous other pre-requisite changes.

## Testing


bidetofevil (approved):
LGTM

PR #310, opened 2024-01-26
Extract tracking of active session to orchestrator

fractalwrench:
## Goal

Extracts the tracking of the active session/background activity state to the `SessionOrchestrator`. The active session is now passed into the services as a parameter, meaning they effectively just create a final payload & instruct where it should be delivered.

I plan on creating a 'transitionState' function in the `SessionOrchestratorImpl` class to better model these state transitions, as the code is getting somewhat complex. That will come in a future changeset.

## Testing

Updated unit test coverage. Some test cases are deleted as they are covered by `SessionOrchestratorTest`, others are no longer necessary due to reduced complexity.


bidetofevil (approved):
Any PR that reduces complexity without reducing functionality is a-ok to me!

PR #311, opened 2024-01-26
Create transitionState function

fractalwrench:
## Goal

Refactors SessionOrchestratorImpl to create a transitionState function that models the logic required for each state transition. This makes it less likely we will forget to perform a step if we ever need to add an extra transition in future, and makes the code easier to read.

## Testing

### DIFF
--- a/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/session/orchestrator/SessionOrchestratorImpl.kt
+++ b/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/session/orchestrator/SessionOrchestratorImpl.kt
@@ -33,11 +33,19 @@ internal class SessionOrchestratorImpl(
         private const val MIN_SESSION_MS = 5000L
     }
 
+    private enum class State {
+        FOREGROUND,
+        BACKGROUND
+    }
+
     /**
      * The currently active session.
      */
     private var activeSession: Session? = null
-    private var inBackground = processStateService.isInBackground
+    private var state = when {
+        processStateService.isInBackground -> State.BACKGROUND
+        else -> State.FOREGROUND
+    }
 
     private val backgroundActivityGate = ConfigGate(backgroundActivityServiceImpl) {
         configService.isBackgroundActivityCaptureEnabled()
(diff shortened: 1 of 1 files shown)


embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/session/orchestrator/SessionOrchestratorImpl.kt
```diff
@@ -51,117 +56,130 @@
...
-                        "many sessions"
-                )
-                return
+        val timestamp = clock.now()
+
+        transitionState(
+            description = "endManual",
+            timestamp = timestamp,
+            clearUserInfo = clearUserInfo,
+            endProcessState = State.FOREGROUND,
```
bidetofevil, line 122: Could this be triggered in the background? That is, can we end background activities manually? 

I'm wondering if this is something we want to derive given the current state rather than stipulate before we transition, as doing this manually seems unrelated to background vs foreground.

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/session/orchestrator/SessionOrchestratorImpl.kt
```diff
@@ -171,6 +189,66 @@
...
+        endProcessState: State,
+        oldSessionAction: (initial: Session) -> Unit,
+        newSessionAction: () -> Session?,
+        earlyTerminationCondition: () -> Boolean = { false },
+        clearUserInfo: Boolean = false,
+    ) {
+        // supplied business logic says that we can't perform a transition yet.
+        // exit early & retain the current state instead.
+        if (earlyTerminationCondition()) {
+            return
```
bidetofevil, line 221: Is there utility in returning whether this succeeded or not? Might be an interesting case to log an internal error/warning for if this is unexpected

bidetofevil (approved):
I have a couple questions about edge cases but looks good otherwise. Feel free to merge away if the questions don't spark anything that you want to change now.

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/session/orchestrator/SessionOrchestratorImpl.kt
fractalwrench: I don't think we can derive this from current state (at least without passing in extra information during `createInitialEnvelope`).

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/session/orchestrator/SessionOrchestratorImpl.kt
fractalwrench: I've added some extra logging for this in this specific class. There are already some existing internal errors logged from ProcessStateService if an unbalanced call happens.

fractalwrench:
Tagging for review again as I needed a rebase & a few fixes for failing tests.

bidetofevil (approved):
Looks good

PR #312, opened 2024-01-26
Return from SDK init only if Spans Service is initialized

bidetofevil:
## Goal

Make it a condition that the `EmbraceSpansService` is initialized before returning from the start method. In practice, this should almost always true by the time the SDK initialization gets to the end.

### DIFF
--- a/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/EmbraceImpl.java
+++ b/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/EmbraceImpl.java
@@ -16,6 +16,7 @@ import java.util.UUID;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.Future;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.regex.Pattern;
 
@@ -363,7 +364,7 @@ final class EmbraceImpl {
 
     private void startImpl(@NonNull Context context,
                            boolean enableIntegrationTesting,
-                           @NonNull Embrace.AppFramework framework) throws ExecutionException, InterruptedException {
+                           @NonNull Embrace.AppFramework framework) throws ExecutionException, InterruptedException, TimeoutException {
         if (application != null) {
             // We don't hard fail if the SDK has been already initialized.
             InternalStaticEmbraceLogger.logWarning("Embrace SDK has already been initialized");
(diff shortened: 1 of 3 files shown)


bidetofevil:
* `master`
  * **PR #312** <a href="https://app.graphite.dev/github/pr/embrace-io/embrace-android-sdk/312?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/embrace-io/embrace-android-sdk/312?utm_source=stack-comment).

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/EmbraceImpl.java
```diff
@@ -704,6 +705,9 @@ private void startImpl(@NonNull Context context,
             internalEmbraceLogger.logDeveloper("Embrace", "Sending startup moment");
             nonNullEventService.sendStartupMoment();
         }
+
+        // This should return immediately given that EmbraceSpansService initialization should be fished at this point
+        spansInitTask.get();
```
fractalwrench, line 710: I'd suggesting bounding this to something like a few seconds, then no-oping/logging an internal error if it exceeds this time? Technically it should happen very rarely but I'm wary of anything that has the theoretical potential to indefinitely hang the SDK

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/EmbraceImpl.java
```diff
@@ -704,6 +705,9 @@ private void startImpl(@NonNull Context context,
             internalEmbraceLogger.logDeveloper("Embrace", "Sending startup moment");
             nonNullEventService.sendStartupMoment();
         }
+
+        // This should return immediately given that EmbraceSpansService initialization should be fished at this point
```
fractalwrench, line 709:
```suggestion
        // This should return immediately given that EmbraceSpansService initialization should be finished at this point
```

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/internal/spans/EmbraceTracer.kt
```diff
@@ -114,4 +112,7 @@ internal class EmbraceTracer(private val spansService: SpansService) : TracingAp
         )
 
     fun getSpan(spanId: String): EmbraceSpan? = spansService.getSpan(spanId = spanId)
+
+    @Deprecated("No required. Use Embrace.isStarted() to know when the full tracing API is available")
```
fractalwrench, line 116:
```suggestion
    @Deprecated("Not required. Use Embrace.isStarted() to know when the full tracing API is available")
```

fractalwrench (approved):
LGTM

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/EmbraceImpl.java
bidetofevil: Haha, yeah, I was going to add that last night but it go too late. Good eye! We don't want this to block the app from starting

PR #319, opened 2024-01-29
Fix failure in version string tests

fractalwrench:
## Goal

Fixes a failure when running unit tests that assert against BuildConfig values using the debug buildType. There seems to have been a recent behavior change that means this differs compared to the debug/release buildType that was failing some test cases locally for me. I have made the assertions less specific to make the test less brittle.


lucaslabari (approved):
LGTM

PR #320, opened 2024-01-29
Resolve test failures in session orchestration layer

fractalwrench:
## Goal

Fixes some test failures in the functional tests of the session orchestration. This was initially caused by altering `ProcessStateService` to set `isInBackground` before the listener was invoked, which broke the startup moment & other callbacks that relied on this order.

My fix is to reinstate the previous behavior & track state internally within the `SessionOrchestrator` class as required.


lucaslabari (approved):
LGTM

PR #322, opened 2024-01-29
Simplifies session orchestration state change logic

fractalwrench:
## Goal

Simplifies the session orchestration state change logic by introducing a `TransitionType` enum that helps calculate the final state of the process for each transition type. This avoids having to manually specify the type for each invocation of `transitionState` and reduces the parameter count.

## Testing


bidetofevil (approved):
Ah nice. LGTM!

PR #323, opened 2024-01-29
Reorganise session code into separate packages

fractalwrench:
## Goal

Alters the session code into distinct sub packages. As part of this I removed all references to `SessionService` and `BackgroundActivityService` within the main sourceSet - this requires eagerly loading the orchestrator class within its module.


bidetofevil (approved):
LGTM

PR #324, opened 2024-01-29
Make session orchestrator responsible for periodic caching

fractalwrench:
## Goal

This changeset makes the `SessionOrchestrator` responsible for periodic caching. This allows the removal of various parameters from the session + BA services & hides implementation details (such as the lock used by the orchestrator). 

## Testing

I've removed redundant test cases & relied on the `PeriodicSessionCacheTest` integration test case.


bidetofevil (approved):
I was so looking forward to this 🔥 🔥 🔥 🔥 🔥

PR #326, opened 2024-01-29
Change release GH actions to create and release from release branch

bidetofevil:
## Goal

Create a release branch to make any version-specific modifications on and make the release from there. The version number of the SDK will incremented when the release branch is cut, and commits after that will be de facto in the next release. The Swazzler will be updated in the pre-release workflow as it does before.

If there are additional changes that need to go into the release branch, they can be done first in master, then cherrypicked into the release branch.

For hotfixes, the same cherrypicking process will be done, but we will need to add additional GH Actions to ensure to automate the release.

### DIFF
--- a/.github/workflows/pre-release-workflow.yml
+++ b/.github/workflows/pre-release-workflow.yml
@@ -12,8 +12,11 @@ env:
 on:
   workflow_dispatch:
     inputs:
-      current_version:
-        description: 'Version to release'
+      version_to_release:
+        description: 'Version to release. A new branch called `release/<version>` will be created for it where the release-specific changes will be committed.'
+        required: true
+      next_version:
+        description: 'Next Version (Do NOT include -SNAPSHOT, will be added automatically)'
         required: true
 
 jobs:
@@ -31,9 +34,11 @@ jobs:
           echo $mavenSigningKeyRingFileEncoded | base64 -di > "$RUNNER_TEMP"/keystore/2DE631C1.gpg
           echo "mavenSigningKeyRingFile=$RUNNER_TEMP/keystore/2DE631C1.gpg" >> $GITHUB_ENV
 
-      - name: Checkout Branch
+      - name: Create and Checkout Branch
         uses: actions/checkout@v3
         with:
+          run: | 
+            git checkout -b release/${{ github.event.inputs.version_to_release }}
           fetch-depth: 0
           token: ${{ secrets.CD_GITHUB_TOKEN }}
 
@@ -83,11 +88,19 @@ jobs:
 
       - name: Gradlew Pre-Release
         run: |
-          sed -i -r "s#version = ([^\']+)#version = ${{ github.event.inputs.current_version }}#" gradle.properties
+          sed -i -r "s#version = ([^\']+)#version = ${{ github.event.inputs.version_to_release }}#" gradle.properties
           git add gradle.properties
-          git commit -m "CI/CD: change version to be released: ${{ github.event.inputs.current_version }}"
+          git commit -m "CI/CD: change version to be released: ${{ github.event.inputs.version_to_release }}"
           git push
           ./gradlew clean publishReleasePublicationToSonatype closeSonatypeStagingRepository -Dorg.gradle.parallel=false --stacktrace
+
+      - name: Set next SDK version
+        run: |
+          git checkout master
+          sed -i -r "s#version = ([^\']+)#version = ${{ github.event.inputs.next_version }}-SNAPSHOT#" gradle.properties
+          git add gradle.properties
+          git commit -m "CI/CD: set next version: ${{ github.event.inputs.next_version }}"
+          git push
 
       - name: Checkout Swazzler
         uses: actions/checkout@v3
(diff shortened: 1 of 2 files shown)


bidetofevil:
* `master`
  * **PR #326** <a href="https://app.graphite.dev/github/pr/embrace-io/embrace-android-sdk/326?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/embrace-io/embrace-android-sdk/326?utm_source=stack-comment).

fractalwrench (approved):
LGTM

.github/workflows/pre-release-workflow.yml
```diff
@@ -83,12 +88,20 @@ jobs:
...
         run: |
-          sed -i -r "s#version = ([^\']+)#version = ${{ github.event.inputs.current_version }}#" gradle.properties
+          sed -i -r "s#version = ([^\']+)#version = ${{ github.event.inputs.version_to_release }}#" gradle.properties
           git add gradle.properties
-          git commit -m "CI/CD: change version to be released: ${{ github.event.inputs.current_version }}"
+          git commit -m "CI/CD: change version to be released: ${{ github.event.inputs.version_to_release }}"
           git push
           ./gradlew clean publishReleasePublicationToSonatype closeSonatypeStagingRepository -Dorg.gradle.parallel=false --stacktrace
 
+      - name: Set next SDK version
```
cesarmax22, line 97: we shouldn't be setting the next version in pre-release-workflow. At least that's how I thought it.

pre-release-workflow will:
- create the staging repository in nexus
- upload the artefacts to it
- close repositories

at this stage, we are able to perform smoke testing. That is why it's called pre-release.

It is important that we do not update gradle.properties with next-version-SNAPSHOT in this step, because we need the actual "pre-released" version for the next workflow (release.yaml)

In release workflow we will:
- find the already created staging repository (this is why we don't need to update gradle.properties in previous step)
- make the actual release from sonatype to maven central

does it make sense? i wasn't able to test the full flow, but it should be working.

If so, we should make the same changes to swazzler

cesarmax22 (changes requested):
left a comment

.github/workflows/pre-release-workflow.yml
bidetofevil: So all the changes should be done to the release branch at `release/version`, so the release stage should be referring to that and not master.

The goal here is to take a snapshot of master and move it to another branch so that master can move along, assuming that everything after that goes into master is part of the next release. 

If there is something wrong in the release branch, we can fix it in master, and cherrypick the change and generate and test another release candidate. As long as we can do that in the release branch without needing to increment the patch version, we should be good.

.github/workflows/pre-release-workflow.yml
```diff
@@ -31,9 +34,11 @@ jobs:
           echo $mavenSigningKeyRingFileEncoded | base64 -di > "$RUNNER_TEMP"/keystore/2DE631C1.gpg
           echo "mavenSigningKeyRingFile=$RUNNER_TEMP/keystore/2DE631C1.gpg" >> $GITHUB_ENV
 
-      - name: Checkout Branch
+      - name: Create and Checkout Branch
         uses: actions/checkout@v3
         with:
+          run: | 
+            git checkout -b release/${{ github.event.inputs.version_to_release }}
```
bidetofevil, line 41: @cesarmax22 this is where I'm creating the release branch and at that point, everything should be running on that and not modifying master.

.github/workflows/pre-release-workflow.yml
```diff
@@ -83,12 +88,20 @@ jobs:
 
       - name: Gradlew Pre-Release
         run: |
-          sed -i -r "s#version = ([^\']+)#version = ${{ github.event.inputs.current_version }}#" gradle.properties
+          sed -i -r "s#version = ([^\']+)#version = ${{ github.event.inputs.version_to_release }}#" gradle.properties
```
bidetofevil, line 91: This is done on `release/<version>`

.github/workflows/pre-release-workflow.yml
bidetofevil: I need to change this so the branch name doesn't have the patch version - will update in a bit after we talk this through.

.github/workflows/pre-release-workflow.yml
```diff
@@ -100,7 +113,7 @@ jobs:
       - name: Swazzler Release
         run: |
           cd ./embrace-swazzler3
-          ./gradlew clean release -Dorg.gradle.parallel=false -Prelease.useAutomaticVersion=true -Prelease.releaseVersion=${{ github.event.inputs.current_version }} -Prelease.newVersion=${{ github.event.inputs.next_version }}-SNAPSHOT --stacktrace
+          ./gradlew clean release -Dorg.gradle.parallel=false -Prelease.useAutomaticVersion=true -Prelease.pushReleaseVersionBranch=release/${{ github.event.inputs.version_to_release }} -Prelease.releaseVersion=${{ github.event.inputs.version_to_release }} -Prelease.newVersion=${{ github.event.inputs.next_version }}-SNAPSHOT --stacktrace
```
bidetofevil, line 116: This is where I modified the command to create a release branch in Swazzler. I followed the docs for the plugin, but I don't know if it actually works, so we'll have to do by trying it out

cesarmax22 (approved):

PR #329, opened 2024-01-30
Combine & rename background activity + session services

fractalwrench:
## Goal

Combines the two services & renames them as `PayloadFactory`, as the class no longer holds any state and is just responsible for creating session objects.

## Testing


bidetofevil (approved):
HELL YEAH

PR #333, opened 2024-01-31
Avoid unnecessary serialisation of background activity

fractalwrench:
## Goal

Avoids serializing if there is back pressure on the background activity snapshotting & something has recently been serialized.


bidetofevil (approved):
Looks good.

PR #334, opened 2024-01-31
EMBR-627 Updated pre-release and release workflow

cesarmax22:
## Goal

Improved steps, checkouts, etc.

We are now not pushing version to be released to remote master; because we don't care about that. Remote master only cares about having the next version to be released (-SNAPSHOT).

We are now publishing and closing Swazzler sonatype repository from this job.

The idea behind this is that we release the complete embrace suite from this job. Simple and easy.

Swazzler has been updated to work with nexus-plugin as well.

Now the complete release process consists of:
1- run Pre-release workflow from SDK actions tab:
  -> it will assemble release artefact for the sdk and swazzler for version to be released
  -> it will run check on sdk and swazzler
  -> it will publish to sonatype repository (all sdk and swazzler artefacts)
  -> it will close sdk and swazzler sonatype repository
we are now ready to smoke test

once tested:
2- run Release workflow from SDK actions tab:
 -> it will update sdk and swazzler version in both gradle.properties and push it to remote master
 -> it will generate documentation for the sdk and push it to docs repository
 -> it will release all sdk and swazzler artefacts to maven central

Done. Embrace will be ready to be used by customers within a few minutes.

## Testing

Partial testing. All this will have to be tested on a real release, on next release.

## Release Notes

**WHAT**:<br>
**WHY**:<br>
**WHO**:<br>


.github/workflows/pre-release-workflow.yml
```diff
@@ -80,45 +75,22 @@ jobs:
...
           ref: master
-          path: ./embrace-swazzler3
           token: ${{ secrets.CD_GITHUB_TOKEN }}
 
-      - name: Swazzler Release
+      - name: Swazzler Pre-Release - Publish and Close repository
         run: |
-          cd ./embrace-swazzler3
-          ./gradlew clean release -Dorg.gradle.parallel=false -Prelease.useAutomaticVersion=true -Prelease.pushReleaseVersionBranch=release/${{ github.event.inputs.version_to_release }} -Prelease.releaseVersion=${{ github.event.inputs.version_to_release }}.0 -Prelease.newVersion=${{ github.event.inputs.next_version }}-SNAPSHOT --stacktrace
+          sed -i -r "s#version = ([^\']+)#version = ${{ github.event.inputs.current_version }}#" gradle.properties
```
bidetofevil, line 87: Third parameter of `sed` needs to start off with `s#version=` instead of `s#version =` (notice the removal of the space before the `=` sign. This is because the file doesn't have that space - this is likely why the version was screwed up in the swazzler.

Alternatively, we can just modify the swazzler properties file to add that space so this replacement will work. Or we can make the regular expression a bit more complicated to take care of both cases. Either way, we need to change.

cesarmax22:
@bidetofevil my intenetion was to make this work; once we're sure this is working we can apply your changes. but i'm open for the other option as well if you want to test everything together

bidetofevil (approved):
Looks good

PR #335, opened 2024-01-31
Improve logging of exceptions in unit tests

fractalwrench:
## Goal

Improve logging of exceptions in unit tests so the full stacktrace is outputted by [specifying `TestLogging`](https://docs.gradle.org/current/dsl/org.gradle.api.tasks.testing.logging.TestLogging.html) for the unit tests. This should help diagnose any flaky behavior on CI that currently we don't log full stacktraces for.


bidetofevil (approved):
nice!

PR #340, opened 2024-02-01
Add class and extension methods to make obtaining nanosecond time easier

bidetofevil:
## Goal

OTel deals in nanoseconds, but Embrace deals in milliseconds, so lets make it as easy as possible to convert the latter to the former

### DIFF
--- a/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/internal/clock/Clock.kt
+++ b/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/internal/clock/Clock.kt
@@ -1,6 +1,7 @@
 package io.embrace.android.embracesdk.internal.clock
 
 import io.embrace.android.embracesdk.annotation.InternalApi
+import java.util.concurrent.TimeUnit
 
 @InternalApi
 public fun interface Clock {
@@ -9,4 +10,15 @@ public fun interface Clock {
      * Returns the current milliseconds from epoch.
      */
     public fun now(): Long
+
+    /**
+     * Returns the current nanoseconds from epoch
+     */
+    public fun nowInNanos(): Long = now().millisToNanos()
 }
+
+/**
+ * Turns a number that specifies a millisecond value to nanoseconds
+ */
+@InternalApi
+public fun Long.millisToNanos(): Long = TimeUnit.MILLISECONDS.toNanos(this)
--- a/embrace-android-sdk/api/embrace-android-sdk.api
+++ b/embrace-android-sdk/api/embrace-android-sdk.api
@@ -219,6 +219,15 @@ public final class io/embrace/android/embracesdk/internal/InternalTracingApi$Def
 
 public abstract interface class io/embrace/android/embracesdk/internal/clock/Clock {
 	public abstract fun now ()J
+	public abstract fun nowInNanos ()J
+}
+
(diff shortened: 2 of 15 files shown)


bidetofevil:
* `master`
  * **PR #340** <a href="https://app.graphite.dev/github/pr/embrace-io/embrace-android-sdk/340?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/embrace-io/embrace-android-sdk/340?utm_source=stack-comment).

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/internal/clock/Clock.kt
```diff
@@ -9,4 +10,15 @@ public fun interface Clock {
      * Returns the current milliseconds from epoch.
      */
     public fun now(): Long
+
+    /**
+     * Returns the current nanoseconds from epoch
+     */
+    public fun nowInNanos(): Long = now().millisToNanos()
```
fractalwrench, line 17:
```suggestion
    public fun now(unit: TimeUnit = TimeUnit.MILLISECONDS): Long
```

I have a slight preference for this syntax as it avoids the proliferation of other functions at a later date such as millisToSeconds etc. Feel free to merge as-is if you disagree.

fractalwrench (approved):
LGTM

fnewberg:
> OTel deals in nanoseconds,

ah yes, another sign of OTel's backend roots....

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/internal/clock/Clock.kt
fnewberg: is the net effect of this that we add 6 zeros to the end of the time value?

fnewberg (approved):

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/internal/clock/Clock.kt
bidetofevil: > is the net effect of this that we add 6 zeros to the end of the time value?

Basically. Are you concerned about perf? If so I can write it in a way that makes the compiler inline this for us. But it's not going to make a huge difference since we don't call this that often.

fnewberg: no, it's fine. i just wanted to understand if i was missing something in terms of precision. i believe the other clock options we may use in the future, [this](https://developer.android.com/reference/android/os/SystemClock#currentGnssTimeClock()) and [this](https://developer.android.com/reference/android/os/SystemClock#currentNetworkTimeClock()), also seem to just have ms precision. the only thing that seems to have ns precision -- not to be confused with ns accuracy :) -- is [this](https://developer.android.com/reference/android/location/GnssClock#getTimeNanos())

bidetofevil: I don't know if the Android clock is even capable of that level of precision. It's certainly not useful for us. Microseconds I can get behind, but this is a bit of an overkill for our use case. Alas, that's the spec.

PR #341, opened 2024-02-01
Move event type to avoid compiler warning

fractalwrench:
## Goal

Moves a class to avoid a compiler warning (and also fix some tech debt by removing a useless wrapper class)


cesarmax22 (approved):
lgtm

PR #343, opened 2024-02-01
Resolve warnings about mockk library

fractalwrench:
## Goal

Rewrites some tests so that they don't trigger warnings about mockk illegally accessing internal JVM fields.


cesarmax22 (approved):
awesome, I like it.

PR #345, opened 2024-02-01
Bump AGP version to 7.4.2

fractalwrench:
## Goal

[AGP 7.4.2](https://developer.android.com/build/releases/past-releases/agp-7-4-0-release-notes) is the last 7.x version supported & is the highest we can go without bumping to Gradle 8 (which has some other roadblocks we need to clear first).

## Testing

Relied on CI + manually testing a publish to maven local.


priettt (approved):
Looks good! Should we test this with our min versions?

fractalwrench:
@priettt I've tested this with our android-test-suite app which has the min compile SDK version - would you suggest anything else?

priettt:
> @priettt I've tested this with our android-test-suite app which has the min compile SDK version - would you suggest anything else?

I'd test it with an app using our min AGP. Was it 4.1.1? I just realized that the bump was minimal, though, updating from 7.3.x to 7.4.x shouldn't break anything 😅

fractalwrench:
Successfully compiled with an app using AGP 4.0.0/Gradle 6.1.1 with all changes in this PR.

PR #346, opened 2024-02-01
Update detekt version

fractalwrench:
## Goal

Updates detekt to 1.23.0, which is the last version we can update to without updating kotlin to 1.9 first.


priettt (approved):
LGTM! Why are you removing clock? It wasn't used?

fractalwrench:
Yep, `clock` was unused and the latest detekt version warns about that.

PR #349, opened 2024-02-01
Use correct kotlin gradle plugin version

fractalwrench:
## Goal

Our build script setup means we have to specify versions in two places. The kotlin gradle plugin version was not updated under `buildSrc` so I changed it to use the correct one. This shouldn't affect our compile-time language support.


priettt:
I'd check with a test app using kotlin 1.4 just in case 😅

bidetofevil (approved):
I thought I had done this already? lol LGTM

PR #354, opened 2024-02-02
Create patch workflow

bidetofevil:
## Goal

Workflow to create a patch RC based on an existing release branch


cesarmax22 (approved):
lgtm

PR #365, opened 2024-02-05
Add initial DataSource interface and DataSourceState class

fractalwrench:
## Goal

Adds an initial interface for the `DataSource`. Currently this just contains the register/unregister calls for dealing with system API callbacks etc.

This also adds the `DataSourceState` class, which is responsible for managing the state of a single data source. This keeps track of whether the feature is currently enabled via config flags, and whether capture should be enabled for the current envelope type. This is done declaratively so that a `DataSource` implementation doesn't need to worry about tracking this state.

## Testing

Added unit test coverage.

### DIFF
--- a/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/arch/DataSourceState.kt
+++ b/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/arch/DataSourceState.kt
@@ -1,0 +1,68 @@
+package io.embrace.android.embracesdk.arch
+
+/**
+ * Holds the current state of the service. This class automatically handles changes in config
+ * that enable/disable the service, and creates new instances of the service as required.
+ * It also is capable of disabling the service if the [EnvelopeType] is not supported.
+ */
+internal class DataSourceState<T : DataSource<R>, R>(
+
+    /**
+     * Provides instances of services. A service must define an interface
+     * that extends [DataSource] for orchestration. This helps enforce testability
+     * by making it impossible to register data capture without defining a testable interface.
+     */
+    factory: () -> DataSource<R>,
+
+    /**
+     * Predicate that determines if the service should be enabled or not, via a config value.
+     */
+    private val configGate: () -> Boolean,
+
+    /**
+     * The type of envelope that contains the data.
+     */
+    private var currentEnvelope: EnvelopeType,
+
+    /**
+     * An envelope type where data capture should be disabled. For example,
+     * background activities capture a subset of sessions.
+     */
+    private val disabledEnvelopeType: EnvelopeType? = null
+) {
+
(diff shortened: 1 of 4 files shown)


embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/arch/DataSourceState.kt
```diff
@@ -0,0 +1,68 @@
...
+     * Provides instances of services. A service must define an interface
+     * that extends [DataSource] for orchestration. This helps enforce testability
+     * by making it impossible to register data capture without defining a testable interface.
+     */
+    factory: () -> DataSource<R>,
+
+    /**
+     * Predicate that determines if the service should be enabled or not, via a config value.
+     */
+    private val configGate: () -> Boolean,
```
bidetofevil, line 20: Are we able to pass in config settings via a similar interface? Perhaps the return value can be a generic object of some sort?

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/arch/DataSourceState.kt
```diff
@@ -0,0 +1,68 @@
...
+
+    /**
+     * The type of envelope that contains the data.
+     */
+    private var currentEnvelope: EnvelopeType,
```
bidetofevil, line 25: We should come up with a new term to the thing that is either a session or a background activity. We use the term envelope to describe the thing we stuff custom payloads into.

Calling it a Session makes the most sense, frankly, but we have it so ingrained that a background activity is not a session that this may be hard. That said, both smell like Sessions to me and probably to a lot of people in the industry - so is this when we bite the bullet and just call this that, and have two types - foreground and background?

This probably requires a bigger discussion than just this review.

bidetofevil (approved):
Looks good as a start. We can iterate on this if we see things we want to improve.

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/arch/DataSourceState.kt
fractalwrench: I'm not sure what this means? My current idea is that we would supply a boolean value from `ConfigService` as a parameter, e.g.:

```
class DataSourceState(
    configGate = { configService.dataCaptureBehavior.isAnrCaptureEnabled() }
...
```

Happy to iterate if you have other ideas

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/arch/DataSourceState.kt
fractalwrench: Agree - I will ask the android guild for some suggestions

PR #367, opened 2024-02-05
Add data capture orchestrator class

fractalwrench:
## Goal

Adds a `DataCaptureOrchestrator` class. This is responsible for relaying config state/envelope changes to all the data sources in the SDK.

## Testing

Added unit test coverage.

### DIFF
--- a/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/arch/DataCaptureOrchestrator.kt
+++ b/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/arch/DataCaptureOrchestrator.kt
@@ -1,0 +1,40 @@
+package io.embrace.android.embracesdk.arch
+
+import io.embrace.android.embracesdk.config.ConfigListener
+import io.embrace.android.embracesdk.config.ConfigService
+import io.embrace.android.embracesdk.logging.InternalEmbraceLogger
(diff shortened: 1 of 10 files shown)


embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/arch/DataCaptureOrchestrator.kt
```diff
@@ -0,0 +1,40 @@
...
+    }
+
+    /**
+     * Callback that is invoked when the envelope type changes.
+     */
+    fun onEnvelopeChange(envelopeType: EnvelopeType) {
+        dataSourceState.forEach { state ->
+            try {
+                // alter the envelope type - some data sources don't capture for background activities.
+                state.onEnvelopeTypeChange(envelopeType)
```
bidetofevil, line 34: Seems like a good place to transition the `CurrentSessionSpan` eh?

bidetofevil (approved):
LGTM

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/arch/DataCaptureOrchestrator.kt
fractalwrench: Yep - we should ultimately do that either here or in the `SessionOrchestrator`

PR #371, opened 2024-02-07
Populate DataCaptureOrchestrator with data source states

fractalwrench:
## Goal

The `DataCaptureOrchestrator` is responsible for propagating envelope/config changes to the data sources. Previously it was just supplied an empty list. This changeset creates a new module named `DataSourceModule` that contains a list of data sources that are used by the Embrace SDK, and hooks it up to the orchestrator.

I have included an empty placeholder implementation to demonstrate what this would look like. In terms of scaling this approach to the many data sources the SDK probably uses, it should be possible to group ~10 data sources into modules based on feature & concatenate the lists.

To alleviate the concern that developers might forget to add a property to the list I created a delegate that will automagically add values to the collection.

## Testing

Updated unit test coverage.


bidetofevil (approved):
Nice. I really like this! We can implicitly control ordering in case there are cascading dependencies where flushing one data source can lead to the modification of another data source. We may want to make that more explicit if we ever to that, but this should make coordination a lot simpler to understand.

PR #372, opened 2024-02-07
Rename envelope type to session type

fractalwrench:
## Goal

Renames envelope type to session type as discussed internally.


bidetofevil (approved):
LGTM

PR #373, opened 2024-02-07
Extract custom breadcrumb capture to separate class

fractalwrench:
## Goal

Extracts custom breadcrumb capture to a separate class that implements the `DataCaptureService` interface. This is an intermediate step - I'd like to refactor this to be the first example of a `DataSource` implementation, but that will come in a separate PR.

## Testing

Relied on existing unit test coverage.


bidetofevil (approved):
LGTM

PR #377, opened 2024-02-07
addSpanExporter api

nelsitoPuglisi:
## Goal

Adds a public api that allows a client app to add a SpanExporter at runtime, through a OpenTelemetryConfiguration class.

## Testing

Added  instrumentation tests to `SpanTests` class.

## Release Notes

New `addSpanExporter` public api, to add many SpanExporters to the OTel SDK.

**WHAT**:We have created an api to add SpanExporters to the OTel SDK
**WHY**:To allow customers to export their data to any OTel Collector.
**WHO**:Any customer with an OTel collector.


embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/injection/InitModule.kt
```diff
@@ -48,6 +48,11 @@ internal interface InitModule {
      */
     val tracer: Tracer
 
+    /**
+     * Composite exporter that can be used to add multiple SpanExporters at runtime
+     */
+    val compositeSpanExporter: CompositeSpanExporter
```
bidetofevil, line 54: Instead of directly exposing this, what do you think about persisting this a `List` somewhere so we can add them later? That way, we won't have to expose the composite exporter outside of `OpenTelemetrySdk`, making it an implementation detail that we can change later, i.e. change to use the default composite once we get this initializing later.

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/Embrace.java
```diff
@@ -515,6 +516,24 @@ public boolean recordCompletedSpan(@NonNull String name, long startTimeNanos, lo
         return false;
     }
 
+    /**
+     * Adds one or many [SpanExporter] to the tracer.
+     *
+     * @param spanExporter the span exporters to add
+     */
+    public void addSpanExporter(SpanExporter... spanExporter) {
```
bidetofevil, line 524: Lets define this in this in the `EmbraceApi`? Everything public should be in one of the interfaces.

And what do you think about just taking one exporter in the parameter? It's unlikely that ppl will want more than one, and if they do, they can create their own composite or just call this method again. This reduces the complexity and surface area of the API which would be preferable.

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/Embrace.java
```diff
@@ -515,6 +516,15 @@ public boolean recordCompletedSpan(@NonNull String name, long startTimeNanos, lo
         return false;
     }
 
+    /**
+     * Adds one [SpanExporter] to the tracer.
+     *
+     * @param spanExporter the span exporter to add
+     */
+    public void addSpanExporter(SpanExporter spanExporter) {
```
bidetofevil, line 524: `@Override`?

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/injection/OpenTelemetryModule.kt
```diff
@@ -13,11 +13,19 @@ import io.embrace.android.embracesdk.internal.spans.SpansSink
...
 /**
  * Module that instantiates various OpenTelemetry related components
  */
 internal interface OpenTelemetryModule {
+    /**
+     * Adds one [SpanExporter] to the tracer.
+     *
+     * @param spanExporter the span exporter to add
+     */
+    fun addSpanExporter(spanExporter: SpanExporter)
```
bidetofevil, line 27: The module shouldn't carry state - it's just there to initialize services and hang objects private to those services off of. I would probably create a new object/service that we can write all of the custom stuff we need to initialize the OTel SDK with, like the SpanExporter, the LogExporter eventually, and additional attributes.

For now, just add support for SpanExporter and we can add to it when we add new features?

embrace-android-sdk/src/integrationTest/kotlin/io/embrace/android/embracesdk/testcases/SpanTest.kt
```diff
@@ -0,0 +1,70 @@
...
+    @JvmField
+    val testRule: IntegrationTestRule = IntegrationTestRule(
+        harnessSupplier = {
+            IntegrationTestRule.newHarness(startImmediately = false)
+        }
+    )
+
+    @Before
+    fun before() {
+        ApkToolsConfig.IS_SDK_DISABLED = false
```
bidetofevil, line 30: You can remove this, I think. This should only be set by tests that need to modify it, and this shouldn't

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/EmbraceImpl.java
```diff
@@ -1459,4 +1460,8 @@ private boolean checkSdkStarted(@NonNull String action, boolean logPublicApiUsag
         }
         return isStarted;
     }
+
+    public void addSpanExporter(SpanExporter spanExporter) {
```
bidetofevil, line 1464: @NonNull annotation?

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/opentelemetry/OpenTelemetryConfiguration.kt
```diff
@@ -0,0 +1,28 @@
...
+import io.embrace.android.embracesdk.internal.spans.SpansSink
+import io.opentelemetry.sdk.trace.SpanProcessor
+import io.opentelemetry.sdk.trace.export.SpanExporter
+
+internal class OpenTelemetryConfiguration(
+    spansSink: SpansSink
+) {
+    val serviceName = BuildConfig.LIBRARY_PACKAGE_NAME
+    val serviceVersion = BuildConfig.VERSION_NAME
+    private var exporters: MutableList<SpanExporter>
```
bidetofevil, line 15: Why not make this `val` and initialize it inline here?

bidetofevil (approved):
Looks good in general. We can change the `OpenTelemetryConfiguration` later when we have more resources to add so it return a list of Pairs to be added as resources, as well as adding an input so we can configure this programmatically.

PR #391, opened 2024-02-10
Add getSpan to public API

bidetofevil:
## Goal

Expose getSpan with a spanId in the public API

PR #392, opened 2024-02-10
Expose SDK clock time in through internal tracing API

bidetofevil:
## Goal

Expose a method in the internal tracing API that returns the locked SDK clock time

### DIFF
--- a/embrace-android-sdk/api/embrace-android-sdk.api
+++ b/embrace-android-sdk/api/embrace-android-sdk.api
@@ -37,6 +37,7 @@ public final class io/embrace/android/embracesdk/Embrace : io/embrace/android/em
 	public fun getInternalInterface ()Lio/embrace/android/embracesdk/internal/EmbraceInternalInterface;
 	public fun getLastRunEndState ()Lio/embrace/android/embracesdk/Embrace$LastRunEndState;
 	public fun getReactNativeInternalInterface ()Lio/embrace/android/embracesdk/ReactNativeInternalInterface;
+	public fun getSdkClockTimeNanos ()J
 	public fun getSessionProperties ()Ljava/util/Map;
 	public fun getSpan (Ljava/lang/String;)Lio/embrace/android/embracesdk/spans/EmbraceSpan;
 	public fun getTraceIdHeader ()Ljava/lang/String;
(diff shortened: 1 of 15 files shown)


bidetofevil:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/embrace-io/embrace-android-sdk/392?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* `master`
  * **PR #391** <a href="https://app.graphite.dev/github/pr/embrace-io/embrace-android-sdk/391?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
    * **PR #392** <a href="https://app.graphite.dev/github/pr/embrace-io/embrace-android-sdk/392?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/embrace-io/embrace-android-sdk/392?utm_source=stack-comment).

fractalwrench (commented):
I'm happy with the internal refactoring but am not fully convinced that exposing this API is helpful. `recordCompletedSpan` will have already existed, so getting time from `getSdkClockTimeNanos` requires resolving any difference between the customer's timestamp & our timestamp. Customers are also very unlikely to account for device time changes etc.

Has a customer requested this API? If not, I'd suggest ditching the public API. If so, I'd suggest allowing a customer to supply their own implementation of the `Clock` interface for use by the Embrace SDK instead.

bidetofevil:
Not explicitly, no, but I figured they would want this in the future, and for cases where they would use this after app startup.

Lets talk about this further in Slack - but for now, I'll remove the public API and stick this in the internal interface. There's already one that returns the time in millis... so maybe I'll just merge them

fractalwrench (approved):
LGTM on the basis that the public API part of these changes should be removed from the changeset

bidetofevil:
Discussed this with Jamie. I am also now convinced that we shouldn't expose this, primarily so users won't have to worry about the complexity that this brings.

We'll find another way to reconcile the difference - even if it's just knowing that internally, the times maybe inconsistent if the start/end times are provided by the API user and not the SDK. We can track this internally too with an attribute. I'll post something in slack to further this discussion.

bidetofevil:
### Merge activity

* **Feb 12, 2:44 PM EST**: @bidetofevil started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/embrace-io/embrace-android-sdk/392).

PR #398, opened 2024-02-12
Add destinations to data source interface

fractalwrench:
## Goal

Alters the `DataSource` interface so that it has 3 potential destinations where it can write data: spans, span events, and logs.

This ties each `DataSource` to capturing only one type of data directly (Span/SpanEvent/Logs). IMO this is preferable and in edge-cases it should be possible to create multiple implementations.

## Testing

Added unit test coverage.

### DIFF
--- a/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/arch/DataSource.kt
+++ b/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/arch/DataSource.kt
@@ -1,10 +1,24 @@
 package io.embrace.android.embracesdk.arch
 
+import io.embrace.android.embracesdk.internal.spans.CurrentSessionSpan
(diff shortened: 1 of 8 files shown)


embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/arch/DataSource.kt
```diff
@@ -16,3 +30,24 @@ internal interface DataSource {
...
+
+/**
+ * A [DataSource] that adds a new log to the log service.
+ */
+internal interface LogDataSource : DataSource<LogService>
+
+/**
+ * Placeholder type for the log service.
+ */
+internal interface LogService
```
fractalwrench, line 53: This is just a placeholder type for now, as we haven't defined the contract for what log capture should look like.

bidetofevil (approved):
Lets give this a go to see if we need more specificity

PR #399, opened 2024-02-12
Rename functions on data source interface

fractalwrench:
## Goal

Renames functions on the `DataSource` interface to `enableDataCapture` and `disableDataCapture`. This is to better reflect their true function.

In this scheme the `DataSourceState` checks whether a single `DataSource` should be enabled, by calculating the current values of the remote config, and the current payload type (session vs background). It then calls `enableDataCapture` or `disableDataCapture` if required.

The `DataSource` implementation should register listeners & start capturing data after `enableDataCapture` is called, and do the opposite when `disableDataCapture` is called. I've added a warning that implementations should _not_ attempt to track the enabled state.


bidetofevil (approved):
LGTM

PR #405, opened 2024-02-13
Use new m1 mac runners for CI

fractalwrench:
## Goal

Updates the CI runner to [use M1 macs](https://github.blog/changelog/2024-01-30-github-actions-introducing-the-new-m1-macos-runner-available-to-open-source/). This should be faster than the existing runner & is free for public repos.

### DIFF
--- a/.github/workflows/ci-gradle.yml
+++ b/.github/workflows/ci-gradle.yml
@@ -10,12 +10,11 @@ on:
 
 jobs:
   gradle-test:
-    runs-on: ubuntu-latest
+    runs-on: macos-14
     timeout-minutes: 30
     strategy:
       matrix:
         jdk-version: ["17"]
-        ndk-version: ["21.4.7075529"]
     steps:
       - name: Checkout Branch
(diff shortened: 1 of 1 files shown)


.github/workflows/functional-tests.yml
```diff
@@ -16,12 +16,12 @@ env:
...
   test:
-    runs-on: macos-latest
+    runs-on: macos-14
     timeout-minutes: 45
     strategy:
       matrix:
         jdk-version: ["17"]
-        ndk-version: ["21.4.7075529"]
+#        ndk-version: ["21.4.7075529"]
         api-level: [29]
```
bidetofevil, line 25: Not for this PR, but should be bump this to the latest? And also test on the minimum?

bidetofevil (approved):
LGTM

.github/workflows/functional-tests.yml
fractalwrench: Emulators beyond API 29 aren't supported by the GH action we use for running emulators yet

PR #406, opened 2024-02-13
[POC] Convert custom breadcrumb capture to use DataSource

fractalwrench:
## Goal

Converts the capture of custom breadcrumbs to use the new `DataSource` interface. The `CustomBreadcrumbDataSource` is the most important class to look at - it shows how `DataSource` can be used to write a `SpanEvent`.

This should not be merged yet as the payload schema is not agreed upon/supported by the backend, but it's useful as an illustration of how this should work.


embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/capture/crumbs/CustomBreadcrumbDataSource.kt
```diff
@@ -1,32 +1,41 @@
...
-        } catch (ex: Exception) {
-            logger.logError("Failed to log custom breadcrumb with message $message", ex)
-        }
+        captureData(
+            inputValidation = {
+                message.isNotEmpty()
+            },
+            captureAction = {
+                val crumb = CustomBreadcrumb(message, timestamp)
+                addEvent(crumb.toSpanEventData())
```
bidetofevil, line 31: I wonder if we can make the `toSpanEventData` call in the `addEvent` method - so the datasource will just need to pass in a breadcrumb that can implicitly convert itself and have that conversion down at a lower layer. 

Are we able to create an interface that covers that so all POJOs that will end up becoming a SpanEvent will just implement that?

bidetofevil (commented):
I like this. We should figure out with the backend folks how we can get this into production incrementally so we don't have to do a big bang merge to switch out the services' implementations and the payloads in one fell swoop.

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/capture/crumbs/CustomBreadcrumbDataSource.kt
fractalwrench: @bidetofevil our hands are tied a little here as the POJOs will be generated from the OpenAPI schema & it's non-trivial to have them directly implement an interface like that. I have however refactored the code so that a mapper function can be passed in, which is probably the nearest we can get.

bidetofevil (approved):
Looks good! I think by the very nature of what we're trying to achieve, there is going to be some amount of complexity so that this stuff can be mostly generalized rather than bespoke. It would be great if we can look into adding more to the boilerplate and providing better default implementations so that most of the complexity can be hidden from the writer of a DataSource - unless the default behaviour doesn't work for them.

fractalwrench:
@bidetofevil can I get a sanity check that the `SpanSanitizer` approach is the way we want to go with spans? I tried using a `SpanProcessor` but figured it would be simpler to sanitize the payload at this point as it avoids having to decorate OTel types.

PR #409, opened 2024-02-13
Store logcat logs as artefact for CI test failures

fractalwrench:
## Goal

When the functional tests fail this alters the workflow so that the logcat output is stored as an artefact. This should make it easier to debug what goes wrong, although we might need to add some additional logging of our own.

I added some extra logging & also increased the timeout for when an API call is expected to fail, as 1s seems a little low for a call to fail, be serialized to disk, and then flushed again.


bidetofevil (approved):
LGTM

PR #410, opened 2024-02-13
Remove failing functional tests

fractalwrench:
## Goal

Removes some functional test cases that have started failing. This seems to be infrastructure related in some respect as I retried a passing job from 2 days ago & it fails now.

The test code that asserts a failed request is stored on disk seems somewhat brittle, so I've decided to try removing those test cases as they're still covered by unit tests elsewhere.


bidetofevil (approved):

fractalwrench:
I will add a KTLO task to the backlog to restore these

PR #413, opened 2024-02-14
Model session span as Embrace Span

fractalwrench:
## Goal

Models the session span as an `EmbraceSpan` rather than a `Span`. This will allow us to use the `EmbraceSpan` internally when capturing span events + attributes from data sources.

This was mostly a 1:1 conversion of switching out function names, with the exception of adding an optional param to `EmbraceSpanImpl` that prevents an undesired attribute from being added to the span. I also purposely avoided tracking the session span in the `SpansRepository` as I assume this might have undesirable consequences.

## Testing

I relied on existing test coverage.

### DIFF
--- a/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/internal/spans/CurrentSessionSpanImpl.kt
+++ b/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/internal/spans/CurrentSessionSpanImpl.kt
@@ -3,8 +3,6 @@ package io.embrace.android.embracesdk.internal.spans
 import io.embrace.android.embracesdk.internal.utils.Provider
 import io.embrace.android.embracesdk.spans.EmbraceSpan
 import io.embrace.android.embracesdk.telemetry.TelemetryService
-import io.opentelemetry.api.common.Attributes
-import io.opentelemetry.api.trace.Span
 import io.opentelemetry.api.trace.Tracer
 import io.opentelemetry.sdk.common.Clock
 import java.util.concurrent.TimeUnit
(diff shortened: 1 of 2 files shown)


embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/internal/spans/CurrentSessionSpanImpl.kt
```diff
@@ -39,7 +37,7 @@ internal class CurrentSessionSpanImpl(
     /**
      * The span that models the lifetime of the current session or background activity
      */
-    private val sessionSpan: AtomicReference<Span?> = AtomicReference(null)
+    private val sessionSpan: AtomicReference<EmbraceSpan?> = AtomicReference(null)
```
bidetofevil, line 40: Ah nice. I didn't even notice this since this was written so long ago! It made a bit more sense when it was part of SpansServiceImpl, but now that this is a separate object, it sticks out lol. Cool cool.

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/internal/spans/EmbraceSpanImpl.kt
```diff
@@ -14,11 +14,14 @@ import java.util.concurrent.atomic.AtomicReference
 internal class EmbraceSpanImpl(
     private val spanBuilder: SpanBuilder,
     override val parent: EmbraceSpan? = null,
-    private val spansRepository: SpansRepository? = null
+    private val spansRepository: SpansRepository? = null,
+    sessionSpan: Boolean = false
```
bidetofevil, line 18: This is fine for now, but I'm wondering if we should model this more explicitly. It's probably OK for now but I don't want to abuse this pattern if there are other specialized spans in the future that we want to wrap similarly.

bidetofevil (approved):
LGTM

PR #414, opened 2024-02-14
Add functions for writing to the session span

fractalwrench:
## Goal

Adds functions that are capable of writing events + attributes onto the session span. This builds on #413.

I have chosen a fairly straightfoward approach of adding two functions: one that writes an event, and another that writes an attribute. These will be exposed via the `SessionSpanWriter` interface. I've chosen this instead of exposing the `CurrentSessionSpan` directly to the `DataSource` because as far as the data source is concerned, the rest of the functions on `CurrentSessionSpan` are an implementation detail.

## Testing

Added unit test coverage.


bidetofevil (approved):
Noice!

PR #415, opened 2024-02-14
No-op data capture functions by default

fractalwrench:
## Goal

Makes the `enableDataCapture/disableDataCapture` functions no-op by default, as there is a lot of telemetry that we capture that isn't possible to disable (e.g. breadcrumbs submitted by developers). It is still possible for data sources to override this behavior by overriding the functions as required.

## Testing


lucaslabari (approved):

bidetofevil (approved):
LGTM

PR #416, opened 2024-02-14
Use tracing api rather than embrace tracer

fractalwrench:
## Goal

Use the `TracingApi` rather than the `EmbraceTracer` directly, as we don't get anything extra from using the concrete implementation.


lucaslabari (approved):

bidetofevil (approved):
Ah nice! I was going to make this switch in other places now that `getSpan()` is also in the interface. I'll make sure the other places are fixed too.

PR #417, opened 2024-02-14
Add function that resets data capture limits

fractalwrench:
## Goal

Adds a function that resets data capture limits for `DataSource` implementations. By default this is a no-op. The limit is reset whenever there is a new session.

## Testing

Added unit test coverage.

### DIFF
--- a/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/arch/DataSource.kt
+++ b/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/arch/DataSource.kt
@@ -36,6 +36,11 @@ internal interface DataSource<T> {
      * You should NOT attempt to track state within the [DataSource] with a boolean flag.
      */
     fun disableDataCapture()
+
+    /**
+     * Resets any data capture limits since the last time [enableDataCapture] was called.
+     */
+    fun resetDataCaptureLimits()
 }
 
 /**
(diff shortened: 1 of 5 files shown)


embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/arch/DataSource.kt
```diff
@@ -36,6 +36,11 @@ internal interface DataSource<T> {
      * You should NOT attempt to track state within the [DataSource] with a boolean flag.
      */
     fun disableDataCapture()
+
+    /**
+     * Resets any data capture limits since the last time [enableDataCapture] was called.
+     */
+    fun resetDataCaptureLimits()
```
lucaslabari, line 43: - When are these limits being set? 
- Are they just `limits` or a broader concept like `attributes`?

fractalwrench: This is a limit that will be applied to data capture, e.g. custom breadcrumbs has a limit of 100. I'd envisioned it being set when `enableDataCapture` is called.

However, perhaps it's worth me looking at creating a lambda or something that gets the limit? That might allow the changeset to be more automatic than it currently is.

bidetofevil: I assume this will be called when we transition between sessions? Lets keep an eye on this and if there's anything else that always need to be done besides limit resetting, we can rename this and do that here too. This is good for now though if there's nothing else!

bidetofevil (approved):
LGTM

PR #418, opened 2024-02-14
Add serializable representation for OTel logs

leandro-godon:
## Goal

Create an Embrace representation of the `LogRecordData` class from OTel.


embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/internal/logs/EmbraceLogRecordData.kt
```diff
@@ -0,0 +1,43 @@
...
+    val severityText: String?,
+
+    @Json(name = "body")
+    val body: EmbraceBody,
+
+    @Json(name = "attributes")
+    val attributes: Map<String, String> = emptyMap()
+) {
+
+    internal constructor(logRecordData: LogRecordData) : this(
```
fractalwrench, line 34: nit: would be good to add a test case that converts this type to `EmbraceLogRecordData`

fractalwrench (approved):
LGTM. Will Fran's OpenAPI work eventually generate classes like this?

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/internal/logs/EmbraceLogRecordData.kt
```diff
@@ -0,0 +1,43 @@
...
+    val severityNumber: Int,
+
+    @Json(name = "severity_text")
+    val severityText: String?,
+
+    @Json(name = "body")
+    val body: EmbraceBody,
+
+    @Json(name = "attributes")
+    val attributes: Map<String, String> = emptyMap()
```
bidetofevil, line 31: This should actually be a List of Pair<String, Any>. The whole list thing is an OTLP decision for performance reasons, and we are moving to allowing types other than Strings as attribute values. The v2 endpoints will support this, so this is something the `EmbraceSpanData` will be be changed to do that as well

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/internal/logs/EmbraceBody.kt
```diff
@@ -0,0 +1,13 @@
+package io.embrace.android.embracesdk.internal.logs
+
+import com.squareup.moshi.Json
+import com.squareup.moshi.JsonClass
+
+/**
+ * Serializable representation of [EmbraceBody]
+ */
+@JsonClass(generateAdapter = true)
+internal data class EmbraceBody(
```
lucaslabari, line 10: The name of this class might be confused with body types in other objects. 
We can consider renaming to `EmbraceLogBody`, or you expect this to be reused in other types?

lucaslabari (approved):
LGTM

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/internal/logs/EmbraceBody.kt
leandro-godon: I used this name because the OTel class is named only `Body`, but it's true that the name can be confusing. It is not used outside of the `opentelemetry.sdk.logs` package, I'll rename it.

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/internal/logs/EmbraceLogRecordData.kt
leandro-godon: Changed to `List<Pair<String, Any>>`.

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/internal/logs/EmbraceLogRecordData.kt
leandro-godon: Added a test case for the constructor.

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/internal/logs/EmbraceLogRecordData.kt
leandro-godon: I've rolled back the change from `Map` to `List<Pair>` because we can not use kotlin.Pair with a generated Moshi class adapter; I think changing that is outside the scope of this PR and part of the work Fran is doing.

PR #420, opened 2024-02-15
Create mapping interface for span events & attributes

fractalwrench:
## Goal

Creates an interface for converting POJOs to span events/attributes. I've created a simple Kotlin class that represents the values required to create a span/event.

## Testing

Updated existing unit test coverage.

### DIFF
--- a/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/arch/SpanAttributeMapper.kt
+++ b/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/arch/SpanAttributeMapper.kt
@@ -1,0 +1,8 @@
+package io.embrace.android.embracesdk.arch
+
+/**
+ * Converts an object of type T to a [SpanAttributeData]
+ */
+internal fun interface SpanAttributeMapper<T> {
+    fun T.toSpanAttributeData(): SpanAttributeData
+}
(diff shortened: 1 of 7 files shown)


embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/arch/SpanAttributeMapper.kt
```diff
@@ -0,0 +1,8 @@
+package io.embrace.android.embracesdk.arch
+
+/**
+ * Converts an object of type T to a [SpanAttributeData]
+ */
+internal fun interface SpanAttributeMapper<T> {
```
lucaslabari, line 6: These two mappers are not being used. Are they gonna be used in another PR?

lucaslabari (approved):
LGTM. Just left a question about the mappers that are not yet being used.

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/arch/SpanAttributeMapper.kt
fractalwrench: Yep - I've got an example of how they work in #406.

bidetofevil (approved):
Really nice!

PR #421, opened 2024-02-15
Add limit strategy interface

fractalwrench:
## Goal

Adds a `LimitStrategy` interface that is a required parameter of the `DataSource`. This will force us to think about how we want to limit data capture for each data source & helps enforce data limits in a standardised way.

I have implemented a simple option of a limit that captures everything up to a certain limit, but there are other strategies we use throughout our SDK that can be implemented via this interface. The limit is reset at the start of each new session boundary.

An example of how this works can be found in #406 

## Testing

Added unit test coverage.

### DIFF
--- a/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/arch/limits/UpToLimitStrategy.kt
+++ b/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/arch/limits/UpToLimitStrategy.kt
@@ -1,0 +1,34 @@
+package io.embrace.android.embracesdk.arch.limits
+
+import io.embrace.android.embracesdk.internal.utils.Provider
+import io.embrace.android.embracesdk.logging.InternalEmbraceLogger
+import io.embrace.android.embracesdk.logging.InternalStaticEmbraceLogger
+
+/**
+ * Allows capturing data up until a limit, then stops capturing.
+ */
+internal class UpToLimitStrategy(
+    private val limitProvider: Provider<Int>,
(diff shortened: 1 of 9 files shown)


embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/arch/limits/UpToLimitStrategy.kt
```diff
@@ -0,0 +1,29 @@
+package io.embrace.android.embracesdk.arch.limits
+
+import io.embrace.android.embracesdk.internal.utils.Provider
+import io.embrace.android.embracesdk.logging.InternalEmbraceLogger
+import io.embrace.android.embracesdk.logging.InternalStaticEmbraceLogger
+
+/**
+ * Allows capturing data up until a limit, then stops capturing.
+ */
+internal class UpToLimitStrategy(
```
bidetofevil, line 10: Do we at least need threadsafety, if not synchronization, at this level?

bidetofevil (approved):
LGTM. Comment about whether we need to ensure threadsafety in the `count`. Feel free to merge if a single instance of this is ever going to be accessed on one thread

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/arch/limits/UpToLimitStrategy.kt
fractalwrench: Added synchronisation around a lock.

PR #422, opened 2024-02-15
Add input validation action

fractalwrench:
## Goal

Adds an `inputValidation` action that returns true if inputs are valid for capturing data, and otherwise returns false. This is a mandatory parameter in `captureData` so should force us to think about how we want to sanitise our inputs, and provide a single codepath for any logic where inputs are invalid.

An example of how this looks in practice can be found in #406.

## Testing

Added unit test coverage.


bidetofevil (approved):
Nice

PR #423, opened 2024-02-15
Add Log Sink

leandro-godon:
## Goal

First implementation of a Sink for OTel logs, based on the SpanSink implementation.


embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/internal/logs/LogSink.kt
```diff
@@ -0,0 +1,27 @@
...
+
+    /**
+     * Stores Logs. Implementations must support concurrent invocations.
+     */
+    fun storeLogs(logs: List<LogRecordData>): CompletableResultCode
+
+    /**
+     * Returns the list of currently stored Logs.
+     */
+    fun logs(): List<EmbraceLogRecordData>
```
fractalwrench, line 20: nit: could this be named `completedLogs`, similar to how `SpanSink` works?

fractalwrench (approved):
LGTM

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/internal/logs/LogSinkImpl.kt
```diff
@@ -0,0 +1,34 @@
...
+
+import io.opentelemetry.sdk.common.CompletableResultCode
+import io.opentelemetry.sdk.logs.data.LogRecordData
+
+internal class LogSinkImpl : LogSink {
+    private val storedLogs: MutableList<EmbraceLogRecordData> = mutableListOf()
+
+    override fun storeLogs(logs: List<LogRecordData>): CompletableResultCode {
+        try {
+            synchronized(storedLogs) {
```
bidetofevil, line 11: This works for now. After we ship, both this and `SpansSink` should figure out a common buffering strategy so we don't have to lock when we are writing to the sink.

embrace-android-sdk/src/test/java/io/embrace/android/embracesdk/internal/logs/LogSinkImpltest.kt
```diff
@@ -0,0 +1,71 @@
...
+import io.opentelemetry.api.logs.Severity
+import io.opentelemetry.sdk.common.CompletableResultCode
+import io.opentelemetry.sdk.logs.data.Body
+import io.opentelemetry.sdk.logs.data.LogRecordData
+import org.junit.Assert.assertEquals
+import org.junit.Assert.assertSame
+import org.junit.Before
+import org.junit.Test
+
+internal class LogSinkImpltest {
```
bidetofevil, line 17: `LogSinkSimplTest`? (capital T)

embrace-android-sdk/src/test/java/io/embrace/android/embracesdk/internal/logs/LogSinkImpltest.kt
```diff
@@ -0,0 +1,71 @@
...
+
+        val flushedLogs = logSink.flushLogs()
+        assertEquals(2, flushedLogs.size)
+        repeat(2) {
+            assertSame(snapshot[it], flushedLogs[it])
+        }
+        assertEquals(0, logSink.completedLogs().size)
+    }
+
+    private fun getLogRecordData(): LogRecordData {
```
bidetofevil, line 54: Either in this PR or another one, can you create a `FakeLogRecordData` that based on the interface so you can use it instead of creating mock? We prefer this so it's easier to work with test data when we use real objects that we can instantiate and won't have to worry about the mocks being always right.

bidetofevil (approved):
LGTM. I have a suggestion to create a fake instead of mocking but that can be done in a separate PR

lucaslabari (approved):
LGTM

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/internal/logs/LogSink.kt
leandro-godon: Renamed.

embrace-android-sdk/src/test/java/io/embrace/android/embracesdk/internal/logs/LogSinkImpltest.kt
leandro-godon: Good catch!

embrace-android-sdk/src/test/java/io/embrace/android/embracesdk/internal/logs/LogSinkImpltest.kt
leandro-godon: `FakeLogRecordData` created.

PR #432, opened 2024-02-16
Add envelope for new endpoints.

lucaslabari:
## Goal

- Added `Envelope` data class and the related classes needed to produce a payload for Logs. 
- This might be replaced once the OpenApi work is done. 

## Testing

They are all data classes that don't need to be tested until we start using them.


fractalwrench (approved):
LGTM

PR #433, opened 2024-02-16
Add new endpoint for OTel logs.

lucaslabari:
## Goal

- Added new endpoint `${BASE_URL}/v2/log/logs` for sending OTel logs. 

- sendLogs() method is not yet finished. This will be added in another PR. The intention of this PR is to add the endpoint and build the new URL for logs.

## Testing

Added and updated unit tests.


fractalwrench (approved):
LGTM

PR #434, opened 2024-02-16
Add OTel Log Orchestrator

leandro-godon:
## Goal

Creates a log orchestrator that flushes the logs to the delivery service if one of these conditions is met:

. There are at least MAX_LOGS_PER_BATCH in the sink
. MAX_INACTIVITY_TIME has passed since the last log was stored
. MAX_BATCH_TIME has passed since the oldest log in the sink was stored

Extracting the batching parameters to an external configuration will be done in a separate PR.


embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/internal/logs/LogOrchestrator.kt
```diff
@@ -0,0 +1,72 @@
...
+        if (storedLogs.isNotEmpty()) {
+            // TBD: Send logs to DeliveryService
+        }
+    }
+
+    private fun scheduleInactivityCheck() {
+        inactivityFuture?.cancel(false)
+        inactivityFuture = logOrchestratorScheduledWorker.schedule<Unit>(
+            {
+                if (clock.now() - lastLogTime > MAX_INACTIVITY_TIME) {
```
fractalwrench, line 49: nit: consider moving this check to `sendLogs`. That would make this function more consistent with how `scheduleBatchTimeCheck` behaves

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/internal/logs/LogOrchestrator.kt
```diff
@@ -0,0 +1,72 @@
...
+        } else {
+            scheduleInactivityCheck()
+            if (batchTimeFuture == null) {
+                scheduleBatchTimeCheck()
+            }
+        }
+    }
+
+    @VisibleForTesting
+    fun sendLogs() {
```
fractalwrench, line 34: I'd suggest adding a `synchronized` block here on an `Any()` lock property. This would avoid logs getting sent multiple times if `scheduleInactivityCheck` and `scheduleBatchTimeCheck` are called at the same time

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/internal/logs/LogOrchestrator.kt
```diff
@@ -0,0 +1,72 @@
...
+        inactivityFuture?.cancel(false)
+        batchTimeFuture?.cancel(false)
+
+        val storedLogs = sink.flushLogs()
+
+        if (storedLogs.isNotEmpty()) {
+            // TBD: Send logs to DeliveryService
```
fractalwrench, line 41: You can probably create the API that Lucas will use on the `DeliveryService` interface & leave it as a no-op, if you want to get this merged quicker

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/internal/logs/LogSinkImpl.kt
```diff
@@ -5,12 +5,14 @@ import io.opentelemetry.sdk.logs.data.LogRecordData
 
 internal class LogSinkImpl : LogSink {
     private val storedLogs: MutableList<EmbraceLogRecordData> = mutableListOf()
+    private var onLogsStored: (() -> Unit)? = null
```
fractalwrench, line 8:
```suggestion
    private val onLogsStored: (() -> Unit)? = null
```

embrace-android-sdk/src/test/java/io/embrace/android/embracesdk/internal/logs/LogOrchestratorTest.kt
```diff
@@ -0,0 +1,110 @@
...
+    }
+
+    @Test
+    fun `send a batch of logs when the batch max size is reached`() {
+        val logs = mutableListOf<EmbraceLogRecordData>()
+        every { logSink.completedLogs() } returns logs
+
+        // Fill the sink with max batch size - 1 logs
+        repeat(49) {
+            logs.add(mockk())
```
fractalwrench, line 53: We can probably get a similar unit test without using mocks & `verify`, by either using real objects, or creating fake objects. These are substantially easier to refactor compared to mocks & are less likely to break when something changes.

fractalwrench (commented):
Excited to see this! Left a few comments inline

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/internal/logs/LogOrchestrator.kt
leandro-godon: Since `sink.flushLogs()` is `synchronized`, logs will be sent only once even if `sendLogs()` is called concurrently. I can add a comment here (and maybe a unit test) to ensure that it can not be accidentally refactored away. Or sync the whole method if we still think that's the best way to go.

fractalwrench: Awesome. A comment explaining that should be more than enough 👍

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/internal/logs/LogSinkImpl.kt
```diff
@@ -5,12 +5,14 @@ import io.opentelemetry.sdk.logs.data.LogRecordData
...
 internal class LogSinkImpl : LogSink {
     private val storedLogs: MutableList<EmbraceLogRecordData> = mutableListOf()
+    private val onLogsStored: (() -> Unit)? = null
 
     override fun storeLogs(logs: List<LogRecordData>): CompletableResultCode {
         try {
             synchronized(storedLogs) {
                 storedLogs += logs.map { EmbraceLogRecordData(logRecordData = it) }
             }
+            onLogsStored?.let { it() }
```
lucaslabari, line 15: nit: this could be clearer to read:
```suggestion
            onLogsStored?.invoke()
```

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/internal/logs/LogSinkImpl.kt
```diff
@@ -31,4 +33,8 @@ internal class LogSinkImpl : LogSink {
             return flushedLogs
         }
     }
+
+    fun callOnLogsStored(onLogsStored: () -> Unit) {
+        this.onLogsStored = onLogsStored
```
lucaslabari, line 38: Should be throwing a compilation error due to the val being assigned a null.

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/internal/logs/LogOrchestrator.kt
lucaslabari, line 49: Is this verification needed? Why calling sendLogs() directly isn't possible given we are cancelling previous jobs and scheduling one for MAX_INACTIVITY_TIME from the time the log arrived.

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/internal/logs/LogOrchestrator.kt
```diff
@@ -0,0 +1,78 @@
...
+import io.embrace.android.embracesdk.comms.delivery.DeliveryService
+import io.embrace.android.embracesdk.internal.clock.Clock
+import io.embrace.android.embracesdk.worker.ScheduledWorker
+import java.util.concurrent.ScheduledFuture
+import java.util.concurrent.TimeUnit
+
+internal class LogOrchestrator(
+    private val logOrchestratorScheduledWorker: ScheduledWorker,
+    private val clock: Clock,
+    private val sink: LogSinkImpl,
```
bidetofevil, line 13: This should be the interface `LogSink` instead so we can provide a fake impl for tests?

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/internal/logs/LogOrchestrator.kt
```diff
@@ -0,0 +1,78 @@
...
+            sendLogs()
+        } else {
+            scheduleInactivityCheck()
+            if (batchTimeFuture == null) {
+                scheduleBatchTimeCheck()
+            }
+        }
+    }
+
+    @VisibleForTesting
```
bidetofevil, line 35: We probably shouldn't expose methods for testing purposes. Instead of observing the internals of this method and what it does, you can provide a fake `LogSink` so you can observe the data that gets written there via this method.

You can have a look at `SpanServiceImplTest` to see how that can be done.

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/internal/logs/LogOrchestrator.kt
```diff
@@ -0,0 +1,78 @@
...
+    fun sendLogs() {
+        inactivityFuture?.cancel(false)
+        inactivityFuture = null
+        batchTimeFuture?.cancel(false)
+        batchTimeFuture = null
+
+        // Sink is synchronized, so even if sendLogs is called concurrently, the logs
```
bidetofevil, line 42: That being true, would we get into a race condition where the following happens?

- A log is added that triggers a flush
- A log is added right after that
- One of the timing job runs and flushes the new log that was added because it wasn't cleared in time.

Given that `sendLogs` isn't synchronized, I think we can get into a situation where that happens?

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/internal/logs/LogOrchestrator.kt
```diff
@@ -0,0 +1,78 @@
...
+        // will be sent only once.
+        val storedLogs = sink.flushLogs()
+
+        if (storedLogs.isNotEmpty()) {
+            deliveryService.sendLogs(LogPayload(logs = storedLogs))
+        }
+    }
+
+    private fun scheduleInactivityCheck() {
```
bidetofevil, line 51: Do we need two jobs? If we only schedule the next time the batch should be checked based on the current state of the batch and the time it was first created, would that be sufficient?

So basically, every time we get a new log, we cancel the next check and schedule a new one that is the minimum of the inactivity time and the time since the first log in the batch was added.

And when the check runs, all it has to do is see if either the inactivity time has elapsed or if the total batch time has been exceed.

If the log results in the max batch size being reached, send the batch cancel the check, and let the next time a log comes in to schedule another one.

bidetofevil (changes requested):
Have look at the checking logic to see if it can be simplified?

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/internal/logs/LogOrchestrator.kt
leandro-godon: Changed to be one job only.

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/internal/logs/LogOrchestrator.kt
bidetofevil: While the flushings themselves will be synchronized, so each flush will not over lap, we may be prematurely calling the second flush - when it checks whether it should flush, it'll say yes because the first flush hasn't completed yet. And if there's a log added after the first flush is finished but before the second flush runs, that log will be sent prematurely.

To put it another way, if the second flush results in an empty list, it shouldn't have been flushed at all - the first flush should've canceled the second flush, but because sendLogs isn't synchronized, we can get into this situation

While the second log will only be

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/internal/logs/LogOrchestrator.kt
```diff
@@ -0,0 +1,74 @@
...
+import java.util.concurrent.ScheduledFuture
+import java.util.concurrent.TimeUnit
+
+internal class LogOrchestrator(
+    private val logOrchestratorScheduledWorker: ScheduledWorker,
+    private val clock: Clock,
+    private val sink: LogSink,
+    private val deliveryService: DeliveryService
+) {
+    private var lastLogTime: Long = 0
```
bidetofevil, line 15: These should be volatile or be Atomic references since they are accesses from multiple threads?

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/internal/logs/LogOrchestrator.kt
```diff
@@ -0,0 +1,74 @@
...
+            sendLogs()
+        } else {
+            scheduleCheck()
+        }
+    }
+
+    private fun logsShouldBeSent(): Boolean {
+        return (
+            (sink.completedLogs().size >= MAX_LOGS_PER_BATCH) ||
+                (clock.now() - lastLogTime > MAX_INACTIVITY_TIME) ||
```
bidetofevil, line 37: Calling `now()` twice could result in different values if we are JUST at the end of the next millisecond. What about calling it once and using the result in both clauses so it's more consistent?

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/internal/logs/LogOrchestrator.kt
```diff
@@ -0,0 +1,74 @@
...
+
+    private fun scheduleCheck() {
+        scheduledCheckFuture?.cancel(false)
+        scheduledCheckFuture = logOrchestratorScheduledWorker.schedule<Unit>(
+            {
+                if (logsShouldBeSent()) {
+                    sendLogs()
+                }
+            },
+            MAX_INACTIVITY_TIME,
```
bidetofevil, line 64: Shouldn't the next check time be less than the inactivity time if the time to the next max batch time is less than that?

embrace-android-sdk/src/test/java/io/embrace/android/embracesdk/internal/logs/LogOrchestratorTest.kt
```diff
@@ -0,0 +1,96 @@
...
+        private const val now = 123L
+    }
+
+    private lateinit var logOrchestrator: LogOrchestrator
+
+    private lateinit var executorService: BlockingScheduledExecutorService
+    private lateinit var scheduledWorker: ScheduledWorker
+    private val clock = FakeClock()
+    private val logSink: LogSinkImpl = LogSinkImpl()
+    private val deliveryService: DeliveryService = mockk(relaxed = true)
```
bidetofevil, line 29: You can use a FakeDeliveryService here and check the actual payload that gets delivered rather than just that a mocked method is invoked.

embrace-android-sdk/src/test/java/io/embrace/android/embracesdk/internal/logs/LogOrchestratorTest.kt
```diff
@@ -0,0 +1,96 @@
...
+        repeat(4) {
+            logSink.storeLogs(listOf(FakeLogRecordData()))
+            moveTimeAhead(timeStep)
+        }
+
+        // Verify no logs have been sent
+        assertFalse(logSink.completedLogs().isEmpty())
+        verify(exactly = 0) { deliveryService.sendLogs(any()) }
+
+        moveTimeAhead(timeStep)
```
bidetofevil, line 85: I think this is sent because of the we've exceeded the 2 second inactivity time. If you only moved this up by 700L, the batch time would have been exceeded (5100L), but I don't think the current implementation would send this because the next scheduled check isn't until 5300L.

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/internal/logs/LogOrchestrator.kt
leandro-godon: Fixed

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/internal/logs/LogOrchestrator.kt
leandro-godon: Set to volatile

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/internal/logs/LogOrchestrator.kt
leandro-godon: You are correct. I changed it to be the min between inactivity time and batch time.

embrace-android-sdk/src/test/java/io/embrace/android/embracesdk/internal/logs/LogOrchestratorTest.kt
leandro-godon: I updated the test so that it catches that situtation.

bidetofevil (approved):
Looks good. Lets merge this down and we can talk about the synchronization separately. It might be easier if I wrote a test to show you what I mean 😅

embrace-android-sdk/src/test/java/io/embrace/android/embracesdk/internal/logs/LogOrchestratorTest.kt
leandro-godon: Switched to FakeDeliveryService

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/internal/logs/LogOrchestrator.kt
leandro-godon: I've merged the conditions checking and the log sending in a single synchronized method so in the case discussed it would return early instead of flushing the sink.

PR #439, opened 2024-02-19
[POC] Convert fragment breadcrumb capture to spans

fractalwrench:
## Goal

Converts the capture of fragment breadcrumbs to use the new `DataSource` interface. The `FragmentBreadcrumbDataSource` is the most important class to look at - it shows how `DataSource` can be used to start & stop `Span` objects.

This should not be merged yet as the payload schema is not agreed upon/supported by the backend, but it's useful as an illustration of how this should work.


embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/capture/crumbs/FragmentBreadcrumbDataSource.kt
```diff
@@ -1,90 +1,82 @@
...
+                inputValidation = NoInputValidation,
+                captureAction = { span.endFragmentSpan(clock.now()) }
+            )
         }
     }
 
-    override fun cleanCollections() {
-        store.cleanCollections()
-        fragmentStack.clear()
+    override fun FragmentBreadcrumb.toSpanEventData(): SpanEventData = SpanEventData(
```
bidetofevil, line 69: Ah, I think I got confused here - `SpanEventData` here is being used to represent the input to create a span as well as to create a SpanEvent (for the custom breadcrumb data source).

Maybe we shouldn't overload this name/class but instead have different data classes for each use case?

bidetofevil (commented):
I think I need to think about this some more and see more examples. I don't think I quite grok the start/stop interfaces at the data source level and how that maps to `SpanService`. Like, what happens if we want to capture spans that are more intricate, like if they contained children, span events, or attribute? Do we just add ad hoc extension methods on the data source that allow that to happen, thus simplifying the capture action blocks?

fractalwrench:
@bidetofevil I've updated this a bit to simplify the interfaces. `captureData` has been renamed to `alterSessionSpan` as it should be called when adding events/attributes to the session span only. I've rolled together `startSpan` and `stopSpan` and named it `captureSpanData`, and have also avoided overloading classes by creating a `StartSpanData` class.

fractalwrench:
>Like, what happens if we want to capture spans that are more intricate, like if they contained children, span events, or attribute?

My gut feel is that we should just let folks hold the `EmbraceSpan` reference and add directly to that for this scenario. It really depends on how complex the data is. The alternative would be creating some sort of DSL/extension functions that performs all those actions?

bidetofevil:
> My gut feel is that we should just let folks hold the `EmbraceSpan` reference and add directly to that for this scenario. It really depends on how complex the data is. The alternative would be creating some sort of DSL/extension functions that performs all those actions?

I think for now, exposing the created span directly would be good, especially since this is internal-only. It's the easiest way to go, and if we want to do the DSL thing in the future for clarity or external usage, we can iterate on that.

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/capture/crumbs/FragmentBreadcrumbDataSource.kt
```diff
@@ -1,90 +1,87 @@
...
-        if (name == null) {
-            return false
-        }
-        synchronized(this) {
-            if (fragmentStack.size >= DEFAULT_VIEW_STACK_SIZE) {
-                return false
+    /**
+     * Called when a fragment is started.
+     */
+    fun startFragment(name: String?): Boolean = captureSpanData(
```
bidetofevil, line 36: Does this assume that the names of the fragment can never repeat? If we never have to show it, does it make sense to get the hashcode of the instance instead to identify it in the map - or something that is unique to that instance of the fragment?

bidetofevil (approved):
This looks good! Part of me is still thinking about the methods names and such, but I'll be thinking about them until the cows come home, so this is good to go IMO.

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/capture/crumbs/FragmentBreadcrumbDataSource.kt
fractalwrench: The developer ultimately supplies any names that come down to this level via our API, so there's not much we can do (short of maybe logging warnings)

PR #440, opened 2024-02-19
Add extra functions to `DataSource` interface

fractalwrench:
## Goal

Adds extra functions to the `DataSource` interface that provide a way to start/stop spans. This is used in #439 but separated out for better reviewability.

## Testing

Added unit test coverage.

### DIFF
--- a/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/arch/DataSource.kt
+++ b/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/arch/DataSource.kt
@@ -2,7 +2,7 @@ package io.embrace.android.embracesdk.arch
 
 import io.embrace.android.embracesdk.internal.spans.CurrentSessionSpan
 import io.embrace.android.embracesdk.internal.spans.EmbraceTracer
-import io.embrace.android.embracesdk.spans.TracingApi
+import io.embrace.android.embracesdk.internal.spans.SpanService
 
 /**
  * Defines a 'data source'. This should be responsible for capturing a specific type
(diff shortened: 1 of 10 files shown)


lucaslabari (approved):
LGTM

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/arch/DataSource.kt
```diff
@@ -51,20 +55,55 @@ internal interface DataSource<T> {
 }
 
 /**
- * A [DataSource] that adds either a [EmbraceSpanEvent] or [EmbraceSpanAttribute]
- * to the current session span.
+ * A [DataSource] that adds or alters a new span on the [SpansService]
  */
-internal typealias EventDataSource = DataSource<SessionSpanWriter>
+internal interface SpanDataSource : DataSource<TracingApi> {
```
bidetofevil, line 60: It seems to make more sense for this to get a `SpanService` instead and have all of its APIs be available? The idea is that these are internal to Embrace (for now), so we can grant them "private" access?

bidetofevil (approved):
I think we should use `SpanService` instead of `TracingApi` so the data sources have full access to the capabilities of the SpansService

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/arch/SessionSpanWriter.kt
```diff
@@ -28,8 +28,8 @@ internal interface SessionSpanWriter {
  * Represents a span event that can be added to the current session span.
  */
 internal class SpanEventData(
-    val name: String,
-    val timeNanos: Long,
+    val spanName: String,
+    val spanStartTimeNanos: Long,
```
bidetofevil, line 32: I must've missed this when I did the initial review - this is actually the span event time, not the start time of the span. so Maybe eventName and eventTimeNanos if we want to make this more clear?

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/arch/SessionSpanWriter.kt
```diff
@@ -28,8 +28,8 @@ internal interface SessionSpanWriter {
  * Represents a span event that can be added to the current session span.
  */
 internal class SpanEventData(
```
bidetofevil, line 30: I got confused about what this is since `SpanEvent` is already a thing.  What you want to capture here are the parameters to startSpan()? If so, it should probably include `parent` in case this data source wants to create child spans as well as parent spans?

PR #443, opened 2024-02-20
Reorganise source files

fractalwrench:
## Goal

Reorganises some of the recently added files so they are in more specific packages.


lucaslabari (approved):
LGTM

PR #447, opened 2024-02-20
Convert LogsApi to kotlin.

lucaslabari:
## Goal

- Convert LogsApi to kotlin

## Testing

- Relied on existing unit tests. 

NOTE: Please review the last 3 params of `logPushNotification`. I had to make them nullable because if I make them non-nullable the implementation needs to be updated to primitive types on those 3 params and we would be changing the API.

### DIFF
--- a/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/LogsApi.kt
+++ b/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/LogsApi.kt
@@ -1,15 +1,9 @@
-package io.embrace.android.embracesdk;
-
-import androidx.annotation.NonNull;
-import androidx.annotation.Nullable;
-
-import java.util.Map;
+package io.embrace.android.embracesdk
 
 /**
  * The public API that is used to send log messages.
  */
-interface LogsApi {
-
+internal interface LogsApi {
     /**
      * Remotely logs a message at the given severity level. These log messages will appear as part of the session
      * timeline, and can be used to describe what was happening at a particular time within the app.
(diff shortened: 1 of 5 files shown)


embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/LogsApi.kt
```diff
@@ -121,8 +125,10 @@ void logException(@NonNull Throwable throwable,
      * @param stacktraceElements the stacktrace to remotely log
      * @param severity           the severity level of the log message
      */
-    void logCustomStacktrace(@NonNull StackTraceElement[] stacktraceElements,
-                             @NonNull Severity severity);
+    fun logCustomStacktrace(
+        stacktraceElements: Array<StackTraceElement?>,
```
bidetofevil, line 129: Do we actually allow nullable elements in the array?

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/LogsApi.kt
```diff
@@ -164,12 +174,14 @@ void logCustomStacktrace(@NonNull StackTraceElement[] stacktraceElements,
...
      */
-    void logPushNotification(@Nullable String title,
-                             @Nullable String body,
-                             @Nullable String topic,
-                             @Nullable String id,
-                             @Nullable Integer notificationPriority,
-                             @NonNull Integer messageDeliveredPriority,
-                             @NonNull Boolean isNotification,
-                             @NonNull Boolean hasData);
+    fun logPushNotification(
```
bidetofevil, line 177: We should look to deprecate this and replace it with a more normal looking method. The method signature is.... not fantastic.

bidetofevil (commented):
I'll defer to Jamie on this, but it looks OK with me as a conversion. We should probably clean these up and deprecate some of the less ideal ones so we can replace them later.

fractalwrench (approved):
LGTM

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/LogsApi.kt
lucaslabari: Good catch... updated.

PR #448, opened 2024-02-22
Add more trace sections to the startup flow

bidetofevil:
## Goal

Add more trace sections to further break down startup. Make the async sections synchronous so it shows up better stacked in their running threads


fractalwrench (approved):
LGTM

PR #450, opened 2024-02-22
Tweak data source implementation

fractalwrench:
## Goal

Makes several tweaks to the [DataSource] implementation to support capturing data. This addresses review comments that were made in #439 and #406 and is separated out to keep the changeset small. Changes include:

- Implementation of a fake for `OpenTelemetryModule`
- Altered interface for adding an event to the session span
- Renamed interfaces/functions to better reflect what they do
- Miscellaneous updates to test code

### DIFF
--- a/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/arch/destination/StartSpanData.kt
+++ b/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/arch/destination/StartSpanData.kt
@@ -1,0 +1,10 @@
+package io.embrace.android.embracesdk.arch.destination
+
+/**
+ * Holds the information required to start a span.
+ */
+internal class StartSpanData(
+    val spanName: String,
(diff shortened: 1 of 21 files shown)


embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/arch/destination/StartSpanData.kt
```diff
@@ -0,0 +1,10 @@
+package io.embrace.android.embracesdk.arch.destination
+
+/**
+ * Holds the information required to start a span.
+ */
+internal class StartSpanData(
+    val spanName: String,
+    val spanStartTimeNanos: Long,
```
bidetofevil, line 8: This should be ms now

bidetofevil (approved):
LGTM! Other than timestamp parameters should be in millis now, this is good to go

PR #451, opened 2024-02-22
Add Log Service

leandro-godon:
## Goal

Create a LogService class that will eventually replace `EmbraceLogMessageService`


embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/internal/logs/LogService.kt
```diff
@@ -0,0 +1,28 @@
...
+ */
+internal interface LogService {
+
+    fun log(
+        message: String,
+        severity: Severity,
+        properties: Map<String, Any>?
+    )
+
+    fun logException(
```
fractalwrench, line 17: Can these functions include KDoc? It should be similar to what already exists on `LogMessageService`.

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/internal/logs/EmbraceExtensions.kt
```diff
@@ -0,0 +1,120 @@
...
+ */
+private const val EMBRACE_ATTRIBUTE_NAME_PREFIX = "emb."
+
+/**
+ * Attribute name for the application state (foreground/background) at the time the log was recorded
+ */
+private const val APP_STATE_ATTRIBUTE_NAME = EMBRACE_ATTRIBUTE_NAME_PREFIX + "state"
+
+/**
+ * Attribute name for the id of the session in progress ]at the time the log was recorded
```
fractalwrench, line 24:
```suggestion
 * Attribute name for the id of the session in progress at the time the log was recorded
```

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/internal/logs/EmbraceExtensions.kt
```diff
@@ -0,0 +1,120 @@
...
+
+/**
+ * Attribute name for the exception library in a log representing an exception
+ */
+private const val EXCEPTION_LIBRARY_ATTRIBUTE_NAME = EMBRACE_ATTRIBUTE_NAME_PREFIX + "exception_library"
+
+/**
+ * Add an attribute representing the application state (foreground/background) at the time the log was recorded
+ */
+internal fun LogRecordBuilder.setAppState(appState: String): LogRecordBuilder {
```
fractalwrench, line 61: Could we model this as an enum instead, and then convert it to the string description within this function?

embrace-android-sdk/src/test/java/io/embrace/android/embracesdk/internal/logs/EmbraceLogServiceTest.kt
```diff
@@ -0,0 +1,119 @@
...
+        private lateinit var metadataService: FakeMetadataService
+        private lateinit var sessionIdTracker: FakeSessionIdTracker
+
+        @BeforeClass
+        @JvmStatic
+        fun beforeClass() {
+            metadataService = FakeMetadataService()
+            sessionIdTracker = FakeSessionIdTracker()
+            mockkStatic(Uuid::class)
+            every { Uuid.getEmbUuid() } returns "id"
```
fractalwrench, line 33: What are these mocks required for? Is there any way we could replace this with fakes?

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/internal/logs/EmbraceLogService.kt
```diff
@@ -0,0 +1,128 @@
...
+            .setBody(message)
+            .setSeverity(otelSeverity)
+            .setSeverityText(otelSeverity.name)
+            .setTimestamp(System.currentTimeMillis(), TimeUnit.MILLISECONDS)
+
+        properties?.forEach {
+            builder.setAttribute(AttributeKey.stringKey(it.key), it.value.toString())
+        }
+
+        // Set these after the custom properties so they can't be overridden
```
fractalwrench, line 107: Would be useful to add a test case to validate this

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/internal/logs/EmbraceLogService.kt
```diff
@@ -0,0 +1,128 @@
...
+    override fun log(
+        message: String,
+        severity: Severity,
+        properties: Map<String, Any>?
+    ) {
+        val sessionId = sessionIdTracker.getActiveSessionId()
+        val appState = metadataService.getAppState()
+
+        backgroundWorker.submit {
+            // TBD: Check if config enabled
```
fractalwrench, line 32: It'd be worth looking at how `ConfigGate` works if we want to no-op for the whole service when a config value changes. That avoids having to know about any config value that disables the entire service from everywhere within the service.

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/internal/logs/EmbraceLogService.kt
```diff
@@ -0,0 +1,128 @@
...
+            // TBD: Check if log should be gated
+            // TBD: Count log and enforce limits
+
+            val attributes = mutableMapOf<String, Any>()
+            properties?.let { attributes.putAll(it) }
```
fractalwrench, line 66: It would be good to add a test for this defensive copy, if it's needed. Also, this is inconsistent with `log` which doesn't make a defensive copy

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/internal/logs/EmbraceLogService.kt
```diff
@@ -0,0 +1,128 @@
...
+        context: String? = null,
+        library: String? = null
+    ) {
+        val otelSeverity = mapSeverity(severity)
+
+        val builder = logger.logRecordBuilder()
+            .setBody(message)
+            .setSeverity(otelSeverity)
+            .setSeverityText(otelSeverity.name)
+            .setTimestamp(System.currentTimeMillis(), TimeUnit.MILLISECONDS)
```
fractalwrench, line 101:
```suggestion
            .setTimestamp(clock.now(), TimeUnit.MILLISECONDS)
```

We probably want to use the `Clock` implementation here for testability

fractalwrench (commented):
Nice! Left a few comments inline

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/internal/logs/LogService.kt
leandro-godon: Added.

embrace-android-sdk/src/test/java/io/embrace/android/embracesdk/internal/logs/EmbraceLogServiceTest.kt
leandro-godon: Removed the mock

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/internal/logs/EmbraceLogService.kt
leandro-godon: Added test case

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/internal/logs/EmbraceLogService.kt
leandro-godon: Removed.

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/internal/logs/EmbraceExtensions.kt
leandro-godon: I will make it in a separate PR since it affects the current clients of `MetadataService`.

lucaslabari (approved):
LGTM

PR #454, opened 2024-02-22
Add detection and fallback to nanos of inappropriate times in millis

bidetofevil:
## Goal

We are changing the API to take in millisecond, but because the method signature hasn't changed, just how we are naming the parameter and how interpret it, there would be no compile time failure to warn folks they are passing in a time in nanos rather than millis.

For now, we'll work around this by detecting that a timestamp is way too big to be a reasonable time in millis in real life (until the year 5000), and assume its' a nanos time and convert it to millis. Once we detect that this isn't done anymore, we can remove this workaround.

## Testing

Added tests at all layers to verify this

### DIFF
--- a/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/internal/clock/Clock.kt
+++ b/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/internal/clock/Clock.kt
@@ -20,11 +20,25 @@ public fun interface Clock {
 /**
  * Turns a number that specifies a millisecond value to nanoseconds
  */
-@InternalApi
-public fun Long.millisToNanos(): Long = TimeUnit.MILLISECONDS.toNanos(this)
+internal fun Long.millisToNanos(): Long = TimeUnit.MILLISECONDS.toNanos(this)
 
 /**
  * Turns a number that specifies a nanosecond value to milliseconds
  */
-@InternalApi
-public fun Long.nanosToMillis(): Long = TimeUnit.NANOSECONDS.toMillis(this)
+internal fun Long.nanosToMillis(): Long = TimeUnit.NANOSECONDS.toMillis(this)
+
+/**
+ * Any epoch timestamp that we detect to be unreasonable to be interpreted as milliseconds, we assume it's an unintended use of nanoseconds
+ * based on an old API version or assumption from OpenTelemetry conventions
+ */
+internal fun Long.normalizeTimestampAsMillis(): Long =
+    if (this < MAX_MS_CUTOFF) {
+        this
+    } else {
+        this.nanosToMillis()
+    }
+
+/**
+ * Equivalent to the epoch time of January 1, 5000 12:00:00 AM GMT
+ */
+internal const val MAX_MS_CUTOFF = 95_617_584_000_000L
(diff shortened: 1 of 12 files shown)


bidetofevil:
* **#454** <a href="https://app.graphite.dev/github/pr/embrace-io/embrace-android-sdk/454?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `master`

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/embrace-io/embrace-android-sdk/454?utm_source=stack-comment).

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/internal/clock/Clock.kt
```diff
@@ -20,11 +20,25 @@ public fun interface Clock {
...
+
+/**
+ * Any epoch timestamp that we detect to be unreasonable to be interpreted as milliseconds, we assume it's an unintended use of nanoseconds
+ * based on an old API version or assumption from OpenTelemetry conventions
+ */
+internal fun Long.normalizeTimestampAsMillis(): Long =
+    if (this < MAX_MS_CUTOFF) {
+        this
+    } else {
+        this.nanosToMillis()
```
fractalwrench, line 38: Can we log an error/warning or something so there's a chance a dev might spot this problem when developing locally? I am not usually a fan of this type of magical behavior but agree it's probably the best solution for this problem...

embrace-android-sdk/src/test/java/io/embrace/android/embracesdk/internal/clock/ClockTest.kt
```diff
@@ -0,0 +1,22 @@
+package io.embrace.android.embracesdk.internal.clock
+
+import org.junit.Assert.assertEquals
+import org.junit.Assert.assertNotEquals
+import org.junit.Test
+
+internal class ClockTest {
+    @Test
+    fun `test timestamp fallback conversion`() {
+        // We need to fix this before the year 5000. I will be dead by then so it'll be someone else's problem.
```
fractalwrench, line 10: Can a user set device year to >5000 & therefore trigger this use-case? What is the consequence of doing this?

fractalwrench (approved):
LGTM pending comments

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/internal/clock/Clock.kt
bidetofevil: Ah yea. Meant to do this in a separate PR. Will do so today,

And for the record, I also don't like this - but to maintain backwards compat at least in the short term, I was OK with it. Once we detect that folks aren't doing this anymore (via the property), we'll remove this

embrace-android-sdk/src/test/java/io/embrace/android/embracesdk/internal/clock/ClockTest.kt
bidetofevil: Actually, I should detect the time based on the SDK's epoch time to determine how far in the future to look. Let me do this in another PR - but that's a good shout!

bidetofevil:
### Merge activity

* **Feb 23, 10:55 AM EST**: @bidetofevil started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/embrace-io/embrace-android-sdk/454).

PR #458, opened 2024-02-23
Harden NDK json serialization

fractalwrench:
## Goal

Hardens the NDK JSON serialization by checking for `NULL` when dynamically allocation memory & checking the `Json_Status` return code when altering the JSON tree. If any of those return `false` then we should return `NULL` from the function & free the JSON tree (which frees anything that was added to the tree).

In rare scenarios where the system is under a lot of memory pressure our calls to `calloc` can fail. We were not checking this appropriately which could lead to a segfault at various points within this function.

Additionally we seemed to be pretty printing the JSON - I've switched this to compact as this usually saves around ~25% memory usage.

## Testing

Regression tested via NDK app & confirmed crash reports are still displayed in dashboard.

### DIFF
--- a/embrace-android-sdk/src/main/cpp/file_writer.c
+++ b/embrace-android-sdk/src/main/cpp/file_writer.c
@@ -60,19 +60,28 @@ static const char *kErrContext = "c";
 static const char *kDefaultNULLFallbackString = "";
 static const char *kCurrentPayloadVersion = "1";
 
+#define RETURN_ON_JSON_FAILURE(func) \
+if ((func) != JSONSuccess) { \
+    return false; \
+}
 
-void emb_add_metadata_to_json(const emb_crash *crash, JSON_Object *root_object);
+bool emb_add_metadata_to_json(const emb_crash *crash, JSON_Object *root_object);
 
-void emb_add_basic_info_to_json(const emb_crash *crash, JSON_Object *root_object);
+bool emb_add_basic_info_to_json(const emb_crash *crash, JSON_Object *root_object);
 
-void emb_add_exc_info_to_json(const emb_crash *crash, JSON_Object *crash_object,
+bool emb_add_exc_info_to_json(const emb_crash *crash, JSON_Object *crash_object,
                               const emb_exception *exception);
 
-void emb_add_frame_dbg_to_json(JSON_Object *frame_object, emb_sframe *frame);
+bool emb_add_frame_dbg_to_json(JSON_Object *frame_object, emb_sframe *frame);
 
-void emb_add_frame_info_to_json(JSON_Object *frame_object, emb_sframe *frame);
+bool emb_add_frame_info_to_json(JSON_Object *frame_object, emb_sframe *frame);
 
-void emb_add_exc_to_json(const emb_exception *exception, JSON_Array *frames_object);
+bool emb_add_exc_to_json(const emb_exception *exception, JSON_Array *frames_object);
+
+bool emb_add_b64_value_to_json(JSON_Object *root_object, const JSON_Value *crash_value);
+
+bool emb_build_crash_json_tree(emb_crash *crash, JSON_Object *root_object,
+                               JSON_Object *crash_object);
 
 bool emb_write_crash_to_file(emb_env *env) {
     int fd = open(env->report_path, O_WRONLY | O_CREAT, 0644);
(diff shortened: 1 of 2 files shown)


embrace-android-sdk/src/main/cpp/file_writer.c
```diff
@@ -158,147 +165,313 @@ emb_error *emb_read_errors_from_file(const char *path) {
...
     EMB_LOGDEV("Serializing final JSON string");
-    serialized_string = json_serialize_to_string_pretty(root_value);
-//    json_free_serialized_string(serialized_string);
-    json_value_free(root_value);
-    json_value_free(crash_value);
+    serialized_string = json_serialize_to_string(root_value);
+
+    Exit:
+    if (root_value != NULL) {
+        json_value_free(root_value);
```
fractalwrench, line 219: This frees all the objects in this JSON tree

embrace-android-sdk/src/main/cpp/file_writer.c
```diff
@@ -158,147 +165,313 @@ emb_error *emb_read_errors_from_file(const char *path) {
...
-
-    JSON_Value *frames_value = json_value_init_array();
-    JSON_Array *frames_object = json_value_get_array(frames_value);
-    emb_add_exc_to_json(exception, frames_object);
-
-    json_object_set_value(crash_object, kFramesKey, frames_value);
+    // initialize JSON objects.
+    root_value = json_value_init_object();
+    if (root_value == NULL) {
+        goto Exit;
```
priettt, line 181: https://xkcd.com/292/ 😆

embrace-android-sdk/src/main/cpp/file_writer.c
```diff
@@ -158,147 +165,313 @@ emb_error *emb_read_errors_from_file(const char *path) {
...
 
-void emb_add_basic_info_to_json(const emb_crash *crash, JSON_Object *root_object) {
+bool emb_add_basic_info_to_json(const emb_crash *crash, JSON_Object *root_object) {
     EMB_LOGDEV("Serializing IDs + payload version.");
-    json_object_set_string(root_object, kReportIDKey, crash->report_id);
-    json_object_set_string(root_object, kVersionKey, kCurrentPayloadVersion);
-    json_object_set_number(root_object, kCrashTSKey, crash->crash_ts);
-    json_object_set_string(root_object, kSessionIDKey, crash->session_id);
-    json_object_set_string(root_object, kAppStateKey, crash->app_state);
+    if (json_object_set_string(root_object, kReportIDKey, crash->report_id) != JSONSuccess) {
```
priettt, line 239: Could we create a function that compares with JSONSuccess and returns true/false

```
bool set_string(object, key, value) {
    if (json_object_set_number(object, key, value) != JSONSuccess) {
        return false;
    }
    return true;
}
```
And then do something like
```
if (!set_string(root_object, kReportIDKey, crash-report_id) || !set_string(root_object, kVersionKey, kCurrentPayloadVersion) || etc...) 
    return false 
```

So we avoid all the `return false;`

priettt (approved):
Looks great!

embrace-android-sdk/src/main/cpp/file_writer.c
fractalwrench: No raptors yet! 🤞 

Goto might seem unorthodox but it's actually a pretty useful pattern for freeing up allocated memory within a C function. It's almost like a try-catch in Java, which is basically implemented as a goto under the hood...

priettt: It looks really clean! But I had to share the comic lol

embrace-android-sdk/src/main/cpp/file_writer.c
fnewberg: or add something like this to wrap functions

```
bool isJSONSuccess(JSON_Status result) {
    return result == JSONSuccess ? true : false;
}
```

fnewberg (approved):

embrace-android-sdk/src/main/cpp/file_writer.c
fractalwrench: I've updated the PR to use a macro

priettt (approved):

embrace-android-sdk/src/main/cpp/file_writer.c
fnewberg: > I've updated the PR to use a macro

looks good!

PR #464, opened 2024-02-23
Fix issue with OkHttp3

lucaslabari:
## Goal

- When using Embrace SDK 6+ with OkHttp3 there was a `NoSuchMethod` exception being thrown because from OkHttp v3 to v4 the `Response` class was converted to kotlin and we were calling the `header` method without passing a default param, but in Java (version 3) that param is mandatory. 

## Testing

Tested manually. 

## Release Notes

- Fixed a compatibility issue with OkHttp3.


bidetofevil:
LGTM

bidetofevil (approved):
LGTM

PR #467, opened 2024-02-25
Add more tracing sections to startup

bidetofevil:
## Goal

Add more tracing sections to startup workflow to better profile SDK startup. 

A misleading log line was also removed regarding loading native files

PR #468, opened 2024-02-25
Return more data from StartupService

bidetofevil:
## Goal

Add additional data exposed by the StartupService

### DIFF
--- a/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/capture/startup/StartupServiceImpl.kt
+++ b/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/capture/startup/StartupServiceImpl.kt
@@ -5,6 +5,7 @@ import io.embrace.android.embracesdk.worker.BackgroundWorker
 
 internal class StartupServiceImpl(
     private val spanService: SpanService,
+    @Suppress("UnusedPrivateMember")
     private val backgroundWorker: BackgroundWorker
 ) : StartupService {
 
@@ -22,13 +23,11 @@ internal class StartupServiceImpl(
 
     override fun setSdkStartupInfo(startTimeMs: Long, endTimeMs: Long) {
         if (sdkStartupDurationMs == null) {
-            backgroundWorker.submit {
-                spanService.recordCompletedSpan(
-                    name = "sdk-init",
-                    startTimeMs = startTimeMs,
-                    endTimeMs = endTimeMs
-                )
-            }
+            spanService.recordCompletedSpan(
+                name = "sdk-init",
+                startTimeMs = startTimeMs,
+                endTimeMs = endTimeMs
+            )
         }
         sdkInitStartMs = startTimeMs
         sdkInitEndMs = endTimeMs
--- a/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/capture/startup/StartupServiceImpl.kt
+++ b/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/capture/startup/StartupServiceImpl.kt
@@ -1,10 +1,18 @@
 package io.embrace.android.embracesdk.capture.startup
 
 import io.embrace.android.embracesdk.internal.spans.SpanService
+import io.embrace.android.embracesdk.worker.BackgroundWorker
 
 internal class StartupServiceImpl(
-    private val spanService: SpanService
+    private val spanService: SpanService,
+    private val backgroundWorker: BackgroundWorker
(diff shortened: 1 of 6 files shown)


bidetofevil:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/embrace-io/embrace-android-sdk/468?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#468** <a href="https://app.graphite.dev/github/pr/embrace-io/embrace-android-sdk/468?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#467** <a href="https://app.graphite.dev/github/pr/embrace-io/embrace-android-sdk/467?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `master`

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/embrace-io/embrace-android-sdk/468?utm_source=stack-comment).

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/EmbraceImpl.java
```diff
@@ -324,16 +325,17 @@ private void startImpl(@NonNull Context context,
...
         deliveryModule.getDeliveryService().sendCachedSessions(
             nativeModule.getNdkService(),
             essentialServiceModule.getSessionIdTracker()
         );
+        Systrace.endSynchronous();
 
         final CrashModule crashModule = moduleInitBootstrapper.getCrashModule();
         loadCrashVerifier(crashModule, moduleInitBootstrapper.getWorkerThreadModule());
 
-        Systrace.startSynchronous("internal-interface-init");
```
fractalwrench, line 336: Is there a reason these are removed?

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/capture/startup/StartupServiceImpl.kt
```diff
@@ -14,17 +22,25 @@ internal class StartupServiceImpl(
 
     override fun setSdkStartupInfo(startTimeMs: Long, endTimeMs: Long) {
         if (sdkStartupDurationMs == null) {
-            spanService.recordCompletedSpan(
-                name = "sdk-init",
-                startTimeMs = startTimeMs,
-                endTimeMs = endTimeMs
-            )
+            backgroundWorker.submit {
```
fractalwrench, line 25: Why is a backgroundworker required here? My first thought is the span service should take care of any perf penalty (if there is any from making this call)

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/EmbraceImpl.java
bidetofevil: Yeah this is actually pretty fast now with a change I did earlier to not always initialize the hosted SDK's respective internal interfaces, so I got rid of it.

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/capture/startup/StartupServiceImpl.kt
bidetofevil: And this should probably live in a different commit/PR because it's not directly related to the POC 😅

fractalwrench: I would lean towards just recording this on the calling thread as the background worker performs operations such as I/O - I'm wary that could throw off the accuracy of the span. I guess that depends on whether we record any other spans during SDK init. If we do, then we're going to pay that penalty anyway. This isn't blocking feedback so I'll leave it up to you

fractalwrench (approved):
LGTM

bidetofevil:
I'll change this to not use the background thread for now and look to change in 6.5 after more testing. I think it'll be an improvement to do on a background thread, but maybe not this one?

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/capture/startup/StartupServiceImpl.kt
bidetofevil: I restored the old behaviour and will improve this in the next release.

PR #469, opened 2024-02-27
Update v2 logs endpoints URL.

lucaslabari:
## Goal

- Updated v2 endpoints suffix from "logs/log" to just "log".

## Testing

- Added test to cover v2 endpoints URLs.


fractalwrench (approved):
LGTM

PR #472, opened 2024-02-27
Add a log count limit

leandro-godon:
## Goal

Add a limit to the amount of logs per session that can be sent. Store the ids of the logs sent and the total number of log attempts. This functionality is functionally equivalent to the current one in `EmbraceLogMessageService`.


lucaslabari (approved):
LGTM

PR #473, opened 2024-02-28
Do disk writes and read on expected threads and stop periodic session caching when we are crashing

bidetofevil:
## Goal

WIP

## Testing

Need to update tests still


bidetofevil:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/embrace-io/embrace-android-sdk/473?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#473** <a href="https://app.graphite.dev/github/pr/embrace-io/embrace-android-sdk/473?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#471** <a href="https://app.graphite.dev/github/pr/embrace-io/embrace-android-sdk/471?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `master`

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/embrace-io/embrace-android-sdk/473?utm_source=stack-comment).

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/comms/delivery/EmbraceDeliveryCacheManager.kt
```diff
@@ -102,33 +116,49 @@ internal class EmbraceDeliveryCacheManager(
         }
     }
 
+    /**
+     * This method will do disk reads so do not run it on the main thread
+     */
     override fun getAllCachedSessionIds(): List<String> {
         val allSessions = cacheService.listFilenamesByPrefix(SESSION_FILE_PREFIX)
         allSessions.forEach { filename ->
             if (filename == OLD_VERSION_FILE_NAME) {
```
bidetofevil, line 125: So we can only ever have one cached session from the old SDK version? In the current SDK, is that basically true as well, but we just have a special timestamp in the file name?

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/comms/delivery/EmbraceDeliveryCacheManager.kt
```diff
@@ -50,6 +51,7 @@ internal class EmbraceDeliveryCacheManager(
     // The session id is used as key for this map
     // This list is initialized when getAllCachedSessions() is called.
     private val cachedSessions = mutableMapOf<String, CachedSession>()
+    private val allowPeriodicSessionCaching = AtomicBoolean(true)
```
fractalwrench, line 54: It is probably worth altering `SessionOrchestratorImpl` so that `processEndMessage` is invoked after the periodic session caching has stopped. The current implementation has a race condition where the periodic caching could happen after the session crash snapshot was written to disk. In this scenario the collections could already be cleaned which would explain why some payloads contained empty span arrays.

I think that moving the call order in `SessionOrchestratorImpl` so periodic caching is always suspended as the first operation would be enough to avoid the need for this check.

https://github.com/embrace-io/embrace-android-sdk/blob/6.3.2/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/session/orchestrator/SessionOrchestratorImpl.kt#L200

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/comms/delivery/EmbraceDeliveryCacheManager.kt
```diff
@@ -181,26 +211,16 @@ internal class EmbraceDeliveryCacheManager(
...
-    }
+    override fun loadPendingApiCalls(): PendingApiCalls =
+        runCatching { cacheService.loadObject(PENDING_API_CALLS_FILE_NAME, PendingApiCalls::class.java) }.getOrNull()
+            ?: loadPendingApiCallsOldVersion()
+            ?: PendingApiCalls()
 
     override fun replaceSession(sessionId: String, mutator: (SessionMessage) -> SessionMessage) {
         val filename = cachedSessions[sessionId]?.filename ?: return
-        cacheService.replaceSession(filename, mutator)
+        backgroundWorker.submit {
```
fractalwrench, line 221: This can be removed as the invocation of `replaceSession` happens on a background thread

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/comms/delivery/EmbraceDeliveryCacheManager.kt
```diff
@@ -228,21 +248,19 @@ internal class EmbraceDeliveryCacheManager(
     }
 
     private fun loadSession(cachedSession: CachedSession): SessionMessage? {
```
fractalwrench, line 250: This and the `override fun loadSession(String)` actually seem unused outside of test code

fractalwrench (commented):
This looks like a good approach to me. I added one comment about how we can tweak `SessionOrchestrator` that could simplify the implementation, and also spotted a couple of extra places we can remove unused code/alter the calling thread when auditing the code myself.

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/comms/delivery/EmbraceDeliveryCacheManager.kt
fractalwrench: From what I remember older versions of the SDK wrote sessions to a single filename, and then on launch converted that file into a session that can be sent to the server. This was not good as it meant we needed to perform I/O at startup to avoid data loss, so we added a timestamp + UUID to the filename to get around this. So yes, there should only be 1 cached session from older SDK versions.

@lucaslabari your knowledge is probably better than mine here, please correct me if I got anything wrong

bidetofevil: Well there's another vector for losing sessions 😅 

BTW, I didn't make the change here, but I think we should defer this until app startup is complete (or do it in the background at least). It's a manner of less than a couple seconds in most cases, but doing this can be quite expensive

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/comms/delivery/EmbraceDeliveryCacheManager.kt
bidetofevil: THAT what was I was looking for! We try to cache a session that is complete, and we should never do that. Nice. I'll have a look

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/comms/delivery/EmbraceDeliveryCacheManager.kt
bidetofevil: Ah... I saw the immediate caller not wrapped in a task, but I didn't see that method's caller wrapping this in a task. Will remove.

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/comms/delivery/EmbraceDeliveryCacheManager.kt
fractalwrench: I agree that'd be sensible as long as the previous process didn't crash, as we don't want to impact deliverability if we're in the middle of a crash loop

fractalwrench: IMO the higher probability is worth it. If the SDK flushes sessions at different points of startup depending on what happened last process, it also makes it easier to measure the production perf impact of initializing later

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/comms/delivery/EmbraceCacheService.kt
```diff
@@ -66,7 +65,7 @@ internal class EmbraceCacheService(
 
     private fun File.writeHalfTheBytes(array: ByteArray): Unit = FileOutputStream(this).use {
         it.write(array.dropLast(array.size / 2).toByteArray())
-        throw RuntimeException("yooooo")
+        throw IllegalAccessException("yooooo")
```
fnewberg, line 68: i'm assuming we're only using this and the previous method in tests. should we at least mark them as such if there is no better alternative place for them to be?

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/comms/delivery/EmbraceDeliveryCacheManager.kt
```diff
@@ -62,9 +62,14 @@ internal class EmbraceDeliveryCacheManager(
             val sessionId = sessionMessage.session.sessionId
             val writeSync = snapshotType == SessionSnapshotType.JVM_CRASH
             val snapshot = snapshotType == SessionSnapshotType.PERIODIC_CACHE
+
             saveBytes(sessionId, writeSync, snapshot) { filename: String ->
                 Systrace.traceSynchronous("serialize-session") {
-                    cacheService.writeSession(filename, sessionMessage)
+                    if (cachedSessions.containsKey(sessionId)) {
```
fnewberg, line 68: does this require synchronization? how do we ensure that there isn't a race between the check if something is in the cache and then writing  to the cache?

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/comms/delivery/EmbraceDeliveryCacheManager.kt
lucaslabari: > From what I remember older versions of the SDK wrote sessions to a single filename, and then on launch converted that file into a session that can be sent to the server. This was not good as it meant we needed to perform I/O at startup to avoid data loss, so we added a timestamp + UUID to the filename to get around this. So yes, there should only be 1 cached session from older SDK versions.

@fractalwrench yeap, you are correct. Apart from performing I/O at startup, the problem with the older approach was that the cache of a session could override a previous one in cases where for example there is a new crash at startup before the previous session was intended to be sent, so it wasn't sent nor saved as a pending API call. Resulting in a session being lost. 
I think we can defer the sent of the session if we want to improve startup time and in case we want to differentiate the behavior in the cases where the previous session crashed, we can use the `crash_marker_file` to know that and do it at an early stage if that file is present.

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/comms/delivery/EmbraceCacheService.kt
bidetofevil: Oh this will be removed before I merge it. None of this stuff get into production code :-)

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/comms/delivery/EmbraceDeliveryCacheManager.kt
bidetofevil: Good call. Will do

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/comms/delivery/EmbraceDeliveryCacheManager.kt
bidetofevil: We can address all of this when we do that work. The native stuff is probably worth deferring more, so we can look at this holistically. I'll create a ticket to track this more formally.

I don't mind startup taking longer if we actually have a session to send - I think for me, we are paying the hit regardless if there's a cached session, so even doing a quick no-op would address this satisfactorily for me.

PR #475, opened 2024-02-29
Pre-requisite changes for migrating breadcrumbs to OTel

fractalwrench:
## Goal

Makes some pre-requisite changes for migrating breadcrumbs to OTel. This changeset adds code to support gating of OTel features in sessions & an integration test to assert it works. This will assert on actual OTel spans in #406. It also alters some test assertions that will be obsolete.

### DIFF
--- a/embrace-android-sdk/src/integrationTest/kotlin/io/embrace/android/embracesdk/session/OtelSessionGatingTest.kt
+++ b/embrace-android-sdk/src/integrationTest/kotlin/io/embrace/android/embracesdk/session/OtelSessionGatingTest.kt
@@ -1,0 +1,142 @@
+package io.embrace.android.embracesdk.session
+
+import androidx.test.ext.junit.runners.AndroidJUnit4
+import io.embrace.android.embracesdk.FakeDeliveryService
+import io.embrace.android.embracesdk.IntegrationTestRule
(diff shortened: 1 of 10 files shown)


priettt (approved):
LGTM!

embrace-android-sdk/src/integrationTest/kotlin/io/embrace/android/embracesdk/session/OtelSessionGatingTest.kt
```diff
@@ -0,0 +1,142 @@
...
+            val payload = harness.getSentSessionMessages().single()
+            verifySessionHappened(payload)
+            assertSessionGating(payload, gated = true)
+        }
+    }
+
+    @Suppress("UNUSED_PARAMETER")
+    private fun assertSessionGating(
+        payload: SessionMessage,
+        gated: Boolean
```
priettt, line 101: Why are we passing this if it's not used? Do you plan to use it in the future?

fractalwrench: Yep - I'm using it in #406. I've broken it down here to make the changeset smaller

PR #476, opened 2024-02-29
Initialize the new log service

leandro-godon:
## Goal

Add LogService and LogOrchestrator to the Log module so they can be used within the SDK.


embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/injection/CustomerLogModule.kt
```diff
@@ -63,4 +69,23 @@ internal class CustomerLogModuleImpl(
...
         )
     }
+
+    override val logService: LogService by singleton {
+        EmbraceLogService(
+            openTelemetryModule.logWriter,
+            initModule.clock,
+            essentialServiceModule.metadataService,
+            essentialServiceModule.sessionIdTracker,
+            workerThreadModule.backgroundWorker(WorkerName.OTEL_LOGGING)
```
fractalwrench, line 79: Generally speaking we want to avoid creating background threads unless it's absolutely essential. Could we reuse the existing remote logger worker for processing logs? Does the `LogOrchestrator` need its own dedicated worker?

fractalwrench (commented):
The addition to the module looks good but I have a question about whether it's necessary to create 2 additional background threads for this functionality

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/injection/CustomerLogModule.kt
leandro-godon: Switched to using the existing log thread.

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/injection/CustomerLogModule.kt
```diff
@@ -63,4 +69,23 @@ internal class CustomerLogModuleImpl(
             workerThreadModule.backgroundWorker(WorkerName.REMOTE_LOGGING)
         )
     }
+
+    override val logService: LogService by singleton {
```
fractalwrench, line 73: I think we might need to initialize this in the `OpenTelemetryModule` instead as it will be used for initializing other data capture sources.

fractalwrench (approved):
LGTM pending one comment

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/injection/CustomerLogModule.kt
leandro-godon: I don't think the `LogService` needs to be used by other data sources, since it deals with Embrace logs only. Other sources can use the `LogWriter` which is initialized in the `OpenTelemetryModule`. On the other hand, `LogOrchestrator` might need to be moved, although its dependency on `DeliveryService` can complicate things.

PR #479, opened 2024-02-29
Resolve some compiler warnings

fractalwrench:
## Goal

Fixes some compiler warnings that have been introduced recently.


priettt (approved):
🙌

PR #460, opened 2024-02-23
Build a LogRecord in the LogWriter.

leandro-godon:
## Goal

Provide a `LogWriter` that can emit logs from a `LogEventData`. A decision still needs to be made on where to populate Embrace-specific attributes like "session_id".

PR #480, opened 2024-02-29
Fix dupe symbol in tests

fractalwrench:
## Goal

Prevents a duplicate symbol from failing test compilation.

#460 seemed to add an implementation of the fake class I wanted, and was merged after #478 ran CI checks (so didn't show up as a failure).


bidetofevil (approved):
Thanks to CI for catching stuff like this lol

PR #485, opened 2024-03-03
Avoid blocking thread when reading RN bundle Id.

lucaslabari:
## Goal

- We were blocking the main thread at the moment of reading the `reactNativeBundleId` value if it was not yet calculated. 

- There is a trade-off between generating ANRs if the value is not there and not getting symbolication on some super early events that could happen before the bundle Id is calculated. We prefer the last option as it might not be a common scenario and we don't want to generate ANRs under any concept. 

- With the change on this PR, we calculate the `reactNativeBundleId` in background and when reading it, we return the value if it's ready or null if't not yet calculated. 

- Refactored the code related to the calculation of the bundleID. 

## Testing

## Release Notes

- Fix ANRs for React Native apps using CodePush.


fractalwrench (approved):
LGTM

PR #486, opened 2024-03-04
Update proguard config to fix R8 warnings introduced by using OTel logs SDK

bidetofevil:
## Goal

Running proguard on the test suite app caused some warnings that this fixed. Not sure if this is the best way to fix this, but the immutable annotation doesn't seem to be that important for this usage, but I'm open to better ideas. Simply keeping that doesn't seem to work

### DIFF
--- a/embrace-android-sdk/embrace-proguard.cfg
+++ b/embrace-android-sdk/embrace-proguard.cfg
@@ -12,5 +12,6 @@
 -keep class java9.** { *; }
 -dontwarn java9.**
 
-## OTel
+## OpenTelemetry Java SDK
 -keep class io.opentelemetry.api.trace.StatusCode { *; }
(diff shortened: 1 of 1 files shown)


bidetofevil:
* **#486** <a href="https://app.graphite.dev/github/pr/embrace-io/embrace-android-sdk/486?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `master`

<h2></h2>

fractalwrench (approved):
LGTM. Perhaps this is a candidate for something we could consider adding back to the OTel SDK if they don't have proguard rules bundled already.

bidetofevil:
Great idea! That said, what do you think about just ignoring this rather than finding a way to make it work by not deobfuscating this instead? I just couldn't get the syntax to work and moved on knowing the implications for us. In general though, others may find this useful

PR #490, opened 2024-03-04
Gate log messages based on remote configuration

leandro-godon:
## Goal

Replicate the message gating functionality currently present in `EmbraceLogMessageService`.

## Testing

Copied unit tests from `EmbraceLogMessageServiceTest`.


embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/internal/logs/EmbraceLogService.kt
```diff
@@ -113,7 +113,9 @@ internal class EmbraceLogService(
         attributes: EmbraceLogAttributes,
     ) {
         backgroundWorker.submit {
-            // TBD: Check if log should be gated
+            if (shouldLogBeGated(severity)) {
```
bidetofevil, line 116: Should we do this before we submit the job or does delaying evaluation to when it runs make more sense?

leandro-godon: No, you are right, it can be done before submit.

PR #493, opened 2024-03-04
Add AEI data source

fractalwrench:
## Goal

Implements the capture of AEI via the `DataSource` interface which allows them to be sent as OTel logs. It's important to note this does _not_ alter any payloads yet. The existing `ApplicationExitInfoService` will continue to add directly to the payload.

A future PR will contain the changeset for enabling the new data capture mechanism. This uncouples work & allows us to merge more without blocking changes.

## Testing

Added unit test coverage.

### DIFF
--- a/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/capture/aei/AeiDataSourceImpl.kt
+++ b/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/capture/aei/AeiDataSourceImpl.kt
@@ -1,0 +1,316 @@
+package io.embrace.android.embracesdk.capture.aei
+
+import android.app.ActivityManager
+import android.app.ApplicationExitInfo
+import android.os.Build.VERSION_CODES
+import androidx.annotation.RequiresApi
+import io.embrace.android.embracesdk.Severity
+import io.embrace.android.embracesdk.arch.datasource.LogDataSourceImpl
+import io.embrace.android.embracesdk.arch.destination.LogEventData
+import io.embrace.android.embracesdk.arch.destination.LogEventMapper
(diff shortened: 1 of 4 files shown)


lucaslabari (approved):
LGTM

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/capture/aei/AeiDataSourceImpl.kt
```diff
@@ -0,0 +1,316 @@
...
+import io.embrace.android.embracesdk.session.id.SessionIdTracker
+import io.embrace.android.embracesdk.worker.BackgroundWorker
+import java.io.IOException
+import java.util.concurrent.Future
+import java.util.concurrent.atomic.AtomicBoolean
+
+@RequiresApi(VERSION_CODES.R)
+internal class AeiDataSourceImpl(
+    private val backgroundWorker: BackgroundWorker,
+    private val configService: ConfigService,
```
bidetofevil, line 36: I thought we had wanted to make the orchestrator push config info to the data sources rather than have them directly depend on them. Is the current state due to how everything is hooked up to the legacy stuff right now, or will this be necessary going forward?

fractalwrench: The orchestrator does control whether a feature is enabled/disabled but that's it right now.

There isn't any such approach for config that alters limits/how data is captured. IMO just passing a `ConfigService` or even a `*Behavior` class is probably the simplest option

bidetofevil (approved):
One comment about the config service usage, but looks good to me in general.

PR #496, opened 2024-03-04
Add method to notify CodePush update into RN internal interface.

lucaslabari:
## Goal

- From RN side we have a way to know if CodePush has recently updated the bundle.
- In this PR I added a method to differentiate CodePush from other OTAs. 
- If a customer is using CodePush, RN SDK will call `setJavaScriptBundleUrlForCodePush` passing also a boolean indicating if this is the first run for that bundle and we need to update the `bundleId`.
- If a customer is using other OTA like Expo, RN SDK will call `setJavaScriptBundleUrl` and the bundleId will need to be calculated each time this method is called with a new bundleUrl, because if it's the same as the persisted one, the bundleId was calculated at startup. 
- At startup, we calculate the bundleId if it's not persisted and if the bundleURL is not persisted either, we use buildInfo.buildId as the default value. 

## Testing

Added unit tests.

## Release Notes

- Optimized Bundle Id calculation for RN Apps using CodePush.

### DIFF
--- a/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/ReactNativeInternalInterface.kt
+++ b/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/ReactNativeInternalInterface.kt
@@ -33,7 +33,7 @@ public interface ReactNativeInternalInterface : EmbraceInternalInterface {
 
     public fun setJavaScriptBundleUrl(context: Context, url: String)
 
-    public fun setJavaScriptBundleUrlForCodePush(context: Context, url: String, didUpdate: Boolean)
+    public fun setCacheableJavaScriptBundleUrl(context: Context, url: String, didUpdate: Boolean)
 
     /**
      * Logs a React Native Redux Action - this is not intended for public use.
--- a/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/ReactNativeInternalInterface.kt
+++ b/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/ReactNativeInternalInterface.kt
@@ -33,6 +33,8 @@ public interface ReactNativeInternalInterface : EmbraceInternalInterface {
 
     public fun setJavaScriptBundleUrl(context: Context, url: String)
 
+    public fun setJavaScriptBundleUrlForCodePush(context: Context, url: String, didUpdate: Boolean)
+
     /**
      * Logs a React Native Redux Action - this is not intended for public use.
      */
--- a/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/ReactNativeInternalInterface.kt
+++ b/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/ReactNativeInternalInterface.kt
@@ -31,8 +31,21 @@ public interface ReactNativeInternalInterface : EmbraceInternalInterface {
 
     public fun setReactNativeVersionNumber(version: String?)
 
+    /**
+     * Sets the React Native Bundle URL.
+     * @param context the context
+     * @param url the JavaScript bundle URL
+     */
     public fun setJavaScriptBundleUrl(context: Context, url: String)
 
+    /**
+     * Sets the React Native Bundle URL, indicating if the bundle was updated or not.
+     * If it was updated, the bundle ID will be recomputed.
+     * If not, the bundle ID will be retrieved from cache.
+     * @param context the context
+     * @param url the JavaScript bundle URL
+     * @param didUpdate if the bundle was updated
+     */
     public fun setCacheableJavaScriptBundleUrl(context: Context, url: String, didUpdate: Boolean)
 
     /**
--- a/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/capture/metadata/EmbraceMetadataService.kt
+++ b/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/capture/metadata/EmbraceMetadataService.kt
@@ -383,20 +383,25 @@ internal class EmbraceMetadataService private constructor(
 
     override fun getEgl(): String? = egl
 
-    override fun setReactNativeBundleId(context: Context, jsBundleUrl: String?) {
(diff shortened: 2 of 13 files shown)


embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/ReactNativeInternalInterface.kt
```diff
@@ -33,6 +33,8 @@ public interface ReactNativeInternalInterface : EmbraceInternalInterface {
 
     public fun setJavaScriptBundleUrl(context: Context, url: String)
 
+    public fun setJavaScriptBundleUrlForCodePush(context: Context, url: String, didUpdate: Boolean)
```
bidetofevil, line 36: Can we just the parameter and get the hosted SDK to call it with and without the parameter for code push?

lucaslabari: We can, it's just a fact of agreeing what's better with Damian... He preferred this way because it's explicitly clear that this call is for CodePush.

bidetofevil: Or alternatively, don't add the extra parameter and have two methods - one for update and one for not - so the API change isn't breaking, if you're worried about that.

bidetofevil: If he prefers it, we should have one method for bundle updated and one for not, and his code can pick and choose which to call instead of passing that boolean down?

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/capture/metadata/EmbraceMetadataService.kt
```diff
@@ -501,11 +506,23 @@ internal class EmbraceMetadataService private constructor(
             if (appFramework == AppFramework.REACT_NATIVE) {
                 reactNativeBundleId = metadataBackgroundWorker.submit<String?> {
                     val lastKnownJsBundleUrl = preferencesService.javaScriptBundleURL
-                    computeReactNativeBundleId(
-                        context,
-                        lastKnownJsBundleUrl,
-                        buildInfo.buildId
-                    )
+                    val lastKnownCodePushJsBundleId = preferencesService.codePushJsBundleId
+                    if (!lastKnownJsBundleUrl.isNullOrEmpty() && !lastKnownCodePushJsBundleId.isNullOrEmpty()) {
```
bidetofevil, line 510: Or there every two bundle Ids for any given app? If we want to distinguish whether they are updatable, can we store a boolean instead to signify that, rather than having two IDs which can be confusing?

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/ReactNativeInternalInterface.kt
bidetofevil: OK, I understand now - there are two pieces of new info: CAN the app update the bundle and DID the app update the bundle since last time this was called.

If that's the case, then this method should be called something like `setUpdateableJavaScriptBundleUrl` - and when this is called for the first time, set the boolean in the prefs service to true, and take the value to see if it needs updating (or if it's the first time, always update)

lucaslabari:
> @lucaslabari what is the reason we are making this change? was the old approach causing ANRs?

The ANRs were solved in another PR. The goal of this PR is to optimize the calculation of the RN bundleId that is sometimes a heavy operation. We are not blocking the main thread anymore, but if we calculate the bundleId each time and the bundle is a heavy file, then the early events could go to the backend with a null value for the bundleId in the payload. So the idea is to minimize this given that we have a way to know if CodePush updated the bundle.

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/ReactNativeInternalInterface.kt
lucaslabari: updated to `setCacheableJavaScriptBundleUrl`

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/capture/metadata/EmbraceMetadataService.kt
```diff
@@ -383,20 +383,25 @@ internal class EmbraceMetadataService private constructor(
 
     override fun getEgl(): String? = egl
 
-    override fun setReactNativeBundleId(context: Context, jsBundleUrl: String?) {
+    override fun setReactNativeBundleId(context: Context, jsBundleUrl: String?, didCodePushBundleUpdate: Boolean?) {
```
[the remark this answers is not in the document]
lucaslabari: Updated to `setCacheableJavaScriptBundleUrl`

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/capture/metadata/EmbraceMetadataService.kt
```diff
@@ -501,11 +506,19 @@ internal class EmbraceMetadataService private constructor(
             if (appFramework == AppFramework.REACT_NATIVE) {
                 reactNativeBundleId = metadataBackgroundWorker.submit<String?> {
                     val lastKnownJsBundleUrl = preferencesService.javaScriptBundleURL
-                    computeReactNativeBundleId(
-                        context,
-                        lastKnownJsBundleUrl,
-                        buildInfo.buildId
-                    )
+                    val lastKnownJsBundleId = preferencesService.javaScriptBundleId
+                    if (!lastKnownJsBundleUrl.isNullOrEmpty() && !lastKnownJsBundleId.isNullOrEmpty()) {
```
bidetofevil, line 510: Do we have to check this? `!lastKnownJsBundleUrl.isNullOrEmpty()`

bidetofevil (approved):
A couple of tweak suggestions but looks good overall

PR #503, opened 2024-03-05
Add emb-type as mandatory attribute for spans, events, and logs

fractalwrench:
## Goal

Enforces that `emb-type` must be specified when creating a span, event, or a log in a `DataSource` implementation. This was requested by the backend to make it easier to distinguish between data types.

## Testing

Added unit test coverage.


lucaslabari (approved):
LGTM

PR #504, opened 2024-03-05
Allow capturing metadata in new envelope

fractalwrench:
## Goal

Adds the ability to capture mutable data in the new `EnvelopeMetadata` class. This will eventually be used to construct a new payload, but right now is only touched in tests.

I have added an interface + implementation as this will make it easier to add fakes for our integration testing. We can follow a similar approach with `EnvelopeResource`. I have inlined functions for timezone/locale that were previously in `MetadataUtils` (as we want to get rid of that class ultimately)

## Testing

Added unit test coverage.


lucaslabari (approved):
LGTM

PR #505, opened 2024-03-05
Add skeleton for creating new payload

fractalwrench:
## Goal

Adds skeleton classes that will capture data in the new `Envelope` classes. This will eventually be used to construct a new payload, but right now is only touched in tests.


embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/capture/envelope/resource/EnvelopeResourceSource.kt
```diff
@@ -0,0 +1,11 @@
+package io.embrace.android.embracesdk.capture.envelope.resource
+
+import io.embrace.android.embracesdk.internal.payload.EnvelopeMetadata
+import io.embrace.android.embracesdk.internal.payload.EnvelopeResource
+
+/**
+ * Creates a [EnvelopeMetadata] object.
```
lucaslabari, line 7:
```suggestion
 * Creates a [EnvelopeResource] object.
```

lucaslabari (approved):

PR #508, opened 2024-03-05
Set the window callback on the mock activity so callback is never null when the activity is created

bidetofevil:
## Goal

The functional tests' mocks are missing the window callback, so the startup code fails. Fixed the mocks as well as deal with the case where the window callback is null, which should not happen, but just in case.....

## Testing

Functional tests pass


bidetofevil:
* **#508** <a href="https://app.graphite.dev/github/pr/embrace-io/embrace-android-sdk/508?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `master`

<h2></h2>

fractalwrench (approved):
LGTM

PR #512, opened 2024-03-06
Add internal error payload models

fractalwrench:
## Goal

Adds model classes for the new internal error payload. This is mostly generated via OpenAPI but tweaked manually for style reasons.

## Testing

Added unit test to confirm new model can be constructed from old one.

### DIFF
--- a/embrace-android-sdk/config/detekt/baseline.xml
+++ b/embrace-android-sdk/config/detekt/baseline.xml
@@ -2,9 +2,11 @@
 <SmellBaseline>
   <ManuallySuppressedIssues></ManuallySuppressedIssues>
   <CurrentIssues>
-    <ID>DataClassContainsFunctions:ExceptionError.kt$ExceptionError$fun addException(ex: Throwable?, appState: String?, clock: Clock)</ID>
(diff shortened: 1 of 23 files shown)


embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/payload/ExceptionInfo.kt
```diff
@@ -35,6 +35,10 @@ internal class ExceptionInfo internal constructor(
     @Json(name = "length")
     val originalLength: Int? = lines.size.takeIf { it > STACK_FRAME_LIMIT }
 
+    fun toNewPayload(): io.embrace.android.embracesdk.internal.payload.ExceptionInfo {
```
priettt, line 38: nit: is the full import needed?

priettt (approved):

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/payload/ExceptionInfo.kt
fractalwrench: Yeah it's needed to disambiguate right now. @priettt how about adding a prefix, like `LegacyExceptionInfo`, to affected classes? I'll roll that out across the other PRs tomorrow if you're happy with the approach

PR #518, opened 2024-03-07
Extract emb.type to separate source file

fractalwrench:
## Goal

Extracts the possible values for `emb.type` into a separate source file. I have organised these as enums implementing a common interface to get a bit of type safety rather than throwing strings around everywhere.

#519 also takes this further with a POC for the rest of the strings.

### DIFF
--- a/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/arch/schema/EmbType.kt
+++ b/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/arch/schema/EmbType.kt
@@ -1,0 +1,38 @@
+package io.embrace.android.embracesdk.arch.schema
+
+internal sealed class EmbType {
+
+    /**
+     * Keys that track how fast a time interval is. Only applies to spans.
+     */
+    internal sealed class Performance : TelemetryType
+
+    /**
+     * Keys that track a point in time & is visual in nature. Applies to spans, logs, and span events.
+     */
+    internal sealed class Ux(subtype: String) : TelemetryType {
+        internal object View : Ux("view")
+
+        override val description = "ux.$subtype"
+    }
+
+    /**
+     * Keys that track a point in time that is not visual in nature. Applies to spans, logs, and span events.
+     */
+    internal sealed class System(subtype: String) : TelemetryType {
+        internal object Breadcrumb : System("breadcrumb")
+        internal object Log : System("log")
+        internal object Exit : System("exit")
+
+        override val description = "system.$subtype"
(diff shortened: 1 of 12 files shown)


embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/arch/schema/EmbType.kt
```diff
@@ -0,0 +1,38 @@
+package io.embrace.android.embracesdk.arch.schema
+
+internal object EmbType {
```
lucaslabari, line 3: wouldn't be better to use a Sealed Class in case we need to evaluate the type at some point?

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/arch/schema/EmbType.kt
```diff
@@ -0,0 +1,38 @@
...
+
+    /**
+     * Keys that track how fast a time interval is. Only applies to spans.
+     */
+    internal enum class Performance : TelemetryType
```
lucaslabari, line 8: We don't have types and description here... I assume it's going to be added later.

lucaslabari (approved):
LGTM

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/arch/schema/EmbType.kt
bidetofevil: There are no subtypes now so this is probably fine until we do

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/arch/schema/EmbType.kt
bidetofevil, line 3: We want the name to be lower-case in the payload - I can't find a way to override `name`, so I used `typeName` before. Is there a way to do that so the default enum name would just work?

bidetofevil (approved):
Looks good. I can refactor `EmbraceAttributes.Type` to refer to this when it's in

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/arch/schema/EmbType.kt
fractalwrench: I'm not sure I follow. This class doesn't serialize/deserialize JSON, it just holds data that is passed to other classes that do.

PR #519, opened 2024-03-07
Enforce data capture types via SchemaType

fractalwrench:
## Goal

Organises telemetry keys using a type system, so we aren't throwing disparate strings around everywhere. `SchemaType`   is a sealed class that always contains the information required to derive `emb.type` (which is itself another type hierarchy), the span/event/log name (which acts as a unique identifier), and the attributes. Attributes are generated from constructor parameters passed to the relevant subclass of `SchemaType`.

This class does not serialize/deserialize JSON directly; it delegates to other classes. I've chosen this approach as I'm under the impression that `emb.type` acts as a namespace - if that's an invalid assumption then it's possible to simplify this further.

### DIFF
--- a/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/arch/schema/SchemaType.kt
+++ b/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/arch/schema/SchemaType.kt
@@ -1,0 +1,51 @@
+package io.embrace.android.embracesdk.arch.schema
+
+import io.embrace.android.embracesdk.internal.logs.EmbraceLogAttributes
+import io.embrace.android.embracesdk.internal.utils.toNonNullMap
+import io.embrace.android.embracesdk.payload.AppExitInfoData
+
+internal sealed class SchemaType(
+    val telemetryType: TelemetryType,
+    val name: String,
+) {
+    abstract val attrs: Map<String, String>
+
+    internal class CustomBreadcrumb(message: String) : SchemaType(
+        EmbType.System.Breadcrumb,
+        "custom-breadcrumb"
+    ) {
+        override val attrs = mapOf("message" to message)
+    }
+
+    internal class ViewBreadcrumb(viewName: String) : SchemaType(
+        EmbType.Ux.View,
+        "view-breadcrumb"
+    ) {
+        override val attrs = mapOf("view.name" to viewName)
+    }
+
+    internal class AeiLog(message: AppExitInfoData) : SchemaType(
+        EmbType.System.Exit,
+        "aei-record"
+    ) {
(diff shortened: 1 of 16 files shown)


embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/arch/schema/SchemaType.kt
```diff
@@ -0,0 +1,44 @@
+package io.embrace.android.embracesdk.arch.schema
+
+import io.embrace.android.embracesdk.internal.utils.toNonNullMap
+import io.embrace.android.embracesdk.payload.AppExitInfoData
+
+internal sealed class SchemaType(
```
bidetofevil, line 6: Doesn't this map one-to-one with TelemetryType.subtype? And if so, we can roll these two together?

fractalwrench: My understand is that `emb.type` was not unique - it's a namespace. E.g. it would be possible to have N sources of data under `system.breadcrumb`. Is that not the case?

bidetofevil: The *schema* is unique per `subtype`, but we can potentially overload a particular one to mean different things depending on the data source. 

For instance, for if the `system.breadcrumb` subtype that has a single message field as the schema, we can have a data source for `CustomBreadcrumbs` that use it - or we can use it to expose a new breadcrumb for, I don't know, "new thread spawned", which just reuses that schema.

Customers might think of that as a new feature (since we are tracking new threads being created), but it's using the same base subtype so we don't have to roll another one. As long as it makes sense for the backend and UI to treat these the same way, there's no need to create a new subtype.

In the future, we might want to create another distinction so the UI can display the same subtype differently, but we can get there when we get there if creating a new subtype all the time for UI diffs isn't viable

bidetofevil (approved):
We can merge this and discuss combining these concepts at the code level later

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/arch/schema/SchemaType.kt
bidetofevil: A good example is actually the Fragment Breadcrumb and a View Breadcrumb. Even if we offer two API methods to do each, the data that backs it is basically the same, so we can reuse the type.subtype (`ux.view`) but have two services be responsible for the capturing.

(Implementation wise, this example could probably be one service if we want, but you can see how services that eventually write to the same object defined by the same schema can/should live in separate services with vastly differently internal logic)

PR #520, opened 2024-03-07
Add span payload models

fractalwrench:
## Goal

Adds model classes for the new span payloads. This is mostly generated via OpenAPI but tweaked manually for style reasons.

## Testing

Added unit test to confirm new model can be constructed from old one.

### DIFF
--- a/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/internal/payload/Attribute.kt
+++ b/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/internal/payload/Attribute.kt
@@ -1,0 +1,22 @@
+package io.embrace.android.embracesdk.internal.payload
+
+import com.squareup.moshi.Json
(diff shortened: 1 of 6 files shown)


embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/internal/payload/Attribute.kt
```diff
@@ -0,0 +1,22 @@
+package io.embrace.android.embracesdk.internal.payload
+
+import com.squareup.moshi.Json
+import com.squareup.moshi.JsonClass
+
+/**
+ * A key-value pair that provides additional context to the span
+ *
+ * @param key The name of the attribute
+ * @param `value` The value of the attribute
```
priettt, line 10: nit: this is now called data right?

priettt (approved):

PR #521, opened 2024-03-07
Fix functional test flakes

fractalwrench:
## Goal

Refactors the functional tests so that they do not require requests to be sent in a specific order. Additionally, I've altered the test cases to ensure they specify what endpoint should be specified for each comparison against a golden file. This should hopefully eliminate one source of the flakes (where requests are sent in different orders due to natural fluctuations in timing & our tests didn't account for that)


bidetofevil:
Nice!

bidetofevil (approved):

PR #527, opened 2024-03-08
Avoid retrying functional tests

fractalwrench:
## Goal

Avoids retrying the functional tests. Since #521 was merged CI is much more reliable, so IMO this isn't required. If CI does fail, it'd be good to know about it the first time so we can dedicate the time to tweak the tests.

### DIFF
--- a/.github/workflows/functional-tests.yml
+++ b/.github/workflows/functional-tests.yml
@@ -73,21 +73,8 @@ jobs:
 
       - name: Run Functional Tests
         uses: embrace-io/android-emulator-runner@v2
-        id: runFunctionalTests1
+        id: runFunctionalTests
         continue-on-error: true
-        with:
-          api-level: ${{ matrix.api-level }}
-          target: ${{ matrix.target }}
-          arch: x86_64
-          profile: Nexus 6
-          # Grab the failures from the device so we can include them in results.
-          # Exit with non-zero so GH checks still fail.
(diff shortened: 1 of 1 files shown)


.github/workflows/functional-tests.yml
```diff
@@ -109,13 +96,13 @@ jobs:
...
         id: slack
-        if: ${{steps.runFunctionalTests1.outcome == 'failure'}}
+        if: ${{steps.runFunctionalTests.outcome == 'failure'}}
         uses: slackapi/slack-github-action@v1.25.0
         with:
           payload: |
             {
-              "functional_test_retried": "${{steps.runFunctionalTests1.outcome == 'failure'}}",
-              "workflow_failed": "${{steps.runFunctionalTests1.outcome == 'failure' && steps.runFunctionalTests2.outcome == 'failure'}}"
+              "functional_test_retried": "${{steps.runFunctionalTests.outcome == 'failure'}}",
```
lucaslabari, line 104: We can remove this line as we are not retrying anymore.

lucaslabari (approved):
LGTM. 
Left a comment to remove a variable. 
I'll reactivate the slack workflow so we get notified again and see if we have correct results.

.github/workflows/functional-tests.yml
fractalwrench: Removed this line.

lucaslabari: Also it would be good to change the comparison to failure and send a positive variable

lucaslabari: Maybe we can rename from `workflow_failed ` to `workflow_succeeded`

PR #528, opened 2024-03-08
Use behavior classes directly rather than config service

fractalwrench:
## Goal

Use behavior classes directly rather than `ConfigService`. This will allow us to eventually get rid of the `ConfigService` if we wish.

## Testing

Updated test coverage.


leandro-godon (approved):
Lgtm

PR #529, opened 2024-03-08
Add remaining parts to new payload

fractalwrench:
## Goal

Copies remaining parts from the OpenAPI generated-code to the new payload classes.


lucaslabari (approved):
LGTM

PR #541, opened 2024-03-11
Preserve moments across session boundaries

fractalwrench:
## Goal

Preserves moments across session boundaries if they are considered 'active' (i.e. they have not ended yet). The current implementation simply cleared all the events in our collections; however, it does not seem this behavior was ever hooked up properly until the recent `SessionOrchestrator` refactor. This meant that moments would be cleared when the session boundary was transitioned - which happened frequently for the app startup moment.

## Testing

Added integration + unit test coverage.

### DIFF
--- a/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/event/EmbraceEventService.kt
+++ b/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/event/EmbraceEventService.kt
@@ -5,7 +5,6 @@ import io.embrace.android.embracesdk.capture.metadata.MetadataService
 import io.embrace.android.embracesdk.capture.user.UserService
 import io.embrace.android.embracesdk.comms.delivery.DeliveryService
 import io.embrace.android.embracesdk.config.ConfigService
-import io.embrace.android.embracesdk.internal.CacheableValue
 import io.embrace.android.embracesdk.internal.EventDescription
 import io.embrace.android.embracesdk.internal.StartupEventInfo
 import io.embrace.android.embracesdk.internal.clock.Clock
@@ -23,10 +22,9 @@ import io.embrace.android.embracesdk.utils.stream
 import io.embrace.android.embracesdk.worker.BackgroundWorker
 import io.embrace.android.embracesdk.worker.WorkerName
 import io.embrace.android.embracesdk.worker.WorkerThreadModule
-import java.util.NavigableMap
 import java.util.concurrent.ConcurrentHashMap
+import java.util.concurrent.ConcurrentLinkedQueue
 import java.util.concurrent.ConcurrentMap
-import java.util.concurrent.ConcurrentSkipListMap
 
 /**
  * Handles the lifecycle of events (moments).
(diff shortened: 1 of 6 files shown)


nelsitoPuglisi (approved):

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/event/EmbraceEventService.kt
```diff
@@ -54,8 +52,7 @@ internal class EmbraceEventService(
     /**
      * Timeseries of event IDs, keyed on the start time of the event.
      */
-    private val eventIds: NavigableMap<Long, String> = ConcurrentSkipListMap()
-    private val eventIdsCache = CacheableValue<List<String>> { eventIds.size }
+    private val eventIds = ConcurrentLinkedQueue<String>()
```
priettt, line 55: just curious, why a queue?

embrace-android-sdk/src/test/java/io/embrace/android/embracesdk/event/EmbraceEventServiceTest.kt
```diff
@@ -353,64 +353,41 @@ internal class EmbraceEventServiceTest {
...
-        assertEquals(2, eventService.findEventIdsForSession(sessionBeginTime, fakeClock.now()).size)
-
-        // After logging another moment, the cache is busted so after the clock ticks forward, we get all 4 moments
-        eventService.startEvent("fourth")
-        fakeClock.setCurrentTime(fakeClock.now() + 1L)
-        assertEquals(4, eventService.findEventIdsForSession(sessionBeginTime, fakeClock.now()).size)
-    }
-
     @Test
     fun `verify close clears the existing events`() {
```
priettt, line 357: I feel like we could remove this test and just call close on the test below. Or maybe improve the naming on this one because "existing events" is a bit ambiguous.

priettt (approved):
LGTM! Love the tests :)

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/event/EmbraceEventService.kt
```diff
@@ -268,13 +263,13 @@ internal class EmbraceEventService(
...
     override fun close() {
         cleanCollections()
-        logDeveloper("EmbraceEventService", "close")
     }
 
     override fun cleanCollections() {
-        eventIds.clear()
-        activeEvents.clear()
-        logDeveloper("EmbraceEventService", "collections cleaned")
+        val activeEventIds = activeEvents.values.map { it.event.eventId }
```
bidetofevil, line 269: Does the active moments list get updated on timeout?

fractalwrench: Yep, the `EventDescription` contains a callback to `endEvent` that is invoked on the timeout. The item is removed from the collection at that point

embrace-android-sdk/src/test/java/io/embrace/android/embracesdk/event/EmbraceEventServiceTest.kt
fractalwrench: Fair enough - removed the old test case.

PR #542, opened 2024-03-11
Populate spans in new payload

fractalwrench:
## Goal

Populates spans in the new payload. This isn't hooked up to the orchestrator yet but will allow replacing the old payload when that happens. I've taken the same logic from `PayloadMessageCollator` for gathering spans & transplanted it here.

## Testing


lucaslabari (approved):
LGTM. 
You can wait for other reviews if you need to validate payload-specific fields.

PR #543, opened 2024-03-11
Populate additional fields on new payload

fractalwrench:
## Goal

Populates some additional fields on the new payload (not hooked up to be delivered yet). Specifically I've constructed the log + session payloads & tested they can be generated using fake implementations.

## Testing

Added unit test coverage.


lucaslabari (approved):
LGTM

PR #544, opened 2024-03-11
[EMBR-2802] Fix race condition on LogSink and LogOrchestrator.

lucaslabari:
## Goal

- Converted the list on LogSink into a `ConcurrentLinkedQueue` to use thread-safe operations. 
- Added a lock on the flush method, so no more than one thread can execute the flush at the same time. 
- Added a param to the flushLogs method to provide the max number of  logs to flush. 
- Converted properties that hold time in LogOrchestrator to AtomicLong. 

## Testing

Added a method created by Hanson to emulate the race condition.

### DIFF
--- a/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/internal/logs/LogSinkImpl.kt
+++ b/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/internal/logs/LogSinkImpl.kt
@@ -2,16 +2,16 @@ package io.embrace.android.embracesdk.internal.logs
 
 import io.opentelemetry.sdk.common.CompletableResultCode
 import io.opentelemetry.sdk.logs.data.LogRecordData
+import java.util.concurrent.ConcurrentLinkedQueue
 
 internal class LogSinkImpl : LogSink {
-    private val storedLogs: MutableList<EmbraceLogRecordData> = mutableListOf()
+    private val storedLogs: ConcurrentLinkedQueue<EmbraceLogRecordData> = ConcurrentLinkedQueue()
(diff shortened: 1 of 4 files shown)


embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/internal/logs/LogSinkImpl.kt
```diff
@@ -21,15 +21,16 @@ internal class LogSinkImpl : LogSink {
...
-    override fun flushLogs(): List<EmbraceLogRecordData> {
-        synchronized(storedLogs) {
-            val flushedLogs = storedLogs.toList()
-            storedLogs.clear()
+    override fun flushLogs(max: Int?): List<EmbraceLogRecordData> {
+        synchronized(flushLock) {
+            val maxIndex = max?.let {
+                minOf(storedLogs.size, it)
+            } ?: storedLogs.size
+            val flushedLogs = storedLogs.toList().subList(0, maxIndex)
```
bidetofevil, line 32: Since the list is ordered, is it possible just to do `storedLogs.take(maxIndex)` for simplicity and efficiency?

bidetofevil (approved):
LGTM! One comment about simplifying the flush logic but feel free to take it or leave it

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/internal/logs/LogSinkImpl.kt
lucaslabari: Nice... updated !!

leandro-godon (approved):

PR #547, opened 2024-03-12
Allow writes and read to not block SpanSinkImpl and only synchronize flush calls

bidetofevil:
## Goal

First step to persisting active spans is to sort out the concurrency story and ensuring that we don't lose data if the app crashes in a state where recorded spans are only in memory.

This change stores the flushed spans in a separate list that we can serialize before clearing the completedSpans queue.

## Testing

Added test to test race condition caused by concurrent access.

### DIFF
--- a/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/internal/spans/SpanSinkImpl.kt
+++ b/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/internal/spans/SpanSinkImpl.kt
@@ -2,15 +2,17 @@ package io.embrace.android.embracesdk.internal.spans
 
 import io.opentelemetry.sdk.common.CompletableResultCode
 import io.opentelemetry.sdk.trace.data.SpanData
+import java.util.Queue
+import java.util.concurrent.ConcurrentLinkedQueue
+import java.util.concurrent.atomic.AtomicReference
 
 internal class SpanSinkImpl : SpanSink {
-    private val completedSpans: MutableList<EmbraceSpanData> = mutableListOf()
+    private val completedSpans: Queue<EmbraceSpanData> = ConcurrentLinkedQueue()
+    private val spansToFlush = AtomicReference<List<EmbraceSpanData>>(listOf())
 
     override fun storeCompletedSpans(spans: List<SpanData>): CompletableResultCode {
         try {
-            synchronized(completedSpans) {
-                completedSpans += spans.map { EmbraceSpanData(spanData = it) }
-            }
+            completedSpans += spans.map { EmbraceSpanData(spanData = it) }
         } catch (t: Throwable) {
             return CompletableResultCode.ofFailure()
         }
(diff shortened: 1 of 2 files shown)


bidetofevil:
* **#547** <a href="https://app.graphite.dev/github/pr/embrace-io/embrace-android-sdk/547?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `master`

<h2></h2>

lucaslabari (approved):
LGTM

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/internal/spans/SpanSinkImpl.kt
```diff
@@ -19,16 +21,17 @@ internal class SpanSinkImpl : SpanSink {
     }
 
     override fun completedSpans(): List<EmbraceSpanData> {
-        synchronized(completedSpans) {
-            return completedSpans.toList()
-        }
+        val spansToReturn = completedSpans.size
+        return completedSpans.take(spansToReturn)
```
lucaslabari, line 25: Why is this needed? Instead of just returning `completedSpans.toList()`

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/internal/spans/SpanSinkImpl.kt
```diff
@@ -19,16 +21,17 @@ internal class SpanSinkImpl : SpanSink {
...
 
     override fun flushSpans(): List<EmbraceSpanData> {
-        synchronized(completedSpans) {
-            val flushedSpans = completedSpans.toList()
-            completedSpans.clear()
-            return flushedSpans
+        synchronized(spansToFlush) {
+            spansToFlush.set(completedSpans())
+            repeat(spansToFlush.get().size) {
+                completedSpans.removeAll(spansToFlush.get().toSet())
```
lucaslabari, line 32: Is this correct? removing all as many times as spans in the list?

lucaslabari (approved):
Left a couple of comments

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/internal/spans/SpanSinkImpl.kt
bidetofevil: Because we're not synchronizing this and and write, we can have the completedSpans queue being modified while we are taking a snapshot. Just to be predictable especially for the flush use case, we are doing this so we aren't trying to flush events that were added while the flush was taking place.

Once the flush is triggered, everything that comes after should not be part of what's being flushed.

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/internal/spans/SpanSinkImpl.kt
bidetofevil: hahaha this is ABSOLUTELY not correct - I will change.

I had a previous implementation that polled N times to remove and add to the flushed spans, but that leaves the the object in a weird state if it crashes in the middle. Doing this atomically works much better even if I have to convert this to a set first.

PR #549, opened 2024-03-12
Create payload module

fractalwrench:
## Goal

Creates a payload module that is responsible for creating classes that generate the new payload. This is hooked up but these objects are never instantiated are used right now.

## Testing

Added unit test coverage.


lucaslabari (approved):
LGTM

PR #551, opened 2024-03-12
Add interface for payload message collator

fractalwrench:
## Goal

Extracts an interface from the `PayloadMessageCollator` class. This will allow creation of fakes in tests & also is a precursor refactoring that will enable the creation of the new payload.

## Testing


lucaslabari (approved):
LGTM

PR #554, opened 2024-03-12
Implement Log Payload Source to build the Log Envelope

lucaslabari:
## Goal

- Added the implementation to `LogPayloadSourceImpl` so we can build the envelope for Logs. 

## Testing

Added unit tests


fractalwrench (approved):
LGTM

PR #555, opened 2024-03-13
Add new fields to SessionMessage class

fractalwrench:
## Goal

Adds new fields to the `SessionMessage` class. This will allow us to send the new style payload whilst maintaining backwards compatibility with the existing payload.

## Testing

Added test coverage.


lucaslabari (approved):
LGTM

PR #556, opened 2024-03-13
Add V2 payload message collator

fractalwrench:
## Goal

Adds the skeleton of a V2 payload message collator. Currently this just calls into the v1 collator but will be updated in a future changeset to populate the new fields.

## Testing

Reused test coverage from old collator.


lucaslabari (approved):
LGTM

PR #557, opened 2024-03-13
Populate fields from new payload

fractalwrench:
## Goal

Populates the `SessionMessage` with fields from the new payload. This is not hooked up yet so won't send any changes to our production API.

## Testing

Added unit test coverage.


bidetofevil (approved):
LGTM

PR #558, opened 2024-03-13
Add ability to create v2 collator from module

fractalwrench:
## Goal

Adds the ability to create a v2 payload collator in the modules. This is not hooked up to production code yet - that will come in a future PR.

## Testing

Updated unit tests.


leandro-godon (approved):
Lgtm

PR #559, opened 2024-03-13
Fix race condition for logs.

lucaslabari:
## Goal

There was a test failing because of a race condition when a log was added by another thread in the middle of the flushing event. 

## Testing

Relied on the existent test.


fractalwrench (approved):
LGTM

PR #560, opened 2024-03-13
Add remote config that controls whether v2 payload is sent or not

fractalwrench:
## Goal

Add remote config that controls whether a v2 session payload is sent or whether the SDK should default to the v1 payload scheme. This changeset defaults to using v1 but this will change soon once the new payload class is fully hooked up.

## Testing

Added unit tests.


lucaslabari (approved):
LGTM

bidetofevil:
LGTM

bidetofevil (approved):

PR #561, opened 2024-03-13
Switch payload message collator depending on config

fractalwrench:
## Goal

Switches the payload message collator depending on config. This is not hooked up to the delivery service yet so won't affect production payloads.

## Testing

Added unit tests.


leandro-godon (approved):
Lgtm

PR #563, opened 2024-03-13
Remove duplicated LogPayload.

lucaslabari:
## Goal

- Removed duplicated `LogPayload` class, so we use only the one in the `payload` package.
- I will remove other duplications in separate PRs.

## Testing

- Relied on existing tests. 
- Updated tests.


fractalwrench (approved):
LGTM

PR #565, opened 2024-03-14
Route new sessions to correct endpoint

fractalwrench:
## Goal

Routes v2 session payloads to the correct endpoint & ensures that the `GatingService` is aware they need to be sanitised. This changeset routes the new v2 session correctly if the remote config field is enabled (currently this is disabled by default).

A future changeset needs to handle cached sessions correctly. I plan on addressing this by encoding the version info in the payload so we don't need to load the session & introspect its contents.

## Testing

Added unit tests.

### DIFF
--- a/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/comms/api/ApiService.kt
+++ b/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/comms/api/ApiService.kt
@@ -3,7 +3,6 @@ package io.embrace.android.embracesdk.comms.api
 import io.embrace.android.embracesdk.config.remote.RemoteConfig
 import io.embrace.android.embracesdk.internal.payload.Envelope
 import io.embrace.android.embracesdk.internal.payload.LogPayload
-import io.embrace.android.embracesdk.internal.payload.SessionPayload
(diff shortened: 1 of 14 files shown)


embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/comms/api/ApiService.kt
```diff
@@ -74,5 +66,10 @@ internal interface ApiService {
      * @param crash the event message containing the crash
      */
     fun sendCrash(crash: EventMessage): Future<*>
-    fun sendSession(action: SerializationAction, onFinish: (() -> Unit)?): Future<*>?
+
+    /**
+     * Sends a session to the API. This can be either a v1 or v2 session - the implementation
+     * is responsible for routing the payload correctly.
+     */
+    fun sendSession(v2Payload: Boolean, action: SerializationAction, onFinish: (() -> Unit)?): Future<*>?
```
bidetofevil, line 74: Since this is no longer only controlling the payload but the endpoint as well, does it make sense to give the parameter a more generic name like `isV2`?

bidetofevil (approved):
LGTM. Minor comment about parameter name but logic is good to go.

PR #566, opened 2024-03-14
Alter where session gating happens

fractalwrench:
## Goal

Alters session gating so that it happens in the `PayloadMessageCollator` implementations. This facilitates the gating of the new session payload, which will happen in future changesets.

## Testing

Added unit tests.


bidetofevil (approved):
LGTM. This seems like a better place.

PR #567, opened 2024-03-14
Implement session gating for user personas in new session payload

fractalwrench:
## Goal

Implements session gating for user personas in the new session payload. This is a steel thread that allows us to expand gating to other fields as we move them over.

## Testing

Added unit tests.


bidetofevil (approved):
LGTM

PR #568, opened 2024-03-14
Encode session payload version in file

fractalwrench:
## Goal

This changeset encodes the session payload version in the file so it's possible to route cached sessions to the correct endpoint. I chose to consolidate most of the existing logic that is related to this change into the existing `CachedSession` class, so it's easier to reason about future changes to information encoded in filenames in future.

## Testing

Added unit tests.

### DIFF
--- a/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/comms/delivery/CachedSession.kt
+++ b/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/comms/delivery/CachedSession.kt
@@ -1,0 +1,74 @@
+package io.embrace.android.embracesdk.comms.delivery
+
+import io.embrace.android.embracesdk.logging.InternalStaticEmbraceLogger
+
+/**
+ * Represents a session that is cached on disk. The cached session encodes information in its
+ * filename such as sessionId, timestamp, and payload version. This allows the SDK to process
+ * cached sessions without deserializing the JSON.
+ */
+internal class CachedSession private constructor(
+    val sessionId: String,
+    val timestampMs: Long,
+    val filename: String,
+    val v2Payload: Boolean
+) {
+
+    companion object {
+        private const val V2_PREFIX = "v2"
+
+        /**
+         * Creates a [CachedSession] from a session id and timestamp.
+         */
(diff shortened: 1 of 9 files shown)


embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/comms/delivery/EmbraceDeliveryCacheManager.kt
```diff
@@ -88,24 +89,30 @@ internal class EmbraceDeliveryCacheManager(
     /**
      * This method will do disk reads so do not run it on the main thread
      */
-    override fun getAllCachedSessionIds(): List<String> {
+    override fun getAllCachedSessionIds(): List<CachedSession> {
         val sessionFileIds = cacheService.normalizeCacheAndGetSessionFileIds()
         sessionFileIds.forEach { filename ->
             val values = filename.split('.')
-            if (values.size != 4) {
+            if (values.size < 4 || values.size > 5) {
```
bidetofevil, line 96: I found this part kind of janky when writing tests for it early. Is it possible to abstract the validation away a bit, and do more validation on the assumptions that are later made (e.g. token 2 is a timestamp, token 3 is a session ID)? Reading this, it would be nice to just see a validation method and have all the gnarly logic in there.

bidetofevil (approved):
LGTM. One comment about the logic to validate the file name - we can probably do more to validate assumptions made later in the parsing, and pull it out to a function so this code is more readable. But it's not entirely related to your change, so feel free to do it later or not at all at your discretion.

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/comms/delivery/CachedSession.kt
```diff
@@ -0,0 +1,71 @@
...
+                timestampMs,
+                filename,
+                v2Payload
+            )
+        }
+
+        /**
+         * Creates a [CachedSession] from a filename.
+         */
+        fun fromFilename(filename: String): CachedSession? {
```
bidetofevil, line 40: Nice! Should we try to log an error or warning if we get a file we think is a session but somehow doesn't fit with the expected format?

bidetofevil (approved):
LGTM x2

PR #569, opened 2024-03-14
Remove duplicate EmbraceLogRecordData class

leandro-godon:
## Goal

Remove `EmbraceLogRecordData` and `EmbraceLogBody` and use the payload classes instead.


lucaslabari (approved):
LGTM. Left a comment.

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/internal/payload/Log.kt
```diff
@@ -53,4 +54,14 @@ internal data class Log(
      If span_id is present, trace_id SHOULD be also present. */
     @Json(name = "span_id")
     val spanId: String? = null
-)
+) {
+    internal constructor(logRecordData: LogRecordData) : this(
```
lucaslabari, line 58: Can we add this, given this file will be autogenerated? If there is risk to be overwritten we can move it to LogMapper.

leandro-godon: I removed it and left an extension function to do the conversion.

PR #570, opened 2024-03-14
Fix functional tests workflow

lucaslabari:
## Goal

- Update the workflow to make it fail when a functional test fails
- Send the URL of the current action to slack. 

## Testing

Tested manually.


bidetofevil (approved):
LGTM

PR #571, opened 2024-03-14
Fix some inconsistencies/bug with internal api

nelsitoPuglisi:
## Goal

Some internal API where writing the preferences service directly, instead of storing the value in-memory using metadata service. 

## Testing

Added some integration test to verify that still the preferences service is used when there is no value.


bidetofevil:
LGTM. I want to refactor this so you don't have to pass the framework into the constructor and just use start(), but that'll be for later

bidetofevil (approved):

PR #572, opened 2024-03-15
[Do not merge] Send v2 session payload by default

fractalwrench:
## Goal

This changeset sends the v2 session payload by default. 

## Testing

### DIFF
--- a/embrace-android-sdk/src/androidTest/assets/golden-files/session-end.json
+++ b/embrace-android-sdk/src/androidTest/assets/golden-files/session-end.json
@@ -98,70 +98,52 @@
       }
     ]
   },
-  "u": {
-    "id": "some id",
-    "em": "user@email.com",
-    "un": "John Doe",
-    "per": [
-      "first_day"
-    ]
+  "resource": {},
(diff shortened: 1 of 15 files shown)


embrace-android-sdk/src/androidTest/assets/golden-files/session-end.json
```diff
@@ -1,167 +1,262 @@
...
+    "email":"user@email.com",
+    "username":"John Doe",
+    "personas":"__EMBRACE_TEST_IGNORE__",
+    "timezone_description":"__EMBRACE_TEST_IGNORE__",
+    "locale":"__EMBRACE_TEST_IGNORE__"
+  },
+  "version":"0.1.0",
+  "type":"session",
+  "data":{
+    "spans":[
```
fractalwrench, line 217: @bidetofevil I spotted that the number of spans is different compared to the legacy payload. I think this is probably related to setting it in 2 places in the same payload. Do you think it's reasonable to only set it in 1 place, or perhaps to extract the flushing of spans (and any similar data) to one place?

PR #574, opened 2024-03-15
Make old fields null where possible

fractalwrench:
## Goal

Makes old fields in the payload null where possible. Some need to remain populated as we haven't fully migrated everything yet.

This only applies to the v2 session payload which is disabled by default, so should have no effect on our production payload.

## Testing

Updated unit tests.


bidetofevil (approved):
LGTM

PR #580, opened 2024-03-17
Fix race condition in Log Sink.

lucaslabari:
## Goal

- There was a race condition with the `take()` method. If `storedLogs` is modified by another thread after calling `take(maxIndex)`, the  `flushedLogs` list could contain more elements than expected. 
- Replaced `take()` with `toList().sublist()` that returns a copy and will always have a fixed number of elements. 

## Testing

- Reproduced with the test and always getting green after the changes.

### DIFF
--- a/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/internal/logs/LogSinkImpl.kt
+++ b/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/internal/logs/LogSinkImpl.kt
@@ -32,7 +32,7 @@ internal class LogSinkImpl : LogSink {
             val maxIndex = max?.let {
                 minOf(currentSize, it)
             } ?: currentSize
-            val flushedLogs = storedLogs.take(maxIndex)
+            val flushedLogs = storedLogs.toList().subList(0, maxIndex)
             storedLogs.removeAll(flushedLogs.toSet())
             return flushedLogs
         }
--- a/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/internal/logs/LogSinkImpl.kt
+++ b/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/internal/logs/LogSinkImpl.kt
@@ -1,5 +1,6 @@
 package io.embrace.android.embracesdk.internal.logs
 
+import io.embrace.android.embracesdk.internal.logs.LogOrchestrator.Companion.MAX_LOGS_PER_BATCH
 import io.embrace.android.embracesdk.internal.payload.Log
 import io.embrace.android.embracesdk.internal.payload.toNewPayload
 import io.opentelemetry.sdk.common.CompletableResultCode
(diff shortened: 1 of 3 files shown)


bidetofevil:
I think this has the same issue because `currentSize` is the thing that is not synchronized because the size() function for a `ConncurrentLinkedQueue` isn't synchronized (see the Javadocs).

So if `max` is null, it'll use a size that is potentially just what the current size is, which could be anything.

If you add a max size to the test when you call `flushLogs`, it should fix this.

To more proactively prevent this, we want want to hardcode the max size here - unless there is a case we don't a max batch size?

lucaslabari:
@bidetofevil that's not the issue that is making the test to fail. 
`flushLogs` is always being called with `MAX_LOGS_PER_BATCH` and never with null. I'll update this so it doesn't accept null or just use the constant and don't need the param. 
But anyway, that isn't the problem, I was able to reproduce almost 100% of the times by modifying the test to run 100 times and reducing the MAX_LOGS_PER_BATCH. 
After reproducing and adding some logs to show the value of `flushedLogs` at different points in time, I was able to see that `flushedLogs` has a greater size than the `maxIndex` passed in the `take(maxIndex)` because of concurrent executions. So, if items are being added while doing `val flushedLogs = storedLogs.take(50)`, it could return a list with more than 50 items. 
That's why doing `val flushedLogs = storedLogs.toList().subList(0, maxIndex)` worked. Because `toList()` makes a copy and then we obtain the sublist. we could also do `val flushedLogs = storedLogs.toList().take(maxIndex)` and it will work. 
The problem is trying to do the `take(maxIndex)` over a list that is being updated by other threads. 
If you want I can send you a branch with the tweaks and logs I did in the test to make it fail most of the times.

bidetofevil:
With `ConcurrentLinkedQueue`, it's ok to read the front of the queue concurrently with insertions - it's even ok with take from the front as long as you know there are enough.

Your solution works because it's doing a snapshot of the state of the list, so it doesn't matter of `maxIndex` is higher than the total number of elements you have. 

But you should probably add a line to validate that `maxIndex` is what you expect. If you call `take(50)` in the original queue without `sublist()` that should also return what you expect.

And perhaps it will help to add a test that passes a `max` that will illustrate what I mean?

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/internal/logs/LogSinkImpl.kt
```diff
@@ -26,13 +27,11 @@ internal class LogSinkImpl : LogSink {
...
-    override fun flushLogs(max: Int?): List<Log> {
+    override fun flushLogs(): List<Log> {
         synchronized(flushLock) {
             val currentSize = storedLogs.size
-            val maxIndex = max?.let {
-                minOf(currentSize, it)
-            } ?: currentSize
-            val flushedLogs = storedLogs.take(maxIndex)
+            val maxIndex = minOf(currentSize, MAX_LOGS_PER_BATCH)
+            val flushedLogs = storedLogs.toList().take(maxIndex)
```
bidetofevil, line 34: If you remove `toList()` this will work as well.

bidetofevil (approved):
So Lucas and I found that `take` is not synchronized for `ConcurrentLinkQueue` and can return more than the N elements requested. Lets merge this as is and we can fix this later so we don't have to take a sublist.

PR #584, opened 2024-03-18
Ensure spans aren't flushed twice

fractalwrench:
## Goal

Ensures that spans aren't flushed twice as the `V1PayloadMessageCollator` is called by the `V2PayloadMessageCollator` when generating the payload. This should be a temporary state of affairs, but the fix is to disable flushing of spans within that class.

## Testing


lucaslabari (approved):
LGTM

PR #586, opened 2024-03-18
Alter new session endpoint

fractalwrench:
## Goal

Alters the new endpoint as we decided to rename it.

## Testing


lucaslabari (approved):
LGTM

PR #588, opened 2024-03-18
Use v1 endpoint for now

fractalwrench:
## Goal

Send v2 payloads to the v1 endpoint for now, as it's capable of processing incomplete payloads whereas the v2 endpoint will not accept legacy fields.


lucaslabari (approved):
LGTM

PR #589, opened 2024-03-18
Remove unused Log method

leandro-godon:
## Goal

Remove a method that was only being used in tests.

## Testing

Updated tests and created a log with the sample app.


fractalwrench (approved):
LGTM

PR #593, opened 2024-03-18
[EMBR-3054] Generate LogEnvelope from LogOrchestrator

lucaslabari:
## Goal

- Logs were passed to the `DeliveryService`, but the envelope was not being generated through the `EnvelopeSource` in order to include metadata, resources, etc. 
- Now the `LogOrchestrator` asks the `EnvelopeSource` for the `envelope` and sends it to the `DeliveryService`. 
- `LogPayloadSource` is now responsible for doing the `flush()` on the `LogSink` and returning the `LogPayload` to form the `envelope`. 
## Testing

- Updated existing tests.


fractalwrench (approved):
LGTM

leandro-godon (approved):

PR #594, opened 2024-03-18
Add app framework value to v2 logs

leandro-godon:
## Goal

Add the "app framework" attribute for logs that was missing.


fractalwrench (approved):
LGTM

PR #595, opened 2024-03-19
Make EmbraceSpan snapshottable

bidetofevil:
## Goal

We need to persist active spans when we have unexpected terminations so we can send the data associated with them in case we can't terminate them in time. 

To do this, I will first make their current state serializable as without needing to end them and make the OTel SDK turn them into `EmbraceSpanData`. 

The tricky part of this is that the `Span` returned by the OTel SDK after a span has started cannot be read - that is, key details like the name, start/end times, in additional to attributes and events, cannot be accessed until a span goes through the export process after it is stopped. Boo!

To get around this, we keep all the data we are interested in in the wrapper, and only add it to the OTel Span object when we are about to stop it. That way, we can always snapshot an `EmbraceSpan`.

## Testing

Add unit tests to verify the snapshot. The logic to move the span data to the wrapper layer is validated by the existing tests.

### DIFF
--- a/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/internal/spans/CurrentSessionSpanImpl.kt
+++ b/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/internal/spans/CurrentSessionSpanImpl.kt
@@ -17,7 +17,7 @@ import java.util.concurrent.atomic.AtomicInteger
 import java.util.concurrent.atomic.AtomicReference
 
 internal class CurrentSessionSpanImpl(
-    private val clock: Clock,
+    private val openTelemetryClock: Clock,
     private val telemetryService: TelemetryService,
     private val spanRepository: SpanRepository,
     private val spanSink: SpanSink,
@@ -81,7 +81,7 @@ internal class CurrentSessionSpanImpl(
             if (appTerminationCause == null) {
                 endingSessionSpan.stop()
                 spanRepository.clearCompletedSpans()
-                sessionSpan.set(startSessionSpan(clock.now().nanosToMillis()))
+                sessionSpan.set(startSessionSpan(openTelemetryClock.now().nanosToMillis()))
             } else {
                 endingSessionSpan.setEmbraceAttribute(appTerminationCause)
                 endingSessionSpan.stop()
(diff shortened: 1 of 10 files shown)


bidetofevil:
* **#597** <a href="https://app.graphite.dev/github/pr/embrace-io/embrace-android-sdk/597?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#596** <a href="https://app.graphite.dev/github/pr/embrace-io/embrace-android-sdk/596?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#595** <a href="https://app.graphite.dev/github/pr/embrace-io/embrace-android-sdk/595?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `master`

<h2></h2>

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/internal/spans/CurrentSessionSpanImpl.kt
```diff
@@ -107,15 +107,21 @@ internal class CurrentSessionSpanImpl(
     private fun startSessionSpan(startTimeMs: Long): EmbraceSpan {
         traceCount.set(0)
 
+        val spanName = "session".toEmbraceSpanName()
```
bidetofevil, line 110: Had to move the logic to prepend `emb` outside side of the SpanBuilder because we need to access it for snapshot purposes, AND we also need to pass it in as a parameter to the friggin builder. I hate this, but not much we can do. The good thing is that this is contained - the logic is here and in SpanServiceImpl.

fractalwrench (approved):
LGTM.

It's worth mentioning that currently OTel implements spans via the `SdkSpan` class which implements the `ReadWrite` interface, which might also achieve what you want. However, these are all non-public APIs so this solution is probably better

bidetofevil:
Yeah, I thought about doing a hard cast and reading it from there, but that seems more brittle in case they do a refactor that changes that relationship - it also ties us explicitly to `SdkSpan`, which we aren't right now.

Doing it this way seems just as good and insulates us from their implementation. It could even form the basis of our own OTel SDK implementation too, but that's more of a future thing.

bidetofevil:
### Merge activity

* **Mar 19, 12:53 PM EDT**: @bidetofevil started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/embrace-io/embrace-android-sdk/595).

PR #598, opened 2024-03-19
Add integration tests for features

fractalwrench:
## Goal

Adds integration tests for custom breadcrumbs, view breadcrumbs, and AEI. These will help us assert that functionality is still working as expected when migrating features to OTel.


lucaslabari (approved):
LGTM

PR #600, opened 2024-03-19
Switch between event and OTel log services based on configuration

leandro-godon:
## Goal

Create a `CompositeLogService` that can switch between `LogMessageService` and `LogService` based on a remote config.

## Testing

Updated tests and ran the test app.


lucaslabari (approved):
LGTM

PR #604, opened 2024-03-20
Add span assertions for tests

fractalwrench:
## Goal

Adds assertions that will be used for writing tests on spans. Additionally refactored the `SchemaType` so that string literals are defined in constants.

### DIFF
--- a/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/arch/schema/SchemaType.kt
+++ b/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/arch/schema/SchemaType.kt
@@ -1,8 +1,19 @@
 package io.embrace.android.embracesdk.arch.schema
 
+import io.embrace.android.embracesdk.arch.schema.SchemaKeys.AEI_RECORD
+import io.embrace.android.embracesdk.arch.schema.SchemaKeys.CUSTOM_BREADCRUMB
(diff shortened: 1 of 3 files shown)


embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/arch/schema/SchemaType.kt
```diff
@@ -1,9 +1,20 @@
...
+internal object SchemaKeys {
+    internal const val CUSTOM_BREADCRUMB = "custom-breadcrumb"
+    internal const val VIEW_BREADCRUMB = "view-breadcrumb"
+    internal const val AEI_RECORD = "aei-record"
+    internal const val LOG = "emb-log"
+}
+
 internal sealed class SchemaType(
     val telemetryType: TelemetryType,
     val name: String,
```
lucaslabari, line 20: nit: maybe we can do this more restrictive and turn it to be of type `SchemaKey` and `SchemaKey` be an enum. 
But it's just a detail... it's ok if you prefer it this way.

fractalwrench: I will keep this as-is for now since `SchemaType` is already a sealed class, but we can maybe reassess if any other functionality ever gets added to the string constants?

PR #605, opened 2024-03-20
Enable custom breadcrumb OTel capture

fractalwrench:
## Goal

Enables the capture of custom breadcrumbs via OTel. Custom breadcrumbs are captured as a span event on `emb-session-span`. The legacy mechanism that adds directly to the session payload is running simultaneously - a future PR will remove this mechanism.

## Testing

Added unit tests.


nelsitoPuglisi (approved):

bidetofevil (approved):
LGTM!

BTW, I think I'm going to take a stab at formally linking SchemaType with `TelemetryType` so that you get to everything about a TelemetryType from it,

PR #607, opened 2024-03-20
Enable view breadcrumb OTel capture

fractalwrench:
## Goal

Enables the capture of view breadcrumbs via OTel. View breadcrumbs are captured as spans. The legacy mechanism that adds directly to the session payload is running simultaneously - a future PR will remove this mechanism.

## Testing

Added unit tests.

### DIFF
--- a/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/arch/schema/SchemaType.kt
+++ b/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/arch/schema/SchemaType.kt
@@ -9,7 +9,7 @@ import io.embrace.android.embracesdk.internal.utils.toNonNullMap
 import io.embrace.android.embracesdk.payload.AppExitInfoData
 
 internal object SchemaKeys {
-    internal const val CUSTOM_BREADCRUMB = "custom-breadcrumb"
+    internal const val CUSTOM_BREADCRUMB = "emb-custom-breadcrumb"
     internal const val VIEW_BREADCRUMB = "view-breadcrumb"
     internal const val AEI_RECORD = "aei-record"
     internal const val LOG = "emb-log"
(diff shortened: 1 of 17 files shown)


embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/arch/schema/SchemaType.kt
```diff
@@ -10,7 +10,7 @@ import io.embrace.android.embracesdk.payload.AppExitInfoData
 
 internal object SchemaKeys {
     internal const val CUSTOM_BREADCRUMB = "emb-custom-breadcrumb"
-    internal const val VIEW_BREADCRUMB = "view-breadcrumb"
+    internal const val VIEW_BREADCRUMB = "screen-view"
```
lucaslabari, line 13: Would be good to be consistent with the usage of the `emb-` prefix for these keys.

bidetofevil: If these are the canonical name of the telemetry generated by this schema, we should look to define them in such a away that we don't have to attach the prefix with each. With that, we can have one place where we decide whether to do it.

bidetofevil: Agreed. Fixing this here one way or the other for now would be good, but long term, we should really define the canonical names sans the prefix and just attach them then necessary.

bidetofevil (approved):
LGTM. Feel free to make the use of the prefix consistent, but no worries if you don't as we'll need to address that more thoroughly later anyway.

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/arch/schema/SchemaType.kt
fractalwrench: I'll skip this for now as this PR has been pretty held up already, but will address in a future changeset (if Hanson's schema type changes don't already address this)

PR #608, opened 2024-03-20
Remove legacy view breadcrumb collection

fractalwrench:
## Goal

Builds on #607 by removing the collection of view breadcrumbs via the legacy mechanism.

## Testing

Verified E2E that view breadcrumbs are still shown in session when delivered only as spans.


lucaslabari (approved):
LGTM. Is backend already processing breadcrumbs as spans?

bidetofevil (approved):
🔥

PR #609, opened 2024-03-20
Enable AEI data capture via OTel

fractalwrench:
## Goal

Enables the capture of AEI via OTel as a log. The legacy mechanism that adds directly to the session payload is also removed in this PR due to the size of the AEI trace.

## Testing

Added unit tests + an integration test.

I tested this E2E in our staging environment & confirmed that AEI show up in the dashboard.


bidetofevil (approved):
LGTM! Anything that we move from the session into Logs is going to be a no-going-back migration, but I think this is specific and save enough, so all good.

PR #617, opened 2024-03-23
Convert EmbraceDeliveryServiceTest to use a real DeliveryCacheManager

bidetofevil:
## Goal

The old test was not testing certain functionality that required the DeliveryCacheManager's implementation because it was using a fake. Convert this to use the real thing so the tests actually test the component in a realistic way.

Seeing as we don't have integration tests that cover this layer, it's crucial that these tests actually very the workflow happening in production

## Testing

Added to the fake ApiService in order to verify the behaviour at that layer.

PR #616, opened 2024-03-23
Use the session start time for the file name used to store a session

bidetofevil:
## Goal

Use the session start time for the file name rather than the time session was first saved. While it doesn't effect how it's used, this will make the file name deterministic and predictable, which wasn't the case before.

## Testing

A test was added in [this PR](https://github.com/embrace-io/embrace-android-sdk/pull/620) that tests that the file name doesn't matter when delivering a cached session file.

PR #618, opened 2024-03-23
Convert snapshots in cached sessions before sending

bidetofevil:
## Goal

Convert snap snapshots found in cached sessions found at SDK startup to failed spans. This ensures that the session span with all the telemetry that will be stored it will always be delivered even if the app process terminates before we could end it.

## Testing

Added tests in various layers to cover this. Codecov doesn't like how much test coverage I added because of the nullability of the new Span payload object and I didn't test those null case. Ah well.

An integration test will be added later that tests the case when the session span terminates. Stay tuned.

### DIFF
--- a/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/comms/delivery/EmbraceDeliveryService.kt
+++ b/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/comms/delivery/EmbraceDeliveryService.kt
@@ -108,10 +108,12 @@ internal class EmbraceDeliveryService(
 
             allSessions.map { it.sessionId }.forEach { sessionId ->
                 cacheManager.transformSession(sessionId = sessionId) { sessionMessage ->
-                    val spansToFail = sessionMessage.spanSnapshots?.map {
-                        it.toFailedSpan(sessionMessage.session.endTime ?: 0L)
-                    } ?: emptyList()
-                    val completedSpans = spansToFail + (sessionMessage.spans ?: emptyList())
+                    val completedSpanIds = sessionMessage.spans?.map { it.spanId }?.toSet() ?: emptySet()
+                    val spansToFail = sessionMessage.spanSnapshots
+                        ?.filterNot { completedSpanIds.contains(it.spanId) }
+                        ?.map { it.toFailedSpan(sessionMessage.session.endTime ?: 0L) }
+                        ?: emptyList()
+                    val completedSpans = (sessionMessage.spans ?: emptyList()) + spansToFail
                     sessionMessage.copy(spans = completedSpans, spanSnapshots = emptyList())
                 }
             }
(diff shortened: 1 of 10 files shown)


bidetofevil:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/embrace-io/embrace-android-sdk/618?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#618** <a href="https://app.graphite.dev/github/pr/embrace-io/embrace-android-sdk/618?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#617** <a href="https://app.graphite.dev/github/pr/embrace-io/embrace-android-sdk/617?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#616** <a href="https://app.graphite.dev/github/pr/embrace-io/embrace-android-sdk/616?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `master`

<h2></h2>

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/comms/delivery/EmbraceDeliveryService.kt
```diff
@@ -99,6 +100,17 @@ internal class EmbraceDeliveryService(
...
                 it.sessionId != sessionIdTracker.getActiveSessionId()
             }
+
+            allSessions.map { it.sessionId }.forEach { sessionId ->
+                cacheManager.transformSession(sessionId = sessionId) { sessionMessage ->
+                    val spansToFail = sessionMessage.spanSnapshots?.map {
+                        it.toFailedSpan(sessionMessage.session.endTime ?: 0L)
+                    } ?: emptyList()
+                    val completedSpans = spansToFail + (sessionMessage.spans ?: emptyList())
+                    sessionMessage.copy(spans = completedSpans, spanSnapshots = emptyList())
```
fractalwrench, line 110: If `transformSession` is called more than once then it looks like this would result in duplicate spans in the payload

fractalwrench (commented):
Mostly looks good, I think the transform could lead to duplicate spans if it ever runs multiple times (e.g. due to the process terminating or crashing halfway through)

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/comms/delivery/EmbraceDeliveryService.kt
bidetofevil: Oooh, good call. I'll fix and write a test for it. Didn't think this could happen but definitely should protect against that. The good thing is the server will reject dupes but that's unnecesary

bidetofevil: Made it so that if a conversion was already done or if a completed span already exists for that spanId, we just drop the snapshot

fractalwrench (approved):
LGTM

bidetofevil:
### Merge activity

* **Mar 26, 3:09 AM EDT**: @bidetofevil started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/embrace-io/embrace-android-sdk/618).

PR #625, opened 2024-03-25
Avoid blocking signals that are not SIGUSR2

fractalwrench:
## Goal

Blocking signals from program errors will probably lead to [undesirable results](https://www.gnu.org/software/libc/manual/html_node/Process-Signal-Mask.html). Therefore our signal handler for Unity thread sampling should only block SIGUSR2 to prevent reentrant calls.

## Testing

Tested that stacktraces are still captured.


fnewberg (approved):
one of the weird things i saw was that the crashes reported via AEI had the signal set to SIGUSR2. do you think this will change so that the real crash reason would show up?

PR #626, opened 2024-03-25
Synchronise access to unity sampler with mutex

fractalwrench:
## Goal

Synchronises access to the Unity sampler with a pthread mutex. While we can't use mutexes in the signal handler, we can use it for any code outside of the handler to ensure starting/stopping the timer doesn't result in an unexpected state with undesirable results. 

## Testing

Manually verified that stacktraces are still captured.


fractalwrench:
Updated changeset to include extra checks for `if (result == 0)`

priettt (approved):
LGTM! If you continue doing changes on this file it might be worth removing the dev logs

PR #627, opened 2024-03-25
Add version and type for Log envelope.

lucaslabari:
## Goal

- Add version and type for Log envelope.

## Testing

Relied on existing tests.


leandro-godon (approved):
Lgtm

PR #628, opened 2024-03-25
Replace LogBody with String.

lucaslabari:
## Goal

- Replace `LogBody` with `String` in Log object. 

## Testing

- Updated unit tests.


leandro-godon (approved):
Lgtm

PR #629, opened 2024-03-25
Send Int value instead of String for AppFramework in EnvelopeResource.

lucaslabari:
## Goal

- We were sending a `String (e.g: "1")` for `NATIVE`, but backend needs an Int (`e.g: 1`)
- Updated adding an adapter to convert to int when serializing and from int when deserializing. 

## Testing

- Added unit tests.


leandro-godon (approved):
Lgtm

PR #630, opened 2024-03-26
Add unit test framework for NDK

fractalwrench:
## Goal

Adds a unit test framework for the NDK. This will run as an instrumentation test so automatically gets picked up by the functional test workflow. I've picked [greatest](https://github.com/silentbicycle/greatest/) as it's simple (1 header file vendored into a project) and I've used it before so we won't be starting from zero for knowledge on our test framework.

For now I've just written one test case that validates a string is copied correctly. In future we can expand this test coverage as desired.

### DIFF
--- a/embrace-android-sdk/src/androidTest/java/io/embrace/android/embracesdk/ndk/EmbraceNativeLayerTest.kt
+++ b/embrace-android-sdk/src/androidTest/java/io/embrace/android/embracesdk/ndk/EmbraceNativeLayerTest.kt
@@ -1,0 +1,26 @@
+package io.embrace.android.embracesdk.ndk
+
+import org.junit.Assert.assertEquals
+import org.junit.Test
+
+internal class EmbraceNativeLayerTest {
+
+    companion object {
+        init {
+            System.loadLibrary("embrace-native")
+            System.loadLibrary("embrace-native-test")
+        }
+    }
+
+    external fun run(): Int
+
+    @Test
+    fun testNativeLayer() {
+        val status = run()
+        assertEquals(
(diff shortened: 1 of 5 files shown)


embrace-android-sdk/src/androidTest/java/io/embrace/android/embracesdk/ndk/EmbraceNativeLayerTest.kt
```diff
@@ -0,0 +1,26 @@
...
+            System.loadLibrary("embrace-native-test")
+        }
+    }
+
+    external fun run(): Int
+
+    @Test
+    fun testNativeLayer() {
+        val status = run()
+        assertEquals(
```
priettt, line 20: why do we expect an error here?

priettt (approved):
heh first time seeing an ndk unit test, nice!

embrace-android-sdk/src/androidTest/java/io/embrace/android/embracesdk/ndk/EmbraceNativeLayerTest.kt
fractalwrench: This is perhaps a little confusing as it's on multiple lines, but the first param is the message used by JUnit if the test fails. We're asserting that the status is 0 (normal termination) as the native code will return 1 if a test failed. I just wanted the error message to be a bit clearer when failures do happen

priettt: ohhhh never used assertEquals like that, sorry! 🤦

PR #631, opened 2024-03-26
Send exceptions as events even if OTel logs are enabled

leandro-godon:
## Goal

The backend is not processing exceptions as OTel logs yet, so send them as events even if the new payload remote config is enabled.


lucaslabari (approved):
LGTM

PR #637, opened 2024-03-27
Include spanId and traceId in OTel logs only if valid.

lucaslabari:
## Goal

- Even when doesn't apply we were sending spanId and traceId in this way
```
        "span_id": "0000000000000000",
        "trace_id": "00000000000000000000000000000000"
```
which is the INVALID value. 
- As these are optional in the Log Model, it's better not to set anything to them. 

## Testing


leandro-godon (approved):
Lgtm

PR #638, opened 2024-03-27
Reorganise how NDK files are structured

fractalwrench:
## Goal

Restructures the NDK project to place source files in more descriptive directories. No code change has been made as part of this.


priettt (approved):
Nice!

PR #639, opened 2024-03-27
Tweak SchemaType to consolidate logic for creating attributes

bidetofevil:
## Goal

Add functionality to SchemaType so data capture services and the infra around adding telemetry can leverage that solely to populate objects. After this change, calling `attributes()` on SchemaType will return all attributes that are specific to this schema, including the `emb.type` the schema is associated with.

## Testing

Modified existing tests to ensure things work and changed some to leverage the new testing infra.

### DIFF
--- a/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/gating/SpanSanitizer.kt
+++ b/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/gating/SpanSanitizer.kt
@@ -1,8 +1,9 @@
 package io.embrace.android.embracesdk.gating
 
-import io.embrace.android.embracesdk.arch.schema.SchemaKeys
+import io.embrace.android.embracesdk.arch.schema.EmbType
 import io.embrace.android.embracesdk.gating.SessionGatingKeys.BREADCRUMBS_CUSTOM
 import io.embrace.android.embracesdk.internal.spans.EmbraceSpanData
+import io.embrace.android.embracesdk.internal.spans.hasEmbraceAttribute
 import io.embrace.android.embracesdk.spans.EmbraceSpanEvent
 
 internal class SpanSanitizer(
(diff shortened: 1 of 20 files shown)


bidetofevil:
* **#639** <a href="https://app.graphite.dev/github/pr/embrace-io/embrace-android-sdk/639?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `master`

<h2></h2>

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/gating/SpanSanitizer.kt
```diff
@@ -38,17 +39,11 @@ internal class SpanSanitizer(
     }
 
     private fun sanitizeSpans(span: EmbraceSpanData): Boolean {
-        if (span.name == "emb-${SchemaKeys.VIEW_BREADCRUMB}" && !shouldAddViewBreadcrumbs()) {
-            return false
-        }
-        return true
+        return !(span.hasEmbraceAttribute(EmbType.Ux.View) && !shouldAddViewBreadcrumbs())
```
fractalwrench, line 42: nit: would marginally prefer the previous pattern. While this is a reasonable thing to change for just two checks, the gating will make dozens of conditional checks

fractalwrench (approved):
LGTM

lucaslabari (approved):

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/gating/SpanSanitizer.kt
bidetofevil: Got it. So you refer to use the early return pattern? I can restore this. Agreed that if there are more than 3 checks, this is a bad idiom to follow.

But given there's another change that builds on this. is it OK if I restored this pattern post-merge?

PR #640, opened 2024-03-27
Add tests for emb_unwind_dlinfo

fractalwrench:
## Goal

Adds unit tests for the `emb_unwind_dlinfo` source file that truncates a stacktrace & symbolicates it using `dladdr`. I have also removed some dead code & moved structs/functions into separate source files where appropriate.


priettt (approved):
Approved 👍

PR #641, opened 2024-03-27
Add session properties to OTel Logs.

lucaslabari:
## Goal

- Added session properties to OTel log attributes with prefix `emb.properties.`

## Testing

- Added unit test to verify session properties are added correctly.


bidetofevil (approved):
LGTM

PR #642, opened 2024-03-27
Replace "active" app state with "foreground"

lucaslabari:
## Goal

- Replace `active` app state with `foreground`. So, instead of having `active` and `background` we have `foreground` and `background`.

## Testing


bidetofevil:
LGTM

bidetofevil (approved):

PR #643, opened 2024-03-28
Reduce the direct use of the static logger so tests functional tests work properly

bidetofevil:
## Goal

Removed most of the direct use of the static logger, and instead rely on an instance that gets passed in that will be created in the InitModule. I also removed a lot of useless dev logging.

The crux of he change involved using the instance that get passed down via the modules init chain, and in places where it's not convenient, rely on the static logger instead. Both are hooked into the same internal error service instance so internal exceptions are reported.

The static logger will be slimmed down then completely removed, but I couldn't get away from having at least one massive change. There was no easy way of teasing these out without risking mistakes, and most of the changes are at least trivial. 

There are other refactors that I wanted to do and they were done in subsequent PRs. 

I'll will point to the right places to look but most of the other changes are just removing dev logging, adding the logger to a constructor, and passing it in when creating instances.

## Testing

Tests were modified as needed. Relying on existing coverage to ensure functionality worked as before, except for adding a test that verifies the static and instance loggers are both hooked up to the same internal error service.

### DIFF
--- a/embrace-android-sdk/src/integrationTest/kotlin/io/embrace/android/embracesdk/testcases/PublicApiTest.kt
+++ b/embrace-android-sdk/src/integrationTest/kotlin/io/embrace/android/embracesdk/testcases/PublicApiTest.kt
@@ -4,8 +4,11 @@ import android.os.Build.VERSION_CODES.TIRAMISU
 import androidx.test.ext.junit.runners.AndroidJUnit4
 import io.embrace.android.embracesdk.Embrace.AppFramework
 import io.embrace.android.embracesdk.IntegrationTestRule
+import io.embrace.android.embracesdk.assertions.assertInternalErrorLogged
 import io.embrace.android.embracesdk.internal.ApkToolsConfig
 import io.embrace.android.embracesdk.internal.TraceparentGeneratorTest.Companion.validPattern
+import io.embrace.android.embracesdk.internalErrorService
+import io.embrace.android.embracesdk.logging.InternalStaticEmbraceLogger
 import io.embrace.android.embracesdk.recordSession
 import org.junit.Assert.assertEquals
 import org.junit.Assert.assertFalse
@@ -134,6 +137,25 @@ internal class PublicApiTest {
             repeat(100) {
                 assertTrue(validPattern.matches(embrace.generateW3cTraceparent()))
             }
+        }
+    }
+
+    @Test
+    fun `static logger and SDK logger instance log to the same internal error service`() {
+        with(testRule) {
+            embrace.start(harness.fakeCoreModule.context)
+            InternalStaticEmbraceLogger.logger.logError("not-used", RuntimeException("static"), true)
+            assertInternalErrorLogged(
+                checkNotNull(internalErrorService()).currentExceptionError,
+                checkNotNull(RuntimeException::class.qualifiedName),
+                "static"
+            )
+            harness.initModule.logger.logError("not-used", RuntimeException("non-static"), true)
+            assertInternalErrorLogged(
+                checkNotNull(internalErrorService()).currentExceptionError,
+                checkNotNull(RuntimeException::class.qualifiedName),
+                "non-static"
+            )
         }
     }
 }
--- a/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/Embrace.java
+++ b/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/Embrace.java
@@ -1,7 +1,5 @@
 package io.embrace.android.embracesdk;
 
-import static io.embrace.android.embracesdk.logging.InternalStaticEmbraceLogger.logger;
-
 import android.annotation.SuppressLint;
 import android.content.Context;
 import android.webkit.ConsoleMessage;
@@ -15,8 +13,6 @@ import java.util.Map;
 import io.embrace.android.embracesdk.annotation.InternalApi;
 import io.embrace.android.embracesdk.internal.EmbraceInternalInterface;
 import io.embrace.android.embracesdk.internal.Systrace;
-import io.embrace.android.embracesdk.logging.InternalEmbraceLogger;
-import io.embrace.android.embracesdk.logging.InternalStaticEmbraceLogger;
 import io.embrace.android.embracesdk.network.EmbraceNetworkRequest;
 import io.embrace.android.embracesdk.payload.PushNotificationBreadcrumb;
 import io.embrace.android.embracesdk.spans.EmbraceSpan;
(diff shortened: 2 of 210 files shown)


bidetofevil:
* **#643** <a href="https://app.graphite.dev/github/pr/embrace-io/embrace-android-sdk/643?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `master`

<h2></h2>

embrace-android-sdk/src/integrationTest/kotlin/io/embrace/android/embracesdk/testcases/PublicApiTest.kt
```diff
@@ -136,4 +139,23 @@ internal class PublicApiTest {
             }
         }
     }
+
+    @Test
+    fun `static logger and SDK logger instance log to the same internal error service`() {
```
bidetofevil, line 144: This will be deleted later on with further refactoring to eliminate having two loggers user the same internal error service, but I left this here in case the PRs aren't merged together.

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/Embrace.java
```diff
@@ -576,8 +569,6 @@
         if (verifyNonNullParameters("trackWebViewPerformance", tag, consoleMessage)) {
             if (consoleMessage.message() != null) {
                 trackWebViewPerformance(tag, consoleMessage.message());
-            } else {
```
bidetofevil, line 579: This never gets hit as the null message cases will not make it pass the null checks.

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/anr/sigquit/SigquitDetectionService.kt
```diff
@@ -51,12 +47,11 @@
 
     private fun setupGoogleAnrTracking() {
         if (configService.anrBehavior.isGoogleAnrCaptureEnabled() && !googleAnrTrackerInstalled.getAndSet(true)) {
-            ThreadUtils.runOnMainThread { setupGoogleAnrHandler() }
+            ThreadUtils.runOnMainThread(logger) { setupGoogleAnrHandler() }
```
bidetofevil, line 50: Worked around the logger being needed in a "static" context by passing it in. We should try to not do this, but to make existing code works, this is fInE in small chunks I guess....

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/capture/metadata/EmbraceMetadataService.kt
```diff
@@ -107,102 +106,111 @@
...
             if (diskUsage == null) {
                 diskUsage = DiskUsage(null, free)
             }
-            logDeveloper("EmbraceMetadataService", "Device disk free: $free")
         }
     }
 
+    @TargetApi(Build.VERSION_CODES.O)
+    @SuppressWarnings("deprecation")
+    fun getDeviceDiskAppUsage(
```
bidetofevil, line 192: Pull this out of the util class because this is the only use case, and putting it here allows me to use the instance's logger. This pattern is repeated a couple times later.

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/capture/thermalstate/EmbraceThermalStatusService.kt
```diff
@@ -72,10 +63,6 @@ internal class EmbraceThermalStatusService(
     override fun getCapturedData(): List<ThermalState> = thermalStates
 
     override fun close() {
-        val pm = powerManager
-        if (pm != null) {
-            logger.logDeveloper("ThermalStatusService", "Removing thermal status listener")
-            pm.removeThermalStatusListener(thermalStatusListener)
-        }
+        powerManager?.removeThermalStatusListener(thermalStatusListener)
```
bidetofevil, line 66: Just followed the IDE refactor suggestions. This should be safe.

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/injection/CoreModule.kt
```diff
@@ -37,11 +36,6 @@
      */
     val appFramework: AppFramework
 
-    /**
-     * Returns an interface that logs messages
-     */
-    val logger: InternalEmbraceLogger
```
bidetofevil, line 43: The logger is required earlier, so put this in the InitModule instead.

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/EmbraceImpl.java
```diff
@@ -210,6 +209,7 @@ final class EmbraceImpl {
     EmbraceImpl(@NonNull ModuleInitBootstrapper bs) {
         moduleInitBootstrapper = bs;
         sdkClock = moduleInitBootstrapper.getInitModule().getClock();
+        internalEmbraceLogger = moduleInitBootstrapper.getInitModule().getLogger();
```
bidetofevil, line 212: get the logger here and use it throughout the init process

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/injection/ModuleInitBootstrapper.kt
```diff
@@ -45,7 +46,8 @@ import kotlin.reflect.KClass
  * A class that wires together and initializes modules in a manner that makes them work as a cohesive whole.
  */
 internal class ModuleInitBootstrapper(
-    val initModule: InitModule = InitModuleImpl(),
+    val logger: InternalEmbraceLogger = InternalEmbraceLogger(),
```
bidetofevil, line 49: This is the one true instance of the logger that is created and used.

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/injection/ModuleInitBootstrapper.kt
```diff
@@ -280,19 +282,21 @@ internal class ModuleInitBootstrapper(
 
                     postInit(SdkObservabilityModule::class) {
                         serviceRegistry.registerService(sdkObservabilityModule.internalErrorService)
+                        initModule.logger.addLoggerAction(sdkObservabilityModule.internalErrorLogger)
```
bidetofevil, line 285: Hook up the internal error service with the instance logger.

fractalwrench (approved):
LGTM. A few more thoughts I've had about logging recently but haven't actioned:

1. We should probably rename it to something less unwieldy than `InternalEmbraceLogger`
2. It's probably worth extracting an interface that allows fakes etc to be passed in
3. We should consider defaulting to a no-op logger in non-debuggable builds for a perf boost
4. Perhaps string construction should be lazy (I'm not sure how this will factor in with anon inner class creation, but we could check what bytecode kotlin generates & perhaps exploit inlining)

bidetofevil:
agreed on all fronts! 1-3 are no-brainers - 4 would be nice too but we should assess the impact. If we really reduced how much logging we do, will this make a difference?

PR #652, opened 2024-04-03
Add tests for stack unwinding

fractalwrench:
## Goal

Adds tests for the NDK stack unwinding that happens during sampling of a Unity ANR. The test cases are not exhaustive & are simply meant as a sanity check that the functionality is working rather than testing the internals of the unwinders.


priettt (approved):
🙌

PR #656, opened 2024-04-04
Rename unwinder function & reorganise source files

fractalwrench:
## Goal

Renames the stack unwinding function to better reflect what it does & extracts a function (that is inlined). No behavior changes were made as part of this changeset.

## Testing

Confirmed NDK crashes are still shown on the dashboard using this changeset.


embrace-android-sdk/src/main/cpp/unwinders/stack_unwinder.cpp
```diff
@@ -0,0 +1,70 @@
+//
+// Created by Eric Lanz on 5/12/20.
+//
+
+#include "stack_unwinder.h"
```
fnewberg, line 5: i assume this file was just moved and not changed, so i didn't review it. LMK if that was a bad assumption

fnewberg (approved):

embrace-android-sdk/src/main/cpp/unwinders/stack_unwinder.cpp
fractalwrench: That is correct, the only change I made was to rename a function

PR #658, opened 2024-04-04
[WIP] Avoid using dladdr with libunwindstack

fractalwrench:
## Goal

Avoids using `dladdr` by using functions available in libunwindstack directly. This avoids the need for extra symbolication code that doesn't necessarily account for edge cases introduced by Android.

The SO load address is different by a few thousand bytes when calculated by libunwindstack. I believe this is because libunwindstack accounts for [how Android reads elf files](https://github.com/embrace-io/embrace-android-sdk/blob/1c8e4abf7d8a7e605250c6aace01d7c13a10d4fe/embrace-android-sdk/src/main/cpp/3rdparty/libunwindstack-ndk/include/unwindstack/MapInfo.h#L78) whereas dladdr does not.

## Testing

I tested 3 representative crashes on an ARM emulator & compared stacktraces (see attached).

The symbolication for libart.so appears to be more complete for libunwindstack (it correctly gets `mterp_op_invoke_static`, whereas dladdr doesn't). However, all the embrace frames seem to incorrectly point at `nullPointerDereference`. I'd assume the different SO load address means our backend symbolication chooses this function.


embrace-android-sdk/src/main/cpp/unwinders/stack_unwinder.cpp
```diff
@@ -44,6 +44,9 @@ static inline void emb_copy_frame_data(unwindstack::AndroidUnwinderData &android
         data->offset = map_info->offset();
         data->flags = map_info->flags();
         emb_strncpy(data->full_name, map_info->GetFullName().c_str(), EMB_FRAME_STR_SIZE);
+
+        // calculate line number
+        data->line_num = data->pc - data->start;
```
fnewberg, line 49: isn't  `line_num` meant to refer to the symbolicated line number? this looks like it's computing the module address offset

embrace-android-sdk/src/main/cpp/serializer/file_writer.c
```diff
@@ -305,22 +305,22 @@ bool emb_add_exc_to_json(const emb_exception *exception, JSON_Array *frames_obje
...
-        RETURN_ON_JSON_FAILURE(json_object_set_string(frame_object, kMethodKey, frame->method));
+        RETURN_ON_JSON_FAILURE(json_object_set_string(frame_object, kMethodKey, frame->function_name));
     }
-    // TODO: lu vs u?
-    RETURN_ON_JSON_FAILURE(json_object_set_number(frame_object, kFrameAddrKey, frame->frame_addr));
-    RETURN_ON_JSON_FAILURE(json_object_set_number(frame_object, kOffsetAddrKey, frame->offset_addr));
-    RETURN_ON_JSON_FAILURE(json_object_set_number(frame_object, kModuleAddrKey, frame->module_addr));
-    RETURN_ON_JSON_FAILURE(json_object_set_number(frame_object, kLineNumKey, frame->line_num));
+
+    RETURN_ON_JSON_FAILURE(json_object_set_number(frame_object, kFrameAddrKey, frame->pc));
```
fnewberg, line 320: i'm not sure i follow these changes. some of the names in `emb_sframe` seem a bit confusing and some of the fields seem like they might be redundant. probably worth a chat to go over that to see if we can clean up some stuff there

PR #659, opened 2024-04-04
Added session ID attribute to CurrentSessionSpan

priettt:


embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/internal/spans/CurrentSessionSpanImpl.kt
```diff
@@ -67,6 +71,16 @@ internal class CurrentSessionSpanImpl(
         }
     }
 
+    override fun getSessionId(): String {
```
bidetofevil, line 74: Actually, lets add a new method to the interface that returns the value of an attribute given a name and have that called here. Getting this via a snapshot seems too heavyweight. If we want to expose the attributes, lets do it directly

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/internal/spans/CurrentSessionSpanImpl.kt
```diff
@@ -32,7 +34,9 @@ internal class CurrentSessionSpanImpl(
     /**
      * The span that models the lifetime of the current session or background activity
      */
-    private val sessionSpan: AtomicReference<EmbraceSpan?> = AtomicReference(null)
+    private val sessionSpan: AtomicReference<PersistableEmbraceSpan?> = AtomicReference(null)
+
+    private val sessionIdAttributeName = "session_id".toEmbraceAttributeName()
```
bidetofevil, line 39: I'm doing something to make us not have to hardcode this at the source of usage. Merge it like this and I'll fix it up

bidetofevil (approved):
LGTM

PR #673, opened 2024-04-08
Add flush to log orchestrator

fractalwrench:
## Goal

Adds a flush function to the `LogOrchestrator` class. This isn't invoked by any production code yet but will be. I primarily need this for writing tests.


lucaslabari (approved):
LGTM

PR #679, opened 2024-04-09
Define OTel schema for capturing ANRs

fractalwrench:
## Goal

Defines the OTel schema that will be used for capturing ANRs.


bidetofevil (approved):
LGTM. This is complex enough that when we draw up a semantic conventions proposal, we'll probably want to see if the community would want this as a span or an Event with a start and end time as attributes. But it's a moot point for us now since we have to ship, so this is good to go for me.

PR #681, opened 2024-04-09
Added functional tests for the new SessionSpan payload

priettt:
I mainly wanted to verify the `emb-session` span, and check if it contains the `emb.session_id` attribute. 

For the rest of the golden file, I just grabbed the resulting JSON and left the fields that seemed like will remain static between builds.

### DIFF
--- a/embrace-android-sdk/src/androidTest/assets/golden-files/session-end-v2.json
+++ b/embrace-android-sdk/src/androidTest/assets/golden-files/session-end-v2.json
@@ -1,0 +1,336 @@
+{
+  "a": {
+    "bi": "default test build id",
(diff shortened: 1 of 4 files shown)


bidetofevil (approved):
LGTM. Being able to be more expressive in gold files is only a good thing!

embrace-android-sdk/src/androidTest/assets/golden-files/session-end-v2.json
```diff
@@ -0,0 +1,298 @@
+{
+  "a": {
```
bidetofevil, line 2: You can probably ignore any v1 attribute when validating the v2 payload?

PR #687, opened 2024-04-10
Map ANR data to OTel data types

fractalwrench:
## Goal

This changeset produces a mapper that converts ANR data captured by Embrace into OTel primitives. These can be serialized in an Embrace payload & sent as part of an Embrace session. This changeset does not hookup the mapper to the payload generation - that will follow in a subsequent change.

@bidetofevil I was unsure about the correct values to generate for spanId/parentSpanId/traceId - I figured `null` was probably best but would be good to get your thoughts on that.

## Testing

Added unit tests.


embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/anr/AnrOtelMapper.kt
```diff
@@ -0,0 +1,79 @@
...
+) : DataCaptureServiceOtelConverter {
+
+    override fun snapshot(): List<Span> {
+        val intervals = anrService.getCapturedData()
+
+        return intervals.map { interval ->
+            val attrs = mapIntervalToSpanAttributes(interval)
+            val events = mapIntervalToSpanEvents(interval)
+            Span(
+                traceId = null,
```
bidetofevil, line 26: Can you use `SpanService` to create this rather than create the payload object directly? This will make it set the right defaults and make this a lot more resilient to change.

bidetofevil:
For you original question, traceId will need to be generated as this is effectively a new trace. For parent, it should be SpanId.getInvalidId() - but all of this will be done for you if you used SpanService.recordCompletedSpan()

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/anr/AnrOtelMapper.kt
bidetofevil, line 26: You should generate a new traceId similar to how you generate a SpanId, i.e. IdGenerator.random().generateTraceId()  as this is a new trace

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/anr/AnrOtelMapper.kt
```diff
@@ -0,0 +1,79 @@
...
+                spanId = IdGenerator.random().generateSpanId(),
+                parentSpanId = null,
```
bidetofevil, line 28: This should be SpanId.getInvalidId()

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/anr/AnrOtelMapper.kt
```diff
@@ -0,0 +1,79 @@
...
+                name = "emb_thread_blockage",
```
bidetofevil, line 29: `emb-thread-blockage` according to the conventions

bidetofevil (approved):
I have some suggested change for the TraceId, Parent SpanId, and name, but should be good to go after that, so I'll approve it and you can merge when ready

fractalwrench:
Updated to use correct values for spanId/parentSpanId/traceId.

PR #691, opened 2024-04-11
Refactor data source module to return data sources

bidetofevil:
## Goal

Datasources needs to be retrievable from the module so they can be used. Make DataSourceState typed so that it can return the appropriate typed DataSource

## Testing

Added tests to check the logic to obtain the data source


fractalwrench (approved):
LGTM - I went with a similar approach in an early iteration, but never needed to retrieve a data source so got rid of the generics.

bidetofevil:
Getting the generics right took me.... more than I wanted to spend lol

PR #696, opened 2024-04-11
Add `AnrOtelMapper` to module & inject into necessary classes

fractalwrench:
## Goal

Add `AnrOtelMapper` to a dependency module & injects it into the necessary classes. This changeset does not hookup OTel capture, that will follow in a future changeset.

I also made a small change to the `AnrService` so that in-progress ANR intervals are not cleared when the session boundary changes.

## Testing

Added unit tests.


bidetofevil (approved):
LGTM

PR #697, opened 2024-04-11
Enable ANR capture via OTel

fractalwrench:
## Goal

Adds ANRs OTel representation to the payload. I've tried to follow the existing logic of when spans should be captured. I've also not added anything to the `span_snapshots` field which I think is correct.

## Testing

Added unit tests.


bidetofevil (approved):
LGTM. Not creating an snapshots is good because we should probably reserve that improvement to when we can actually "start" a span when an interval start and stop it when it ends, because then you don't have to manually do it, but the system will do that automatically. For now, this make sense.

PR #699, opened 2024-04-11
[DNM] Remove legacy ANR fields from payload

fractalwrench:
## Goal

Removes legacy ANR fields from the payload.


bidetofevil (approved):
Can't wait until we ca validate and merge this.

PR #701, opened 2024-04-11
Add network calls as Spans

priettt:


embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/network/logging/EmbraceNetworkLoggingService.kt
```diff
@@ -243,6 +246,30 @@ internal class EmbraceNetworkLoggingService(
     private fun storeNetworkCall(callId: String, networkCall: NetworkCallV2) {
         callsStorageLastUpdate.incrementAndGet()
         sessionNetworkCalls[callId] = networkCall
+        spanService.recordCompletedSpan(
+            name = "emb-network-request-$callId",  //TODO: What name goes here?
```
bidetofevil, line 250: Check with IOS to see what they do

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/network/logging/EmbraceNetworkLoggingService.kt
```diff
@@ -243,6 +246,30 @@ internal class EmbraceNetworkLoggingService(
...
+            startTimeMs = networkCall.startTime,
+            endTimeMs = networkCall.endTime,
+            errorCode = null,
+            parent = null,
+            attributes = getNetworkSpanAttributes(networkCall),
+            type = EmbType.Performance.Network,
+        )
+    }
+
+    private fun getNetworkSpanAttributes(networkCall: NetworkCallV2): Map<String, String> {
```
bidetofevil, line 260: Can we define this mapping in SchemaType and along with the keys either there or in OpenTelemetryAttributeKeys? We should be using the semantic convention keys for network requests anyway and these don't appear to be the right names 😅

PR #702, opened 2024-04-11
Send exceptions to v2.

lucaslabari:
## Goal

- Send exceptions to v2 endpoint when the feature flag is on, as backend is now supporting it. 

## Testing

Added unit test.


leandro-godon (approved):
Lgtm

PR #707, opened 2024-04-12
Use correct keys for anr data capture

fractalwrench:
## Goal

Adds `emb.type` to ANR OTel capture which was missing from both the span + events. Additionally removes the `emb.` prefix which wasn't in the agreed schema.


bidetofevil (approved):
LGTM. Later, we should move the definition of the keys to SchemaType and make them `EmbraceAttributeKey` so they can be shared and not hardcoded, and leverage the helpers in there for adding and testing.

PR #709, opened 2024-04-12
Use shutdown rather than shutdownNow

fractalwrench:
## Goal

Use `shutdown` rather than `shutdownNow`. This is because `shutdown` allows previously submitted tasks to complete, whereas `shutdownNow` attempts to shut everything down. I believe this was a cause of flaky behavior in the `cleanCollections` test case.


bidetofevil (approved):
LGTM

PR #710, opened 2024-04-12
Create new remote config for OTel feature flags.

lucaslabari:
## Goal

- Created a new remote config for OTel feature flags.
- Created 3 feature flags:
   - dev
   - beta
   - stable

## Testing

Added unit tests.

### DIFF
--- a/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/config/behavior/OTelBehavior.kt
+++ b/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/config/behavior/OTelBehavior.kt
@@ -1,0 +1,32 @@
+package io.embrace.android.embracesdk.config.behavior
+
+import io.embrace.android.embracesdk.config.local.SessionLocalConfig
+import io.embrace.android.embracesdk.config.remote.RemoteConfig
+import io.embrace.android.embracesdk.internal.utils.Provider
+
+/**
+ * Provides the behavior that functionality relating to OTel should follow.
+ */
+internal class OTelBehavior(
+    thresholdCheck: BehaviorThresholdCheck,
+    remoteSupplier: Provider<RemoteConfig?>
+) : MergedConfigBehavior<SessionLocalConfig, RemoteConfig>(
+    thresholdCheck,
+    { null },
+    remoteSupplier
+) {
+    /**
+     * Whether to use the V2 payload format for sending sessions.
+     */
+    fun useV2SessionPayload() = remote?.oTelConfig?.useV2SessionPayload ?: false
+
+    /**
+     * Whether to use the V2 payload format for sending logs.
+     */
+    fun useV2LogPayload() = remote?.oTelConfig?.useV2LogPayload ?: false
+
+    /**
+     * Whether to use the V2 payload format for sending crashes.
+     */
+    fun useV2CrashPayload() = remote?.oTelConfig?.useV2CrashPayload ?: false
+}
--- a/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/config/behavior/OTelBehavior.kt
+++ b/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/config/behavior/OTelBehavior.kt
@@ -18,7 +18,7 @@ internal class OTelBehavior(
     /**
      * Returns whether OTel Stable is enabled.
      */
-    fun isStableEnabled() = remote?.oTelConfig?.isStableEnabled ?: false
+    fun isStableEnabled() = remote?.oTelConfig?.isStableEnabled ?: true
 
     /**
      * Returns whether OTel Beta is enabled.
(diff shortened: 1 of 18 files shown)


bidetofevil:
I like having 3 flags. I wonder if it might work better if we controlled behaviour in the following way instead of tying these to a feature:

 - Stable - on by default
 - Beta - off by default
 - Dev - off by default

So we can put embrace logs in Stable, and maybe AEI if we can verify that. Beta can control Crashes, so we can flip it on when ready. Then Dev can control the rest - sessions and Network request capture.

So if we ship 6.6, we can have some things going to v2/logs by default (and we can turn off in prod if necessary), some not on by default but can be turned on (i.e. crashes), and others can can turn on for testing but won't affect the others

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/config/behavior/OTelBehavior.kt
```diff
@@ -0,0 +1,32 @@
...
+    remoteSupplier: Provider<RemoteConfig?>
+) : MergedConfigBehavior<SessionLocalConfig, RemoteConfig>(
+    thresholdCheck,
+    { null },
+    remoteSupplier
+) {
+    /**
+     * Returns whether OTel Stable is enabled.
+     */
+    fun isStableEnabled() = remote?.oTelConfig?.isStableEnabled ?: false
```
bidetofevil, line 21: The default for this should be true so we can ship some OTel features default on.

embrace-android-sdk/src/test/java/io/embrace/android/embracesdk/session/message/PayloadFactoryImplTest.kt
```diff
@@ -98,7 +98,7 @@ internal class PayloadFactoryImplTest {
 
     @Test
     fun `v2 payload generated`() {
-        sessionConfig = SessionRemoteConfig(useV2Payload = true)
+        oTelConfig = OTelRemoteConfig(isStableEnabled = true)
```
bidetofevil, line 101: This should be using dev because we don't be turning this on for 6.6

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/session/message/PayloadFactoryImpl.kt
```diff
@@ -17,7 +17,7 @@ internal class PayloadFactoryImpl(
 
     private val collator: PayloadMessageCollator
         get() = when {
-            configService.sessionBehavior.useV2Payload() -> v2payloadMessageCollator
+            configService.oTelBehavior.isStableEnabled() -> v2payloadMessageCollator
```
bidetofevil, line 20: This should be dev - this is the least stable right now and will require us migration everything previously in SessionMessage to be able to turn on 😅

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/internal/logs/CompositeLogService.kt
```diff
@@ -24,7 +24,7 @@ internal class CompositeLogService(
 ) : LogMessageService {
 
     private val useV2LogService: Boolean
-        get() = configService.sessionBehavior.useV2Payload()
+        get() = configService.oTelBehavior.isBetaEnabled()
```
bidetofevil, line 27: Have we tested this in production? If we are confident that this should mostly work, we should put this is stable. In fact, this is the one that I would most like to be in stable.

bidetofevil:
Structure looks good - we should just decide which features are in what bucket, and also have one bucket (stable) that we will enable by default.

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/config/behavior/OTelBehavior.kt
lucaslabari: Done

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/session/message/PayloadFactoryImpl.kt
lucaslabari: Updated

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/internal/logs/CompositeLogService.kt
lucaslabari: Let's leave it in Beta for now, as we need to fix some tests once we enable this.

bidetofevil: Ah got it. Make sense.

bidetofevil (approved):
LGTM! We can change the baseTest to turn on stable when we have features attached to that

PR #713, opened 2024-04-12
Replace emb.state with android.state

lucaslabari:
## Goal

- Replaced `emb.state` with `android.state` that is what the Opentelemetry documentation specifies that should be used. 

## Testing

Rely on existing unit tests.


embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/internal/logs/EmbraceLogService.kt
```diff
@@ -199,7 +199,7 @@ internal class EmbraceLogService(
 
         attributes.setAttribute(logRecordUid, Uuid.getEmbUuid())
         sessionIdTracker.getActiveSessionId()?.let { attributes.setAttribute(embSessionId, it) }
-        metadataService.getAppState()?.let { attributes.setAttribute(embState, it) }
```
bidetofevil, line 202: Have the iOS and the backend signed off on using this? I also added this but that was just going to be a proposal

lucaslabari: I just talked with backend and iOS and we agree on not using it yet. They argued that we are considering just 2 states instead of all the possible states defined in the Opentelemetry documentation. 
Let's postpone the discussion and make the change once all platforms are ready.

PR #723, opened 2024-04-15
Flush logs on crash

leandro-godon:
## Goal

Call `LogOrchestrator.flush()` when processing a crash.

## Testing

Tested used the test suite against the mock api. Changed the log orchestrator parameters to ensure it would not flush due to inactivity, then sent a couple of logs and make the app to crash.


fractalwrench (changes requested):
This is better than not calling `flush()` but still doesn't ensure that logs are properly persisted on a crash. `EmbraceDeliveryService#sendLogs` is asynchronous & makes a HTTP request so is unlikely to succeed in the context of an uncaught exception.

IMO two changes are necessary before merging:

1. Integration test confirming that a payload is actually received by the `DeliveryService`
2. Tweak `EmbraceDeliveryService#sendLogs` (or create a new function) so that if we're in a crashing context, the log payload is written directly to disk rather than attempting a HTTP request

leandro-godon:
I split point (2) into a different PR here: https://github.com/embrace-io/embrace-android-sdk/pull/728 so we can discuss if that is the approach we want to take.

PR #724, opened 2024-04-15
Add OTel schema for Unity ANRs

fractalwrench:
## Goal

Adds the schema that will be used for capturing Unity ANRs (native thread blockages) as OTel objects.


bidetofevil (approved):
LGTM

PR #728, opened 2024-04-15
Flush and save log requests on crash

leandro-godon:
## Goal

Call `LogOrchestrator.flush()` when processing a crash. Instead of trying to send those logs to the backend, save them as pending requests to be sent on restart.

This would be a simpler alternative to implement a caching mechanism for the `LogSink`, the way sessions do.


fractalwrench (commented):
I'm happy with this approach & think it's the most pragmatic solution for now. Happy to approve when the lines with missing test coverage are covered

leandro-godon:
Added test coverage. I will create the integration test in a separate PR.

fractalwrench (approved):
LGTM

PR #731, opened 2024-04-15
Memory warning datasource

leandro-godon:
## Goal

Send the Low Memory Warnings as span events.

Removal of warnings in PerformanceInfo will be in a different PR.

## Testing

Tested against the production dashboard, verified that the events were visible.


leandro-godon:
Added an integration test.

fractalwrench (approved):
LGTM

PR #732, opened 2024-04-15
Remove memory warnings collection in the Session

leandro-godon:
## Goal

Remove the collection of Low Memory Warnings in the session, leaving them as Span events.

## Testing

Tested against the production dashboard, verified that the events were visible.


fractalwrench (approved):
LGTM

PR #733, opened 2024-04-15
Switch between v1 or v2 crash service.

lucaslabari:
## Goal

- Added CompositeCrashService to switch between v1 and v2 crash payloads. 

## Testing

Added unit tests.


bidetofevil (commented):
This is a good base to work from - one comment about instantiation, but we can iterate from this point.

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/injection/CrashModule.kt
```diff
@@ -35,13 +38,8 @@ internal class CrashModuleImpl(
         CrashFileMarker(markerFile, initModule.logger)
     }
 
-    override val lastRunCrashVerifier: LastRunCrashVerifier by singleton {
-        LastRunCrashVerifier(crashMarker, initModule.logger)
-    }
-
-    override val crashService: CrashService by singleton {
+    private val crashServiceV1: CrashService by singleton {
```
fractalwrench, line 41: nit: could these be named `crashService` and `crashDataSource` respectively?

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/capture/crash/CrashDataSourceImpl.kt
```diff
@@ -0,0 +1,131 @@
...
+/**
+ * Intercept and track uncaught Android Runtime exceptions
+ */
+internal class CrashDataSourceImpl(
+    configService: ConfigService,
+    private val sessionOrchestrator: SessionOrchestrator,
+    private val sessionProperties: EmbraceSessionProperties,
+    private val anrService: AnrService?,
+    private val ndkService: NdkService,
+    @Suppress("UnusedPrivateMember") private val gatingService: GatingService,
```
fractalwrench, line 40: We need to retain similar behavior for gating on any crashes that are captured via OTel

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/capture/crash/CrashDataSourceImpl.kt
```diff
@@ -0,0 +1,131 @@
...
+    }
+
+    private fun getSchemaType(attributes: TelemetryAttributes): SchemaType =
+        if (attributes.getAttribute(embAndroidReactNativeCrashJsExceptions) != null) {
+            SchemaType.ReactNativeCrash(attributes)
+        } else {
+            SchemaType.Crash(attributes)
+        }
+
+    private fun getCause(throwable: Throwable?): String {
```
fractalwrench, line 121: This might iterate infinitely if an exception in the `cause` chain is the same object as the root throwable. This sounds unlikely but it will happen in production! To handle this scenario I'd keep a collection of all the previously added throwables & return early from the function if we spot one we've already seen

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/capture/crash/CrashDataSourceImpl.kt
```diff
@@ -0,0 +1,131 @@
...
+    @Suppress("UnusedPrivateMember") private val gatingService: GatingService,
+    private val preferencesService: PreferencesService,
+    private val crashMarker: CrashFileMarker,
+    private val logger: InternalEmbraceLogger,
+    private val logWriter: LogWriter,
+) : CrashDataSource,
+    LogDataSourceImpl(
+        destination = logWriter,
+        logger = logger,
+        limitStrategy = OnOffLimitStrategy { configService.autoDataCaptureBehavior.isUncaughtExceptionHandlerEnabled() }
```
fractalwrench, line 49: I'd suggest replacing this with `NoopLimitStrategy` as we don't ever want to limit the capture of uncaught exceptions. Control of data capture can happen in `DataSourceModule`, where this lambda could be supplied as a value to `configGate`

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/injection/CrashModule.kt
lucaslabari: Updated to `legacyCrashService` and `crashDataSource` because `crashService` is already used by the `CompositeCrashService` to keep references from other components.

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/capture/crash/CrashDataSourceImpl.kt
lucaslabari: Yeap. We'll do this in a separate PR. The goal of this one is just to add the ability to switch between both crash payloads.

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/capture/crash/CompositeCrashService.kt
```diff
@@ -0,0 +1,46 @@
+package io.embrace.android.embracesdk.capture.crash
+
+import io.embrace.android.embracesdk.config.ConfigService
+import io.embrace.android.embracesdk.internal.ApkToolsConfig
+import io.embrace.android.embracesdk.logging.InternalEmbraceLogger
+import io.embrace.android.embracesdk.payload.JsException
+
+internal class CompositeCrashService(
```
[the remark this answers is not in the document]
lucaslabari: As far as I know, by using `by singleton` it's being initialized lazily, so the service that's not referenced will never be instantiated. Lmk if I'm missing something.

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/capture/crash/CrashDataSourceImpl.kt
lucaslabari: @bidetofevil ^

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/capture/crash/CrashDataSourceImpl.kt
lucaslabari: Thanks for the feedback Jamie. I'll attack this in a separate PR as the intention of this one is just to add the ability to switch between both services. This was a previous code that was moved and still need iteration.

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/capture/crash/CompositeCrashService.kt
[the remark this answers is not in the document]
bidetofevil: By passing it in as an argument you are effectively referencing it and making it initialize. That was why I suggested a Provider that will only get invoked if it's actually used in this class. That will prevent the initialization on reference.

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/capture/crash/CrashDataSourceImpl.kt
bidetofevil: Yup, that makes sense! I was wondering why something like this wasn't there - configGate is much better.

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/capture/crash/CompositeCrashService.kt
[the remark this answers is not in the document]
lucaslabari: > By passing it in as an argument you are effectively referencing it and making it initialize. 
I thought it's initialized just when accessed or when calling a method on it, but not when passing it as param.

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/capture/crash/CrashDataSourceImpl.kt
lucaslabari: Updated.

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/capture/crash/CompositeCrashService.kt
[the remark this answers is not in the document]
lucaslabari: Updated

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/capture/crash/CrashDataSourceImpl.kt
lucaslabari: I replaced with `NoopStrategy` and will move the CrashDataSource into DataSourcesModule in another PR, because dependencies make it hard to do it in an easy way within this PR. At that point, we can put it into `configGate`.

bidetofevil (approved):
Looks good. Merge it down and we can iterate on the rest

PR #737, opened 2024-04-16
Define session span attribute keys

fractalwrench:
## Goal

Defines keys that will be used on the session span. This populates the `emb.cold_start` attribute but leaves the others for now.


bidetofevil (approved):
LGTM. I have some questions about how these are used and whether they are the correct implementation, ultimately, but we can change that once the migration is done.

PR #740, opened 2024-04-16
If sdk not started, let's not add embrace in the middle of requests

cesarmax22:
## Goal

if embrace is not started then create a default UrlStreamHandler, otherwise create an EmbraceUrlStreamHandler

## Release Notes

**WHAT**:<br>
**WHY**:<br>
**WHO**:<br>


bidetofevil:
As discussed, this happens at class loading time, so once it's set, it's permanent, so we can't really do that

PR #743, opened 2024-04-17
Write session-related attributes to session span

fractalwrench:
## Goal

Updates the session orchestration so that session-related attributes are written to the session span. This will allow the old session object to be removed from the payload eventually. I've implemented the following in this changeset:

- `emb.cold_start`
- `emb.session_number`
- `emb.state`
- `emb.clean_exit`
- `emb.heartbeat_time_unix_nano`
- `emb.crash_id`

## Testing

Added unit test coverage & updated functional tests.


bidetofevil (approved):
LGTM

PR #744, opened 2024-04-17
Add gating for log and session properties in logs and crashes.

lucaslabari:
## Goal

- Added gating for log and session properties on crash events. 
- Added gating for log and session properties on log events. 

## Testing

Added unit tests.


nelsitoPuglisi (approved):

fractalwrench (approved):
LGTM

PR #748, opened 2024-04-17
Record logged network requests as completed spans

priettt:
- Record network requests as spans.

- Consolidate logNetworkCall and logNetworkError methods

EmbraceNetworkLoggingService had two methods: logNetworkCall and logNetworkError. They were both doing essentially the same, but with a different NetworkCallV2 object: strip the URL, sanitize the TraceId, logging the networkCapturedData, and processing the network call.

We don't need a NetworkCallV2 object anymore, and the EmbraceNetworkRequest is the same for both network call and network error, so we can just use one method.

- Remove unused methods.

### Questions:
- Why do we need to strip the URL? 
- Why are we stripping the URL for the span, but not for the networkCapturedData?

### DIFF
--- a/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/network/logging/EmbraceNetworkLoggingService.kt
+++ b/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/network/logging/EmbraceNetworkLoggingService.kt
@@ -1,20 +1,20 @@
 package io.embrace.android.embracesdk.network.logging
 
+import io.embrace.android.embracesdk.arch.schema.EmbType
+import io.embrace.android.embracesdk.arch.schema.SchemaType
 import io.embrace.android.embracesdk.config.ConfigService
-import io.embrace.android.embracesdk.internal.network.http.NetworkCaptureData
+import io.embrace.android.embracesdk.internal.spans.SpanService
 import io.embrace.android.embracesdk.logging.InternalEmbraceLogger
 import io.embrace.android.embracesdk.network.EmbraceNetworkRequest
 import io.embrace.android.embracesdk.network.logging.EmbraceNetworkCaptureService.Companion.NETWORK_ERROR_CODE
-import io.embrace.android.embracesdk.payload.NetworkCallV2
 import io.embrace.android.embracesdk.payload.NetworkSessionV2.DomainCount
 import io.embrace.android.embracesdk.session.MemoryCleanerListener
 import io.embrace.android.embracesdk.utils.NetworkUtils.getDomain
-import io.embrace.android.embracesdk.utils.NetworkUtils.getValidTraceId
+import io.embrace.android.embracesdk.utils.NetworkUtils.getUrlPath
 import io.embrace.android.embracesdk.utils.NetworkUtils.isIpAddress
 import io.embrace.android.embracesdk.utils.NetworkUtils.stripUrl
 import java.util.concurrent.ConcurrentHashMap
 import java.util.concurrent.atomic.AtomicInteger
-import kotlin.math.max
 
 /**
  * Logs network calls according to defined limits per domain.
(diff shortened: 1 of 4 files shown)


embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/network/logging/EmbraceNetworkLoggingService.kt
```diff
@@ -49,113 +45,40 @@ internal class EmbraceNetworkLoggingService(
...
+        if (networkRequest.networkCaptureData != null) {
             networkCaptureService.logNetworkCapturedData(
-                url,
-                httpMethod,
-                NETWORK_ERROR_CODE,
-                startTime,
-                endTime,
-                networkCaptureData,
-                errorMessage
+                networkRequest.url, //TODO: This used the non-stripped URL, is that correct?
```
bidetofevil, line 56: Yeah, we want to log the entire URL when capturing the entire request/response in case there are parameters are that important. We aren't aggregating by this so we don't need to reduce the cardinality by tweaking the URL

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/network/logging/EmbraceNetworkLoggingService.kt
```diff
@@ -49,113 +45,40 @@ internal class EmbraceNetworkLoggingService(
...
+    private fun processNetworkRequest(networkRequest: EmbraceNetworkRequest) {
         // Get the domain, if it can be successfully parsed. If not, don't log this call.
-        val domain = networkCall.url?.let {
-            getDomain(it)
-        } ?: return
+        val domain = getDomain(
+            stripUrl(networkRequest.url)
+        ) ?: return
 
         synchronized(callsStorageLastUpdate) {
```
bidetofevil, line 78: we can remove this synchronization and only block if we are counting against the same domain

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/network/logging/EmbraceNetworkLoggingService.kt
```diff
@@ -207,9 +130,29 @@ internal class EmbraceNetworkLoggingService(
         }
     }
 
-    private fun storeNetworkCall(callId: String, networkCall: NetworkCallV2) {
+    private fun storeNetworkRequest(networkRequest: EmbraceNetworkRequest) {
         callsStorageLastUpdate.incrementAndGet()
-        sessionNetworkCalls[callId] = networkCall
+
+        //TODO: Why do we want to strip the URL?
```
bidetofevil, line 136: We strip because we want to aggregate all the calls to this base URL without parameters and hash section links (or whenever they are called). I guess the idea is that foo.com/#first and foo.com/#second is close enough that they should be looked at together.

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/utils/NetworkUtils.kt
```diff
@@ -74,11 +74,7 @@ internal object NetworkUtils {
      * @return the URL with the hash fragment and query string parameters removed
      */
     @JvmStatic
-    fun stripUrl(url: String?): String? {
-        if (url == null) {
-            return null
-        }
-
+    fun stripUrl(url: String): String {
```
bidetofevil, line 77: Can move this. method to the logging service? It's only called there and in the test, and the test could just validate the URL is stripped without using that method to get the expected match?

embrace-android-sdk/src/test/java/io/embrace/android/embracesdk/NetworkUtilsTest.kt
```diff
@@ -140,7 +140,7 @@ internal class NetworkUtilsTest {
 
     @Test
     fun stripUrl() {
```
bidetofevil, line 142: rip it out!

bidetofevil:
### Merge activity

* **Apr 18, 7:50 PM EDT**: @bidetofevil started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/embrace-io/embrace-android-sdk/748).

PR #767, opened 2024-04-18
Fix crash attributes serialization.

lucaslabari:
## Goal

- Fix serialization of some crash attributes.

## Testing


leandro-godon (approved):
Lgtm

PR #768, opened 2024-04-18
OTel network request capture

leandro-godon:
## Goal

Send captured network requests as otel logs instead of events.

## Testing

Tested against the mock API. Verified that the data reaches Kafka.


embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/arch/schema/SchemaType.kt
```diff
@@ -233,4 +234,32 @@ internal sealed class SchemaType(
     ) {
         override val schemaAttributes = emptyMap<String, String>()
     }
+
+    internal class NetworkCapturedRequest(networkCapturedCall: NetworkCapturedCall) : SchemaType(
+        telemetryType = EmbType.System.NetworkCapturedRequest,
+        fixedObjectName = "network-captured-request"
```
bidetofevil, line 240: We shouldn't need this for logs

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/network/logging/NetworkCaptureDataSourceImpl.kt
```diff
@@ -0,0 +1,42 @@
...
+        logger = logger,
+        limitStrategy = NoopLimitStrategy,
+    ) {
+
+    /**
+     * Creates a log with data from a captured network request.
+     *
+     * @param networkCapturedCall the captured network information
+     */
+    override fun logNetworkCapturedCall(networkCapturedCall: NetworkCapturedCall?) {
```
bidetofevil, line 28: Should we even allow a null here? We might just not want call this if the network capture object is null

bidetofevil (approved):
LGTM. One comment about whether we can just not support a null object to pass to the data source but if we can't, no big deal.

We'll finalize the attribute names later and since the backend doesn't use them yet, this is totally fine.

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/arch/schema/SchemaType.kt
leandro-godon: Removed

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/network/logging/NetworkCaptureDataSourceImpl.kt
leandro-godon: It was copied from the `LogService` interface but you are right that it doesn't make sense in the data source, changed to non-null

PR #770, opened 2024-04-18
Fix RN crash attribute serialization.

lucaslabari:
## Goal

- Fix RN crash attribute serialization.
- Updated key `emb.crash_number` to `emb.android.crash_number`
- Made `android.react_native_crash.js_exception` singular, as we were always sending only one `JsException`.

## Testing

Added unit test.


leandro-godon (approved):
Lgtm

PR #785, opened 2024-04-19
Add log.record.uid to all logs generated

bidetofevil:
## Goal

Add `log.record.uid` to every OTel log written that uses `LogWriterImpl`, similar to how we add other common attributes.

Kept the existing logic in JVM crashes and Embrace logs because of some extra things they are doing - limiting the size of the change and will consolidate that later.

## Testing
Add test to LogWriterImpl

### DIFF
--- a/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/arch/destination/LogWriterImpl.kt
+++ b/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/arch/destination/LogWriterImpl.kt
@@ -2,8 +2,10 @@ package io.embrace.android.embracesdk.arch.destination
 
 import io.embrace.android.embracesdk.capture.metadata.MetadataService
 import io.embrace.android.embracesdk.internal.spans.toOtelSeverity
+import io.embrace.android.embracesdk.internal.utils.Uuid
 import io.embrace.android.embracesdk.opentelemetry.embSessionId
 import io.embrace.android.embracesdk.opentelemetry.embState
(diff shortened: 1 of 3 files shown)


bidetofevil:
* **#785** <a href="https://app.graphite.dev/github/pr/embrace-io/embrace-android-sdk/785?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `master`

<h2></h2>

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/arch/destination/LogWriterImpl.kt
```diff
@@ -28,6 +30,8 @@ internal class LogWriterImpl(
             .setSeverity(logEventData.severity.toOtelSeverity())
             .setSeverityText(logEventData.severity.name)
 
+        builder.setAttribute(logRecordUid, Uuid.getEmbUuid())
```
lucaslabari, line 33: what about crashes? that get the unityCrashId? will be overwritten, right?

lucaslabari (approved):
LGTM. left a question

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/arch/destination/LogWriterImpl.kt
bidetofevil: Yup - anythign from teh schema will overwrite anything that is set here. The native crash actually overrides the sessionId because we want the old one, not the current one

PR #787, opened 2024-04-19
Only log network request when getWrappedStream is called

bidetofevil:
## Goal

We were prematurely trying to log a span for a request sent using HttpUrlConnection - we used to overwrite this but since we don't do that anymore, it got used.

Instead of relying on that, we streamline the logging - putting in the "cacheNetworkCall" to force request invocation down the path of calling getResponseCode(). This took care of some edge cases that we previously didn't cover.

If we are not wrapping the stream, we log the call after the request is executed. If we are wrapping the stream, we do the logging when the stream is read and the bytes are counted. If there's an error during the execution or if the request wasn't executed at all because of the lack of a network connection, the logging happens at disconnect.

HttpUrlConnection - you got me again!

## Testing

Existing unit tests pass.


bidetofevil:
* **#787** <a href="https://app.graphite.dev/github/pr/embrace-io/embrace-android-sdk/787?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `master`

<h2></h2>

nelsitoPuglisi (approved):

PR #789, opened 2024-04-19
 Only log network request when getWrappedStream is called

nelsitoPuglisi:
## Goal

Cherry pick #787 into 6.6


bidetofevil (approved):
LGTM

PR #799, opened 2024-04-22
[WIP] Capture sigquit as span event

fractalwrench:
## Goal

This changeset captures SIGQUIT as a span event on the OTel session span.

## Testing

Added unit test coverage.


embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/injection/DataSourceModule.kt
```diff
@@ -56,6 +58,7 @@ internal class DataSourceModuleImpl(
     systemServiceModule: SystemServiceModule,
     androidServicesModule: AndroidServicesModule,
     workerThreadModule: WorkerThreadModule,
+    anrModule: AnrModule
```
nelsitoPuglisi, line 61: Doesn't require a provider instead of the instance? Be careful with initialization.

nelsitoPuglisi (approved):

PR #801, opened 2024-04-23
Remove logging functional tests

fractalwrench:
## Goal

Removes the functional tests for the logging API. This is because the `LoggingApiTest` integration test covers this functionality & is much faster to execute.


lucaslabari (approved):
LGTM

PR #803, opened 2024-04-23
Convert session message functional test

fractalwrench:
## Goal

Converts the session message functional test to an integration test instead. I have altered the integration test infrastructure so the payload is serialized into JSON, and rewritten the JSON validator to include clearer error messages for debugging what is wrong with the payload.


embrace-android-sdk/src/integrationTest/kotlin/io/embrace/android/embracesdk/testcases/V1SessionApiTest.kt
```diff
@@ -0,0 +1,38 @@
...
+    @Test
+    fun sessionEndMessageTest() {
+        with(testRule) {
+            val message = harness.recordSession {
+                embrace.setUserIdentifier("some id")
+                embrace.setUserEmail("user@email.com")
+                embrace.setUsername("John Doe")
+
+                // add webview information
+                val msg = ResourceReader.readResourceAsText("expected-webview-core-vital.json")
```
lucaslabari, line 32: `expected-webview-core-vital.json` is the `expected` in this naming correct? AFAIK this is not what we expect, but what we inject instead, right?

embrace-android-sdk/src/test/resources/v1_session_expected.json
```diff
@@ -0,0 +1,66 @@
...
+    "st": "__EMBRACE_TEST_IGNORE__",
+    "sm": "s",
+    "sd": "__EMBRACE_TEST_IGNORE__",
+    "sdt": "__EMBRACE_TEST_IGNORE__",
+    "ue": 0,
+    "lwc": 0,
+    "nc": [],
+    "wl": [],
+    "e": "__EMBRACE_TEST_IGNORE__",
+    "wvi_beta": "__EMBRACE_TEST_IGNORE__"
```
lucaslabari, line 55: We were validating this in the previous implementation. Did you remove it because it's also the input for the test? In that case, It seems right to avoid the validation for this case.

lucaslabari (approved):
LGTM. Left a couple of comments/questions.

embrace-android-sdk/src/integrationTest/kotlin/io/embrace/android/embracesdk/testcases/V1SessionApiTest.kt
fractalwrench: Yep - this is what we inject

PR #804, opened 2024-04-23
Convert v2 session test to functional test

fractalwrench:
## Goal

Converts the v2 session message functional test to an integration test instead. This mostly remains the same except I chose to validate span_snapshots outside of JSON as it was failing on that key & very hard to debug.


lucaslabari (approved):
LGTM

PR #806, opened 2024-04-24
Add integration test for ANRs

fractalwrench:
## Goal

Adds an integration test for the ANR functionality. This is more reliable than the functional test as we have the ability to control the clock + executor, meaning we can make more specific assertions (and also write more test cases as they execute much faster).


lucaslabari (approved):
LGTM

PR #813, opened 2024-04-29
Improve error message when desugaring not enabled

fractalwrench:
## Goal

Improves the error message when desugaring has not been enabled in a project by catching a `NoClassDefError` thrown when initializing OTel. We then rethrow it with instructions on how to resolve the problem.

I've chosen `LinkageError` as we catch `Exception` but not `Error`, and it felt appropriate to rethrow with a similar type.

## Testing

Verified by running our test-suite on Android 6 & saw the following error message:

```
java.lang.LinkageError: Please enable library desugaring in your project to use the Embrace SDK. This is required if you target API levels below 24. For instructions, please see https://developer.android.com/studio/write/java8-support#library-desugaring
at io.embrace.android.embracesdk.injection.OpenTelemetryModuleImpl$openTelemetrySdk$2.invoke(OpenTelemetryModule.kt:107)
at io.embrace.android.embracesdk.injection.OpenTelemetryModuleImpl$openTelemetrySdk$2.invoke(OpenTelemetryModule.kt:78)
```

### DIFF
--- a/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/injection/OpenTelemetryModule.kt
+++ b/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/injection/OpenTelemetryModule.kt
@@ -98,10 +98,19 @@ internal class OpenTelemetryModuleImpl(
 
     private val openTelemetrySdk: OpenTelemetrySdk by lazy {
         Systrace.traceSynchronous("otel-sdk-wrapper-init") {
-            OpenTelemetrySdk(
-                openTelemetryClock = initModule.openTelemetryClock,
-                configuration = openTelemetryConfiguration
-            )
+            try {
+                OpenTelemetrySdk(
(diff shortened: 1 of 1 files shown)


priettt (approved):

priettt:
LGTM! There are two approaches I'd like to discuss, maybe on the Friday meeting or when you come back from vacation: 
- Detect if they have the desugaring library dependency and crash at build time in the swazzler so it doesn't fail on run time. 
- Maybe inject the dependency ourselves? Is that feasible? Maybe it's too invasive 😅

fractalwrench:
@priettt detecting this at build-time sounds like a great idea. It should be possible to check the minSdkVersion + whether desugaring is enabled via swazzler, and perhaps fail the build in that instance. @bidetofevil can help move this forward while I'm on vacation & coordinate priorities with Eliab

bidetofevil (approved):
LGTM. Having this and the build time thing that @cesarmax22 is working on will be great.

PR #814, opened 2024-04-29
Filter private logs out of external exporters

leandro-godon:
## Goal

Filter out OTel logs with the "emb.private" attribute when exporting to a customer-provided log exporter.

## Testing

Modified unit tests. Note that currently there are no logs that are marked as private within the SDK.


embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/arch/destination/LogWriterImpl.kt
```diff
@@ -40,6 +42,10 @@ internal class LogWriterImpl(
             builder.setAttribute(embState.attributeKey, appState)
         }
 
+        if (isPrivate) {
+            builder.setFixedAttribute(PrivateSpan)
```
leandro-godon, line 46: We may want to rename `PrivateSpan` to `Private` to avoid confusion here.

nelsitoPuglisi (approved):

PR #820, opened 2024-05-07
Use onCreate invocation to denote Activity init start time

bidetofevil:
## Goal

In production, we are seeing a small minority of cases where there is a big gap between when onActivityPreCreated was invoked and when onActivityStart() was invoked. This could be a case of these being two separate activities or instances, or there are cases when the activity creation and starting stage can be paused.

In order to improve the trace and diagnose what's really going on, I changed the trace to use the onCreate time rather than the preCreate time, which is how it works on Android 9 or earlier, as preCreate is not a thing on those API versions. We also explicitly log the pre-create and post-create times as attributes for further debugging if this doesn't end up addressing the problem.

As well as the fix, I tweaked the implementation of the trace logging component (AppStartupTraceEmitter) so it's more streamlined, logs the name of the activity being displayed, removed the emb- prefix on attributes, and extracted the public interface into an interface so i can more easily test the StartupTracker by faking that.

In additional, I added more tests to verify the startup trace being logged properly as well as the `StartupTracker` consuming lifecycle events properly. I couldn't add test to verify the render due to the limitations of Roboletric, so I've left that alone for now. I've spent way too much time on this already....

## Testing

Updated Roboletric to the latest stable version so I can test with U as well as added the modified the appropriate tests.

### DIFF
--- a/embrace-android-sdk/build.gradle
+++ b/embrace-android-sdk/build.gradle
@@ -97,7 +97,7 @@ dependencies {
     testImplementation "io.mockk:mockk:1.12.2"
     testImplementation "androidx.test:core:1.4.0"
     testImplementation "androidx.test.ext:junit:1.1.3"
-    testImplementation "org.robolectric:robolectric:4.10.3"
+    testImplementation "org.robolectric:robolectric:4.12.1"
     testImplementation project(path: ":embrace-android-sdk")
     testImplementation "com.squareup.okhttp3:mockwebserver:4.9.3"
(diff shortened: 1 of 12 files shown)


bidetofevil:
* **#820** <a href="https://app.graphite.dev/github/pr/embrace-io/embrace-android-sdk/820?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `master`

<h2></h2>

fractalwrench (approved):
LGTM

PR #821, opened 2024-05-08
Migrate thermal states to OTel

leandro-godon:
## Goal

Send thermal state changes as spans.

Removal of the old service will be done in a separate PR.

## Testing

The feature is not yet ready on the backend. It was tested against the mock api.


fractalwrench (approved):
LGTM. We should hold off merging until the backend is ready & has been tested E2E.

PR #822, opened 2024-05-08
Remove ThermalStatusService and BetaFeatures.

leandro-godon:
## Goal

Remove the thermal services replaced by `ThermalStateDataSource` and the `BetaFeatures` object (which only carried the thermal state data) from the v1 payload.

The "do not merge" label will be removed once the backend enables thermal states as spans.


fractalwrench (approved):
LGTM

PR #823, opened 2024-05-09
Remove strict mode config

fractalwrench:
## Goal

Removes the concept of 'strict mode' for internal error logging. In practice this has not been used & complicates the implementation, so we should remove it.


bidetofevil (approved):
Yessssss. LGTM

PR #825, opened 2024-05-09
Remove dead code

fractalwrench:
## Goal

Renames a function to use lower case acronyms, removes the unused `threshold` concept, and gets rid of a couple of log messages + functions that we aren't using for AEI.

## Testing


bidetofevil (approved):
LGTM

PR #826, opened 2024-05-09
Inline logcat action

fractalwrench:
## Goal

Inlines the action that sends a message to logcat.

## Testing


bidetofevil (approved):
LGTM

PR #827, opened 2024-05-09
Inline error service action

fractalwrench:
## Goal

Inlines the call to the `InternalErrorService` into `InternalEmbraceLogger`, which simplifies the code & avoids a call to an iterator each time the log function is called. 

## Testing


embrace-android-sdk/src/integrationTest/kotlin/io/embrace/android/embracesdk/testcases/LogRecordExporterTest.kt
```diff
@@ -60,10 +63,7 @@ internal class LogRecordExporterTest {
 
                 sleep(3000)
             }
-            Assert.assertTrue((fakeLogRecordExporter.exportedLogs?.size ?: 0) == 0)
-            Assert.assertTrue(fake.msgQueue.any {
```
bidetofevil, line 64: If there were previous test logic that was inspecting what internal errors were logged, how do we do this now?

fractalwrench: This can be achieved with changes in #828 as it replaces the logger with an interface. IMO the check isn't super helpful here as it's just asserting that a message was logged. I would like to make it clearer that we are logging internal errors & make it easier to test, but that come in additional refactoring

bidetofevil: Got it. There should be other side effects that can be observed, eh? I suppose if logging is the only thing that isn't verified, it's not that bad.

bidetofevil (approved):
LGTM

PR #828, opened 2024-05-09
Create interface for internal logging

fractalwrench:
## Goal

Extracts an interface named `EmbLogger` that will be used for internal logging. I have renamed `InternalEmbraceLogger` as `EmbLoggerImpl` for now but am open to naming suggestions. Extracting this code to an interface will make it easier to verify that specific messages are logged during testing. It also opens up the possibility of using a no-op implementation in production builds to save CPU time.

## Testing


bidetofevil (approved):
Phew. Massive and much needed change. LGTM!

PR #829, opened 2024-05-09
ReactNativeAction DataSource

nelsitoPuglisi:
## Goal

The purpose is to capture ReactNative actions as spans instead of breadcrumbs.

## Testing

Integration and Manual tests

### DIFF
--- a/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/arch/schema/EmbType.kt
+++ b/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/arch/schema/EmbType.kt
@@ -85,6 +85,8 @@ internal sealed class EmbType(type: String, subtype: String?) : TelemetryType {
             val embAndroidReactNativeCrashJsException = EmbraceAttributeKey("android.react_native_crash.js_exception")
         }
 
+        internal object ReactNativeAction : System("rn_action", true)
+
         internal object NativeCrash : System("android.native_crash", true) {
             /**
              * Exception coming from the native layer
--- a/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/arch/schema/SchemaType.kt
+++ b/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/arch/schema/SchemaType.kt
@@ -331,7 +331,7 @@ internal sealed class SchemaType(
         )
             .plus(
                 properties
-                    .mapKeys { it.toString().toSessionPropertyAttributeName() }
+                    .mapKeys { it.key.toString().toSessionPropertyAttributeName() }
                     .mapValues { it.value.toString() }
             )
             .toNonNullMap()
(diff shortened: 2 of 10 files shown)


nelsitoPuglisi:
* **#829** <a href="https://app.graphite.dev/github/pr/embrace-io/embrace-android-sdk/829?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `master`

<h2></h2>

 Join @nelsitoPuglisi and the rest of your teammates on <a href="https://graphite.dev?utm-source=stack-comment">Graphite <b>Graphite</b></a>

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/arch/schema/EmbType.kt
```diff
@@ -85,6 +85,8 @@ internal sealed class EmbType(type: String, subtype: String?) : TelemetryType {
             val embAndroidReactNativeCrashJsException = EmbraceAttributeKey("android.react_native_crash.js_exception")
         }
 
+        internal object ReactNativeAction : System("rn_action", true)
```
bidetofevil, line 88: Should all React Native telemetry go into the same name space? So this becomes `react_native.action` or something?

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/arch/schema/SchemaType.kt
```diff
@@ -313,4 +314,26 @@ internal sealed class SchemaType(
             "emb.webview_info.tag" to tag
         ).toNonNullMap()
     }
+
+    internal class ReactNativeAction(
+        name: String,
+        outcome: String,
+        payloadSize: Int,
+        properties: Map<String?, Any?>,
```
bidetofevil, line 322: Can we just make this Map<String, String> since that's what we require anyway?

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/injection/DataSourceModule.kt
```diff
@@ -211,6 +213,18 @@ internal class DataSourceModuleImpl(
         )
     }
 
+    override val rnActionDataSource: DataSourceState<RnActionDataSource> by dataSourceState {
```
bidetofevil, line 216: Should we only create this data source if the framework is react native or is it assumed that any platform can log one of these?

Come to think of it, maybe we should think about moving this to the RN layer - not now, but it feels weird to have to do this in the Android/iOS layer now we know this is just a Span.

bidetofevil (approved):
Some non-blocking comments about naming and initialization, but logic LGTM overall.

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/arch/schema/SchemaType.kt
nelsitoPuglisi: We need "Any" in the public API. This place is the proper one to convert the properties to "schema attributes" to the expected format.

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/injection/DataSourceModule.kt
nelsitoPuglisi: I agree that we can move this to another layer. But for now I don't see a problem. This is created only after being called, and it's only used by the internal interface.

PR #831, opened 2024-05-09
Remove code related to "live debug" sessions.

lucaslabari:
## Goal

- `live debug` feature is no longer supported and `data-dev` endpoint is now a mirror of the `data` endpoint. 
- removed code that uses `data-dev` endpoint.
- deprecated `Embrace.start()` methods that receives as param the `isDevMode` flag which was used to go through the `data-dev` endpoint. 
- added an `Embrace.start()` method receiving the `context` and the `AppFramework` without the `isDevMode` flag.

## Testing

Updated unit tests.


embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/Embrace.java
```diff
@@ -70,21 +70,36 @@
...
 
     @Override
+    public void start(@NonNull Context context, @NonNull AppFramework appFramework) {
+        if (verifyNonNullParameters("start", context, appFramework)) {
+            impl.start(context, appFramework);
+        }
+    }
+
+    /**
+     * @deprecated Use {@link #start(Context)} instead.
```
bidetofevil, line 85: I would mention that the `isDevMode` parameter does nothing now so folks can feel safe about changing this

bidetofevil (approved):
OOOOOOH this feels so good. LGTM!

fractalwrench (approved):
LGTM

PR #832, opened 2024-05-10
Add explicit function for tracking internal errors

fractalwrench:
## Goal

Adds an explicit function for tracking internal errors. This will eventually be called when we want to report something to our internal telemetry, rather than implicitly capturing throwables.

## Testing


bidetofevil (approved):
LGTM

PR #834, opened 2024-05-10
Refactor internal error capture so that it is explicit

fractalwrench:
## Goal

Refactors the capture of internal errors so that it's necessary to explicitly call `trackInternalError` for anything to show up in our telemetry dashboards. This avoids spamming our telemetry with non-useful data.

I've attempted to enforce this by adding an enum type that represents the possible states we want to capture. I've chosen this rather than a string as IMO the focus should be on capturing a specific event & its stacktrace. Using a string as we were before tends to lead to high cardinality (e.g. when timestamps + IDs are included in error messages).

I manually reviewed all the log statements & only added tracking when I thought something was actionable by us, and important enough that we actually want telemetry on it.

## Testing


bidetofevil (approved):
I WHOLEHEARTED agree with this. Using an exception is convenient, but we should only be recording an internal exception when something is really wrong, and making this explicit is definitely the way to go. If the number of categories become too big, maybe it's time to start creating subcategories. I like this a lot.

PR #835, opened 2024-05-10
Remove logStacktrace parameter from EmbLogger

fractalwrench:
## Goal

Removes the `logStacktrace` parameter from `EmbLogger` as it's confusing & not well-used throughout the codebase. This changes the SDK so that if a Throwable is passed into a log function, it gets logged out via logcat. As part of this change I removed log messages that could end up being noisy.


bidetofevil (approved):
LGTM

PR #840, opened 2024-05-13
Convert internal error service to implement data capture service

fractalwrench:
## Goal

Converts the internal error service to implement the data capture service interface. This standardises how data is cleared between sessions & reduces the complexity of the previous implementation.

I have also removed logic for ignoring throwables as part of this changeset. Originally we searched the `Throwable` chain to see if any exceptions were network-connectivity errors, and skipped sending them if so. As we manually curate the exceptions we deliver to the backend now, this is not necessary anymore.

## Testing


bidetofevil (approved):
LGTM

PR #841, opened 2024-05-13
Update lifecycle owner dependency

fractalwrench:
## Goal

Updates the androidx lifecycle dependency to the latest version.

## Testing

I checked the [release notes](https://developer.android.com/jetpack/androidx/releases/lifecycle) and it seems to be only bug fixes + Kotlin conversions which shouldn't affect our use-case. I tested in our example app to confirm a session was reported to the dashboard.


bidetofevil (approved):
Very nice

PR #842, opened 2024-05-13
Remove app state from internal errors

fractalwrench:
## Goal

Removes app state from internal errors as it's unused. This allowed moving the `InternalErrorService` a bit earlier in the initialization order so we can capture errors as soon as possible.

## Testing


bidetofevil (approved):
LGTM

PR #843, opened 2024-05-13
Use non deprecated api for lifecycle callbacks

fractalwrench:
## Goal

Avoids using a deprecated API in the [androidx lifecycle library](https://developer.android.com/jetpack/androidx/releases/lifecycle) that we rely upon to start sessions & background activities. I have used `LifecycleEventObserver` rather than `DefaultLifecycleObserver` as this is available since 2.1.0 rather than 2.4.0 of the library.

## Testing

Added unit test coverage & confirmed that sessions + background activities are reported to the dashboard.

## Release Notes

**WHAT**: Updated implementation of androidx lifecycle callbacks to use non-deprecated APIs<br>
**WHY**: Avoids use of a deprecated API<br>
**WHO**: Everyone<br>


bidetofevil (approved):
Nice

PR #844, opened 2024-05-14
Convert EmbraceImpl to kotlin

fractalwrench:
## Goal

Converts `EmbraceImpl` to Kotlin. This should make it easier to refactor in future & avoids having a major component of our SDK in Java unnecessarily. I mostly just used the auto-conversion, with a few tweaks for our style.

## Testing


embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/EmbraceImpl.kt
```diff
@@ -0,0 +1,1072 @@
...
+    @Volatile
+    var application: Application? = null
+        private set
+
+    /**
+     * The type of application being instrumented by this SDK instance, whether it's directly used by an Android app, or used via a hosted
+     * SDK like Flutter, React Native, or Unity.
+     */
+    @Volatile
+    private var appFramework: Embrace.AppFramework? = null
```
bidetofevil, line 100: In a subsequent change, I want to get rid of these potentially nullable things with a default implementation used before the SDK initializes. These are a scourge

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/EmbraceImpl.kt
```diff
@@ -0,0 +1,1072 @@
...
+            null,
+            LogExceptionType.HANDLED,
+            null,
+            null,
+            null,
+            message
+        )
+    }
+
+    @JvmOverloads
```
bidetofevil, line 714: I hope we can get rid of these in a subsequent change because our code calling into this shouldn't be Java anymore

bidetofevil (approved):
Looks good!

PR #845, opened 2024-05-14
Detect and convert nanosecond timestamps for all timestamp parameters

bidetofevil:
## Goal

Use nanosecond fallback for all timestamp parameters. This is because the hosted SDKs are passing in nanosecond time because that's what iOS expects, so we'll just support it on every parameter, rather than just the ones that had previously been in nanoseconds so customers might be using it.

## Testing

Added tests in every layer


bidetofevil:
* **#845** <a href="https://app.graphite.dev/github/pr/embrace-io/embrace-android-sdk/845?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `master`

<h2></h2>

nelsitoPuglisi (approved):

PR #847, opened 2024-05-16
Downgrade lifecycle library

fractalwrench:
## Goal

Downgrades the androidx lifecycle library to the previous version to avoid a compatibility warning on older JDK versions.


cesarmax22 (approved):
lgtm

PR #848, opened 2024-05-16
Refactor EmbraceImpl to use property delegates

fractalwrench:
## Goal

Refactors EmbraceImpl to use property delegates, rather than relying on vars.

## Testing


bidetofevil (approved):
LGTM

PR #849, opened 2024-05-16
Split EmbraceImpl implementation into API delegates

fractalwrench:
## Goal

Splits the `EmbraceImpl` implementation so the user API is implemented as a delegate. Future changesets will factor other related APIs into appropriate delegates.

## Testing

Added unit test coverage.


bidetofevil (approved):
Nice. Keep them coming!

PR #850, opened 2024-05-16
Extract session api to delegate

fractalwrench:
## Goal

Splits the `EmbraceImpl` implementation so the session API is implemented as a delegate. Future changesets will factor other related APIs into appropriate delegates.

## Testing


bidetofevil (approved):
LGTM

PR #851, opened 2024-05-16
Extract network request api delegate

fractalwrench:
## Goal

Splits the `EmbraceImpl` implementation so the network request API is implemented as a delegate. Future changesets will factor other related APIs into appropriate delegates.

## Testing

Added unit test coverage.


bidetofevil (approved):
LGTM.

PR #853, opened 2024-05-16
Create logs api delegate

fractalwrench:
## Goal

Splits the `EmbraceImpl` implementation so the logs API is implemented as a delegate. Future changesets will factor other related APIs into appropriate delegates.

## Testing

Added unit test coverage.


bidetofevil (approved):
LGTM

PR #857, opened 2024-05-17
Refactor EmbraceImpl so it implements TracingApi

fractalwrench:
## Goal

Refactor EmbraceImpl so it implements TracingApi via a property delegate.

## Testing


bidetofevil (approved):
LGTM

PR #859, opened 2024-05-17
Break inheritance hierarchy in interfaces

fractalwrench:
## Goal

Breaks an inheritance hierarchy in internal interfaces that would complicate our approach of API delegates. This shouldn't be a breaking change as the actual API implementation remains on `Embrace.java`, and the interface was previously non-public.

## Testing


bidetofevil (approved):
LGTM

PR #860, opened 2024-05-17
Extract view tracking delegate

fractalwrench:
## Goal

Extracts a delegate for the view tracking API calls.

## Testing

Added unit test coverage.


bidetofevil (approved):
LGTM

PR #861, opened 2024-05-17
Extract network calls into api delegate

fractalwrench:
## Goal

Extracts a few more network calls functions into an api delegate

## Testing

Added unit tests


bidetofevil (approved):
LGTM

PR #863, opened 2024-05-17
Set warm start delay heuristic to 2 seconds

bidetofevil:
## Goal

At 60 seconds, the heuristic we are using to determine whether we are experiencing a warm start vs a cold start, which previously measured the time between the last event before activity onCreate is called to the activity onCreate of the Activity we consider to be the activity whose rendering  will end startup, is too generous, leaving too much time for background app startups followed by a user-invoked launcher startup to be considered a cold startup. 

As such, we are going to shrink that to a more reasonable 2 seconds, which is quite a bit of time already if that last even is the end of the Application onCreate. We are also counting the FIRST activity onCreate being invoked, event if it is not the startup activity, because that indicates the startup is happening.

Even with this fix, though, the implementation is still imperfect, because if we don't know when application onCreate finishes (which we don't know out of the box), we simply use the end of the Embrace SDK's init time as that. We should look to improve this heuristic by getting customers to set this information, much like how they call `reportFullyDrawn()` to sent the end of time startup for GPC. 

Ideally, we don't even use this heuristic, but that is a longer term goal that we can't quite do it without some major time investment.

While this is not perfect, p99 should be plenty good enough - with a few false negatives leading to an actual cold start being tracked as a warm start. But that would still be better than the state of the startup traces now.

## Testing

Added appropriate tests. I also did some consolidation so there is less repeats in the tests.

### DIFF
--- a/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/capture/startup/AppStartupTraceEmitter.kt
+++ b/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/capture/startup/AppStartupTraceEmitter.kt
@@ -404,10 +404,10 @@ internal class AppStartupTraceEmitter(
     companion object {
         /**
          * The gap between the end of the Embrace SDK initialization to when the first activity loaded after startup happens.
-         * If this gap is greater than 1 minute, it's likely that the app process was not created because the user tapped on the app icon,
+         * If this gap is greater than 2 seconds, it is assumed that the app process was not created because the user tapped on the app,
          * so we track this app launch as a warm start.
          */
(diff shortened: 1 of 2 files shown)


bidetofevil:
* **#863** <a href="https://app.graphite.dev/github/pr/embrace-io/embrace-android-sdk/863?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `master`

<h2></h2>

bidetofevil:
### Merge activity

* **May 20, 5:28 AM EDT**: [Graphite](https://app.graphite.dev/github/pr/embrace-io/embrace-android-sdk/863) rebased this pull request as part of a merge.

PR #865, opened 2024-05-17
WebView Info - OTel Migration

lucaslabari:
## Goal

- Migrated `webview_info` to OTel
- Created `WebViewDataSource`

## Testing

Added unit tests and tested manually


nelsitoPuglisi (approved):

embrace-android-sdk/src/test/java/io/embrace/android/embracesdk/session/SessionHandlerTest.kt
```diff
@@ -72,13 +72,13 @@ internal class SessionHandlerTest {
         private val logMessageService: LogMessageService = FakeLogMessageService()
         private val clock = FakeClock()
         private val internalErrorService = EmbraceInternalErrorService(clock)
-        private const val now = 123L
+        private const val NOW = 123L
```
nelsitoPuglisi, line 75: Is this change expected?

fractalwrench: I think so - it's `const val` so convention in Kotlin is to have an uppercase name for constants

PR #869, opened 2024-05-20
Extract otel exporter to api delegate

fractalwrench:
## Goal

Extracts the otel exporter API to a delegate.

## Testing

Added unit tests.


bidetofevil (approved):
LGTM

PR #870, opened 2024-05-20
Move package of internal interfaces

fractalwrench:
## Goal

Moves the internal interfaces out of the default package. I've avoided moving any internal interfaces that we know are called from Unity/RN/Flutter.


bidetofevil (approved):
Nice. LGTM

PR #873, opened 2024-05-21
Fix services stopping logic

bidetofevil:
## Goal

Fix the logic to stopping registered services when the SDK is stopped.

## Testing

Add basic tests

### DIFF
--- a/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/injection/ModuleInitBootstrapper.kt
+++ b/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/injection/ModuleInitBootstrapper.kt
@@ -478,7 +478,7 @@ internal class ModuleInitBootstrapper(
     }
 
     fun stopServices() {
-        if (isInitialized()) {
+        if (!isInitialized()) {
             return
         }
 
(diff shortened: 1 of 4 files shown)


bidetofevil:
* **#873** <a href="https://app.graphite.dev/github/pr/embrace-io/embrace-android-sdk/873?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `master`

<h2></h2>

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/injection/ModuleInitBootstrapper.kt
```diff
@@ -478,7 +478,7 @@ internal class ModuleInitBootstrapper(
     }
 
     fun stopServices() {
-        if (isInitialized()) {
```
fnewberg, line 481: i'm guessing we check this since you can't have `asyncInitTask` be `null` when it is used with `synchronized`, but this i think it would be more readable if that was check was done directly. or add a comment explaining that that is why this is checked here and then again below.

that said, is this 100% safe? in theory, couldn't `asyncInitTask` become `null` after the initial check? should we just get rid of the initial check on L481 and use some other object in the `synchronized` statement?

fractalwrench (approved):
LGTM

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/injection/ModuleInitBootstrapper.kt
bidetofevil: I added to prevent unnecessary locking under high contention situations. In practice, it shouldn't matter in this specific case, but it's an idiom I follow whenever I lock.

The first check isn't necessarily threadsafe, but it's not supposed to be - it's just an optimization to do an early return and not have to lock. The second check is threadsafe, and that's all that matters for that.

PR #874, opened 2024-05-21
Remove requirement for supplying appId

fractalwrench:
## Goal

Removes the requirement for supplying appId. This will allow folks to send data from Embrace to 3rd party OTel exporters without needing to send anything to Embrace's servers.

I have taken the approach of making the `DeliveryService` no-op if an appId is not supplied. There is probably some additional work that we could cancel (such as construction of payload messages etc), but I feel this is a good first approach. Perhaps ultimately we can aim to export as pure OTel & therefore kill a lot of SDK code.

I would appreciate in-depth reviews on this as it substantially changes how the SDK works & I want to be sure I haven't forgotten anything important.

### DIFF
--- a/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/config/LocalConfigParser.kt
+++ b/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/config/LocalConfigParser.kt
@@ -1,5 +1,6 @@
 package io.embrace.android.embracesdk.config
 
+import android.content.res.Resources.NotFoundException
 import android.util.Base64
 import io.embrace.android.embracesdk.config.local.LocalConfig
 import io.embrace.android.embracesdk.config.local.SdkLocalConfig
@@ -7,13 +8,14 @@ import io.embrace.android.embracesdk.internal.AndroidResourcesService
 import io.embrace.android.embracesdk.internal.ApkToolsConfig
 import io.embrace.android.embracesdk.internal.serialization.EmbraceSerializer
 import io.embrace.android.embracesdk.logging.EmbLogger
+import io.embrace.android.embracesdk.opentelemetry.OpenTelemetryConfiguration
 
 internal object LocalConfigParser {
 
     /**
      * Build info app id name.
      */
-    public const val BUILD_INFO_APP_ID = "emb_app_id"
+    const val BUILD_INFO_APP_ID = "emb_app_id"
 
     /**
      * Build info sdk config id name.
@@ -23,12 +25,12 @@ internal object LocalConfigParser {
     /**
      * Build info ndk enabled.
      */
-    public const val BUILD_INFO_NDK_ENABLED = "emb_ndk_enabled"
+    const val BUILD_INFO_NDK_ENABLED = "emb_ndk_enabled"
 
     /**
      * The default value for native crash capture enabling
      */
-    public const val NDK_ENABLED_DEFAULT = false
+    const val NDK_ENABLED_DEFAULT = false
 
     /**
      * Loads the build information from resources provided by the config file packaged within the application by Gradle at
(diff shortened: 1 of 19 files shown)


bidetofevil:
I think exporting in pure OTel is unlikely to happen in the short term given how hard it'll be to get away from needing to send session data in a single payload, but I think no-op-ing the delivery service is, as you said, a good first approach. 

I think we still need to keep the payload object construction for caching reasons, but we can probably optimize what we store in that case.

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/config/LocalConfigParser.kt
```diff
@@ -80,14 +75,35 @@ internal object LocalConfigParser {
...
         appId: String?,
         ndkEnabled: Boolean,
         sdkConfigs: String?,
         serializer: EmbraceSerializer,
         logger: EmbLogger
     ): LocalConfig {
-        require(!appId.isNullOrEmpty()) { "Embrace AppId cannot be null or empty." }
+        logger.logWarning(
+            "No appId supplied to Embrace. This is required if you want to " +
+                "send data to Embrace, unless you have configured an OTel exporter."
```
fnewberg, line 105: can we detect if a user has configured an OTel exporter? if they haven't, i.e. they are sending data to our backend, it would probably be better if this threw an exception rather than issuing a warning log

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/config/behavior/SdkEndpointBehavior.kt
```diff
@@ -23,10 +23,10 @@ internal class SdkEndpointBehavior(
     /**
      * Data base URL.
      */
-    fun getData(appId: String): String = local?.data ?: "https://a-$appId.$DATA_DEFAULT"
+    fun getData(appId: String?): String = local?.data ?: "https://a-$appId.$DATA_DEFAULT"
```
fnewberg, line 26: i think it would be better if this returned an empty string if `appId` is not set

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/config/behavior/SdkEndpointBehavior.kt
```diff
@@ -23,10 +23,10 @@ internal class SdkEndpointBehavior(
...
 
     /**
      * Config base URL.
      */
-    fun getConfig(appId: String): String = local?.config ?: "https://a-$appId.$CONFIG_DEFAULT"
+    fun getConfig(appId: String?): String = local?.config ?: "https://a-$appId.$CONFIG_DEFAULT"
```
fnewberg, line 31: same as above

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/config/LocalConfigParser.kt
bidetofevil: Maybe in debug? I'm wearying about a bad SDK configuration crashing the app if we threw in prod. Like passing a null into a Kotlin String or annotated @NonNull parameter, even though what they did isn't correct, I don't think we should do that.

bidetofevil: FWIW, an OTel exporter has to be configured before the SDK starts, so if we want to detect this case, we probably an.

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/injection/EssentialServiceModule.kt
```diff
@@ -226,7 +226,7 @@
                 thresholdCheck = thresholdCheck,
                 localSupplier = localConfig.sdkConfig::baseUrls,
             )
-
+            checkNotNull(appId)
```
bidetofevil, line 229: Will the thrown exception be caught?

bidetofevil (approved):
LGTM in general. I like Fredric's suggestion of returning empty strings instead of invalid URLs - but nothing blocking from my side

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/config/LocalConfigParser.kt
fnewberg: > Maybe in debug? I'm wearying about a bad SDK configuration crashing the app if we threw in prod.

yeah, fair point

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/config/behavior/SdkEndpointBehavior.kt
fractalwrench: Updated to return empty string

PR #876, opened 2024-05-21
Tweak data source warning

fractalwrench:
## Goal

Tweaks the log level & text shown when the data capture limit is reached in `DataSourceImpl`. This caused a bit of confusion so a lower severity level & clearer message is desirable. I've also altered the message to include the class name that has run into the limit.

## Testing

Relied on existing test coverage & also tested out message manually in the test app:

>Data capture limit reached for io.embrace.android.embracesdk.capture.crumbs.BreadcrumbDataSource. Ignoring to keep payload size reasonable - other data types will still be captured normally.

### DIFF
--- a/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/arch/datasource/DataSourceImpl.kt
+++ b/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/arch/datasource/DataSourceImpl.kt
@@ -37,7 +37,10 @@ internal abstract class DataSourceImpl<T>(
     ): Boolean {
         try {
             if (enforceLimits && !limitStrategy.shouldCapture()) {
-                logger.logWarning("Data capture limit reached.")
+                logger.logInfo(
+                    "Data capture limit reached for ${this.javaClass.name}." +
+                        " Ignoring to keep payload size reasonable - other data types will still be captured normally."
+                )
                 return false
             }
             if (!inputValidation()) {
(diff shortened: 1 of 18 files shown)


embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/arch/datasource/DataSourceImpl.kt
```diff
@@ -37,7 +37,10 @@ internal abstract class DataSourceImpl<T>(
     ): Boolean {
         try {
             if (enforceLimits && !limitStrategy.shouldCapture()) {
-                logger.logWarning("Data capture limit reached.")
+                logger.logInfo(
```
bidetofevil, line 40: Do we want to output this multiple times or just once per session?

bidetofevil (approved):
LGTM. One comment about whether we want to keep logging this once we've done it once for a particular data source, but not a blocking comment

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/arch/datasource/DataSourceImpl.kt
fractalwrench: IMO let's do it multiple times for now, then following feedback from customers change to once per session if it's too spammy. I'm hoping that setting the log level & increasing descriptiveness of message will be enough of a change

PR #878, opened 2024-05-22
Break dependency on v1 collator

fractalwrench:
## Goal

Breaks the dependency on the v1 collator when generating a v2 payload. I've achieved this by inlining the implementation & also retaining it in the v1 collator (as this will be deleted soon enough).

I've chosen not to refactor this any further beyond just inlining functions. This will make review easier & future changesets will refactor some of the gating logic etc.

## Testing


bidetofevil (approved):
Duplicating this make sense since the v1 stuff will be deleted soon anyway.

PR #880, opened 2024-05-23
Update to gradle 8.7

fractalwrench:
## Goal

Updates gradle to the [latest version](https://docs.gradle.org/8.7/release-notes.html).


gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,6 +1,7 @@
-#Mon Jul 22 17:43:50 PDT 2019
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
+distributionUrl=https\://services.gradle.org/distributions/gradle-8.7-bin.zip
+networkTimeout=10000
```
bidetofevil, line 4: What is this for? This timeout value seems too small?

bidetofevil (approved):
LGTM. One question about the network timeout value - if it's actually applicable to fetch tasks that we execute, we may want to bump that up to 30 or 60s so we don't have slow networks nuking builds?

gradle/wrapper/gradle-wrapper.properties
fractalwrench: This was added automatically by Gradle when I ran the wrapper task. I'll remove it

PR #884, opened 2024-05-24
Add fallback end time for spans in a terminated session

bidetofevil:
## Goal

If a cached session is killed by the user, the endTime of the session object will not be populated. In that case, use the terminationTime, followed by the lastHeartbeatTime, followed by the start time to ensure that the span always has an end time

## Testing

Add to unit test


embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/comms/delivery/EmbraceDeliveryService.kt
```diff
@@ -175,6 +175,12 @@ internal class EmbraceDeliveryService(
         }
     }
 
+    private fun getFailedSpanEndTimeMs(sessionMessage: SessionMessage) =
+        sessionMessage.session.endTime
+            ?: sessionMessage.session.terminationTime
+            ?: sessionMessage.session.lastHeartbeatTime
+            ?: sessionMessage.session.startTime
```
fractalwrench, line 182: This feels like it's going to surprise someone 6 months down the line when they ask why a span has the same start + end time. Would it be better to set a null value, or are there constraints that mean that's not possible?

fractalwrench (approved):
LGTM pending one comment on the start time/end time potentially being the same for some sessions

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/comms/delivery/EmbraceDeliveryService.kt
bidetofevil: Hopefully this is not actually needed as terminationTime should always be set if forceQuit is set to true. That said, i added lastHeartbeat and start times just in case.

I agree that start time isn't great, but it's better than 0 (which leads to a negative duration on the backend) and null (ended spans should have an endTime). I'll ping the backend folks when I ping them about why endTime for the session is null for a forceQuit about this as well.

PR #886, opened 2024-05-24
Add otel log data source for internal errors

fractalwrench:
## Goal

Implements our internal telemetry using OTel logs. This is not hooked up to data capture yet - that will be done in a separate PR.

## Testing

Added unit tests.


embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/arch/schema/EmbType.kt
```diff
@@ -57,6 +57,8 @@ internal sealed class EmbType(type: String, subtype: String?) : TelemetryType {
 
         internal object Exception : System("exception")
 
+        internal object InternalError : System("internal")
```
bidetofevil, line 60: why not a name that indicates that it's some sort of error or warning?

bidetofevil (approved):
LGTM. Question about why it's only called `sys.internal` and doesn't indicate that there's some sort of unexpected error, but that's not blocking

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/capture/internal/errors/InternalErrorDataSourceImpl.kt
```diff
@@ -0,0 +1,33 @@
...
+        destination = logWriter,
+        logger = logger,
+        limitStrategy = UpToLimitStrategy { 10 },
+    ) {
+
+    override fun handleInternalError(throwable: Throwable) {
+        alterSessionSpan(NoInputValidation) {
+            this.addLog(throwable, true) {
+                val schemaType = SchemaType.InternalError(throwable)
+                LogEventData(schemaType, Severity.ERROR, "")
```
bidetofevil, line 29: Should be make these logs private so they are not exported to the generic exporter?

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/arch/schema/EmbType.kt
fractalwrench: This is what was suggested with Pablo/Austin

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/capture/internal/errors/InternalErrorDataSourceImpl.kt
fractalwrench: I thought that supplying `true` to `addLog` was enough to do that?

PR #887, opened 2024-05-24
Replace internal errors with OTel logs

fractalwrench:
## Goal

Replaces internal errors with OTel logs.

## Testing


bidetofevil (approved):
LGTM

PR #888, opened 2024-05-24
Use a threadsafe map to make getActiveSpans threadsafe

bidetofevil:
## Goal

The underlying snapshot taking operation isn't thread safe, so we have to use a thread safe data structure to ensure we can snapshot the current active spans while is being modified. This leaves room for some race conditions as to which session a span is started or stopped in, which is probably fine given the asynchronous nature of tracking spans anyway.

What I don't want is to lock everything every time a span is started or stopped, which could potentially result in some pretty bad performance. I can take a look to see if we can have some greater guarantees to reduce the aforementioned race condition, but this will work fine for now as long as we don't mind a bit of timestamp misalignment between span and session boundaries.


bidetofevil:
* **#888** <a href="https://app.graphite.dev/github/pr/embrace-io/embrace-android-sdk/888?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `master`

<h2></h2>

PR #889, opened 2024-05-28
Remove legacy user info field

fractalwrench:
## Goal

Removes the legacy user info field & migrates tests to the new one. This is targeting an integration branch so won't be merged down to master just yet.

## Testing


bidetofevil (approved):
LGTM

PR #893, opened 2024-05-28
Remove performance info field

fractalwrench:
## Goal

Removes the legacy performance info field. This is targeting an integration branch so won't be merged down to master just yet.

## Testing


bidetofevil (approved):
LGTM

PR #894, opened 2024-05-28
Send payload to v2 endpoint

fractalwrench:
## Goal

Sends the payload to the new v2 endpoint. This is targeting an integration branch so won't be merged down to master just yet.

## Testing

### DIFF
--- a/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/comms/api/ApiRequestMapper.kt
+++ b/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/comms/api/ApiRequestMapper.kt
@@ -68,9 +68,8 @@ internal class ApiRequestMapper(
         return requestBuilder(url)
     }
 
-    @Suppress("UNUSED_PARAMETER")
-    fun sessionRequest(v2Payload: Boolean): ApiRequest {
-        val url = Endpoint.SESSIONS // send to v1 endpoint for now.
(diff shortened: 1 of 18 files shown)


embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/comms/api/ApiRequestMapper.kt
```diff
@@ -68,9 +68,8 @@ internal class ApiRequestMapper(
         return requestBuilder(url)
     }
 
-    @Suppress("UNUSED_PARAMETER")
-    fun sessionRequest(v2Payload: Boolean): ApiRequest {
-        val url = Endpoint.SESSIONS // send to v1 endpoint for now.
+    fun sessionRequest(): ApiRequest {
+        val url = Endpoint.SESSIONS_V2
```
bidetofevil, line 72: Do we want to hook this logic up with the feature flag (default to on) so we can roll this back in production in case something goes wrong?

fractalwrench: The v1 endpoint doesn't support fields like `metadata` etc so this is a one-way migration unless we do extra work on the backend. IMO we can mitigate this by performing testing before any of this code ships.

bidetofevil (approved):
LGTM

PR #895, opened 2024-05-28
Added EmbraceTracedConverterFactory

priettt:
## Goal

This allows the user to wrap their Retrofit converter factories with our class, and generate spans that measure request and response body serialization.

For example, if a user is using kotlinx.serialization, they should wrap their converter factory like this when they add it to Retrofit:

```
Retrofit.Builder()
    .baseUrl(CAT_API_BASE_URL)
    .addConverterFactory(
        EmbraceTracedConverterFactory(
            Json { ignoreUnknownKeys = true }.asConverterFactory("application/json".toMediaType()) //kotlinx.serialization converter factory
        )
    )
    .client(okHttpClient)
    .build()
```

This PR goes along with a PR in swazzler, that will inject the embrace-android-retrofit module: https://github.com/embrace-io/embrace-swazzler3/pull/584

## Testing

Tested manually with a test app that uses retrofit (https://github.com/priettt/CatJokes)

### DIFF
--- a/embrace-android-retrofit/CREDITS.md
+++ b/embrace-android-retrofit/CREDITS.md
@@ -1,0 +1,13 @@
+In an effort to provide fair attribution, the following list contains the names and licensing notices for all sources that have provided significant direct or indirect contributions to this project.
+
+--------------------------------------------------------------------------------
+
+The MIT License (MIT)
+Copyright (c) 2016 Raygun Limited
+
+Permission is hereby granted, free of charge, to any person obtaining a copy of this software and associated documentation files (the "Software"), to deal in the Software without restriction, including without limitation the rights to use, copy, modify, merge, publish, distribute, sublicense, and/or sell copies of the Software, and to permit persons to whom the Software is furnished to do so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in all copies or substantial portions of the Software.
+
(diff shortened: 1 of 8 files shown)


embrace-android-retrofit/CREDITS.md
```diff
@@ -0,0 +1,13 @@
+In an effort to provide fair attribution, the following list contains the names and licensing notices for all sources that have provided significant direct or indirect contributions to this project.
+
+--------------------------------------------------------------------------------
+
+The MIT License (MIT)
```
bidetofevil, line 5: Is this the right license? I thought we were Apache 2.0 now?

embrace-android-retrofit/README.md
```diff
@@ -0,0 +1,3 @@
+# Embrace.io - Retrofit Networking Android SDK
```
bidetofevil, line 1: The company name is just Embrace. `io` is just part of the website :)

embrace-android-retrofit/README.md
```diff
@@ -0,0 +1,3 @@
...
+
+This optional add-on SDK exposes some additional networking related APIs that extends the functionality of the core Embrace.io Android SDK.
```
bidetofevil, line 3: This is an "add-on module" rather an an SDK? And can we be more specific about what it does, e.g. provide components to record traces around Retrofit operations?

embrace-android-retrofit/src/main/java/io/embrace/android/embracesdk/retrofit/swazzle/callback/retrofit/EmbraceTracedConverterFactory.kt
```diff
@@ -0,0 +1,74 @@
...
+
+    private class ResponseBodyTracedConverter<T>(
+        private val delegate: Converter<ResponseBody, T>,
+        private val url: String
+    ) : Converter<ResponseBody, T> {
+
+        override fun convert(value: ResponseBody): T? {
+            var result: T? = null
+            Embrace.getInstance().internalInterface.recordSpan(
+                "Retrofit response body serialization",
```
bidetofevil, line 26: Trace names are used for aggregation, so lets use it's URL are part of the name for now?

bidetofevil:
We should not make this installable right now given the volume of network requests that we get. If these all become `emb.type` perf traces, the volume will be tremendous - we need to figure out how to safely do this so we don't overwhelm the spans table 😅

PR #900, opened 2024-05-29
Migrate span snapshot field

fractalwrench:
## Goal

Migrates the span snapshots to the new position in the payload. This is targeting an integration branch & will not be merged to master yet.

## Testing


bidetofevil (approved):
LGTM

PR #903, opened 2024-05-29
Add ability to rename spans that have not ended

bidetofevil:
## Goal

Add the ability to rename a span that hasn't been stopped. This is so we can expose this functionality in our OTel Span API wrapper.

## Testing
Added unit tests


fractalwrench (approved):
LGTM as an implementation but I have a question about whether this is necessary as a feature. Are customers likely to actually do this?

bidetofevil:
It's part of the OTel Span API and something iOS already supports. I don't know if customers will use it, but it's part of the API and it's relatively easy to implement (without needing anything on the server), so I just did it to reduce on area of potential differentiate between our implementation and the official Java OTel SDK's.

PR #912, opened 2024-05-30
V2 migration integration branch

fractalwrench:
## Goal

Integration branch for v2 payload migration.


bidetofevil (approved):
Lets. Effing. GO!

PR #914, opened 2024-05-30
Removes legacy span field

fractalwrench:
## Goal

Removes the legacy field in the payload that was used to populate spans & replaced it with the new one instead. I've tried to split this changeset out into smaller PRs where possible but by virtue of the impact of this change, it's quite a large diff.

This is targeting an integration branch so won't be released into master until E2E testing is complete.

## Testing


embrace-android-sdk/src/integrationTest/kotlin/io/embrace/android/embracesdk/assertions/SpanAssertions.kt
```diff
@@ -24,38 +23,38 @@ import org.junit.Assert.assertTrue
  * Assert the [EmbraceSpanData] is as expected
  */
 internal fun assertEmbraceSpanData(
-    span: EmbraceSpanData?,
+    span: Span?,
```
bidetofevil, line 26: After this, can we rename this to `SpanData` to be more inline with the OTel interface that this extends. We should probably look at all the naming of these internal classes once the migration is done - and also the wrapper stuff I'm doing to implement the OTel API too

bidetofevil (approved):
NICE! One comment about the nullability of parentSpanId but non-blocking and LGTM

embrace-android-sdk/src/integrationTest/kotlin/io/embrace/android/embracesdk/assertions/SpanAssertions.kt
fractalwrench: Sounds good. I will probably wait to rename until the integration branch gets merged around to avoid making merge conflicts even worse than they will be

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/internal/payload/Span.kt
```diff
@@ -30,7 +31,7 @@ internal data class Span(
 
     /* A value that uniquely identifies the parent span */
     @Json(name = "parent_span_id")
-    val parentSpanId: String? = null,
+    val parentSpanId: String? = SpanId.getInvalid(),
```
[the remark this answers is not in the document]
fractalwrench: @bidetofevil the old code was always setting `parentSpanId` in this way if no parent span was applied, but the new implementation was not. This was the easiest way I could find to default the parent span ID

PR #921, opened 2024-06-04
Remove old implementation of Crashes

priettt:
- Removed EmbraceCrashService.

This was the legacy way of handling crashes for Embrace. We now use CrashDataSource, which will add the crashes to the current session span and emit an OTel log.

- Removed CompositeCrashService

This allowed us to use the legacy or the new crashes implementation, depending on a config flag (`configService.oTelBehavior.isBetaEnabled()`).
Most importantly, it registered the EmbraceUncaughtExceptionHandler when the class was initialized. This was moved to the CrashDataSourceImpl.

### Questions:

- Should we remove the LegacyException[Info/Error/ErrorInfo] classes? They are used in the new implementation, so I'm not sure how.
- Should we still use the CrashService interface? Should we replace it with CrashDataSource?


fractalwrench (approved):
LGTM, but it'd probably be sensible to target this at `integration/v2-migration` unless you've already verified that this works fine E2E and are happy to ship this to production.

For classes like `LegacyExceptionInfo` etc, we could remove the serialization annotations & rename the class. It'd be preferable to do that in another changeset though

PR #922, opened 2024-06-05
Misc style updates

fractalwrench:
## Goal

I've broken style updates out of #914 to reduce the size of the changeset.


bidetofevil (approved):
LGTM

PR #923, opened 2024-06-05
Specify end time for in progress anrs

fractalwrench:
## Goal

Specifies an end time for ANRs that are in-progress if one does not already exist. Ordinarily the OTel infrastructure would take care of this, but we've chosen to go outside of that for ANRs.

## Testing

Updated unit test coverage.


bidetofevil (approved):
LGTM. 

It really feels like we should find a nice way to handling cases like this, e.g. send a bunch of telemetry to a holding area to give us a chance to redact logged telemetry, then "release" them when we're done. Building some plumbing around that feels like a reasonable thing to do for mobile apps

PR #924, opened 2024-06-05
Fixing logRnView

priettt:
### Fix `enable_automatic_activity_capture` 

ViewDataSource is in charge of creating view tracking spans for two situations:
- When our public API is called: (startView, endView, logRnView)
- When we automatically detect an activity change.

It is only created when enable_automatic_activity_capture is not disabled in the local config. If that flag is disabled in the local config, we don't only disable the automatic activity tracking, we also disable the public API tracking. This was fixed in this PR.

### Fix logRnView

The changedView logic was wrong. I fixed that, and now if logRnView is called twice with the same name, two different spans will be created. This is subject to changes in the future, and we are analyzing deleting this method altogether.
**Note**: The last log sent with logRnView is going to be sent as a span snapshot.

### Minor changes:
- Renamed automatic_activity_capture flag.
- Removed unused force flag from ViewDataSource.changeView.
- Removed unused timestamp parameter from EmbraceBreadcrumbService.logView.

### DIFF
--- a/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/capture/crumbs/EmbraceBreadcrumbService.kt
+++ b/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/capture/crumbs/EmbraceBreadcrumbService.kt
@@ -3,6 +3,7 @@ package io.embrace.android.embracesdk.capture.crumbs
 import android.app.Activity
 import io.embrace.android.embracesdk.config.ConfigService
 import io.embrace.android.embracesdk.injection.DataSourceModule
+import io.embrace.android.embracesdk.internal.clock.Clock
 import io.embrace.android.embracesdk.internal.utils.Provider
 import io.embrace.android.embracesdk.payload.Breadcrumbs
 import io.embrace.android.embracesdk.payload.PushNotificationBreadcrumb.NotificationType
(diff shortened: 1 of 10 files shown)


embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/capture/crumbs/EmbraceBreadcrumbService.kt
```diff
@@ -24,13 +23,12 @@ import io.embrace.android.embracesdk.session.lifecycle.ActivityLifecycleListener
...
  */
 internal class EmbraceBreadcrumbService(
-    private val clock: Clock,
     private val configService: ConfigService,
     private val dataSourceModuleProvider: Provider<DataSourceModule?>
 ) : BreadcrumbService, ActivityLifecycleListener, MemoryCleanerListener {
 
-    override fun logView(screen: String?, timestamp: Long) {
-        dataSourceModuleProvider()?.viewDataSource?.dataSource?.changeView(screen, false)
+    override fun logView(screen: String?) {
```
bidetofevil, line 30: Do we not require the ability to specify a time in case the logging is delayed?

priettt: we always use the current clock time, at least for this flow!

bidetofevil: So it wasn't passed in or was it just never actually used? If it's the latter, we may have a bug, albeit a subtle one

bidetofevil: Looking at the code, I think we should actually respect this and pass it down. We should use the time when the API method is called instead of when we are processing the event, so I think it's right for the RN call to pass it down - and we should probably do the same for the manual `startView` and `endView` calls as well

bidetofevil (commented):
I would keep the interface the same with the startTime parameter and make it so that it's respected, but feel free to leave the other changes to `startView` for another PR

bidetofevil (approved):
LGTM!

PR #926, opened 2024-06-06
Remove unused session fields

fractalwrench:
## Goal

Removes fields from the `Session` object that are unused or already replaced in the new payload schema.


bidetofevil (approved):
LGTM

PR #927, opened 2024-06-06
Remove old payload collator

fractalwrench:
## Goal

Removes the v1 payload collator. This will not be required anymore as we'll want to send everything in the v2 payload format.


bidetofevil (approved):
WOW ❤️

PR #929, opened 2024-06-06
Create session zygote class

fractalwrench:
## Goal

Creates a `SessionZygote` class. This holds the state information at the very start of a session & is used when the session ends to create a full payload. Separating the concept of state holding & serialization will make it easier to fully remove `Session` from the payload in future.

## Testing

### DIFF
--- a/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/payload/SessionZygote.kt
+++ b/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/payload/SessionZygote.kt
@@ -1,0 +1,45 @@
+package io.embrace.android.embracesdk.payload
+
+/**
+ * A precursor object that holds state associated with a newly started session.
+ */
+internal data class SessionZygote(
+
+    /**
+     * A unique ID which identifies the session.
+     */
+    val sessionId: String,
+
+    /**
+     * The time that the session started.
+     */
+    val startTime: Long,
+
+    /**
+     * The ordinal of the session, starting from 1.
+     */
+    val number: Int,
+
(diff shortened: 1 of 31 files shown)


embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/payload/SessionZygote.kt
```diff
@@ -0,0 +1,43 @@
...
+     */
+    val isColdStart: Boolean,
+
+    /**
+     * The type of start event that triggered this session.
+     */
+    val startType: Session.LifeEventType
+) {
+
+    fun toSession() = Session(
```
bidetofevil, line 39: Should data classes have functions? Or should Sessions have this responsibility of being initialized via a zygote?

bidetofevil: Or maybe just an extension function would work to keep the class completely method free.

fractalwrench: This is a temporary shim that I should be able to remove entirely when we no longer need a `Session` object in the payload. There are a few more fields to remove until I get to that point.

FWIW we have a detekt rule that stops arbitrary functions in data classes but makes an allowance for conversion functions like this. I felt like that was a reasonable default but happy to adopt different style if you feel strongly

bidetofevil (approved):
I love this. A question about the session-create factory method in the data class and whether is should be a class function, but LGTM otherwise.

PR #931, opened 2024-06-07
Remove the population of emb.type from schema attributes

bidetofevil:
## Goal

Remove the population `emb.type` via an object's schema attributes. Instead, do it when the schema is being processed, which is already being done for spans so that it can't be overwritten by users who have a reference to the span object.

Correspondingly, changed where it's populated for logs and span events on the session span as well, which are the two other places that reads schema attributes.

## Testing
Fixed unit tests looking at DataSource *Data objects' attributes for the emb.type, and instead look at the schema itself to check right value. Integration test correctly check the attributes and therefore are still passing.


bidetofevil:
* **#931** <a href="https://app.graphite.dev/github/pr/embrace-io/embrace-android-sdk/931?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `master`

<h2></h2>

fractalwrench (approved):
LGTM

PR #932, opened 2024-06-07
Remove startup fields from session object

fractalwrench:
## Goal

Removes the startup fields from the session object as these should be calculated from spans instead.


bidetofevil (approved):
LGTM

PR #933, opened 2024-06-07
Remove end type from session object

fractalwrench:
## Goal

`endType` is no longer required on the session object as it is set as an attribute on the session span.


bidetofevil (approved):
LGTM

PR #934, opened 2024-06-07
Remove crash id

fractalwrench:
## Goal

Remove crash ID from the session object as it is now an attribute of the session span.

## Testing


bidetofevil (approved):
LGTM

PR #935, opened 2024-06-07
Remove log ids

fractalwrench:
## Goal

Removes IDs from the session object as these are no longer required.


embrace-android-sdk/src/integrationTest/kotlin/io/embrace/android/embracesdk/session/MomentBoundaryTest.kt
```diff
@@ -55,9 +55,6 @@ internal class MomentBoundaryTest {
             assertEquals(MOMENT_NAME, myEndMoment.event.name)
             assertEquals(EventType.END, myEndMoment.event.type)
             assertEquals(message.session.sessionId, myEndMoment.event.sessionId)
-
-            val expectedEventIds = listOf(startMoment.event.eventId, myStartMoment.event.eventId)
-            assertEquals(expectedEventIds, message.session.eventIds)
```
bidetofevil, line 60: We still have moments which are not OTel logs, so wouldn't we still need these?

fractalwrench: I've checked with Pablo & `s.ss` is not used by the backend

bidetofevil (approved):
LGTM

PR #937, opened 2024-06-09
Flesh out synchronization of session span transition

bidetofevil:
## Goal

Reviewing synchronization of the current span object and tweaked  it to prevent potentially double init. While it really wouldn't happen now, who knows what the future brings.

### DIFF
--- a/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/internal/spans/CurrentSessionSpanImpl.kt
+++ b/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/internal/spans/CurrentSessionSpanImpl.kt
@@ -36,8 +36,12 @@ internal class CurrentSessionSpanImpl(
     private val sessionSpan: AtomicReference<PersistableEmbraceSpan?> = AtomicReference(null)
 
     override fun initializeService(sdkInitStartTimeMs: Long) {
-        synchronized(sessionSpan) {
-            sessionSpan.set(startSessionSpan(sdkInitStartTimeMs))
+        if (sessionSpan.get() == null) {
+            synchronized(sessionSpan) {
+                if (sessionSpan.get() == null) {
+                    sessionSpan.set(startSessionSpan(sdkInitStartTimeMs))
+                }
+            }
         }
     }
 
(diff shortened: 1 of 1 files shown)


bidetofevil:
* **#937** <a href="https://app.graphite.dev/github/pr/embrace-io/embrace-android-sdk/937?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `master`

<h2></h2>

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/internal/spans/CurrentSessionSpanImpl.kt
```diff
@@ -36,8 +36,12 @@ internal class CurrentSessionSpanImpl(
     private val sessionSpan: AtomicReference<PersistableEmbraceSpan?> = AtomicReference(null)
 
     override fun initializeService(sdkInitStartTimeMs: Long) {
-        synchronized(sessionSpan) {
-            sessionSpan.set(startSessionSpan(sdkInitStartTimeMs))
+        if (sessionSpan.get() == null) {
+            synchronized(sessionSpan) {
+                if (sessionSpan.get() == null) {
```
fractalwrench, line 41: Is it worth formalising this pattern in a separate PR? It's raised a couple of questions in the past - I understand the rationale, but others might not

```
private inline fun doubleCheckLock(lock: Any, predicate: () -> Boolean, action: () -> Unit) {
        if (predicate()) {
            return
        }
        synchronized(lock) {
            if (predicate()) {
                return
            }
            action()
        }
    }
```

fractalwrench (approved):
LGTM

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/internal/spans/CurrentSessionSpanImpl.kt
bidetofevil: Yeah we should! We'll have to be careful how we implement these - we don't want a dangling reference in an `AtomicReference` causing leaks.

PR #939, opened 2024-06-10
Consolidate construction of session message

fractalwrench:
## Goal

Consolidates the construction of fake session messages to use mostly one fake. This will make future refactoring easier as there'll be primarily one place that sets the ID & other attributes in test code.


bidetofevil (approved):
Hell yeah LFG!

PR #940, opened 2024-06-10
startView with an existing name ends the previous one before starting a new one

priettt:
Also added unit tests


fractalwrench (approved):
LGTM

PR #941, opened 2024-06-10
Stop auto closing the last spans for automatic activity capture

priettt:
I want to make this change for consistency. When the session ends:
- If we start a view and don't end it, it is sent as a span snapshot.
- If we used the changeView API, the last span, the one from the view that hasn't changed yet, is sent as a span snapshot.
- **Why should we send a closed span for the last activity tracked if we could also send it as a span snapshot?**

Any span that hasn't ended is sent as a span snapshot and is shown correctly in our session timeline in the dashboard. I think we should make the different APIs in the BreadcrumbService behave in the same way.

### DIFF
--- a/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/capture/crumbs/ViewDataSource.kt
+++ b/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/capture/crumbs/ViewDataSource.kt
@@ -1,6 +1,5 @@
 package io.embrace.android.embracesdk.capture.crumbs
 
-import io.embrace.android.embracesdk.arch.datasource.NoInputValidation
 import io.embrace.android.embracesdk.arch.datasource.SpanDataSourceImpl
 import io.embrace.android.embracesdk.arch.datasource.startSpanCapture
 import io.embrace.android.embracesdk.arch.destination.StartSpanData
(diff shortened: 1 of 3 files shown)


embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/capture/crumbs/ViewDataSource.kt
```diff
@@ -32,12 +31,14 @@ internal class ViewDataSource(
...
 
     /**
-     * Called when a view is started.
+     * Called when a view is started. If a view with the same name is already running, it will be ended.
      */
     fun startView(name: String?): Boolean = captureSpanData(
         countsTowardsLimits = true,
         inputValidation = { !name.isNullOrEmpty() },
         captureAction = {
+            viewSpans[name]?.stop() // End the last view if it exists.
```
bidetofevil, line 40: This actually stops the span, so it shouldn't be a span shot anymore.

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/capture/crumbs/ViewDataSource.kt
```diff
@@ -65,21 +66,6 @@ internal class ViewDataSource(
         }
     )
 
-    /**
-     * Called when the activity is closed (and therefore all views are assumed to close).
-     */
-    fun onViewClose() {
```
bidetofevil, line 71: Why take this out? Leaving it here seems reasonable

bidetofevil (changes requested):
I don't think we can ignore the stop call - that will change the end time to when the session ends, which wouldn't work if an explicit `endView` call is made.

PR #942, opened 2024-06-10
Remove session id + start time

fractalwrench:
## Goal

Replaces the session ID with a span attribute & start time with the span start time.

## Testing


embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/comms/delivery/EmbraceDeliveryService.kt
```diff
@@ -190,7 +192,8 @@
         sessionMessage.session.endTime
             ?: sessionMessage.session.terminationTime
             ?: sessionMessage.session.lastHeartbeatTime
-            ?: sessionMessage.session.startTime
+            ?: sessionMessage.getSessionSpan()?.startTimeNanos?.nanosToMillis()
+            ?: -1
```
bidetofevil, line 196: We should get the backend to have a metric to detect an endTime of -1 so this basically impossible condition is caught

bidetofevil (approved):
LGTM

PR #943, opened 2024-06-10
Extract enums from session source file

fractalwrench:
## Goal

Extracts enums to separate source files.

## Testing


bidetofevil (approved):
Ah, this is the stuff

PR #945, opened 2024-06-11
Populate span attributes with end time as required

fractalwrench:
## Goal

The Android SDK now sends the following fields that are all that is necessary to calculate end time for the backend:

1. `emb.terminated` as a boolean
2. `emb.clean_exit` as a boolean
3. The session span end time

## Testing

Added unit test coverage.


bidetofevil (approved):
LGTM

PR #948, opened 2024-06-11
Remove dead code used for constructing session payloads

fractalwrench:
## Goal

Removes dead code used for constructing session payloads. These conditions are no longer required after the migration to OTel so the code can be simplified.

## Testing


priettt (approved):
LGTM!

PR #950, opened 2024-06-11
Hook up session gating

fractalwrench:
## Goal

Session gating was previously relying on the v1 implementation. This hooks up the gating service so that spans in the v2 payload are sanitized.

## Testing


priettt (approved):
Looks good! 

We might want to rethink the `crashId` string parameter. We never really use the id, so maybe a boolean like sessionHasCrashes might be more accurate.

fractalwrench:
Updated to use a bool parameter instead.

PR #951, opened 2024-06-11
Remove OTel flags

priettt:
- Removed LoggingApiTest, as that is now tested by OTelLoggingApiTest
- NativeCrashService is always wrapped in a DataSource
- CompositeLogService will always return the v2 log service, but I will clean that class in another PR


fractalwrench (approved):
LGTM

PR #952, opened 2024-06-11
Remove legacy breadcrumb classes

fractalwrench:
## Goal

Removes the legacy breadcrumb classes by circumventing the need to create them when adding span events/starting spans. This required a small change to the span writing internal API but is covered by existing integration tests for each feature.

## Testing

Updated existing test coverage where necessary.


priettt (approved):
LGTM!

PR #953, opened 2024-06-11
Remove old payload classes

fractalwrench:
## Goal

Removes legacy payload classes by circumventing the need to create them when adding span events/starting spans. This required a small change to the span writing internal API but is covered by existing integration tests for each feature.

## Testing

Updated existing test coverage where necessary.


embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/anr/sigquit/SigquitDataSource.kt
```diff
@@ -36,18 +34,12 @@ internal class SigquitDataSource(
 
     fun saveGoogleAnr(timestamp: Long) {
```
priettt, line 35: should we call this sigquit too?

fractalwrench: I'll rename it.

Sigquit is the C signal that the Android OS sends when an ANR happens. 'Google ANR' is just internal jargon that we've created - IMO it'd be better to be precise with our meaning & use sigquit terminology instead when discussing this.

PR #954, opened 2024-06-11
Remove dead logging code

priettt:
This ended up being huge 😞 I tried to divide it by commits, so you can see what the steps were, but here are the overall milestones:

- Removed EmbraceLogMessageService.
- Removed CompositeLogService, as we only use LogService now. Some of the logic present in this class was moved to EmbraceLogService.
- Removed unused methods from the LogService interface.


fractalwrench (approved):
LGTM

PR #955, opened 2024-06-12
Remove unnecessary layer

fractalwrench:
## Goal

Removes an unnecessary layer in the data source implementation that was preventing us from getting rid of some otherwise unused payload classes.

## Testing


bidetofevil (approved):
Awesome!

PR #956, opened 2024-06-12
Remove unused payload classes

fractalwrench:
## Goal

Removes some payload classes that are no longer used to capture data but were still left in the codebase.

## Testing


bidetofevil (approved):
Ah, I guess we never ported the heartbeat stuff over eh? This FEELS like something that could be useful in the future. If so, we can always revive this from history 😅

PR #957, opened 2024-06-12
Use envelope for session payload

fractalwrench:
## Goal

Uses the `Envelope<SessionPayload>` class rather than `SessionMessage`. This will allow further reuse of code in future as all types are now stored using `Envelope`.

## Testing


bidetofevil (approved):
I didn't see anything in this that would've caused or exacerbate the race condition that led to the flakey test, so I'm good with merging this.

PR #958, opened 2024-06-12
Remove json serializers

fractalwrench:
## Goal

Removes some unused JSON serializers that are now unused. The classes themselves currently still need to be kept as they are used for holding data.


bidetofevil (approved):
We may want to move these out of the `payload` package into some sort of intermediate POJO package in a different PR given they are no longer being used as payload. But not a big deal.

fractalwrench:
Yep - my thinking was it makes sense to move them to their respective packages (e.g. `capture.aei` and `capture.anr` for AEI/ANR classes respectively). Would prefer to do that once this is merged to master though

PR #959, opened 2024-06-12
Migrate disk usage to otel payload

fractalwrench:
## Goal

Migrates disk usage to the new payload schema. This is implemented as:

`resource.disk_space`: gives the total disk space available on the device
`emb.disk_free_bytes`: session span attribute that gives the free bytes on the device

## Testing

Added unit test coverage.


bidetofevil (approved):
LGTM

PR #963, opened 2024-06-13
Add semantic conventions dependency

fractalwrench:
## Goal

Adds the OTel semantic conventions repo as a dependency. This will allow us to use OTel's built-in conventions rather than reinventing the wheel with our own.

This should not be merged until the corresponding swazzler PR has been raised.

## Testing

## Release Notes

**WHAT**:Added a build warning for users that have minSdk <24 and are using AGP <8.3<br>
**WHY**:There is a desugaring bug on old AGP versions that can cause runtime crashes<br>
**WHO**:Anyone with minSdk <24 using AGP <8.3<br>


bidetofevil (approved):
Provisional approval assuming we can up the AGP version support with minSdk 21

PR #964, opened 2024-06-14
Use semantic convention constants

fractalwrench:
## Goal

Replaces existing OTel semantic convention keys with the constants defined in the semantic convention dependency.

Note: as part of this I noticed that `os.model.identifier` and `os.model.name` were mistakenly set as a resource value - they should have been prefixed with `device`. I've chosen to fix this as I believe it will only affect OTel export & doesn't need changes on our backend.

## Testing


bidetofevil (approved):
LGTM

bidetofevil (commented):
Oops! Yeah, those keys are not used by us and in fact the resource object isn't even sent to our backend via the envelope. We probably should though - the `resource` object in the payload envelope should be a superset of this.

The reason this wasn't done was because the the attribute naming. Our backend doesn't like the `.` in the key names, so we never merged the two. There's an outstanding item to resolve this that we were going to do after we shipped our first version.... which is now.

More details about this here: https://www.notion.so/embraceio/How-do-we-name-payload-envelope-fields-that-are-also-OTel-attributes-and-resources-given-the-conflic-4e8bc1ea3a4c457c9243774ada475c90?pvs=4

bidetofevil (approved):
LGTM

PR #965, opened 2024-06-14
Use further semantic convention constants

fractalwrench:
## Goal

I searched through the [semantic convention attribute registry](https://opentelemetry.io/docs/specs/semconv/attributes-registry/) and replaced string literals with constants where there was a direct match.

## Testing


bidetofevil (approved):
LGTM

PR #968, opened 2024-06-17
Fix flaky test

fractalwrench:
## Goal

`EmbraceCacheServiceConcurrentAccessTest` started failing sporadically because the `TestPlatformSerializer` did not override all serializer functions. The serialization code was recently changed to use a different function signature & therefore execution wasn't paused, resulting in a race condition that failed some assertions.

I ran this test case ~10 times locally to verify it now passes & also avoided inheritance by delegation to ensure this problem is more obvious if it ever happens in future.


bidetofevil (approved):
Ugh. Of course! Thanks for fixing this!

PR #969, opened 2024-06-17
Prevent lateinit property exception

fractalwrench:
## Goal

The `NetworkConnectivityService` registers for network changes during its initialization. As this service has a bootstrapped reference to the `DataSourceModule` which is initialized after it, in rare circumstances where startup is slow + a network connectivity change happens this could result in access to a lateinit var that isn't initialized, which was captured as an internal error.

I've fixed this by ensuring that connectivity change registration only happens after the `DataSourceModule` is initialized.


bidetofevil (approved):
LGTM

PR #970, opened 2024-06-17
Remove unnecessary parameters from log counter

fractalwrench:
## Goal

The `LogCounter` class will on rare ocassions throw an exception complaining about an inconsistent range from the when finding a submap of log IDs. This is reported as an internal error. It's unclear under what circumstances this happens but we don't need to supply those parameters any more in the v2 payload so we may as well just get rid of them.


bidetofevil (approved):
LGTM

PR #973, opened 2024-06-19
Remove unnecessary warning

fractalwrench:
## Goal

Removes a misleading error message. `isInBackground` relies on the current state of the `LifecycleOwner`. The internal implementation of this sets the new state & then invokes the callback, meaning it's possible for an 'unbalanced' call to happen on the first launch, although practically speaking this makes no difference.


bidetofevil (approved):
LGTM. 

FWIW, I wouldn't mind adding an internal error log for when we encounter cases where background/foreground isn't what's expected when these callbacks are invoked AFTER the init call. But removing this and adding that later as separate tasks seem like the right thing to do.

PR #974, opened 2024-06-19
Add limit for internal span capture

fractalwrench:
## Goal

Adds a limit of 1000 for the number of internal spans that we capture. I'm open to discussion on the exact limit we enforce, but noticed during my validation of this migration that the number of internal spans was unbounded.

Technically we should always bound the number of spans in the `DataSource` implementation, but in practice I'd like to add this as a backstop to prevent the payload from getting overloaded.

## Testing

Added unit tests.

### DIFF
--- a/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/internal/spans/SpanServiceImpl.kt
+++ b/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/internal/spans/SpanServiceImpl.kt
@@ -140,6 +140,7 @@ internal class SpanServiceImpl(
     }
 
     companion object {
+        const val MAX_INTERNAL_SPANS_PER_SESSION = 5000
         const val MAX_NON_INTERNAL_SPANS_PER_SESSION = 500
     }
 }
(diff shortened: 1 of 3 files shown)


embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/internal/spans/SpanServiceImpl.kt
```diff
@@ -140,6 +140,7 @@ internal class SpanServiceImpl(
     }
 
     companion object {
+        const val MAX_INTERNAL_SPANS_PER_SESSION = 1000
```
bidetofevil, line 143: I think we should bump this limit - network requests will easily push past this, and we allow 1000 per session per domain, IIRC. I'd go 3000 to start off with, and maybe look at existing data to inform this limit?

bidetofevil (approved):
LGTM. We may want to bump the limit though because network requests will blow past this easily

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/internal/spans/SpanServiceImpl.kt
fractalwrench: Good point. I bumped this to 5000 & we can adjust from there. Really it's just a backstop to avoid stupid amounts of spans getting logged & in practice I don't think this will be reached for the vast majority of payloads

PR #975, opened 2024-06-19
Remove disk field

fractalwrench:
## Goal

Removes the `disk_space` field in `Resource` as it's already covered by `disk_total_capacity`.


bidetofevil (approved):
LGTM

PR #976, opened 2024-06-19
Avoid dropping sessions

fractalwrench:
## Goal

Resolves an issue with the integration branch where sessions weren't persisted to disk as the session ID couldn't be found in the payload. The fix is to look first in the `spans` field, and then in the `span_snapshots` field for the session span.

I've also improved the error messaging around failed HTTP requests as it was initially very hard to tell that a 400 status code was received by the server. Finally, I've tweaked the `type` value to `spans` as this seems to be what the mock api expects, although the field is unused by the actual implementation at this time.

### DIFF
--- a/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/internal/payload/Envelope.kt
+++ b/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/internal/payload/Envelope.kt
@@ -41,6 +41,7 @@ internal data class Envelope<T>(
 
 internal fun Envelope<SessionPayload>.getSessionSpan(): Span? {
     return data.spans?.singleOrNull { it.hasFixedAttribute(EmbType.Ux.Session) }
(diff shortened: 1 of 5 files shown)


embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/internal/payload/Envelope.kt
```diff
@@ -41,6 +41,7 @@ internal data class Envelope<T>(
 
 internal fun Envelope<SessionPayload>.getSessionSpan(): Span? {
     return data.spans?.singleOrNull { it.hasFixedAttribute(EmbType.Ux.Session) }
+        ?: data.spanSnapshots?.singleOrNull { it.hasFixedAttribute(EmbType.Ux.Session) }
```
bidetofevil, line 44: Oh yeah this would be a thing

bidetofevil (approved):
LGTM

PR #979, opened 2024-06-21
Update code coverage to include results from all modules

fractalwrench:
## Goal

Updates Kover to the latest version & configures it so that code coverage is included for all current modules.

## Testing

Looked at the code coverage report locally & will check CI too.


bidetofevil (approved):
still above 80% 💛

PR #980, opened 2024-06-21
Convert build script to Kotlin

fractalwrench:
## Goal

Converts the majority of the build script to use Kotlin instead of Gradle. This gives us the advantage of IDE auto completion & is now the default for Android projects.


bidetofevil (approved):
Very nice!

PR #981, opened 2024-06-21
Set device free disk bytes to null when not available

fractalwrench:
## Goal

Avoid serializing `"null"` when `deviceDiskFree` is not available.


bidetofevil (approved):
LGTM

PR #982, opened 2024-06-21
Use correct method signature for sigquit

fractalwrench:
## Goal

We were using the wrong JDI signature for our SIGQUIT capture which led to silent failure. This only affects the migration branch.


bidetofevil (approved):
LGTM

PR #984, opened 2024-06-24
Use version catalog for lib dependencies

fractalwrench:
## Goal

Uses a [version catalog](https://developer.android.com/build/migrate-to-catalogs) to specify library dependencies.


bidetofevil (approved):
Yeah!

PR #985, opened 2024-06-24
Convert binary plugin to convention plugin

fractalwrench:
## Goal

Converts our existing binary plugin to a convention plugin which is pre-compiled & applied to each module. This is more appropriate as we don't exhibit any behavior that is controlled via an extension & it also uses a more declarative syntax which will be more familiar to folks used to configuring gradle.


bidetofevil (approved):
LGTM

PR #986, opened 2024-06-24
Only start a new session span if it's appropriate

bidetofevil:
## Goal

Pass down whether or not ending a session in the CurrentSessionSpan should start a new session, and make the check appropriately based on end type as well as whether background sessions are enabled.

## Testing
Added new tests for the condition

### DIFF
--- a/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/internal/spans/CurrentSessionSpanImpl.kt
+++ b/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/internal/spans/CurrentSessionSpanImpl.kt
@@ -83,7 +83,7 @@ internal class CurrentSessionSpanImpl(
         return sessionSpan.get()?.getSystemAttribute(embSessionId) ?: ""
     }
 
-    override fun endSession(appTerminationCause: AppTerminationCause?): List<EmbraceSpanData> {
+    override fun endSession(appTerminationCause: AppTerminationCause?, startNewSession: Boolean): List<EmbraceSpanData> {
         synchronized(sessionSpan) {
             val endingSessionSpan = sessionSpan.get()
             return if (endingSessionSpan != null && endingSessionSpan.isRecording) {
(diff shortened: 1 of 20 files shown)


bidetofevil:
* **#986** <a href="https://app.graphite.dev/github/pr/embrace-io/embrace-android-sdk/986?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `master`

<h2></h2>

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/internal/spans/CurrentSessionSpanImpl.kt
```diff
@@ -98,7 +98,9 @@ internal class CurrentSessionSpanImpl(
                 if (appTerminationCause == null) {
                     endingSessionSpan.stop()
                     spanRepository.clearCompletedSpans()
-                    sessionSpan.set(startSessionSpan(openTelemetryClock.now().nanosToMillis()))
+                    if (startNewSession) {
```
fractalwrench, line 101: This introduces undesirable behavior where calling `sessionSpan.get()` afterwards means folks could attempt to write to the span after the session has ended. Setting the reference to `null` if no new session is created should be enough to fix it?

bidetofevil: Good call. Will fix tomorrow!

bidetofevil: This brings up an interesting point about what happens if we log traces when there is no session span - or do anything when there is no session span.

Before OTel, that data is just lost, right?

bidetofevil: I'll put up the fix for this but lets figure this out before I merge this - we should have tests to verify this case and ensure we don't crash

bidetofevil: Added tests to verify no-session-span state. This should be good to go.

bidetofevil:
Differentiated between initialization and whether a session span exists, as there doesn't necessarily have to be one now, if we are in the background and background activity isn't on.

fractalwrench (approved):
LGTM

PR #987, opened 2024-06-25
Run dependabot on gradle dependencies

fractalwrench:
## Goal

Use the [gradle package ecosystem](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#package-ecosystem) to search for dependency updates in our gradle build.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,6 +1,6 @@
 version: 2
 updates:
-  - package-ecosystem: "github-actions"
+  - package-ecosystem: "gradle"
     directory: "/"
     schedule:
       interval: "daily"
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -1,6 +1,6 @@
 version: 2
 updates:
-  - package-ecosystem: "github-actions"
+  - package-ecosystem: "gradle"
     directory: "/"
```
vitaliyf, line 4: no, keep actions, but add gradle as another - package ecosystem below

github-actions is used to update /.github/workflows/*.yaml so that you don't end up getting stuck using old versions of things like "actions/checkout@v4" in there since they update/deprecate them pretty frequently

vitaliyf (approved):

bidetofevil (approved):
LGTM

PR #989, opened 2024-06-25
Migrate to OTel semconv keys

priettt:
- Alter the keys used in the payload to match OTel’s semantic conventions.
- Deleted unused SchemaType objects.

Backend ready! Tested and ready to merge :)

### DIFF
--- a/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/anr/AnrOtelMapper.kt
+++ b/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/anr/AnrOtelMapper.kt
@@ -10,6 +10,8 @@ import io.embrace.android.embracesdk.payload.AnrInterval
 import io.embrace.android.embracesdk.payload.AnrSample
 import io.opentelemetry.api.trace.SpanId
 import io.opentelemetry.sdk.trace.IdGenerator
+import io.opentelemetry.semconv.JvmAttributes
+import io.opentelemetry.semconv.incubating.ExceptionIncubatingAttributes
 
 /**
  * Maps captured ANRs to OTel constructs.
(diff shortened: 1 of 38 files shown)


embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/anr/AnrOtelMapper.kt
```diff
@@ -67,12 +69,12 @@ internal class AnrOtelMapper(
             attrs.add(Attribute("sample_code", it.toString()))
         }
         sample.threads?.singleOrNull()?.let { thread ->
-            attrs.add(Attribute("thread_state", thread.state.toString()))
+            attrs.add(Attribute(JvmAttributes.JVM_THREAD_STATE.key, thread.state.toString()))
```
bidetofevil, line 72: Does the value here match the definition in the semantic conventions? I want to make sure what we are writing is what's expected - otherwise, we should keep using our custom one.

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/arch/schema/SchemaType.kt
```diff
@@ -272,23 +191,23 @@ internal sealed class SchemaType(
...
+            HttpIncubatingAttributes.HTTP_REQUEST_BODY_SIZE.key to networkCapturedCall.requestBodySize.toString(),
             "request-query" to networkCapturedCall.requestQuery,
-            "request-query-headers" to networkCapturedCall.requestQueryHeaders.toString(),
+            "http.request.header" to networkCapturedCall.requestQueryHeaders.toString(),
             "request-size" to networkCapturedCall.requestSize.toString(),
             "response-body" to networkCapturedCall.responseBody,
-            "response-body-size" to networkCapturedCall.responseBodySize.toString(),
-            "response-headers" to networkCapturedCall.responseHeaders.toString(),
+            HttpIncubatingAttributes.HTTP_RESPONSE_BODY_SIZE.key to networkCapturedCall.responseBodySize.toString(),
+            "http.response.header" to networkCapturedCall.responseHeaders.toString(),
```
bidetofevil, line 204: Are there no conventions for headers?

bidetofevil (approved):
A couple of comments, but looks good otherwise.

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/anr/AnrOtelMapper.kt
priettt: yeah, I think so. Sample values include `runnable` and `blocked`, and those are part of the states in `Thread.java` (we get the values from there)

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/arch/schema/SchemaType.kt
priettt: HttpAttributes.HTTP_REQUEST_HEADER is a stringArrayKeyTemplate, so we can't get the key. 

The docs say this: 

> The attribute value MUST consist of either multiple header values as an array of strings or a single-item array containing a possibly comma-concatenated string, depending on the way the HTTP library provides access to headers.

We are doing the second option, but there is no way of getting the key, at least with the stringArrayKeyTemplate class.

bidetofevil: Got it. Cool.

PR #998, opened 2024-06-26
Explicitly set the default for span creation to non-private

bidetofevil:
## Goal

Make interfaces default to creating spans that are non-private. This is already the case for the create and start span APIs, but the other methods still defaulted to pairing private with internal, which can be confusing.

In practice, the only change is that network request spans are now non-private, which they should've been anyway. All other changes are removing explicit parameters that match the default, which doesn't change existing behaviour

## Testing
Modified existing tests to verify the default. Did not add additional tests for network spans because that's just using the default now.


bidetofevil:
* **#998** <a href="https://app.graphite.dev/github/pr/embrace-io/embrace-android-sdk/998?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `master`

<h2></h2>

priettt (approved):

PR #1001, opened 2024-06-26
Added missing proguard rules

priettt:
## Goal

Avoiding apps with `minifyEnabled = true` failing on compilation.


bidetofevil (approved):
LGTM

PR #1020, opened 2024-07-02
Setup common module for test code

fractalwrench:
## Goal

Creates a new common module that is added as a `testImplementation` dependency for all modules in the project. This will make it easier to share test code between modules. Initially I've just moved `ResourceReader` over but will do a pass of other classes in a separate changeset.

@bidetofevil I think there's also a bit of a ideological question here - do we want the `embrace-test-common` module to have zero dependencies on any other module? Or should it have a dependency on the `embrace-android-sdk` module so that it can host most of the fakes? Perhaps this is something that will become more apparent later on in the modularisation?


bidetofevil (approved):
I like the hosting fakes idea, but I'm not thrilled about at test module sitting on top of the actual SDK module.

What do you think about creating an interface/data classes module that sits at the very bottom, and both the SDK module and the test-common module can depend on that? That way, the module with the actual implementations won't need to be a dependency, but we also allow it to contain all the fake implementations?

PR #1021, opened 2024-07-02
Delete unused test resources

fractalwrench:
## Goal

Deletes some unused resources that weren't deleted when old unit tests were.


bidetofevil (approved):
LGTM

PR #1026, opened 2024-07-03
Rename function on data source

fractalwrench:
## Goal

Renames `alterSessionSpan` to `captureData` as IMO this better reflects what the function is doing.


bidetofevil (approved):
LGTM

PR #1035, opened 2024-07-03
Create registry that tracks what data source states are used in the SDK

fractalwrench:
## Goal

Currently most features in the SDK are built using a `DataSourceState`. These were passed into the `DataCaptureOrchestrator` as a list which then informs them when new data needs capturing in response to changes in the session boundary/config.

This PR alters this so a `DataSourceState` is added directly to the `DataCaptureOrchestrator`. I've also tweaked `DataSource` so that consumers of the internal API don't have the ability to set the `currentSessionType` state. Finally, this also adds support for adding `DataSourceState` values _after_ the session orchestration has started by populating the initial session state. Currently this doesn't happen but is likely to do in future.

In a future changeset I'll investigate adding a way to initialize data sources on a background thread as that should help reduce startup impact substantially.

## Testing

Updated unit test coverage.

### DIFF
--- a/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/arch/DataCaptureOrchestrator.kt
+++ b/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/arch/DataCaptureOrchestrator.kt
@@ -4,16 +4,16 @@ import io.embrace.android.embracesdk.arch.datasource.DataSourceState
 import io.embrace.android.embracesdk.capture.internal.errors.InternalErrorType
 import io.embrace.android.embracesdk.config.ConfigService
 import io.embrace.android.embracesdk.logging.EmbLogger
+import java.util.concurrent.CopyOnWriteArrayList
 
 /**
  * Orchestrates all data sources that could potentially be used in the SDK. This is a convenient
  * place to coordinate everything in one place.
  */
 internal class DataCaptureOrchestrator(
-    private val dataSourceState: List<DataSourceState<*>>,
-    private val logger: EmbLogger,
-    configService: ConfigService
-) {
+    configService: ConfigService,
+    private val logger: EmbLogger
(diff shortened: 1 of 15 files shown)


embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/arch/DataCaptureOrchestrator.kt
```diff
@@ -4,25 +4,31 @@ import io.embrace.android.embracesdk.arch.datasource.DataSourceState
...
+    private val logger: EmbLogger
+) : EmbraceFeatureRegistry {
 
     init {
         configService.addListener {
             onConfigChange()
         }
     }
 
+    private val dataSourceStates = CopyOnWriteArrayList<DataSourceState<*>>()
```
fractalwrench, line 24: Chosen for synchronisation strategy

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/arch/EmbraceFeatureRegistry.kt
```diff
@@ -0,0 +1,15 @@
+package io.embrace.android.embracesdk.arch
+
+import io.embrace.android.embracesdk.arch.datasource.DataSourceState
+
+/**
+ * Registry for all features whose instrumentation should be orchestrated by the Embrace SDK.
+ */
+internal interface EmbraceFeatureRegistry {
```
fractalwrench, line 8: I anticipate exposing this as part of the internal API between modules that allows us to add automatic instrumentation to the Embrace SDK

bidetofevil (approved):
Nice. This seems right. I've been somewhat cavalier when it comes to internal object relationships, thinking we'll clean it up after. Glad we're doing this now!

PR #1041, opened 2024-07-03
Fix lint warning in agp 8.5

fractalwrench:
## Goal

Fixes a lint warning in AGP 8.5 that showed up in #1032. Basically the supplied parameter value is not used & we can use the default value instead.


bidetofevil (approved):
LGTM

bidetofevil (approved):
LGTM

PR #1042, opened 2024-07-03
Prevent eager initialization of data sources

fractalwrench:
## Goal

Prevents the eager initialization of data sources until the `configGate` is `true` and a session is under way.


bidetofevil (approved):
LGTM

PR #1043, opened 2024-07-03
Proguard rules improvements

priettt:
## Goal

Customers using R8 to obfuscate and shrink their app weren't able to shrink Embrace, because we had proguard rules instructing to keep all classes. This PR changes that. 

In `android-test-app`, turning this on reduced Embrace size by 70%, comparing against master (from 800kb to 250kb, it's not like we were too heavy in the first place 😅)

- Remove unnecessary rules:
    - java9.**
    - okhttp3.** and okio.**
    - Attributes: Exceptions, InnerClasses, Signature.  

- Removed keep rule for the whole embrace package
    - Anyone using r8 to either obfuscate or shrink their apk can now remove unused Embrace code.
 
- Added rules to keep classes used by hosted SDKs, native code, okhttp, swazzled and serialized classes.

## Testing

- Tested in these apps: 
    - https://github.com/embrace-io/small-android-test-app
    - https://github.com/embrace-io/android-test-suite
    - https://github.com/embrace-io/embrace-react-native-sdk/tree/master/examples/react-native-test-suite
    - https://github.com/embrace-io/embrace-unity-sdk-internal/tree/main/UnityProjects/2023 (created a project from the Unity editor and tested in Android Studio with that project)

- Ran swazzler tests

### DIFF
--- a/embrace-android-sdk/embrace-proguard.cfg
+++ b/embrace-android-sdk/embrace-proguard.cfg
@@ -1,8 +1,13 @@
 -keepattributes LineNumberTable, SourceFile
 
-## Proguard configuration for Embrace
--keep class io.embrace.android.embracesdk.** { *; }
--dontwarn io.embrace.android.embracesdk.**
+## Keep classes used by hosted SDKs
+-keep class io.embrace.android.embracesdk.Embrace { *; }
+-keep class io.embrace.android.embracesdk.Embrace$AppFramework { *; }
+-keep class io.embrace.android.embracesdk.Embrace$LastRunEndState { *; }
+-keep class io.embrace.android.embracesdk.Severity { *; }
+-keep public class * implements io.embrace.android.embracesdk.internal.EmbraceInternalInterface {
+    *;
+}
 
 ## OpenTelemetry Java SDK
 -keep class io.opentelemetry.api.trace.StatusCode { *; }
--- a/embrace-android-sdk/embrace-proguard.cfg
+++ b/embrace-android-sdk/embrace-proguard.cfg
@@ -5,9 +5,12 @@
 -keep class io.embrace.android.embracesdk.Embrace$AppFramework { *; }
 -keep class io.embrace.android.embracesdk.Embrace$LastRunEndState { *; }
 -keep class io.embrace.android.embracesdk.Severity { *; }
--keep public class * implements io.embrace.android.embracesdk.internal.EmbraceInternalInterface {
-    *;
-}
+-keep public class * implements io.embrace.android.embracesdk.internal.EmbraceInternalInterface { *; }
+
+## Keep classes used from native code
+-keep class io.embrace.android.embracesdk.payload.NativeThreadAnrSample { *; }
+-keep class io.embrace.android.embracesdk.payload.NativeThreadAnrStackframe { *; }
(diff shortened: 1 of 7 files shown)


bidetofevil (approved):
Nice! It'd be cool if we can manage this using annotations rather than a proguard file, but this is great!

embrace-android-sdk/embrace-proguard.cfg
```diff
@@ -1,20 +1,33 @@
...
--dontwarn okhttp3.**
--dontwarn okio.**
-
-## Proguard configuration for Arrow
--keep class java9.** { *; }
--dontwarn java9.**
+## Keep classes used from native code
+-keep class io.embrace.android.embracesdk.payload.NativeThreadAnrSample { *; }
+-keep class io.embrace.android.embracesdk.payload.NativeThreadAnrStackframe { *; }
+-keep class io.embrace.android.embracesdk.anr.sigquit.SigquitDataSource { *; }
```
fractalwrench, line 13: Do we also need to keep `NativeThreadSamplerNdkDelegate`, or will that work ok with R8's obfuscation? It's probably worth global-searching for any `external fun` in the project & checking they work.

embrace-android-sdk/embrace-proguard.cfg
```diff
@@ -1,20 +1,33 @@
--keepattributes Exceptions, InnerClasses, Signature, LineNumberTable, SourceFile
```
fractalwrench, line 1: What's the effect of removing the rule to keep these attributes? Instinctively I'd have thought we would want to retain information about inner classes/exception metadata

fractalwrench (approved):
LGTM pending comments

embrace-android-sdk/embrace-proguard.cfg
priettt: I verified that they work, but I'm not sure if it's because they're referenced by kept classes or if it's because obfuscation doesn't affect JNI calls. I'll check just in case because I haven't tested every external fun.

The classes referenced from the native code must be kept because they're searched for by name.

embrace-android-sdk/embrace-proguard.cfg
priettt: We only need LineNumberTable and SourceFile.
- `Exceptions` keeps info on what exceptions a method might throw (the `throws` part of a function signature). We don't use it.
- `Signature` keeps generics type information. If not present, instead of <T> it'll use Object. We don't use it.
- `InnerClasses` keeps the relationship between classes. If it's not present, inner classes might be renamed and their relationship with their outer class might be lost. This doesn't affects us per se, but some stacktraces might change. Keeping it wouldn't be a problem, but I don't think it's necessary.

embrace-android-sdk/embrace-proguard.cfg
priettt: The classes are always being kept.

Not because we are actually adding any rules for them, but **I think** it's because they are being initialized in modules that override `kotlin.jvm.functions.Function0.invoke()`, and those are also called in different places in ComponentActivity, and every activity of any app is added  because of some aapt_rules.

It's super weird, and I don't want to lose much time on this, so I'll just add rules for the classes that contain `external fun` methods just in case 😅

PR #1053, opened 2024-07-05
Support async init of features

fractalwrench:
## Goal

Supports the async initialization of features in Embrace. While this changeset does not alter any feature to be initialized async it creates the infrastructure to do so.

## Testing

Added unit test coverage.


bidetofevil (approved):
LGTM. Can't wait to see the init perfetto trace

PR #1056, opened 2024-07-05
Create interface for internal apis

fractalwrench:
## Goal

Creates an interface that is used for accessing the unity/flutter/react native internal interfaces. This change will ultimately make it easier to create a shim interface around `Embrace/EmbraceImpl`


bidetofevil (approved):
LGTM

PR #1060, opened 2024-07-08
Preparatory refactor for making EmbraceImpl an interface

fractalwrench:
## Goal

Preparatory refactor that will make it easier to create an interface out of `EmbraceImpl`, which will make modularisation possible.

### DIFF
--- a/embrace-android-sdk/api/embrace-android-sdk.api
+++ b/embrace-android-sdk/api/embrace-android-sdk.api
@@ -245,6 +245,18 @@ public abstract interface class io/embrace/android/embracesdk/internal/api/Inter
 	public abstract fun getUnityInternalInterface ()Lio/embrace/android/embracesdk/UnityInternalInterface;
 }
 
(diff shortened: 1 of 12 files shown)


embrace-android-api/src/main/kotlin/io/embrace/android/embracesdk/internal/api/NetworkRequestApi.kt
```diff
@@ -15,7 +15,7 @@ public interface NetworkRequestApi {
      */
     public fun recordNetworkRequest(networkRequest: EmbraceNetworkRequest)
 
-    public fun getTraceIdHeader(): String
+    public val traceIdHeader: String
```
bidetofevil, line 18: Given that we have NSF as well as OTel stuff to consider, this naming is pretty problematic. 

We shouldn't change it in this PR, but we should probably change this when we bump our SDK version to 7.0

bidetofevil (approved):
LGTM

PR #1061, opened 2024-07-08
Hide embrace impl behind interface

fractalwrench:
## Goal

Creates an interface for `EmbraceImpl`. This will allow us to set the embrace implementation on the facade without directly referencing it.


bidetofevil (approved):
LGTM

PR #1062, opened 2024-07-08
Alter internal api visibility

fractalwrench:
## Goal

Alters the internal API interface visibility in preparation for moving them to the API module.


bidetofevil (approved):
LGTM. 

Are you planning to add any annotations to these public-but-not-public interfaces? We should definitely add `@hide` but maybe also `@InternalOnly` or something to make it explicit?

fractalwrench:
@bidetofevil `@hide` doesn't do anything for non-framework classes but I added `@InternalApi` to all these interfaces. Tagging for re-review as it altered a reasonable number of source files.

bidetofevil:
I believe @hide removes it from the generated documented we have. Will these be similarly hidden?

bidetofevil (approved):
Looks good, if these internal public classes can be removed from the generated docs.

fractalwrench:
Addressed outstanding docs issues in #1081. I also added a task for us to do a sweep at the end of the project so we don't unintentionally include unwanted symbols in the docs.

PR #1063, opened 2024-07-08
Support app startup initialization

fractalwrench:
## Goal

Initializes the SDK using the [App Startup library](https://developer.android.com/topic/libraries/app-startup) rather than manual invocation. This happens earlier in the process lifecycle & requires less manual intervention from a user during the integration process.

## Testing

Manually tested via our test app.


bidetofevil (approved):
LGTM

PR #1064, opened 2024-07-08
Allow setting app framework via embrace-config.json

fractalwrench:
## Goal

Adds the option to set the AppFramework via embrace-config.json and makes this the preferred method of doing so. The old way of setting AppFramework via `Embrace.getInstance.start(context, AppFramework)` is still supported for now. `embrace-config.json` takes precedence, then anything supplied in the `start` function, followed by the SDK default of `AppFramework.NATIVE`.

## Testing

Added unit test coverage.


bidetofevil (approved):
LGTM. The idea is then to have all folks just older versions to add the appropriate value in the config file eventually? We can probably do this as part of the bump to 7.0

fractalwrench:
Yep, that's the idea. We can add instructions in the changelog for the next release, and the migration guide for 7

PR #1066, opened 2024-07-08
Remove automatic verification from EmbraceSamples

priettt:
## Goal

Automatic verification is seldom used, and it causes a bit of confusion, as some customers are actually keeping it on release builds. This PR removes it altogether.


bidetofevil (approved):
Will this require the UI to be updated? If so, we may not want to do this quite yet - we can remove the email thing, cherrypick that, and then do this so it gives the UI folks more time to change the integration flow.

fractalwrench (approved):
LGTM but we should not merge this until UI changes to our dash are deployed. Can you coordinate those changes with the relevant folks (if you haven't already)?

PR #1067, opened 2024-07-09
Setup embrace-android-core & embrace-android-features modules

fractalwrench:
## Goal

Adds new modules named `embrace-android-core` and `embrace-android-features`. These modules will eventually contain Embrace's implementation details that are required to capture OTel data & send it to a backend, and Embrace's built-in instrumentation features respectively.

`embrace-android-sdk` adds a `implementation` dependency on both, therefore this should effectively have zero breaking change compared to the approach in #1055.

I have not migrated any code over to these modules yet.

### DIFF
--- a/embrace-android-core/README.md
+++ b/embrace-android-core/README.md
@@ -1,0 +1,4 @@
+# embrace-android-core
+
+Contains implementation details that Embrace does not consider to be part of its public API.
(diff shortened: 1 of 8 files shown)


bidetofevil (approved):
LGTM

PR #1071, opened 2024-07-09
fix small typo in README setup step

jpmunz:
Small typo in the README file that tripped me up when I was setting this up, brings it inline with the code snippet from https://embrace.io/docs/android/integration/session-reporting/#add-the-start-call


bidetofevil (approved):
LGTM

PR #1081, opened 2024-07-12
suppress unwanted docs

fractalwrench:
## Goal

Suppresses unwanted symbols that are internal-only from our KDoc. `@hide` should not be used - [`@supress` should be used instead](https://kotlinlang.org/docs/kotlin-doc.html#suppress). Anything inside an `internal.**` package is automatically hidden by our Dokka setup.

This hides the vast majority of unwanted symbols. We may want to convert some Java classes as part of the major bump to have a mechanism for hiding them (and also get rid of Java code).


fractalwrench:
There isn't a way to get Dokka to understand arbitrary annotations as far as I can see unfortunately. However, given that we're moving most of the internal classes to other modules eventually this should be a much smaller problem in future.

PR #1082, opened 2024-07-12
Create interfaces for config service behaviors

fractalwrench:
## Goal

The `ConfigService` needs to be public so that it is accessible by the feature module. Its `Behavior` fields need to be public, but we don't want to expose the payload classes that are referenced by these unnecessarily. Therefore I have created interfaces for each of these behavior classes & made the relevant interfaces public.

I've marked each class with the `@InternalApi` annotation. In a future changeset I'll also move these to a package under the `internal` namespace.


bidetofevil (approved):
LGTM

PR #1083, opened 2024-07-12
Move classes into separate source files

fractalwrench:
## Goal

Moves some internal arch classes into separate source files so there's only one class per file. This should make subsequent refactoring easier to follow.


bidetofevil (approved):
LGTM

PR #1084, opened 2024-07-12
Make internal api symbols visible

fractalwrench:
## Goal

Makes classes public & marks them as an `@InternalApi` where the feature module needs to access them. In a future changeset I'll also move these to a package under the `internal` namespace.


bidetofevil (approved):
LGTM. We'll add `@suppress` to any place marked as `@InternalApi` in another PR?

PR #1085, opened 2024-07-12
Setup correct dependencies for core module

fractalwrench:
## Goal

Alters the dependency graph so code in the core module can be accessed by the feature module.


bidetofevil (approved):
LGTM

PR #1086, opened 2024-07-12
Migrate arch to internal package

fractalwrench:
## Goal

Migrates the `arch` package so that it is `internal.arch`. Eventually I'd like to do this with all non-public packages in our SDK  to make it absolutely clear what is part of our public API & what isn't.


bidetofevil (approved):
Just a move, so I didn't look at each file carefully. If there are files I should take a look at specifically, let me know, but I'm going to rubberstamp this otherwise.

PR #1087, opened 2024-07-12
Migrate config package

fractalwrench:
## Goal

Migrates the `config` package so that it is `internal.config`. Eventually I'd like to do this with all non-public packages in our SDK  to make it absolutely clear what is part of our public API & what isn't.


bidetofevil (approved):
Rubberstamping large move

PR #1090, opened 2024-07-12
Remove otel dep in core module

fractalwrench:
## Goal

Removes the OTel dependency in the core module. After discussion our aim is to keep this module free of dependencies such as OTel/Moshi & confine these to other modules.


bidetofevil (approved):
LGTM

PR #1091, opened 2024-07-12
Move behavior interfaces to core module

fractalwrench:
## Goal

Moves some additional interfaces to the `embrace-android-core` module


bidetofevil (approved):
LGTM

PR #1092, opened 2024-07-12
Remove api checks for core + feature modules

fractalwrench:
## Goal

The core/feature modules are implementation dependencies of `embrace-android-sdk` so are not exposed to customers. Therefore it doesn't make sense to run api checks on them.


bidetofevil (approved):
Cool. Is there a way to just opt out the entire module out of Dokka then?

fractalwrench:
Dokka is only applied to the `embrace-android-sdk` module right now

PR #1098, opened 2024-07-15
Migrate session package to internal

fractalwrench:
## Goal

Migrates the session package under `internal`. Eventually I'd like to do this with all non-public packages in our SDK to make it absolutely clear what is part of our public API & what isn't.


bidetofevil (approved):
LGTM

PR #1099, opened 2024-07-15
Migrate capture package

fractalwrench:
## Goal

Migrates the capture package under `internal`. Eventually I'd like to do this with all non-public packages in our SDK to make it absolutely clear what is part of our public API & what isn't.


bidetofevil (approved):
Fantastic! It'd be great if we can validate in test that there aren't any `internal` packages in our generated docs, in the future.

PR #1100, opened 2024-07-15
Move injection files

fractalwrench:
## Goal

Migrates the injection package under `internal`. As part of this I moved each class to a separate source file.


bidetofevil (approved):
Rubberstamping code move PR

PR #1103, opened 2024-07-15
Migrate payload files

fractalwrench:
## Goal

Migrates the remaining payload files to the `internal` package. This completes moving everything that is not part of our public API surface into this package, which should make it easier to know what needs modularising.


bidetofevil (approved):
Rubberstamping code move PR

PR #1104, opened 2024-07-15
Move arch types to core module

fractalwrench:
## Goal

Moves some initial architecture types to the core module. These mostly didn't require any refactor apart from moving a couple of extension functions here & there.


bidetofevil (approved):
LGTM

PR #1106, opened 2024-07-16
Move arch classes to core module

fractalwrench:
## Goal

Moves some additional architecture classes to the `embrace-android-core` module. This should be most of the classes required in order to register a `DataSourceModule`.

## Testing


buildSrc/src/main/kotlin/embrace-defaults.gradle.kts
```diff
@@ -85,6 +85,7 @@ android {
 }
 
 dependencies {
+    implementation("androidx.startup:startup-runtime:1.1.1")
```
bidetofevil, line 88: Is there a way we can get versions in this into a toml? Having that and `Versions.kt` is kind of annoying...

bidetofevil (approved):
LGTM. One comment about version definition but more like a rant rather than a useful comment

buildSrc/src/main/kotlin/embrace-defaults.gradle.kts
fractalwrench: Unfortunately not that I'm aware of. This is a deficiency of gradle - this script is effectively precompiled at the same time as the version catalog, so it's not possible to use the version catalog as there would be a circular dependency. Hopefully this oversight is fixed in newer versions of Gradle.

PR #1107, opened 2024-07-16
Move additional arch files to core module

fractalwrench:
## Goal

Moves some additional architecture files to the core module which allow capturing log/event data types. This required some minor refactoring to get around the public API `Severity` enum not being accessible from that module.

## Testing


bidetofevil (approved):
LGTM

PR #1108, opened 2024-07-16
Migrate memory warning data source to features module

fractalwrench:
## Goal

Migrates the memory warning data source to the features module. As part of this I consolidated the legacy memory services so that they are implemented as part of the data source instead.

## Testing

Updated feature tests to include a limit test & removed obsolete unit tests. Note that the feature test discovered the global span event limit of 10 means no more than 10 memory warnings can be recorded. This seems a little low but is global behavior. Also verified in the dashboard that memory warnings are still shown.


bidetofevil (approved):
Migration looks good, but we really shouldn't have a limit to this. The reason it behaves this way is because we are not bypassing typical span validation in the data source, which we should. We don't have a consistent story for this but we really should. I'll enter a task to track this and we can remedy this soon. For now, I wouldn't bother with the test, but I'm OK with leaving it in for now and ripping it out later. Up to yuou.

PR #1109, opened 2024-07-16
Enable build cache

fractalwrench:
## Goal

Enables the [gradle build cache](https://docs.gradle.org/current/userguide/build_cache.html), which promises faster builds. We can see what effect this has on performance when developing locally.

Right now enabling configuration cache won't do anything for us because we use kapt, but once we migrate Kotlin there should be some local perf benefits from enabling that.

## Testing

Relying on CI.


bidetofevil (approved):
This could help a bunch now that we are getting more modularized. If things don't work, clear your cache :-)

PR #1110, opened 2024-07-16
Alter CI cache keys

fractalwrench:
## Goal

Alters the keys used for calculating a cache on CI as it didn't account for all the things in the build script that could potentially alter the build.


bidetofevil (approved):
LGTM

PR #1111, opened 2024-07-16
Use latest action for gradle wrapper validation

fractalwrench:
## Goal

Use a non-deprecated action for validating our gradle wrapper: https://github.com/gradle/wrapper-validation-action/releases


bidetofevil (approved):
LGTM

PR #1112, opened 2024-07-16
Optimize tasks run on CI

fractalwrench:
## Goal

Attempts to optimize the tasks that are run on CI by being more specific rather than running for every build type, and avoiding a clean.


bidetofevil (approved):
LGTM

PR #1114, opened 2024-07-16
Modularise payload classes

fractalwrench:
## Goal

Modularises some of the payload classes into the payload module.

## Testing


bidetofevil (approved):
LGTM

PR #1117, opened 2024-07-17
Move span arch classes over

fractalwrench:
## Goal

Starts migrating classes related to capturing spans to the core module.


bidetofevil (approved):
LGTM

PR #1118, opened 2024-07-17
Replace kapt with ksp

fractalwrench:
## Goal

Replace usage of kapt with ksp now that our Kotlin version has been updated.

## Testing

Relied on existing test coverage.

PR #1119, opened 2024-07-17
Enable configuration cache

fractalwrench:
## Goal

Enables Gradle's [configuration cache](https://docs.gradle.org/current/userguide/configuration_cache.html#config_cache) now that #1118 addresses the last remaining problems. This should allow substantial speed improvements when developing locally.


bidetofevil (approved):
LGTM

PR #1126, opened 2024-07-18
Add linker flag to make library compatible with 16 kb page sizes.

priettt:
Adding this allows our library to work on devices with a 16kb memory page size. 

If we don't do this, even if the customer has ndk_enabled set to false, the app will crash with a SIGSEGV when we try to load the native library.

This was done following the steps here: https://developer.android.com/guide/practices/page-sizes

I verified the correct alignment of the test apk after the changes using the alignment script here: https://developer.android.com/guide/practices/page-sizes#test


fractalwrench (approved):
LGTM

bidetofevil (approved):
LGTM. 

This doesn't change the NDK version customers an use right? The deps are all built into individual .so files and there's no dynamic linking at runtime or cross .so calls?

PR #1127, opened 2024-07-19
Run subset of swazzler tests on every pull request

cesarmax22:
## Goal

Run a subset of swazzler tests on every PR. This is to ensure that we don't break anything at compilation time.

## Testing

Manually tested.


fractalwrench (approved):
LGTM

PR #1130, opened 2024-07-19
Modularise span data source

fractalwrench:
## Goal

Moves `SpanService`, `SpanDataSource` and related classes to the core module. This will be required for use by any data source that manipulates spans.


priettt (approved):
👍

PR #1136, opened 2024-07-22
Fix cpuinfo.c JNI call to address the correct package

priettt:
## Goal

EmbraceCpuInfoDelegate is in a new package now, so I updated the JNI method to address it correctly.


fractalwrench (approved):
LGTM

PR #1137, opened 2024-07-24
Allow null for traceparent generator

fractalwrench:
## Goal

Allows `null` as a valid value when generating a traceparent. This is because the feature may not be enabled in which case we want to avoid generating a traceparent.

## Testing

Updated unit test coverage.


priettt (approved):

bidetofevil (approved):
LGTM

PR #1139, opened 2024-07-25
Migrate to OTel semantic convention keys

priettt:
- Altered the keys used in the payload to match OTel’s semantic conventions.
- Deleted unused SchemaType objects.

Backend already supports the new keys.


bidetofevil (approved):
LGTM

PR #1140, opened 2024-07-26
Remove apk tools config

fractalwrench:
## Goal

Removes the unused `ApkToolsConfig` class. In practice we are not using this so we should get rid of the complexity.


bidetofevil (approved):
OH FINALLY!

PR #1141, opened 2024-07-26
Move init module to core

fractalwrench:
## Goal

Moves the `InitModule` class and implementation to `embrace-android-core`. This will ultimately allow it to be accessed from `embrace-android-features` so features can be instantiated declaratively. I needed to move a few extension functions around to achieve this, and also:

- Moved `openTelemetryClock` from `InitModule` to `OpenTelemetryModule`
- Refactored serialization so that `EmbraceUrl` doesn't require an adapter, as the alternative is dragging a bunch of unrelated HTTPUrlConnection code over to the core module
- Tweaked serialization function in NDK code

## Testing


bidetofevil (approved):
Very nice! I don't recall why the OTel clock has to be that low, but if you can move it to later and it still compiles, then awesome!

PR #1143, opened 2024-07-26
Move declaration of data sources to feature module

fractalwrench:
## Goal

Moves the declaration of 2 data sources so that their instantiation is dealt with in the feature module.

In future changesets we'll leverage the app startup library to automatically add these `DataSourceState` references when initializing the SDK, so that the main SDK module/core module don't need to know anything about the features.

## Testing


bidetofevil (approved):
Nice. LGTM

PR #1145, opened 2024-07-29
Modularise data source classes

fractalwrench:
## Goal

Begins moving the data source classes to the `embrace-android-features` module.

## Testing

Relied on existing integration tests.


bidetofevil (approved):
LGTM

PR #1146, opened 2024-07-29
Move worker module

fractalwrench:
## Goal

Moves the worker module & implementation to `embrace-android-core`.


bidetofevil (approved):
LGTM

PR #1148, opened 2024-07-29
Extract further features

fractalwrench:
## Goal

Moves 2 additional features to the `embrace-android-features` module.


bidetofevil (approved):
LGTM

PR #1149, opened 2024-07-29
Modularise further data sources

fractalwrench:
## Goal

Moves 3 additional features to the `embrace-android-features` module.


bidetofevil (approved):
LGTM

PR #1150, opened 2024-07-29
Modularise payload classes

fractalwrench:
## Goal

Moves the remaining payload classes to the `embrace-android-payload` module. This required extracting a few extension functions.

## Testing


bidetofevil (approved):
Didn't look too carefully as this is just a move, but rubberstamped

PR #1152, opened 2024-07-29
Create test module that holds fake implementations

fractalwrench:
## Goal

Creates a new module that contains fake implementations used in our test code. Access to fakes is becoming a bit of an obstacle in migrating code over as unit/integration tests typically use fake implementations of a real interface. Currently these are all in `embrace-android-sdk` under the `unitTest` sourceSet, so they become inaccessible if we wish to move unit tests anywhere else.

The solution I've chosen is to create a new module that implements the fakes & added it as a test dependency. I've also taken the opportunity to standardise the quality checks for test modules, as we also have `embrace-test-common` which holds _no_ references to other embrace code & is intended for pure util code.


bidetofevil (approved):
LGTM

PR #1154, opened 2024-07-30
Modularise api interfaces

fractalwrench:
## Goal

Moves some additional interfaces to the `embrace-android-api` module.


embrace-android-api/src/main/kotlin/io/embrace/android/embracesdk/LogExceptionType.kt
```diff
@@ -10,7 +10,7 @@ import io.embrace.android.embracesdk.annotation.InternalApi
  * @suppress
  */
 @InternalApi
-public enum class LogExceptionType(internal val value: String) {
+public enum class LogExceptionType(public val value: String) {
```
bidetofevil, line 13: Doesn't the parameter's visibility just inherent from the class's?

bidetofevil (approved):
LGTM

embrace-android-api/src/main/kotlin/io/embrace/android/embracesdk/LogExceptionType.kt
fractalwrench: Normally, but we have explicit API mode enabled on all our prod modules which requires explicitly setting visibility.

PR #1155, opened 2024-07-30
Modularise spans package

fractalwrench:
## Goal

Moves some classes from the spans package into the `embrace-android-core` module.


bidetofevil (approved):
LGTM

PR #1157, opened 2024-07-30
Modularise envelope creation

fractalwrench:
## Goal

Modularises creation of the envelope by moving classes to `embrace-android-core`. This involved moving a few classes into separate source files which created a big diff, but there isn't any change to the classes beyond altering modifiers.


bidetofevil (approved):
LGTM

PR #1159, opened 2024-07-30
Reduce visibility into embrace-android-features

fractalwrench:
## Goal

Initial implementation for reducing the visibility of classes inside `embrace-android-features` that don't need to be visible. I've added a `registerFeatures` function on `FeatureModule` that will be invoked when the SDK initializes. This means classes such as `MemoryWarningDataSource` that aren't directly referenced anywhere can be added to the orchestration layer indirectly, allowing us to set the visiblity as internal.

There are a few more roadblocks before other classes can be made non-public, mostly around test fakes being in the wrong module.

## Testing

Relied on integration test.


bidetofevil (approved):
LGTM

bidetofevil (approved):
Nice. LGTM

PR #1160, opened 2024-07-30
Modularise ANR code

fractalwrench:
## Goal

Modularises ANR code by moving it to `embrace-android-features`. There are a few minor edits (deleting an `enforceThread` function that is only present in debug builds, altering visibility, and moving classes to separate source files), but otherwise this a straightforward move over.

## Testing


bidetofevil (approved):
LGTM

PR #1161, opened 2024-07-30
Modularise crash implementations

fractalwrench:
## Goal

Moves the crash implementation to `embrace-android-features`.


bidetofevil (approved):
LGTM

PR #1164, opened 2024-07-30
Remove unnecessary MetadataService functions

fractalwrench:
## Goal

Removes unnecessary functions on the `MetadataService` that are just called from test code.


bidetofevil (approved):
Great!

PR #1165, opened 2024-07-30
Passing a time of zero to startSpan will now use openTelemetryClock.now()

priettt:
This was done to match how the OTel library works. Also, it fixes an inconsistency between span snapshots and actual spans.


bidetofevil (approved):
LGTM

PR #1168, opened 2024-07-31
Move session id tracking

fractalwrench:
## Goal

Refactors session ID tracking so that a direct reference on the NDK service isn't required, which will aid modularisation by breaking a dependency.

## Testing

Updated existing unit tests.


bidetofevil (approved):
LGTM

PR #1169, opened 2024-07-31
Modularise metadata service

fractalwrench:
## Goal

Moves `MetadataService` to `embrace-android-core` and removes its responsibility for giving the appId/appState, which are both handled by other services.

## Testing


bidetofevil (approved):
LGTM

PR #1170, opened 2024-07-31
Modularise session prop implementation

fractalwrench:
## Goal

Modularises session properties into `embrace-android-features` and avoids directly referencing `NdkService` from the class.


bidetofevil (approved):
LGTM

PR #1171, opened 2024-07-31
Modularises some more OTel symbols

fractalwrench:
## Goal

Moves some additional OTel symbols to the core module.


bidetofevil (approved):
LGTM. Now I can go rename some of these :-)

PR #1174, opened 2024-08-01
Span snapshots are now sent with a null end time

priettt:
Span snapshots denote a span that was in process when the session was sent. It makes more sense for the end time to be null, so we don't cause any issues in the backend


fractalwrench (approved):
LGTM

PR #1179, opened 2024-08-02
Rearrange source files

fractalwrench:
## Goal

Separates classes into their own source file where possible.


bidetofevil (approved):
LGTM

PR #1181, opened 2024-08-02
Modularise envelope source classes

fractalwrench:
## Goal

Modularises some of the envelope construction classes. This required refactoring & hiding some services behind an interface.


embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/capture/envelope/session/SessionPayloadSourceImpl.kt
```diff
@@ -62,17 +50,13 @@ internal class SessionPayloadSourceImpl(
...
                     )
-                    if (appTerminationCause == null) {
-                        sessionPropertiesServiceProvider().populateCurrentSession()
-                    }
                     spans.map(EmbraceSpanData::toNewPayload)
                 }
 
                 else -> spanSink.completedSpans().map(EmbraceSpanData::toNewPayload)
             }
             // add ANR spans if the payload is capturing spans.
```
bidetofevil, line 58: ANR spans are added else where now?

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/internal/capture/envelope/session/OtelPayloadMapperImpl.kt
```diff
@@ -0,0 +1,34 @@
...
+
+    override fun getSessionPayload(endType: SessionSnapshotType, crashId: String?): List<Span> {
+        val cacheAttempt = endType == SessionSnapshotType.PERIODIC_CACHE
+        val appTerminationCause = when {
+            crashId != null -> AppTerminationCause.Crash
+            else -> null
+        }
+        if (!cacheAttempt) {
+            if (appTerminationCause == null) {
+                sessionPropertiesServiceProvider().populateCurrentSession()
```
bidetofevil, line 28: We may want to do this even if we're crashing. I'll have to take a look at which properties we are actually including - I think previously it was diagnostic attributes, but if we are putting our literally session properties there, we should always do that.

Don't worry about this for this PR. I'll take care of it

bidetofevil (approved):
A comment about a comment, and another that is just an FYI. Good to go!

PR #1183, opened 2024-08-02
Migrate tests

fractalwrench:
## Goal

Migrates the breadcrumb test to the features module. This necessitates altering some of the build dependencies & moving fakes around. This is an intermediate step to making some data sources + implementation details internally visible.


bidetofevil (approved):
LGTM

PR #1188, opened 2024-08-05
Move test code to appropriate module

fractalwrench:
## Goal

Moves some of the test fakes + test cases to the appropriate module.


bidetofevil (approved):
Just a move, so rubberstamp it is.

PR #1189, opened 2024-08-05
Modularise config behaviors

fractalwrench:
## Goal

Moves some config classes to the `embrace-android-core` module.


bidetofevil (approved):
LGTM

PR #1190, opened 2024-08-05
Delete serialization tests

fractalwrench:
## Goal

Deletes some tests that were asserting serialization/deserialization. These are fairly tedious to maintain & are effectively just testing Kotlin's type system + Moshi's code generator, which we can assume work as expected. Therefore I think there's a benefit to deleting them.


bidetofevil (approved):
Cool. The important ones should already be tested by the integration tests, so we have some implicit coverage already.

PR #1193, opened 2024-08-06
Added sensitive_keys_denylist to the LocalConfig

priettt:
I will use those keys to redact properties in a future PR

### DIFF
--- a/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/internal/config/behavior/SensitiveKeysBehaviorImpl.kt
+++ b/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/internal/config/behavior/SensitiveKeysBehaviorImpl.kt
@@ -1,0 +1,17 @@
+package io.embrace.android.embracesdk.internal.config.behavior
+
+import io.embrace.android.embracesdk.internal.config.local.SdkLocalConfig
+
+private const val SENSITIVE_KEY_MAX_LENGTH = 128
+private const val SENSITIVE_KEYS_LIST_MAX_SIZE = 10000
+
+internal class SensitiveKeysBehaviorImpl(
+    localConfig: SdkLocalConfig
+) : SensitiveKeysBehavior {
+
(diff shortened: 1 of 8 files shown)


embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/internal/config/behavior/SensitiveKeysBehaviorImpl.kt
```diff
@@ -0,0 +1,16 @@
...
+
+import io.embrace.android.embracesdk.internal.config.local.SdkLocalConfig
+
+private const val SENSITIVE_KEY_MAX_SIZE = 128
+
+internal class SensitiveKeysBehaviorImpl(
+    private val localConfig: SdkLocalConfig
+) : SensitiveKeysBehavior {
+
+    // TODO: max list length
```
fractalwrench, line 11: `private val list = localConfig.sensitiveKeysDenylist.take(n)` would be a good option for this

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/internal/config/behavior/SensitiveKeysBehaviorImpl.kt
```diff
@@ -0,0 +1,16 @@
...
+    override fun isSensitiveKey(key: String): Boolean {
+        val keysList = localConfig.sensitiveKeysDenylist ?: return false
+        return keysList.any { it.take(SENSITIVE_KEY_MAX_SIZE).toRegex().matches(key) }
```
fractalwrench, line 14: This is potentially quite expensive as it constructs a regex for every entry in the list every time this function is called. This would be a good candidate for memoisation if we needed to support regex, but thankfully we don't according to the spec. Therefore It should be sufficient to just check whether the key equals any object in the list.

priettt: we don't need to support regex? that's great 😅

PR #1195, opened 2024-08-06
Move payload tests

fractalwrench:
## Goal

Moves some tests to the payload module & reorganises some of the test package structure inside the SDK module


bidetofevil (approved):
LGTM

PR #1197, opened 2024-08-06
Begin modularising injection classes

fractalwrench:
## Goal

Begins modularising the classes used for injection. This PR moves the `AnrModule` to `embrace-android-features` and takes the approach of an `AnrModule` interface that defines the type`, an `AnrModuleSupplier` that defines the constructor & exposes a function to construct an object, and `AnrModuleImpl`, which is hidden by the module boundary & contains all the concrete details. Future changesets will roll this approach out to other modules.


bidetofevil (approved):
Very nice!

PR #1198, opened 2024-08-06
Extract additional modules & create suppliers

fractalwrench:
## Goal

Builds on the approach detailed in #1197 by extracting suppliers so that it will be possible to hide the implementation class, and moving a few dependency injection classes to their new modules.


bidetofevil (approved):
LGTM

PR #1199, opened 2024-08-06
Extract data + feature modules

fractalwrench:
## Goal

Moves the `FeatureModule` and `DataSourceModule` out of the `embrace-android-sdk` module. As part of this it was required to refactor the `FeatureModule` so that it's solely responsible for creating `DataSourceState` objects, and reduce the responsibilities of `DataSourceModule` to just creating the registry where features are added.

## Testing


bidetofevil (approved):
Looks good but one general comment:

Given that we are relying a lot more on runtime suppliers, do we have any centralized internal error and logging to verify that whenever we try to invoke a supplier before it's ready, we get notified? Or is this a problem that we would discover without needing production telemetry?

fractalwrench:
There isn't any centralised telemetry that would show us this. I think it'd make sense to add an internal error whenever the SDK fails to initialize as that'll cover the vast majority of cases, I'll do that in another PR

PR #1203, opened 2024-08-06
Increase span attribute limit at the OTel layer

bidetofevil:
## Goal

OTel imposes a limit on how many attributes are allowed on a span, which is 128 by default. For the session span, we require more than that because you can theoretically have 100 session properties.

To support this, we bump the total limit set in the OTel layer to 300, and validate that our custom limit of 50 for SDK users is not enforced when the SDK adds attributes via the internal `addSystemAttribute()` method. This will also bump the total number of other attributes the SDK can add to the session span to a safer ceiling.

## Testing

Added and modified appropriate unit tests to ensure the expected maximums can be written.

### DIFF
--- a/embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/arch/destination/SessionSpanWriter.kt
+++ b/embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/arch/destination/SessionSpanWriter.kt
@@ -21,15 +21,11 @@ public interface SessionSpanWriter {
 
     /**
      * Add the given key-value pair as an Attribute to the session span
-     *
-     * Returns true if the attribute was added, otherwise false.
      */
-    public fun addCustomAttribute(attribute: SpanAttributeData): Boolean
+    public fun addSystemAttribute(attribute: SpanAttributeData)
 
     /**
      * Remove the attribute with the given key
-     *
-     * Returns true if attribute was removed, otherwise false.
      */
-    public fun removeCustomAttribute(key: String): Boolean
+    public fun removeSystemAttribute(key: String)
 }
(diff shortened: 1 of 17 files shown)


bidetofevil:
* **#1203** <a href="https://app.graphite.dev/github/pr/embrace-io/embrace-android-sdk/1203?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `master`

<h2></h2>

embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/arch/destination/SessionSpanWriter.kt
```diff
@@ -21,15 +21,11 @@ public interface SessionSpanWriter {
 
     /**
      * Add the given key-value pair as an Attribute to the session span
-     *
-     * Returns true if the attribute was added, otherwise false.
      */
-    public fun addCustomAttribute(attribute: SpanAttributeData): Boolean
+    public fun addSystemAttribute(attribute: SpanAttributeData)
```
bidetofevil, line 25: This always succeeds now because we are adding to a map without additional validation.

embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/spans/EmbraceSpanImpl.kt
```diff
@@ -42,8 +42,8 @@ public class EmbraceSpanImpl(
     private var updatedName: String? = null
     private val systemEvents = ConcurrentLinkedQueue<EmbraceSpanEvent>()
     private val customEvents = ConcurrentLinkedQueue<EmbraceSpanEvent>()
-    private val systemAttributes = ConcurrentHashMap<AttributeKey<String>, String>().apply {
-        putAll(spanBuilder.getFixedAttributes().associate { it.key.attributeKey to it.value })
+    private val systemAttributes = ConcurrentHashMap<String, String>().apply {
+        putAll(spanBuilder.getFixedAttributes().associate { it.key.attributeKey.key to it.value })
```
bidetofevil, line 46: Some system attributes now have arbitrary string keys (i.e. session properties), so we make the key a string and reference the string inside the `AttributeKey` instead of requiring that

fractalwrench (approved):
LGTM

bidetofevil:
### Merge activity

* **Aug 7, 2:10 PM EDT**: @bidetofevil started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/embrace-io/embrace-android-sdk/1203).

PR #1205, opened 2024-08-07
Track internal error if SDK fails to start

fractalwrench:
## Goal

Tracks an internal error if the SDK fails to start. This should catch any problems with the `ModuleInitBootstrapper`.


bidetofevil (approved):
LGTM. We should probably review our logging to make sure we are indicating the right things. The instrumentation is definitely disabled, no "mays" about it

PR #1206, opened 2024-08-07
Add test for clearing session props

fractalwrench:
## Goal

Adds a test for clearing session properties.


bidetofevil (approved):
This takes care of the clearing, but I think we still need change where the initial population happens so we take care of the background activity disabled case.

We can merge this and deal with that in a separate PR.

embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/capture/session/SessionPropertiesServiceImpl.kt
```diff
@@ -35,19 +40,30 @@ public class EmbraceSessionPropertiesService(
...
+    override fun getProperties(): Map<String, String> = props.get()
+
+    override fun populateCurrentSession(): Boolean {
+        return dataSourceProvider()?.addProperties(getProperties()) ?: false
+    }
 
-    override fun populateCurrentSession(): Boolean = dataSourceProvider()?.addProperties(getProperties()) ?: false
+    override fun clearTemporary() {
+        val removedProps = props.clearTemporary()
+        removedProps.forEach { (key, _) ->
```
bidetofevil, line 59: I don't think this is necessary anymore. Now that we've moved the population to after the temp clearing in the service happens, the new session span have the temporary props that were there before, so clearing the service will be sufficient

PR #1207, opened 2024-08-07
Decouple crash data source from other services

fractalwrench:
## Goal

Decouples the `CrashDataSource` from several unrelated services by creating a listener that is invoked when a crash happens.


embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/internal/injection/ModuleInitBootstrapper.kt
```diff
@@ -419,15 +419,17 @@ internal class ModuleInitBootstrapper(
...
-                            androidServicesModule,
-                            customerLogModule,
+                            androidServicesModule
                         )
                     }
 
                     postInit(CrashModule::class) {
-                        serviceRegistry.registerService(crashModule.crashService)
+                        serviceRegistry.registerService(crashModule.crashDataSource)
+                        with(crashModule.crashDataSource) {
```
bidetofevil, line 428: Is there a way to guarantee execution order here? I feel like what was there may need to be invoked in order for things to work properly, and it'd be good if we had more than call order to maintain that order.

If it's too complicate to put in a priority system of some sort, we should at least add some comments?

bidetofevil (approved):
One comment about handler invocation ordering, but looks good otherwise.

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/internal/injection/ModuleInitBootstrapper.kt
fractalwrench: I can't see any reason why invocation order would matter but have restored it back to the original order (apart from the crash file marker, which now comes first). I agree that deterministic order is important so updated the implementation to use a list instead of a set

PR #1208, opened 2024-08-07
Decouple ndk service

fractalwrench:
## Goal

Decouples the NDK service from needing to know directly about the user service by adding a listener.


bidetofevil (approved):
Very nice. This may improve SDK startup time by itself because this was one of the things that dragged user service init earlier than it strictly needed to be.

PR #1209, opened 2024-08-07
Rename customer log module + reorg crash module

fractalwrench:
## Goal

Renames the customer log module more appropriately & moves one field from the `CrashModule` to `NativeModule`


embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/internal/api/delegate/LogsApiDelegate.kt
```diff
@@ -19,7 +19,7 @@ internal class LogsApiDelegate(
 ) : LogsApi {
 
     private val logger = bootstrapper.initModule.logger
-    private val logService by embraceImplInject(sdkCallChecker) { bootstrapper.customerLogModule.logService }
+    private val logService by embraceImplInject(sdkCallChecker) { bootstrapper.logModule.logService }
```
bidetofevil, line 22: Something to consider: this is essentially the Embrace logging service, where we use the OTel logger to create embrace logs. When we go full on supporting the OTel logger interface, we may want to rename this to be more specific, and actually create a LogService analogous to SpanService at a higher level that can be used by all log-creating services like this one. But that'll be for later.

bidetofevil (approved):
LGTM. One aside that doesn't affect this PR which is good to go.

PR #1210, opened 2024-08-07
Sensitive keys denylist

priettt:
## Goal

If a `sensitive_keys_denylist` is provided in `embrace-config.json`, the following components will be scanned:

- [x] Moments
- [x] Session properties
- [x] Log attrs + log event attributes
- [x] Span attrs + span event attributes
- [x] Event attributes

If the key in any of those properties is present in the denylist, the value will be changed to`<redacted>`.

### DIFF
--- a/embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/injection/OpenTelemetryModuleImpl.kt
+++ b/embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/injection/OpenTelemetryModuleImpl.kt
@@ -2,8 +2,7 @@ package io.embrace.android.embracesdk.internal.injection
 
 import io.embrace.android.embracesdk.internal.OpenTelemetryClock
 import io.embrace.android.embracesdk.internal.Systrace
-import io.embrace.android.embracesdk.internal.config.behavior.SensitiveKeysBehaviorImpl
-import io.embrace.android.embracesdk.internal.config.local.SdkLocalConfig
+import io.embrace.android.embracesdk.internal.config.behavior.SensitiveKeysBehavior
 import io.embrace.android.embracesdk.internal.logs.LogSink
 import io.embrace.android.embracesdk.internal.logs.LogSinkImpl
 import io.embrace.android.embracesdk.internal.opentelemetry.EmbOpenTelemetry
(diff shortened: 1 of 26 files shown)


embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/injection/OpenTelemetryModuleImpl.kt
```diff
@@ -69,11 +70,19 @@ internal class OpenTelemetryModuleImpl(
         openTelemetrySdk.sdkTracer
     }
 
+    private var sensitiveKeysBehavior: SensitiveKeysBehavior? = null
+
+    override fun setupSensitiveKeysBehavior(sensitiveKeysBehavior: SensitiveKeysBehavior) {
+        this.sensitiveKeysBehavior = sensitiveKeysBehavior
+        embraceSpanFactory.setupSensitiveKeysBehavior(sensitiveKeysBehavior)
+    }
```
fractalwrench, line 78: I'm not convinced this is the right pattern & think it might be better to use a `() -> SensitiveKeysBehavior` parameter on the constructor of `OpenTelemetryModuleImpl` instead to get around any circular dependency. What was the reason `setupSensitiveKeysBehavior` was needed in the first place?

embrace-android-core/src/test/java/io/embrace/android/embracesdk/internal/spans/SpanServiceImplTest.kt
```diff
@@ -53,6 +55,11 @@ internal class SpanServiceImplTest {
                 tracer = initModule.openTelemetryModule.sdkTracer,
                 openTelemetryClock = initModule.openTelemetryModule.openTelemetryClock,
                 spanRepository = initModule.openTelemetryModule.spanRepository,
+                sensitiveKeysBehavior = SensitiveKeysBehaviorImpl(
+                    SdkLocalConfig(
+                        sensitiveKeysDenylist = listOf("password")
+                    )
+                )
             )
```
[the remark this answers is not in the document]
priettt: good catch! this change doesn't really make sense, I ended up testing it from integration tests

embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/injection/OpenTelemetryModuleImpl.kt
priettt: Summing up our chat today here so Hanson can see it too: 

- setupSensitiveKeysBehavior is needed because OpenTelemetryModuleImpl is created before the LocalConfig.
- We can't pass a function that provides SensitiveKeysBehavior because the local config isn't initialized yet when the OpenTelemetryModuleImpl is constructed.
- This approach is good enough for now, but we should examine how the OpenTelemetryModule is initialized and see if it's worth deconstructing it into different parts.

bidetofevil: We really need to find a way for `LocalConfig` to be created before the initialization of the OTel SDK. It would not only help this, but also we need to properly populate the Resource attributes with the name and version of the app using the SDK, which we aren't do now.

With the `CodeGenConfig` work that @fractalwrench is playing with, we'll be able to do this, right?

embrace-android-core/src/test/java/io/embrace/android/embracesdk/internal/spans/SpanServiceImplTest.kt
```diff
@@ -52,7 +52,7 @@ internal class SpanServiceImplTest {
             embraceSpanFactory = EmbraceSpanFactoryImpl(
                 tracer = initModule.openTelemetryModule.sdkTracer,
                 openTelemetryClock = initModule.openTelemetryModule.openTelemetryClock,
-                spanRepository = initModule.openTelemetryModule.spanRepository,
```
bidetofevil, line 55: Trailing commas are recommended, so you don't have to remove this next time

bidetofevil (approved):
LGTM. Lets try to remove the setup method by getting `LocalConfig` initialized before everything.

embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/injection/OpenTelemetryModuleImpl.kt
fractalwrench: Yep - assuming we choose to go down that route.

PR #1211, opened 2024-08-07
Extract interface for SessionSpanAttrPopulator

fractalwrench:
## Goal

Extracts an interface for the `SessionSpanAttrPopulator` class. This is an initial refactor that lays the groundwork for simplifying our session property implementation.


bidetofevil (approved):
LGTM

PR #1212, opened 2024-08-07
Add integration tests to check JNI interface - CPU Info

priettt:
The JNI interface relies on signatures that might break if a class is moved to another package or renamed. This is only detected in runtime, as compilation goes fine and our tests don't verify this behavior. 

The goal of this PR is to create tests that, at the very least, break when a class targeted by a JNI signature is moved or renamed. If the class is not too complex, the tests will also verify additional functionality.

Classes to test:

- [x] EmbraceCpuInfoNdkDelegate
- [ ] SigquitDataSource
- [ ] NdkDelegateImpl
- [x] NativeThreadSamplerNdkDelegate


bidetofevil (approved):
LGTM. 

Right now, we are only verifying that the interfaces exists and there are no crashes when invoked, correct?

priettt:
> LGTM.
> 
> Right now, we are only verifying that the interfaces exists and there are no crashes when invoked, correct?

@bidetofevil yes, but for CPUInfo I'll also get the actual CPU getting returned, as that shouldn't be hard to get

PR #1215, opened 2024-08-07
Add JNI interface tests for NdkDelegate

priettt:
The JNI interface relies on signatures that might break if a class is moved to another package or renamed. This is only detected in runtime, as compilation goes fine and our tests don't verify this behavior.

The goal of this PR is to create tests that, at the very least, break when a class targeted by a JNI signature is moved or renamed. If the class is not too complex, the tests will also verify additional functionality.

Classes to test:

- [x]  EmbraceCpuInfoNdkDelegate
- [x]  SigquitDataSource
- [x]  NdkDelegateImpl
- [x]  NativeThreadSamplerNdkDelegate


bidetofevil (approved):
LGTM. Same comments as in previous PR about whether it makes sense to consolidate verification in one test.

PR #1218, opened 2024-08-08
Use Android Test Orchestrator to run tests in individual Instrumentation instances

priettt:
## Goal

Some of the JNI interface tests were affecting other integration tests. We needed a way to run every component in isolation so it wouldn't affect the outcome of other tests. Android Test Orchestrator allows that to happen.


bidetofevil (approved):
LGTM

fractalwrench (commented):
Why exactly is `pm clear` necessary? When using `pm clear` with Appium in the past it slowed down test execution notably, is there potential for this to happen here?

PR #1220, opened 2024-08-08
Update proguard rules with new delegates location 

priettt:
Some proguard rules were pointing to a wrong location. We should also try to add some swazzler-tests or integration tests to avoid outdated proguard rules.


fractalwrench (approved):
LGTM - nice catch!

PR #1221, opened 2024-08-09
Use session properties service directly

fractalwrench:
## Goal

Uses the `SessionPropertiesService` directly throughout the codebase rather than relying on a mix of this & the `EmbraceSessionProperties` object. This is the first step in a refactor of the session properties implementation that will address some bugs.

I've also deprecated `getSessionProperties` as iOS have removed this function & ultimately getting rid of this API will allow us to simplify our implementation too.

## Testing


bidetofevil (approved):
LGTM

bidetofevil (commented):
I think my change to fix the populate will eliminate a bunch of code that this modifies. I'll important your new test and make sure this case is covered.

PR #1222, opened 2024-08-09
Rollback app startup library changes

fractalwrench:
## Goal

Avoids shipping the app startup library changes as part of 6.10.0 so we can consider how to support the span/log exporter use-case & how to behave when users forget to supply `app_framework` in their config.


bidetofevil (approved):
LGTM

PR #1223, opened 2024-08-09
Populate session properties in session span during transition

bidetofevil:
## Goal

We were populating session properties in a session span as a side effect of constructing the end payload of the previous session. This caused a couple of problems: session-only properties not being cleared before the population, and if background activity is disabled, session properties population fails.

Changing where we do the population fixes both of these, as well as allow us to remove the extra population call during SDK startup. 

This change also supersedes #1206 as literal removal is no longer necessary - session only properties will be added or removed from the current session's session span by the `SessionPropertiesService` 

## Testing

Add integration tests for all cases, as well as the one added in #1206 to verify that session-only props don't bleed into the next session. Unit tests were also modified accordingly.

### DIFF
--- a/embrace-android-core/src/test/java/io/embrace/android/embracesdk/internal/capture/session/SessionPropertiesServiceImplTest.kt
+++ b/embrace-android-core/src/test/java/io/embrace/android/embracesdk/internal/capture/session/SessionPropertiesServiceImplTest.kt
@@ -60,8 +60,9 @@ internal class SessionPropertiesServiceImplTest {
             ::dataSource
         )
         assertEquals(0, fakeCurrentSessionSpan.attributeCount())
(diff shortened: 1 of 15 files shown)


bidetofevil:
* **#1223** <a href="https://app.graphite.dev/github/pr/embrace-io/embrace-android-sdk/1223?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `master`

<h2></h2>

priettt (approved):

bidetofevil:
### Merge activity

* **Aug 9, 6:52 PM EDT**: @bidetofevil started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/embrace-io/embrace-android-sdk/1223).

PR #1224, opened 2024-08-09
Return an error when configure_reporting fails for Sigquit capture

priettt:
When we called FindClass on SigquitDataSource and it was not found, we were returning 0 as the result of installGoogleAnrHandler. This means we would've logged that Google ANR was correctly installed, but it wouldn't have worked.

This PR fixes that, and also adds a comment in some interface tests that are also testing classes targeted by JNI FindClass


bidetofevil (approved):
Nice. LGTM

fractalwrench (approved):

PR #1228, opened 2024-08-09
Get the bundle id from resource file

TheMutsi:
## Goal

Stop using buildId to identify sourcemap. In order to do that, I had to get the bundle id from the resource file


embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/BuildInfo.kt
```diff
@@ -19,13 +19,19 @@ public class BuildInfo(
...
+     * The ID of the particular js bundle, generated at compile-time.
+     */
+    public val bundleId: String?,
 ) {
 
     public companion object {
         public const val BUILD_INFO_BUILD_ID: String = "emb_build_id"
         public const val BUILD_INFO_BUILD_TYPE: String = "emb_build_type"
         public const val BUILD_INFO_BUILD_FLAVOR: String = "emb_build_flavor"
+        public const val BUILD_INFO_BUNDLE_ID: String = "emb_bundle_id"
```
bidetofevil, line 34: We should make clear what this is - if it's the RN JS bundle ID, we should called it `emb_js_bundle_id` or something that makes it clear it is not the Android build bundle's ID. Ideally the swazzler attribute should be changed too, but we should change at least the constant in the Android SDK.

bidetofevil (commented):
I would change line 95 in `EmbraceMetadataService` as well to reference the newly passed in bundleId. I don't know what that is doing, but whatever it's doing, it should reference the right ID. If that's not doing anything, maybe we should just remove it

TheMutsi:
> I would change line 95 in `EmbraceMetadataService` as well to reference the newly passed in bundleId. I don't know what that is doing, but whatever it's doing, it should reference the right ID. If that's not doing anything, maybe we should just remove it

As we talked I dont think that line is correct, since its setting bundle id as build id in no React Native apps (probably this file won't be executed if its a No RN app).

I will delete it

bidetofevil (approved):
LGTM

PR #1229, opened 2024-08-11
Fall back to heartbeat time when converted span snapshots

bidetofevil:
## Goal

Use the last heartbeat time of the session span if the end time was missing, like if the session itself wasn't terminated properly, like during a native crash or a user kill while the app is in background.

Previously we were assuming that there's always an end time, and we even tested for that case too, which yield and end of 0 in those cases. The backend is able to work with that in most case, as it just figures it out by looking at other timestamps. So while customers don't see this problem for the most part, like for sessions associated with native crashes, the session span is wrong. I was also seeing some 0-length sessions for background kills where the session span contains a heartbeat time, so it's better than we do this population properly, and test it properly as well.

## Testing
Modify and extend unit test on the `EmbraceDeliveryService`. Tweaking some testing infra to make this easier too.

### DIFF
--- a/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/internal/comms/delivery/EmbraceDeliveryService.kt
+++ b/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/internal/comms/delivery/EmbraceDeliveryService.kt
@@ -7,6 +7,7 @@ import io.embrace.android.embracesdk.internal.injection.SerializationAction
 import io.embrace.android.embracesdk.internal.logging.EmbLogger
 import io.embrace.android.embracesdk.internal.ndk.NativeCrashService
 import io.embrace.android.embracesdk.internal.opentelemetry.embCrashId
+import io.embrace.android.embracesdk.internal.opentelemetry.embHeartbeatTimeUnixNano
 import io.embrace.android.embracesdk.internal.payload.Attribute
 import io.embrace.android.embracesdk.internal.payload.Envelope
 import io.embrace.android.embracesdk.internal.payload.EventMessage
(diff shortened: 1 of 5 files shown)


bidetofevil:
* **#1229** <a href="https://app.graphite.dev/github/pr/embrace-io/embrace-android-sdk/1229?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `master`

<h2></h2>

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/internal/comms/delivery/EmbraceDeliveryService.kt
```diff
@@ -191,8 +194,18 @@ internal class EmbraceDeliveryService(
...
 
-    private fun getFailedSpanEndTimeMs(envelope: Envelope<SessionPayload>) =
-        envelope.getSessionSpan()?.endTimeNanos?.nanosToMillis() ?: -1
+    /**
+     * To approximate the time of any snapshot to be converted into a failed span, we look to the session span of the payload and take
+     * either the end time or the last heartbeat time, whichever exists and is later. If the session span itself is a snapshot, it will
+     * not have an end time, in which case it will fall back to the last heartbeat time. If either exists, it means we can't find a better
+     * time, so we just leave it at 0.
+     */
+    private fun getFailedSpanEndTimeMs(envelope: Envelope<SessionPayload>): Long {
```
fractalwrench, line 203: I found this a bit hard to follow, would suggest breaking it down with local variables. E.g.

```
    private fun getFailedSpanEndTimeMs(envelope: Envelope<SessionPayload>): Long {
        val sessionSpan = envelope.getSessionSpan()
        val endTime = sessionSpan?.endTimeNanos ?: 0L
        val heartbeatTime = sessionSpan?.attributes?.findAttributeValue(embHeartbeatTimeUnixNano.attributeKey.key)?.toLongOrNull() ?: 0L
        return max(endTime, heartbeatTime).nanosToMillis()
    }
```

fractalwrench (approved):
LGTM

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/internal/comms/delivery/EmbraceDeliveryService.kt
bidetofevil: yeah good idea. I'll do that!

I also want to pull this out of this class because it's not really a delivery concern, but I didn't want to do that now, or when I rejiggered the logic initially. I should do it soon....

bidetofevil:
### Merge activity

* **Aug 12, 12:20 PM EDT**: @bidetofevil merged this pull request with [Graphite](https://app.graphite.dev/github/pr/embrace-io/embrace-android-sdk/1229).

PR #1233, opened 2024-08-13
Begin simplifying MetadataService

fractalwrench:
## Goal

Replaces the `ofContext` initializer with a constructor & reduces the number of functions required to get app/device info.

## Testing


bidetofevil (approved):
LGTM

PR #1234, opened 2024-08-13
Split metadata services responsibilities

fractalwrench:
## Goal

Attempts to reduce the complexity of `EmbraceMetadataService` by extracting responsibilities out to other classes & delegating.

The `RnBundleIdTracker` tracks the React Native bundle ID state. It also has functions associated with constructing a bundle ID hash.

The `EmbraceMetadataService` has been updated to delegate to the `EnvelopeResourceSource` and `EnvelopeMetadataSource`. These are already called when sending OTel payloads & duplicated some of the data capture. I've removed the concept of a 'lightweight' capture because the delegates do not support that right now.

## Testing


bidetofevil (approved):
Very nice

PR #1235, opened 2024-08-13
Rename dependency injection modules

fractalwrench:
## Goal

Renames a few modules to better describe the properties they contain.


bidetofevil (approved):
LGTM

PR #1236, opened 2024-08-13
Move payload generation out of essential service module

fractalwrench:
## Goal

Moves payload generation out of `EssentialServiceModule` into the `PayloadSourceModule`.

## Testing


bidetofevil (approved):
LGTM

PR #1237, opened 2024-08-13
Extracts some further properties from essential service module

fractalwrench:
## Goal

Extracts some further properties from `EssentialServiceModule` to the `PayloadSourceModule`.


bidetofevil (approved):
Nice

PR #1239, opened 2024-08-13
Extract config service to separate module

fractalwrench:
## Goal

Extracts the config service to a separate module so that `EssentialServiceModule` is broken further up. This also allows moving the config fully into `embrace-android-core`.

## Testing

Updated existing unit tests.


embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/injection/ConfigModuleImpl.kt
```diff
@@ -0,0 +1,50 @@
...
+import io.embrace.android.embracesdk.internal.payload.AppFramework
+import io.embrace.android.embracesdk.internal.worker.WorkerName
+
+internal class ConfigModuleImpl(
+    initModule: InitModule,
+    coreModule: CoreModule,
+    openTelemetryModule: OpenTelemetryModule,
+    workerThreadModule: WorkerThreadModule,
+    androidServicesModule: AndroidServicesModule,
+    customAppId: String?,
```
bidetofevil, line 16: I think we should seriously think about getting rid of this in 7.0

bidetofevil (approved):
Oh this is excellent! We can probably simplify the overrides in the test harness with this. And this makes a lot more sense, to decouple the fetch from the core service, given that not every instance of the SDK will need to talk to the embrace server, so this makes so much sense.

embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/injection/ConfigModuleImpl.kt
fractalwrench: Agree

PR #1243, opened 2024-08-13
Improved EmbraceLogService

priettt:
I started adding unit tests for EmbraceLogService and ended up modifying some things 😅 I wanted to add some tests for the denylist work, so this will serve as a base for that. 

What's in this PR:
- Added tests and simplified LogCounter: we were only using the error log count, regardless of the stored log ids, so I removed those.
- Added tests and simplified EmbraceLogService: fixed trimming logic, removed unhandledExceptionsCount as it was unused.

For the trimming logic, there was an edge case where the limit was smaller than 3, which is the length of the ellipsis we added at the end. The log was lost in this case because we were calling substring with a wrong end index. 

I fixed it, and now, if the max length is smaller than 3, we add the ellipsis.

### DIFF
--- a/embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/logs/EmbraceLogService.kt
+++ b/embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/logs/EmbraceLogService.kt
@@ -11,7 +11,6 @@ import io.embrace.android.embracesdk.internal.arch.schema.SchemaType.FlutterExce
 import io.embrace.android.embracesdk.internal.arch.schema.SchemaType.Log
 import io.embrace.android.embracesdk.internal.arch.schema.TelemetryAttributes
 import io.embrace.android.embracesdk.internal.capture.session.SessionPropertiesService
-import io.embrace.android.embracesdk.internal.clock.Clock
 import io.embrace.android.embracesdk.internal.config.ConfigService
 import io.embrace.android.embracesdk.internal.config.behavior.LOG_MESSAGE_MAXIMUM_ALLOWED_LENGTH
 import io.embrace.android.embracesdk.internal.logging.EmbLogger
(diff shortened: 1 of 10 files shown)


embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/logs/EmbraceLogService.kt
```diff
@@ -286,25 +270,23 @@ public class EmbraceLogService(
...
-        return if (message.length > maxLength) {
+        if (message.length > maxLength) {
+            logger.logWarning("Truncating message of ${message.length} characters to $maxLength characters")
             val endChars = "..."
 
-            // ensure that we never end up with a negative offset when extracting substring, regardless of the config value set
-            val allowedLength = when {
-                maxLength >= endChars.length -> maxLength - endChars.length
-                else -> LOG_MESSAGE_MAXIMUM_ALLOWED_LENGTH - endChars.length
+            if (maxLength <= endChars.length) {
```
bidetofevil, line 283: This takes care of the edge case, but effectively erases the value being set. Why not just return first the X characters you can based on max length and forget about the ...?

bidetofevil (approved):
Looks good. One comment about how to handle the edge case of a really low limit, but otherwise this is good to go

embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/logs/EmbraceLogService.kt
priettt: I mainly wanted to leave the ... because it indicates that the value is being truncated. Your suggestion also makes sense though

priettt: fixed!

PR #1251, opened 2024-08-14
Reduce visibility of classes in ANR feature

fractalwrench:
## Goal

Reduces the visibility of classes in the ANR feature now that most of the modularisation has been done


bidetofevil (approved):
LGTM

PR #1260, opened 2024-08-16
Begin refactoring api code

fractalwrench:
## Goal

Extracts some of our HTTP API code to the payload/core modules.


embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/comms/api/Endpoint.kt
```diff
@@ -0,0 +1,14 @@
+package io.embrace.android.embracesdk.internal.comms.api
+
+public enum class Endpoint(
+    public val path: String,
+    public val version: String
+) {
+    EVENTS("events", "v1"),
+    LOGGING("logging", "v1"),
```
bidetofevil, line 8: Many of these can be deleted because they're no longer used?

bidetofevil (approved):
Comment about some endpoints that can be deleted but looks good otherwise.

embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/comms/api/Endpoint.kt
fractalwrench: I'll address this in a separate PR because while some of the endpoints might not be used by the SDK any more, there is a non-trivial amount of code associated with them

PR #1263, opened 2024-08-16
Modularise essential service module

fractalwrench:
## Goal

Moves the `EssentialServiceModule` to the core module. Most of this changeset is the essential refactoring the session properties implementation so that it isn't referenced in the feature module due to a circular dependency. I've removed the `SessionPropertiesDataSource` & we now write directly to a `SessionSpanWriter` from `EmbraceSessionProperties`. A future step could be to merge `EmbraceSessionProperties` & the `SessionPropertiesService` files, but I've avoided that to keep this changeset small.

## Testing


embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/capture/session/EmbraceSessionProperties.kt
```diff
@@ -82,7 +94,23 @@ internal class EmbraceSessionProperties(
 
     private fun size(): Int = permanentProperties().size + temporary.size
 
-    fun clearTemporary(): Unit = temporary.clear()
+    fun clearTemporary() {
+        synchronized(permanentPropertiesReference) {
+            temporary.clear()
+            addPermPropsToSessionSpan()
```
bidetofevil, line 100: I think we need to add the permanent session properties at the beginning, when the span is started, because the session could be killed before it ends, and we need to send the cached version over. Is there another place we can do this population that doesn't bring back to circular dependency?

fractalwrench: The `permanentProperties()` function (L29) should take care of properties being added to the very first session of a process. Unless I've misunderstood something?

bidetofevil (approved):
Since it's passing with the new integration test, LGTM

PR #1264, opened 2024-08-16
Move test code to core module

fractalwrench:
## Goal

Moves some additional unit tests to the core module.


bidetofevil (approved):
Rubberstamped

PR #1265, opened 2024-08-16
Modularise moments implementation

fractalwrench:
## Goal

Moves the moments implementation to the core module.


bidetofevil (approved):
LGTM

PR #1266, opened 2024-08-16
Move session classes to core module

fractalwrench:
## Goal

Moves some additional session classes to the core module.


bidetofevil (approved):
LGTM

PR #1267, opened 2024-08-19
Delete unused endpoint code

fractalwrench:
## Goal

Deletes some unused endpoint code as requested in another PR.


bidetofevil (approved):
Oh hell yeah. The fact that this reduced test coverage probably means we don't have enough tests for Moments 😓 

But I think it's fine. LGTM

PR #1268, opened 2024-08-19
Alter network code to avoid direct embrace reference

fractalwrench:
## Goal

Alters our networking code to avoid direct references to the `Embrace` singleton facade.

## Testing


bidetofevil (approved):
Perfect. I had meant to do that 😅

PR #1269, opened 2024-08-19
Move class to feature module

fractalwrench:
## Goal

Moves the `CrashModule` implementation to the feature module.


bidetofevil (approved):
LGTM

PR #1270, opened 2024-08-19
Move network logging to core module

fractalwrench:
## Goal

Moves network logging to the core module. 

## Testing


bidetofevil (approved):
Rubberstamped

PR #1271, opened 2024-08-19
Move the session module to core module

fractalwrench:
## Goal

Move the session module implementation to the core module.

## Testing


bidetofevil (approved):
LGTM

PR #1272, opened 2024-08-19
Move internal interfaces

fractalwrench:
## Goal

Moves the internal interfaces to the API module.

## Testing


bidetofevil (approved):
LGTM

PR #1273, opened 2024-08-19
Modularise remaining injection

fractalwrench:
## Goal

Modularises the remaining injection classes that will be moved out of the SDK module.


bidetofevil (approved):
LGTM

PR #1279, opened 2024-08-20
Delete unbalanced call detector

fractalwrench:
## Goal

Deletes the unbalanced call detector as the log messages are noisy & unactionable.


bidetofevil (approved):
LGTM

PR #1280, opened 2024-08-20
Reduce visibility of symbols in core module

fractalwrench:
## Goal

Reduces the visibility of symbols in the core module to the minimum necessary level.

## Testing


bidetofevil (approved):
Pulling the limits out. Nice. LGTM

PR #1286, opened 2024-08-21
Fix miscellaneous build warnings

fractalwrench:
## Goal

Fix miscellaneous build warnings. I've also enabled warnings as errors since the Kotlin version we use no longer blocks us from enabling that.


bidetofevil (approved):
Nice. LGTM

PR #1294, opened 2024-08-28
Only try to send failed API requests if the ApiClient is initialized

bidetofevil:
## Goal

The method used to send failed requests is being invoked before the ApiService is initialized, before `setSendMethod()` is ever called. This leads to a crash in a background thread because a `lateinit var` was being accessed before it was initialized. This leads to `failed_api_calls..json` to re overridden, which leads to lost data.

To fix this, ditch the lateinit var and use an atomic reference to hold the queue. Before trying to access it, check its nullness, and skip if it's null. This will force the api request to retry at the next interval.

A better fix for this will be to delay the retry attempt until after ApiClient is done, but this change is a more appropriate, surgical fix, so we can use this until we make the bigger fix later.

## Testing

Added unit test to verify the state of the pending API calls queue depending on whether the SendMethod is set.


bidetofevil:
* **#1294** <a href="https://app.graphite.dev/github/pr/embrace-io/embrace-android-sdk/1294?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `master`

<h2></h2>

bidetofevil:
### Merge activity

* **Aug 29, 11:41 AM EDT**: [Graphite](https://app.graphite.dev/github/pr/embrace-io/embrace-android-sdk/1294) rebased this pull request as part of a merge.

PR #1298, opened 2024-08-30
Ensure logs are persisted sync in case of crash

fractalwrench:
## Goal

Ensures that logs are persisted synchronously if a crash happens.

## Testing

Added unit tests.


priettt (approved):
🙌

bidetofevil (approved):
LGTM. This will fix the no network case, so lets merge this now, but it still leaves a hole in the case the network is technical available but is bad (this would've been a problem before even OTel migration tbf).

In `EmbraceApiService.handleApiRequest()`, we should save the request first before we try to send it, and instead delete it if it succeeds to clean this up.

And since this is all happening in a background thread, I wonder if we lose some amount of crashes because the save doesn't happen in time. We may want to examine whether we are truly the "persist everything synchronously when we crash" business with all the recent changes. Sessions we do, but OTel logs? Not sure.

PR #1301, opened 2024-08-30
Refactor how metadata is initialized in the NDK service

fractalwrench:
## Goal

Refactor how metadata is initialized in the NDK service so that it's not coupled to installing signal handlers. This is a pre-requisite refactor for initializing NDK code asynchronously.

## Testing


embrace-android-sdk/src/main/cpp/jnibridge/emb_ndk_manager.c
```diff
@@ -101,10 +99,6 @@ JNIEXPORT void JNICALL
 Java_io_embrace_android_embracesdk_internal_ndk_NdkDelegateImpl__1updateMetaData(JNIEnv *env,
                                                                         jobject thiz,
                                                                         jstring _device_meta_data) {
-    if (!__emb_env) {
```
fractalwrench, line 104: These checks are redundant because `__emb_env` is now statically allocated rather than dynamically allocated so is guaranteed not to fail

priettt (approved):

embrace-android-sdk/src/main/cpp/jnibridge/emb_ndk_manager.c
bidetofevil: Do we need changes in the backend to account for this change?

fractalwrench: This shouldn't affect any payload that gets sent to the backend. `emb_env` is always initialized so I just wanted to remove a redundant check - in practice I don't think these will get called until the SDK is fully initialized.

PR #1303, opened 2024-09-02
Preparatory refactor for initializing NDK error handling on background thread

fractalwrench:
## Goal

This changeset contains several preparations for initializing NDK error handling on a background thread:

- Removed `isServiceInitDeferred` config as the current implementation breaks signal handler installation, and we should also IMO just default to the correct behavior on this rather than giving a choice
- Remove redundant lock/installed checks within `EmbraceNdkService` as the JNI functions can be called as soon as the SO file is loaded
- Attempted to simplify `ModuleInitBootstrapper` section for NDK/native thread sampling installation, without making any functional changes (yet)

## Testing


bidetofevil (approved):
Very nice! Glad we can remove that setting now that it's proven to be fine :-)

PR #1304, opened 2024-09-02
Initialize ndk on background thread

fractalwrench:
## Goal

Alters the SDK so that NDK error detection is initialized on a background thread, which offloads the loading of a SO file (and other I/O) away from the main thread.

This involved two parts: firstly the NDK service is initialized async from `ModuleInitBootstrapper`, and posts the signal handler installation to the front of the message queue so that it is installed on the main thread asap. Secondly, the Unity ANR thread sampling is called directly from the Unity interface. It appears the Unity SDK already [appears to be doing this](https://github.com/embrace-io/embrace-unity-sdk/blob/2eb7231547bb7a63c69eb1f4f52bef6fa4f4b3e8/io.embrace.sdk/Scripts/Native/Embrace_Android.cs#L511) so it shouldn't affect functionality.

## Profiling

This change reduces startup time from ~16ms to ~14ms on a OnePlus 7 running Android 10. Additionally it removes all but one direct call to I/O on the main thread as part of our initialization sequence.

Screenshot 2024-09-02 at 11 29 01
Screenshot 2024-09-02 at 11 29 09

## Testing

Updated test coverage & manually verified that NDK crashes are still reported to the dashboard.

### DIFF
--- a/embrace-android-features/src/main/kotlin/io/embrace/android/embracesdk/internal/injection/NativeFeatureModuleImpl.kt
+++ b/embrace-android-features/src/main/kotlin/io/embrace/android/embracesdk/internal/injection/NativeFeatureModuleImpl.kt
@@ -45,7 +45,6 @@ internal class NativeFeatureModuleImpl(
                 embraceNdkServiceRepository,
                 NdkDelegateImpl(),
                 workerThreadModule.backgroundWorker(WorkerName.BACKGROUND_REGISTRATION),
-                workerThreadModule.backgroundWorker(WorkerName.SERVICE_INIT),
                 payloadSourceModule.deviceArchitecture,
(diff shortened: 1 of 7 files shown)


bidetofevil (approved):
Do we want to put any telemetry in place to see much time elapses between the signal handler installation and the SDK startup? While I think it's definitely the right trade off to make, it would be good to know how large this gap is in production. It may even just be a span event on a session span so we can find this on a case by case basis.

Anyway, this looks good, but be might want to considering add telemetry like this once the startup improvement work is finished.

fractalwrench:
Agree - it'd be good to add that sort of telemetry to anything important that happens in the background once other improvements are made

PR #1305, opened 2024-09-02
Optimise system service initialization

fractalwrench:
## Goal

Optimizes SDK initialization by delaying an unnecessary binder transaction.

## Profiling

Screenshot 2024-09-02 at 11 50 02

Screenshot 2024-09-02 at 11 49 56


bidetofevil (approved):
Keep kicking the can down the road until it's outside of the startup window. Nice!

PR #1306, opened 2024-09-02
Optimize process id generation

fractalwrench:
## Goal

Generating a process ID for the OTel configuration object was taking a surprisingly long period of time (~1ms). I've offloaded this to the background thread as it didn't need to be on the main thread.

## Profiling
Screenshot 2024-09-02 at 14 19 49
Screenshot 2024-09-02 at 14 19 40


bidetofevil (approved):
Wow. I wonder if this is the classloading?

PR #1307, opened 2024-09-02
Update profile installer dependency

fractalwrench:
## Goal

The profile installer dependency used in our SDK is using an [old version](https://developer.android.com/jetpack/androidx/releases/profileinstaller) that doesn't have support for newer versions of Android. In this changeset I've generated a baseline profile again using the latest stable version, which will be supported up to API 34. (API 35 support is in a beta version but we can update to that soon).

## Profiling

This seems to make a large impact on a OnePlus 7 running Android 10!

Screenshot 2024-09-02 at 14 55 37
Screenshot 2024-09-02 at 14 55 30


bidetofevil (approved):
Fantastic! I was wondering how much this will help with the new versions that we didn't support - the answer seems to be a lot!

bidetofevil (commented):
FWIW, I hope this doesn't actually impact too many customers because they'd be generating their own baseline profiles to optimize their app startups, and our stuff should be included. But this will help everyone else :-)

PR #1308, opened 2024-09-02
Remove unused startup status

fractalwrench:
## Goal

The SDK startup status is not used so can be removed. This is read on a background thread so won't have a major impact on startup time.


bidetofevil (approved):
LGTM

PR #1309, opened 2024-09-02
Initialize network connectivity listeners async

fractalwrench:
## Goal

Offloads network connectivity listeners so that they are registered async. These are not essential so can afford to wait a few milliseconds until startup is done.

## Profiling
Screenshot 2024-09-02 at 16 24 29
Screenshot 2024-09-02 at 16 24 23


bidetofevil (approved):
Totally. Nice

PR #1310, opened 2024-09-03
Add lazy properties through startup

fractalwrench:
## Goal

Initializes several properties lazily where they were expensive (mostly due to binder transactions). MetadataService was a particularly expensive example shown in the profiling below.

## Profiling

Screenshot 2024-09-03 at 09 10 23
Screenshot 2024-09-03 at 09 10 32


embrace-android-sdk/lint-baseline.xml
```diff
@@ -67,6 +67,17 @@
             column="20"/>
     </issue>
 
+    <issue
```
bidetofevil, line 70: Is this something we should get rid of?

bidetofevil (approved):
Nice!

embrace-android-sdk/lint-baseline.xml
fractalwrench: Yep - updated to make the property non-private

PR #1311, opened 2024-09-03
Avoid unnecessary HTTP request for crashes

fractalwrench:
## Goal

The code appears to be making a HTTP request to the old `v1/log/events` endpoint for NDK crashes as well as to the new OTel log endpoint. This PR removes the legacy approach which was the last part of startup that performed I/O on the main thread.

## Profiling

This seems like a modest improvement but I've previously seen this call have very high variance as it performs I/O.

Screenshot 2024-09-03 at 10 03 55
Screenshot 2024-09-03 at 10 04 06

## Testing

Manually verified that crashes still show in the dashboard.

### DIFF
--- a/embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/comms/delivery/EmbraceDeliveryService.kt
+++ b/embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/comms/delivery/EmbraceDeliveryService.kt
@@ -22,8 +22,6 @@ import io.embrace.android.embracesdk.internal.session.id.SessionIdTracker
 import io.embrace.android.embracesdk.internal.session.orchestrator.SessionSnapshotType
 import io.embrace.android.embracesdk.internal.spans.findAttributeValue
 import io.embrace.android.embracesdk.internal.utils.Provider
-import io.embrace.android.embracesdk.internal.worker.BackgroundWorker
-import io.embrace.android.embracesdk.internal.worker.TaskPriority
 import java.io.OutputStream
 import java.util.concurrent.TimeUnit
 import kotlin.math.max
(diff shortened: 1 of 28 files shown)


embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/comms/delivery/EmbraceDeliveryService.kt
```diff
@@ -88,59 +84,33 @@ internal class EmbraceDeliveryService(
...
-            }
-        }
-    }
-
     override fun sendCachedSessions(
         nativeCrashServiceProvider: Provider<NativeCrashService?>,
         sessionIdTracker: SessionIdTracker
     ) {
-        sendCachedCrash()
-        backgroundWorker.submit(TaskPriority.HIGH) {
```
priettt, line 107: is it okay to stop using the backgroundWorker for this?

fractalwrench: It would not be, this still needs to run on a background thread. I've refactored it so `EmbraceDeliveryService` doesn't know anything about the worker: https://github.com/embrace-io/embrace-android-sdk/pull/1311/files#diff-c270387adce360d4187ce687c82e19785f634d47e6989038139d343ac00da9b4R207

priettt (approved):

embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/comms/delivery/EmbraceDeliveryService.kt
priettt: awesome, I hadn't reached that part yet 😅

PR #1313, opened 2024-09-03
Avoid binder transaction when starting session

fractalwrench:
## Goal

Avoids a binder transaction when starting a session. The network connectivity service no longer needs to be informed of a session start as it uses spans, whereas in the legacy data model it did.

## Profiling

Screenshot 2024-09-03 at 13 41 23
Screenshot 2024-09-03 at 13 41 17


bidetofevil (approved):
Oh that sucks we were doing that. LGTM

PR #1314, opened 2024-09-03
Avoid unnecessary service registration

fractalwrench:
## Goal

Avoids some unnecessary service registration & moves some metadata service functionality off the main thread.

## Testing


bidetofevil (approved):
LGTM

PR #1315, opened 2024-09-03
Register services lazily

fractalwrench:
## Goal

Adds objects to the `ServiceRegistry` using lazy initialization so they are only created at the end of startup, to make their impact more obvious. I also removed various objects that don't need to be registered anymore.

## Profiling

Screenshot 2024-09-03 at 15 04 21
Screenshot 2024-09-03 at 15 04 15

### DIFF
--- a/embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/config/EmbraceConfigService.kt
+++ b/embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/config/EmbraceConfigService.kt
@@ -316,10 +316,6 @@ internal class EmbraceConfigService(
         return clock.now() > lastRefreshConfigAttempt + configRetrySafeWindow * 1000
     }
 
-    override fun close() {
-        logger.logDebug("Shutting down EmbraceConfigService")
-    }
-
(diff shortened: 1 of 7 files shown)


embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/config/EmbraceConfigService.kt
```diff
@@ -316,10 +316,6 @@ internal class EmbraceConfigService(
         return clock.now() > lastRefreshConfigAttempt + configRetrySafeWindow * 1000
     }
 
-    override fun close() {
-        logger.logDebug("Shutting down EmbraceConfigService")
```
bidetofevil, line 320: lol that's all it's doing?

bidetofevil (approved):
Oh this is going to be so impactful! LGTM

embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/config/EmbraceConfigService.kt
fractalwrench: Yep! Screaming into the void :D

PR #1316, opened 2024-09-03
Moved scheduled retry after send method setup

priettt:
Defer the submission of the api calls delivery to the scheduledWorker until `sendMethodRef` (which we will use to execute the delivery) is set


bidetofevil (approved):
LGTM

PR #1322, opened 2024-09-05
Save crashes synchronously and process other unbatched log requests in background thread

bidetofevil:
## Goal

Changed the `SendImmediately` designation to a `SendMode` enum to additionally allow logs to be saved instead of sent, and use this to save crash logs rather than send them. This ensures they will be delivered on the next app startup in a consistent matter, rather than rely on our existing, flakey mechanism to attempt to send synchronously, and only if it fails, save the payload.

Bundled with this change is the need to serialize the log payload and create the request for other unbatched logs off the calling thread. Right now, they are not being done on a background thread - only sent through it - so this could be problematic depending on where it's being invoked. Instead, we will run them on the same scheduler that we use to run the log batching job.

## Testing

Added unit tests everywhere to ensure this SendMode concept is passed down. Also added integration test to ensure that this works.

### DIFF
--- a/embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/arch/schema/EmbType.kt
+++ b/embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/arch/schema/EmbType.kt
@@ -3,7 +3,7 @@ package io.embrace.android.embracesdk.internal.arch.schema
 public sealed class EmbType(type: String, subtype: String?) : TelemetryType {
     override val key: EmbraceAttributeKey = EmbraceAttributeKey(id = "type")
     override val value: String = type + (subtype?.run { ".$this" } ?: "")
-    override val sendImmediately: Boolean = false
+    override val sendMode: SendMode = SendMode.DEFAULT
 
     /**
      * Keys that track how fast a time interval is. Only applies to spans.
@@ -48,7 +48,7 @@ public sealed class EmbType(type: String, subtype: String?) : TelemetryType {
      */
     public sealed class System(
         subtype: String,
-        override val sendImmediately: Boolean = false
+        override val sendMode: SendMode = SendMode.DEFAULT
     ) : EmbType("sys", subtype) {
 
         public object Breadcrumb : System("breadcrumb")
@@ -71,11 +71,11 @@ public sealed class EmbType(type: String, subtype: String?) : TelemetryType {
             public val embFlutterExceptionLibrary: EmbraceAttributeKey = EmbraceAttributeKey("exception.library")
         }
 
-        public object Exit : System("exit", true)
+        public object Exit : System("exit", SendMode.IMMEDIATE)
 
         public object PushNotification : System("push_notification")
 
-        public object Crash : System("android.crash", true) {
+        public object Crash : System("android.crash", SendMode.DEFER) {
             /**
              * The list of [Throwable] that caused the exception responsible for a crash
              */
@@ -83,7 +83,7 @@ public sealed class EmbType(type: String, subtype: String?) : TelemetryType {
                 EmbraceAttributeKey("android.crash.exception_cause")
         }
 
-        public object ReactNativeCrash : System("android.react_native_crash", true) {
+        public object ReactNativeCrash : System("android.react_native_crash", SendMode.DEFER) {
             /**
              * The JavaScript unhandled exception from the ReactNative layer
              */
@@ -92,9 +92,9 @@ public sealed class EmbType(type: String, subtype: String?) : TelemetryType {
             )
         }
 
-        public object ReactNativeAction : System("rn_action", true)
+        public object ReactNativeAction : System("rn_action")
 
-        public object NativeCrash : System("android.native_crash", true) {
+        public object NativeCrash : System("android.native_crash", SendMode.DEFER) {
             /**
              * Exception coming from the native layer
              */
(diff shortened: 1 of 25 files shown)


bidetofevil:
* **#1322** <a href="https://app.graphite.dev/github/pr/embrace-io/embrace-android-sdk/1322?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `master`

<h2></h2>

embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/arch/schema/SendMode.kt
```diff
@@ -0,0 +1,17 @@
+package io.embrace.android.embracesdk.internal.arch.schema
+
+import io.embrace.android.embracesdk.internal.arch.schema.SendMode.DEFAULT
+import io.embrace.android.embracesdk.internal.arch.schema.SendMode.DEFER
+import io.embrace.android.embracesdk.internal.arch.schema.SendMode.IMMEDIATE
+
+public enum class SendMode {
+    DEFAULT, IMMEDIATE, DEFER
```
fractalwrench, line 8: Some Kdoc here would be helpful in explaining what each mode does

embrace-android-core/src/test/java/io/embrace/android/embracesdk/internal/logs/LogOrchestratorTest.kt
```diff
@@ -178,13 +179,24 @@ internal class LogOrchestratorTest {
...
+    fun `logs with IMMEDIATE SendMode are sent immediately`() {
         logSink.storeLogs(listOf(unbatchableLogRecordData))
+        executorService.runCurrentlyBlocked()
         // Verify the logs are sent
         assertNull(logSink.pollNonbatchedLog())
         verifyPayload(1)
     }
 
+    @Test
+    fun `logs with DEFER SendMode are saved by not sent`() {
```
fractalwrench, line 191:
```suggestion
    fun `logs with DEFER SendMode are saved but not sent`() {
```

fractalwrench (approved):
LGTM

embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/arch/schema/EmbType.kt
```diff
@@ -92,9 +92,9 @@ public sealed class EmbType(type: String, subtype: String?) : TelemetryType {
             )
         }
 
-        public object ReactNativeAction : System("rn_action", true)
```
priettt, line 95: This was being sent as "true" before, and now it's using the default sendMode. Is this expected?

embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/envelope/log/LogEnvelopeSource.kt
```diff
@@ -1,10 +1,11 @@
...
 
+import io.embrace.android.embracesdk.internal.logs.LogRequest
 import io.embrace.android.embracesdk.internal.payload.Envelope
 import io.embrace.android.embracesdk.internal.payload.LogPayload
 
 public interface LogEnvelopeSource {
     public fun getBatchedLogEnvelope(): Envelope<LogPayload>
 
-    public fun getNonbatchedEnvelope(): List<Envelope<LogPayload>>
+    public fun getNonbatchedEnvelope(): List<LogRequest<Envelope<LogPayload>>>
```
priettt, line 10: no need to change it now as it's not part of your PR but I don't like "nonbatched" lol

Unbatched is an English word right?

priettt (approved):

embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/arch/schema/SendMode.kt
bidetofevil: Yeah good call. I was going to wait until we do the refactor but I shouldn't wait for that.

embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/arch/schema/EmbType.kt
bidetofevil: Yes - this is a span, and this doesn't apply anyway. I think it was a mistake to supply it before.

embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/envelope/log/LogEnvelopeSource.kt
bidetofevil: Lol I actually changed it to something else but it sounded weird so I just reverted it. Let me take another stab at it. Do you think `SingleLogBatch` is better?

bidetofevil: Did a pass on the names and got rid of any "nonbatch" references. I kept one "unbatched", but the rest use `SingleLog` or something that denotes that.

bidetofevil:
### Merge activity

* **Sep 5, 12:08 PM EDT**: @bidetofevil started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/embrace-io/embrace-android-sdk/1322).

PR #1323, opened 2024-09-05
Add instrumented config classes

fractalwrench:
## Goal

Adds new classes that will hold values from a deserialized `embrace-config.json`. These classes will have functions that return a default value, thus simplifying the existing `*Behavior` class implementations. If a customer specifies their own key within `embrace-config.json` the swazzler will instrument the relevant class and substitute the appropriate return value.

I have not implemented this bytecode instrumentation in full yet as I'd like feedback on the class structure in this changeset first. There are a few things to note:

1. Each class defines functions rather than properties. This simplifies the swazzling as it provides a clear target to alter. I first attempted using properties & ran into difficulties with the Kotlin compiler optimizing values away
2. I attempted to reorganise the config into more logical groupings. I don't intend on changing `embrace-config.json` schema yet, but this may form a useful starting point if we move over to supporting yaml config
3. I deliberately chose not to supply an interface/testable constructor because this adds more code. I'm assuming that we can always test what we want by supplying a fake `Behavior` interface
4. Mappings to the `embrace-config.json` schema are in the docs for each function


bidetofevil (commented):
I like the approach. If the functions return constants, the compiler should just inline them at build time so we don't incur lots of unnecessary function calls?

Doing this way allows the code gen to be done with any input format, and also not necessarily in the swazzler (e.g. swazzler invokes a binary that does this), which will make it a lot more adaptable and testable.

Aside: looking at this, are we getting close to creating a common repo/module that has dependencies between the Swazzler and SDK? We have artificially created this by requiring the SDK and Swazzler to be of the same version, so why not make it explicit?

Another way to approach this - what about adding the swazzler to the SDK repo so we can manage dependencies like deps between any two SDK modules?

fractalwrench:
>I like the approach. If the functions return constants, the compiler should just inline them at build time so we don't incur lots of unnecessary function calls?

I'm unsure of how the internals of the Kotlin compiler will behave (particularly as it'll differ between versions) but we instrument bytecode before any optimizations occur, so it seems likely the compiler would make any necessary optimizations.

>Another way to approach this - what about adding the swazzler to the SDK repo so we can manage dependencies like deps between any two SDK modules?

I think this would simplify quite a few things & would be worth doing as the long term solution.

bidetofevil (approved):
LGTM

PR #1324, opened 2024-09-05
Added concurrency policy to pull_request Github actions

priettt:
This concurrency policy cancels in-progress runs when pushing new code to a specific branch.

Example: 

- Let's say we create a `branch_A`, and create a pull request. This will trigger the actions in this PR. 
- If we create a `branch_B`, the actions will also run, simultaneously.
- If we push new code to `branch_A`, the actions that were running on that branch will be canceled, and new ones will be triggered. 

This avoids running actions on stale code.

### DIFF
--- a/.github/workflows/ci-gradle.yml
+++ b/.github/workflows/ci-gradle.yml
@@ -8,6 +8,10 @@ on:
   release:
     types: [ released ]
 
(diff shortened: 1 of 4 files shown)


bidetofevil (approved):
Nice

PR #1326, opened 2024-09-06
Add moshi rules to prevent adapters failing in minified build

fractalwrench:
## Goal

Moshi requires enums to be annotated to prevent R8/ProGuard from stripping their values. Several enums in our codebase were not annotated in this way, which worked fine until we changed the ProGuard rules to be more specific in #1043.

This change adds the `JsonClass` annotation to enums [as recommended](https://github.com/square/moshi?tab=readme-ov-file#r8--proguard).

## Testing

Manually verified against the `release` build variant of a minified test app that no errors show up in Logcat related to serialization.

### DIFF
--- a/embrace-android-sdk/embrace-proguard.cfg
+++ b/embrace-android-sdk/embrace-proguard.cfg
@@ -1,7 +1,15 @@
 -keepattributes LineNumberTable, SourceFile
 
-## Keep classes used by hosted SDKs
+## Keep public API including enums
 -keep class io.embrace.android.embracesdk.Embrace { *; }
+-keep class io.embrace.android.embracesdk.network.http.HttpMethod { *; }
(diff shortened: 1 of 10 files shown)


priettt (approved):

embrace-android-sdk/embrace-proguard.cfg
```diff
@@ -1,7 +1,15 @@
 -keepattributes LineNumberTable, SourceFile
 
-## Keep classes used by hosted SDKs
+## Keep public API including enums
```
bidetofevil, line 3: After this, should we move all enums to the same package so we can just specify that instead and make sure future ones are also protected?

bidetofevil (approved):
LGTM

PR #1327, opened 2024-09-06
Retry session delivery where possible

fractalwrench:
## Goal

Alters the session delivery logic so that for cases where a payload is still deliverable (e.g. if there is no network connection or server response) the SDK attempts delivery again rather than dropping the payload.

## Testing

Added unit tests for new retry logic & manually verified an error is no longer logged in a test app with no network connection.


priettt (approved):

bidetofevil (approved):
LGTM

PR #1329, opened 2024-09-06
Remove unnecessary public modifiers

fractalwrench:
## Goal

Builds on #1328 by removing unnecessary public modifiers from modules that don't use Kotlin's explicit API mode. I achieved this by running an inspection inside Android Studio & accepting the auto-fixes.

## Testing

Ensured existing tests pass.


priettt (approved):
heh I'll trust android studio

PR #1330, opened 2024-09-06
Optimize log init

fractalwrench:
## Goal

Defers instantiation of the log orchestrator until the first log is created.

## Profiling

This reduces the log-init section but offloads quite a bit to the session orchestrator, which can be optimized in future.

Screenshot 2024-09-06 at 14 43 01

Screenshot 2024-09-06 at 14 43 06


bidetofevil (approved):
Nice. LGTM

PR #1331, opened 2024-09-06
Improved generate baseline profile workflow

priettt:
- Removed unnecessary matrix.
- Removed fetch-depth: 0. I don't think we need that.
- Added setup-gradle action, which handles caching and wrapper validation.
- Testing if AVD cache is actually needed.


bidetofevil (approved):
I assume this just works, so LGTM

PR #1332, opened 2024-09-06
Add extra tracing

fractalwrench:
## Goal

Adds some additional tracing for startup benchmarking, and also tweaks a couple of places that were using references instead of lambdas & slowing things down.


bidetofevil (approved):
LGTM

PR #1340, opened 2024-09-09
Github Actions improvements

priettt:
- Stop using matrix for jdk-version
- Use setup-gradle instead of cache and wrapper_validation
- Use ubuntu-latest instead of macos
- Don't use fetch-depth: 0 when it's not necessary
- Add KVM group permissions for emulator
- Use working-directory instead of cd
- Remove unnecessary refs to default branches
- Replaced references to "master" with "main"


fractalwrench (approved):
LGTM once tests are passing. I think we also need to update the GH mandatory checks because the name of the test job has changed.

bidetofevil (approved):
LGTM

PR #1342, opened 2024-09-10
Create fake behavior implementations

fractalwrench:
## Goal

Test code is quite often constructing a real instance of the behavior implementations, when in the majority of cases the test only wants to disable/enable a feature or tweak one setting. I've gone through our test code & created fakes that will make it easier. Where tests are genuinely testing the existing config implementation I've left things as is.


priettt (approved):
🙌

PR #1343, opened 2024-09-10
Rename config functions

fractalwrench:
## Goal

Renames functions in the behavior classes for more consistency with the `InstrumentedConfig` naming scheme.


priettt (approved):
👍

PR #1344, opened 2024-09-10
Use instrumented config classes to drive behavior

fractalwrench:
## Goal

Updates the SDK implementation so that instead of `LocalConfig` the `InstrumentedConfig` drives the behavior of features. This builds on #1323 which adds the new class that will be swazzled with the contents of the config file.

## Testing


priettt (approved):

bidetofevil (approved):
Skimmed this only but this seems right

PR #1349, opened 2024-09-11
Remove old config classes

fractalwrench:
## Goal

The final piece of the puzzle started in #1323. This changeset removes the now unused `LocalConfig` and associated classes in favour of `InstrumentedConfig`, and performs final tweaks to get the SDK working.

## Testing


priettt (approved):
nice 🙌

bidetofevil (approved):
LGTM. Not related to this directly, but we should get rid of custom app IDs. We should have that now to tell us if folks are actually using it.

PR #1350, opened 2024-09-11
Load the native library in NdkService before accessing the delegate methods

priettt:
_updateAppState was accessed before the native library was loaded. I added some guards to the delegate methods so we always verify that the native library is loaded before we call them.


bidetofevil (commented):
For now, can we just check whether the native lib is loaded an return early if it's not ready? Ultimately, we want the usage, perhaps even init, of this class to be dependent on the .so object load succeeding, but all that is too much to do right now if this is something we want to get out today

bidetofevil (approved):
LGTM. We may want to consider just checking the load status when one of these methods are called, but whatever you think is best works.

PR #1351, opened 2024-09-11
Add cherrypicked fix for NdkService

priettt:


bidetofevil (approved):
LGTM. After we kick off the build, we should verify in a test that the crashing condition no longer results in a crash - but I don't want that to block the release.

PR #1355, opened 2024-09-12
Prevent JNI error on SIGQUIT when Google ANR capture enabled

fractalwrench:
## Goal

Prevents a JNI error that is triggered by SIGQUIT when Google ANR capture is enabled (i.e. when an ANR happens). This feature is disabled by default & users would either need to set `sdk_config.anr.capture_google` to `true` in `embrace-config.json` _or_ have a remote config set that enables this behavior. The error looks like:

>JNI DETECTED ERROR IN APPLICATION: can't call void io.embrace.android.embracesdk.internal.anr.sigquit.SigquitDataSourceImpl.saveSigquit(long) on instance of io.embrace.android.embracesdk.internal.anr.sigquit.EmbraceSigquitNdkDelegate

This was been introduced in #1214 so will be present in 6.10.0 upwards. The `anr_manager.c` code assumed that the caller would always be the object that would contain the `saveSigquit` function, but that assumption was invalidated in #1214. I've fixed this by explicitly passing a reference to the object that should be called from the JNI.

Additionally, I've added a Proguard keep rule as the class doesn't seem to have a suitable keep rule.

## Testing

Manually confirmed in a debug + release app that triggering an ANR with this feature enabled doesn't crash the process.


priettt (approved):
I never realized we were using the `thiz` object 🤦

bidetofevil (approved):
NDK!

PR #1356, opened 2024-09-12
Remove unused build id function from project config

fractalwrench:
## Goal

This function is not instrumented by the swazzler yet & `BuildInfo` is used in the SDK still. This changeset removes it for now to avoid confusion.


cesarmax22 (approved):
lgtm

PR #1357, opened 2024-09-12
Replace most resource injection with config instrumentation

fractalwrench:
## Goal

Adds extra fields that will be instrumented by the swazzler rather than using build resource injection.

## Testing

Manually verified by logging out values in test app:

>getBuildId: 905E7A601CD7437EB7FF9AC145BE45FC
>getBuildType: debug
>getBuildFlavor: demoDevelopment


bidetofevil (approved):
Hell yeah

PR #1360, opened 2024-09-12
Download and cache NDK

priettt:
Use a specific NDK version so we don't depend on the github runners having it pre installed


bidetofevil (approved):
LGTM

PR #1362, opened 2024-09-13
Minor tweaks to perf

fractalwrench:
## Goal

Some minor tweaks to startup performance & tracing. In particular I've simplified the ANR service initialization as the configService is now immutable & won't change.


bidetofevil (approved):
LGTM

PR #1363, opened 2024-09-13
[DNM] Remove moments implementation

fractalwrench:
## Goal

Removes the implementation of moments as the Tracing API is its superior replacement. This should only be merged as part of 7.0 as it contains breaking changes.

The startup tracking in `SessionSpanAttrPopulatorImpl` is one class I'd like to draw attention to. I've removed the `embSessionStartupThreshold` attribute as I don't believe it's used. I'm also now relying on the `AppStartupTraceEmitter` to supply startup information rather than the startup moment. @bidetofevil I think you probably have the most context on this part of the codebase - does that sound like a correct approach?

## Testing


UPGRADING.md
```diff
@@ -1,8 +1,14 @@
 # Upgrade guide
 
+# Upgrading from 6.x to 7.x
+
+Version 7 of the Embrace Android SDK has breaking changes:
+
+- The `startMoment/endMoment` API has been removed. Use `startSpan/recordSpan` instead.
```
bidetofevil, line 7: We should probably write a similar upgrade guide that iOS has for moving from Moments to Traces

embrace-android-api/api/embrace-android-api.api
```diff
@@ -76,7 +76,6 @@ public abstract interface class io/embrace/android/embracesdk/internal/EmbraceIn
 	public abstract fun recordIncompleteNetworkRequest (Ljava/lang/String;Ljava/lang/String;JJLjava/lang/String;Ljava/lang/String;Ljava/lang/String;Lio/embrace/android/embracesdk/internal/network/http/NetworkCaptureData;)V
 	public abstract fun recordIncompleteNetworkRequest (Ljava/lang/String;Ljava/lang/String;JJLjava/lang/Throwable;Ljava/lang/String;Lio/embrace/android/embracesdk/internal/network/http/NetworkCaptureData;)V
 	public abstract fun recordNetworkRequest (Lio/embrace/android/embracesdk/network/EmbraceNetworkRequest;)V
-	public abstract fun setProcessStartedByNotification ()V
```
bidetofevil, line 79: I was going to advocate for keeping this, but I see precisely 0 calls were made in the last 7 days, I don't think anyone actually uses it.

UPGRADING.md
fractalwrench: Added docs here: https://github.com/embrace-io/embrace-docs/pull/43

fractalwrench:
@bidetofevil this is now ready for re-review

embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/capture/startup/StartupServiceImpl.kt
```diff
@@ -1,12 +1,9 @@
 package io.embrace.android.embracesdk.internal.capture.startup
 
 import io.embrace.android.embracesdk.internal.spans.SpanService
-import io.embrace.android.embracesdk.internal.worker.BackgroundWorker
 
-internal class StartupServiceImpl(
-    private val spanService: SpanService,
-    @Suppress("UnusedPrivateMember")
-    private val backgroundWorker: BackgroundWorker
+class StartupServiceImpl(
```
bidetofevil, line 5: In the future, we can probably streamline and simplify this quite a bit and all its moments responsibilities are gone. Seems excess to have an entire service just providing startup info and creating a trace.

embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/session/orchestrator/SessionSpanAttrPopulatorImpl.kt
```diff
@@ -54,20 +51,10 @@ internal class SessionSpanAttrPopulatorImpl(
...
-                addSystemAttribute(
-                    SpanAttributeData(
-                        embSdkStartupDuration.name,
-                        sdkStartupDurationProvider(coldStart).toString()
-                    )
-                )
-                addSystemAttribute(SpanAttributeData(embSessionStartupDuration.name, info.duration.toString()))
-                addSystemAttribute(SpanAttributeData(embSessionStartupThreshold.name, info.threshold.toString()))
+            if (coldStart) {
+                startupService.getSdkStartupDuration()?.let { duration ->
```
bidetofevil, line 55: This is fine as long as the service is populated before then, when `StartupService.setSdkStartupInfo()` is called during init.

I will say though - is this actually used? I can confirm with the backend folks, but the `emb-sdk-startup` span records the same information but provides much more context. 

If it's init ordering is OK, this should be OK. But if we can rip it out, then it's probably best to do that. I'll confirm.

bidetofevil (approved):
Looks good. Top 3 favourite Embrace PR for me lol.

BTW, we can probably remove/deprecate `@StartupActivity` and replace it with a better-named annotation to denote Activities we want to skip when tracking app startup, i.e. `@StartupTrampolineActivity`

PR #1371, opened 2024-09-17
[DNM] Delete sigquit capture

fractalwrench:
## Goal

Removes the implementation of sigquit capture. This should only be merged as part of 7.0 as it contains breaking changes.


bidetofevil (approved):
LGTM

PR #1384, opened 2024-09-18
Add additional worker to test in FakeWorkerModule 

bidetofevil:
## Goal

Add an additional worker that we can tweak the execution via a blocking executor. This is needed for testing the delivery service components

This PR was hijacked for this as the actual scheduling service code in previous revisions that were under discussion will come in subsequent PRs.

### DIFF
--- a/embrace-android-core/src/test/java/io/embrace/android/embracesdk/internal/event/EmbraceEventServiceTest.kt
+++ b/embrace-android-core/src/test/java/io/embrace/android/embracesdk/internal/event/EmbraceEventServiceTest.kt
@@ -90,7 +90,7 @@ internal class EmbraceEventServiceTest {
         sessionPropertiesService = FakeSessionPropertiesService()
         gatingService = FakeGatingService(EmbraceGatingService(configService, FakeLogService(), FakeEmbLogger()))
         val initModule = FakeInitModule(clock = fakeClock)
-        fakeWorkerThreadModule = FakeWorkerThreadModule(fakeInitModule = initModule, name = Worker.Background.NonIoRegWorker)
+        fakeWorkerThreadModule = FakeWorkerThreadModule(fakeInitModule = initModule, testWorkerName = Worker.Background.NonIoRegWorker)
         eventHandler = EventHandler(
             metadataService = metadataService,
             sessionIdTracker = sessionIdTracker,
--- a/embrace-android-core/src/test/java/io/embrace/android/embracesdk/internal/injection/DataSourceModuleImplTest.kt
+++ b/embrace-android-core/src/test/java/io/embrace/android/embracesdk/internal/injection/DataSourceModuleImplTest.kt
@@ -18,7 +18,7 @@ internal class DataSourceModuleImplTest {
             FakeConfigService(),
             FakeWorkerThreadModule(
                 fakeInitModule = fakeInitModule,
-                name = Worker.Background.NonIoRegWorker
+                testWorkerName = Worker.Background.NonIoRegWorker
             ),
         )
         assertSame(module.dataCaptureOrchestrator, module.embraceFeatureRegistry)
--- a/embrace-android-core/src/test/java/io/embrace/android/embracesdk/internal/session/SessionOrchestrationModuleImplTest.kt
+++ b/embrace-android-core/src/test/java/io/embrace/android/embracesdk/internal/session/SessionOrchestrationModuleImplTest.kt
@@ -23,7 +23,7 @@ internal class SessionOrchestrationModuleImplTest {
     private val configService = FakeConfigService()
     private val workerThreadModule = FakeWorkerThreadModule(
         fakeInitModule = initModule,
-        name = Worker.Background.NonIoRegWorker
+        testWorkerName = Worker.Background.NonIoRegWorker
     )
 
     @Test
--- a/embrace-android-features/src/test/java/io/embrace/android/embracesdk/internal/crash/LastRunCrashVerifierTest.kt
+++ b/embrace-android-features/src/test/java/io/embrace/android/embracesdk/internal/crash/LastRunCrashVerifierTest.kt
@@ -25,7 +25,8 @@ internal class LastRunCrashVerifierTest {
     fun setUp() {
         mockCrashFileMarker = mockk()
         lastRunCrashVerifier = LastRunCrashVerifier(mockCrashFileMarker, EmbLoggerImpl())
-        fakeWorkerThreadModule = FakeWorkerThreadModule(fakeInitModule = FakeInitModule(), name = Worker.Background.NonIoRegWorker)
+        fakeWorkerThreadModule =
+            FakeWorkerThreadModule(fakeInitModule = FakeInitModule(), testWorkerName = Worker.Background.NonIoRegWorker)
         worker = fakeWorkerThreadModule.backgroundWorker(Worker.Background.NonIoRegWorker)
     }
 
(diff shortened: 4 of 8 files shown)


bidetofevil:
* **#1384** <a href="https://app.graphite.dev/github/pr/embrace-io/embrace-android-sdk/1384?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/comms/delivery/v2/SchedulingService.kt
```diff
@@ -0,0 +1,111 @@
...
+
+    private val retryQueue: Queue<File> = ConcurrentLinkedQueue()
+
+    override fun checkPendingPayloads() {
+        schedulingWorker.submit {
+            var ready = readyToSend()
+            while (ready) {
+                runCatching {
+                    getNextReadyPayload()?.let { payload ->
+                        val response = send(payload).get()
```
fractalwrench, line 33: Not keen on this `.get()`. I think the approach of delegating to another worker is good & enables future multi-threaded request sending, but I don't think it makes sense to block this thread here

embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/comms/delivery/v2/CacheService2.kt
```diff
@@ -0,0 +1,8 @@
+package io.embrace.android.embracesdk.internal.comms.delivery.v2
+
+import java.io.File
+
+interface CacheService2 {
+    fun deletePayload(file: File)
```
fractalwrench, line 6: Is it worth creating an abstraction over `File`, such as `Payload`, that contains the internal implementation that loads a file for sending?

One of the problems with our current approach is that it's fairly simple to just delete or transform a `File` once a reference is obtained & this breaks assumptions made by other components. I can't decide whether this is over-engineering or justified paranoia so would be interesting to hear your thoughts

embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/comms/delivery/v2/CacheService2.kt
```diff
@@ -0,0 +1,8 @@
...
+    fun getPayloads(priority: Int): List<File>
```
fractalwrench, line 7: What are the possible values of `priority`? Could an enum better represent these states?

embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/comms/delivery/v2/ExecutionService.kt
```diff
@@ -0,0 +1,8 @@
+package io.embrace.android.embracesdk.internal.comms.delivery.v2
+
+import io.embrace.android.embracesdk.internal.comms.api.ApiResponse
+import java.io.File
+
+interface ExecutionService {
+    fun sendPayload(payload: File): ApiResponse
```
fractalwrench, line 7: Regardless of whether we choose to create an abstraction over `File`, I think it'd be good to be explicit in naming on whether this refers to a JSON payload or a HTTP payload.

embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/comms/delivery/v2/SchedulingService.kt
```diff
@@ -0,0 +1,111 @@
...
+                        if (response is ApiResponse.TooManyRequests) {
+                            blockEndpoint(response)
+                        }
+                    }
+                }
+
+                ready = readyToSend()
```
fractalwrench, line 40: Is this necessary?

embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/comms/delivery/v2/SchedulingService.kt
```diff
@@ -0,0 +1,111 @@
...
+            }
+            scheduleNextRetry()
+        }
+    }
+
+    private fun getNextReadyPayload(): File? {
```
fractalwrench, line 46: Should this be the priority of the scheduling service or should it be in the cache service?

bidetofevil: This service needs to be aware of more than just priority, but also if a particular endpoint is being 429'd or if the highest priority payload needs to be retried.

So if we deferred this to the CacheService, we'd have to make multiple reads disk. 

Now, I do think it's valuable to get a list of Files in sorted order by priority. I'll modify the interface to specify that in the comments and method name.

embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/comms/delivery/v2/SchedulingService.kt
bidetofevil: Ah, I messed up the logic here. I meant to not block on the current thread, but instead do that in the `send()` and the the future that it returns. I'll fix that up

embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/comms/delivery/v2/SchedulingService.kt
bidetofevil: The intent here is to check to see if the SDK and the execute environment is now ready to send requests. Thinking more about it, if we can just have the component react to any state changes that brings the SDK back on line, we don't need to check at each iteration. 

Let me take a look to see if I can remove.

embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/comms/delivery/v2/CacheService2.kt
bidetofevil: Oh yeah, we should do that. I just wanted something quick here but it should be a durable way to uniquely referencing a payload. Frankly, the file name is probably good enough if we expect the file name no never change, i.e. the timestamp is always when it was first written or something.

embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/comms/delivery/v2/CacheService2.kt
bidetofevil: oh definitely enum. Again, I was just trying to clobbering something up for testing purposes. I should've specified as much

embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/comms/delivery/v2/ExecutionService.kt
bidetofevil: Right. Definitely a JSON payload, i.e. the thing we stick inside the Body of an HTTP request

fractalwrench (approved):
LGTM

bidetofevil:
### Merge activity

* **Sep 25, 10:59 AM EDT**: @bidetofevil started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/embrace-io/embrace-android-sdk/1384).

PR #1385, opened 2024-09-18
Setup new module for delivery code

fractalwrench:
## Goal

Creates the skeleton of a new module for delivery-related code. I haven't moved anything over to this module as of yet.


bidetofevil (approved):
LGTM

PR #1387, opened 2024-09-18
Set correct version in android-sdk-benchmark repo when running generate_baseline_profile.yml

priettt:
- Grab the version from gradle.properties in embrace-android-sdk
- Use that version when running the baseline-profile.

### DIFF
--- a/.github/workflows/generate_baseline_profile.yml
+++ b/.github/workflows/generate_baseline_profile.yml
@@ -23,9 +23,6 @@ jobs:
           sudo udevadm control --reload-rules
           sudo udevadm trigger --name-match=kvm
 
-      - name: Checkout Branch
-        uses: actions/checkout@v4
-
       - name: Setup Java
         uses: actions/setup-java@v4
         with:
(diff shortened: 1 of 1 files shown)


.github/workflows/generate_baseline_profile.yml
```diff
@@ -43,6 +47,12 @@ jobs:
           path: ./android-sdk-benchmark
           token: ${{ secrets.GH_ANDROID_SDK_TOKEN || secrets.token }} # NOTE: read android-sdk-benchmark
 
+      - name: Update gradle.properties with the proper SDK version
+        working-directory: android-sdk-benchmark
+        run: |
+          sed -i "s/swazzler_version=.*/swazzler_version=${{ env.ANDROID_SDK_VERSION }}/" gradle.properties
```
fractalwrench, line 53: As an FYI you can supply this as a [command-line arg to Gradle](https://docs.gradle.org/current/userguide/build_environment.html#priority_for_configurations) if you want a solution that doesn't create a dirty working tree

fractalwrench (approved):
LGTM

.github/workflows/generate_baseline_profile.yml
priettt: that's a great idea I'll use that 🙌

PR #1388, opened 2024-09-18
Add functionality for shutting down worker & waiting for tasks to complete

fractalwrench:
## Goal

Adds a function to `BackgroundWorker` that calls `shutdown` and then `awaitTermination` on the `ExecutorService` implementation. This is intended for use when payloads are being persisted but the process is about to terminate due to a JVM crash. However, it will likely become useful in other scenarios.

## Testing

Added unit test coverage.


bidetofevil (approved):
LGTM

PR #1392, opened 2024-09-18
update-main-versions.yml now creates PRs with the new versions

priettt:


bidetofevil (approved):
LGTM

Lets do a follow up to also change the test suite and benchmark apps?

PR #1396, opened 2024-09-19
Create intake service interface

fractalwrench:
## Goal

Creates an initial interface for the `IntakeService`, an initial implementation for the file metadata, and a fake service implementation.

## Discussion

1. Should we distinguish between different log payloads (log/crash/network) in the file metadata?
2. Do we want to rename the supported log payload types in the file metadata at all?
3. Are there additional types to support (e.g. AEI)?
4. Does the filename metadata structure look correct? Is there anything else we want to encode?

## Testing

Added unit test coverage

### DIFF
--- a/embrace-android-delivery/src/main/kotlin/io/embrace/android/embracesdk/internal/delivery/StoredTelemetryMetadata.kt
+++ b/embrace-android-delivery/src/main/kotlin/io/embrace/android/embracesdk/internal/delivery/StoredTelemetryMetadata.kt
@@ -1,0 +1,38 @@
+package io.embrace.android.embracesdk.internal.delivery
+
+import kotlin.Result.Companion.failure
+
+/**
+ * Metadata about the telemetry payload.
+ *
+ * This information is encoded in the filename.
+ */
+class StoredTelemetryMetadata(
+    val timestamp: Long,
+    val uuid: String,
+    val type: SupportedEnvelopeType,
+    val filename: String = "v1_${timestamp}_${type.description}_$uuid.json"
+) {
+
+    companion object {
+
+        /**
+         * Parses a filename and constructs a [StoredTelemetryMetadata] object. This returns a
+         * [Result] because the filename may be invalid.
+         */
+        fun fromFilename(filename: String): Result<StoredTelemetryMetadata> {
+            val parts = filename.split("_")
+            if (parts.size != 4) {
+                return failure(IllegalArgumentException("Invalid filename: $filename"))
+            }
+            val timestamp = parts[1].toLongOrNull() ?: return failure(
+                IllegalArgumentException("Invalid timestamp: $filename")
+            )
+            val type = SupportedEnvelopeType.fromDescription(parts[2]) ?: return failure(
+                IllegalArgumentException("Invalid type: $filename")
+            )
+            val uuid = parts[3].removeSuffix(".json")
+            return Result.success(StoredTelemetryMetadata(timestamp, uuid, type, filename))
+        }
+    }
+}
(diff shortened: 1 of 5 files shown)


embrace-android-delivery/src/main/kotlin/io/embrace/android/embracesdk/internal/delivery/SupportedEnvelopeType.kt
```diff
@@ -0,0 +1,28 @@
+package io.embrace.android.embracesdk.internal.delivery
+
+import io.embrace.android.embracesdk.internal.payload.Envelope
+import java.lang.reflect.Type
+
+/**
+ * Enumerates the different types of telemetry that are supported when persisting to disk.
+ */
+enum class SupportedEnvelopeType(
```
bidetofevil, line 9: We only have two types of payloads that we wrap in our envelope - `SessionPayload` and `LogPayload`. The latter can contain a different mixture of logs, but they have the same shape. What is different is how each is prioritized and whether they are batched (and batching isn't really a concern of the delivery layer). 

So instead of having different types based on content, perhaps we should divide these along the lines of the actual payload type, and the priority of delivery? Right now, we have Crash, Log, and Network all being essentially LogPayloads with different priorities - but if we have another one later, say InternalErrors, I don't think we need another enum value if we just had various priorities, e.g. 

```
enum class DeliveryPriority(
  payloadType: Class
) {
  HIGHEST(LogPayload::class.java),
  HIGH(SessionPayload::class.java),
  NORMAL(LogPayload::class.java),
  LOW(LogPayload::class.java),
}
```

I wouldn't mind separating the payload type and the priority concepts all together as well, but not mapping one to one to the "types of telemetry in the thing" is what I'm going for.

embrace-android-delivery/src/main/kotlin/io/embrace/android/embracesdk/internal/delivery/StoredTelemetryMetadata.kt
```diff
@@ -0,0 +1,38 @@
+package io.embrace.android.embracesdk.internal.delivery
+
+import kotlin.Result.Companion.failure
+
+/**
+ * Metadata about the telemetry payload.
+ *
+ * This information is encoded in the filename.
+ */
+class StoredTelemetryMetadata(
```
bidetofevil, line 10: For future proofing, do we want to leave room to add additional tokens, or put in some sort of thing that allows us to version?

bidetofevil (approved):
Couple of comments, but feel free to merge first and tweak later when we iterate.

embrace-android-delivery/src/main/kotlin/io/embrace/android/embracesdk/internal/delivery/StoredTelemetryMetadata.kt
fractalwrench: Prefixing the filename with `v1` feels like a sensible addition. I don't think it makes sense to add additional tokens now as we won't know what the format will look like, but my thinking is that if the version prefix can be renamed then we can parse the filename differently without too much hassle.

embrace-android-delivery/src/main/kotlin/io/embrace/android/embracesdk/internal/delivery/SupportedEnvelopeType.kt
fractalwrench: Is there a difference in that approach from what already exists other than semantics? I'll merge for now & we can discuss it in further detail

embrace-android-delivery/src/main/kotlin/io/embrace/android/embracesdk/internal/delivery/StoredTelemetryMetadata.kt
bidetofevil: If it's all the same, I'd prefer a post-fix, so "${timestamp}_${type.description}_${uuid}_v1.json"?

embrace-android-delivery/src/main/kotlin/io/embrace/android/embracesdk/internal/delivery/SupportedEnvelopeType.kt
bidetofevil: Yeah I'll ping you. The idea is that if we have 10 variations of payloads with the same two structures, we won't have have 10 different constants to know how to sort between.

PR #1397, opened 2024-09-19
Implement prioritisation of stored telemetry

fractalwrench:
## Goal

Implements rules for how stored telemetry should be prioritised. This will be used when sending telemetry, but IMO we should also use this when persisting data to avoid the queue getting saturated. In both scenarios this should be possible by supplying a comparator to the executor in `BackgroundWorker` - I'll make this change in a follow-up PR.

## Testing

Added unit tests.


bidetofevil (approved):
Do we need to synchronize this prioritization with how we persist the payloads? Or are we thinking we can do a full read every time and have logic on our side to sort after obtaining the files?>

fractalwrench:
@bidetofevil I don't follow entirely, but my thought process right now is:

1. When persisting to disk, submit each payload to an executor with a priority queue using this comparator
2. When reading from disk, read the entire directory & sort with an executor in the same way
3. If files are added/deleted when we're processing them, read the entire directory after X seconds have passed to get up-to-date

PR #1398, opened 2024-09-19
Rename workers and alter queue for some tasks

fractalwrench:
## Goal

Renames `WorkerName` in preparation for a future changeset that will separate the background/scheduled workers into a sealed class hierarchy. This will be necessary as we want to supply a comparator to the priority queue in the executor implementation, but that wouldn't make sense for `ScheduledWorker`.

In this changeset I altered the case of the constants & also renamed several to better reflect what they are doing.

Additionally I took tasks submitted to the background registration worker that performed I/O & put them on a separate queue.

## Testing


bidetofevil (approved):
Fantastic. We should've done this a while ago

PR #1399, opened 2024-09-19
Avoid using direct executor

fractalwrench:
## Goal

Replaces `MoreExecutors` use in tests with `fakeBackgroundWorker`.


bidetofevil (approved):
Perfect. Can we get rid of all guava usage, you think?

bidetofevil:
Ohhhhh I think we can :-)

PR #1400, opened 2024-09-19
Extract prioritised executor

fractalwrench:
## Goal

Extracts a `PrioritizedWorker` class that is capable of setting a priority, and removes the possibility to set `TaskPriority` on `BackgroundWorker` as the field was ignored for every worker apart from the network request queue.

This is an intermediate stage in the refactor - I plan on combining `BackgroundWorker` and `ScheduledWorker` next as they now effectively fulfil the same function. 

## Testing


bidetofevil (approved):
LGTM. And +1 to combining the workers. They are backed by the same executor type, so we are really just hiding some additional functionality from one and not the other, which really isn't worth all the trouble.

PR #1401, opened 2024-09-19
Replace worker implementation

fractalwrench:
## Goal

Alters `BackgroundWorker` to use `ScheduledExecutorService` and updates tests also. This should not functionally impact anything as it was always used under the hood anyway. This will allow combining BackgroundWorker/ScheduledWorker in future PRs.


bidetofevil (approved):
LGTM

PR #1402, opened 2024-09-20
Fix integration test infra for validating saved background activities

bidetofevil:
## Goal

Background activities are saved in a background thread, triggered explicitly by certain public API method calls. In order to control the order of execution and taking care of both the case when the background thread save runs before or after the validation code hits on the main test thread, I added `checkNextSavedBackgroundActivity` which takes in a lambda that will trigger an action that in turn triggers a save, as well as one that runs some validation code on the NEXT background activity that is saved. 

Along the way, I tweaked the implementation of `returnIfConditionMet` to separate the data being used to check the condition that is being continuously checked. TBH, I'm not sure if that's really needed - the initial failures were due to the background save not being triggered because there wasn't enough time that have lapsed before the previous save. But this is working and I don't want to change it back again, so lets go with this to see if the races are addresed.

### DIFF
--- a/embrace-android-sdk/src/integrationTest/kotlin/io/embrace/android/embracesdk/IntegrationTestRuleExtensions.kt
+++ b/embrace-android-sdk/src/integrationTest/kotlin/io/embrace/android/embracesdk/IntegrationTestRuleExtensions.kt
@@ -1,6 +1,7 @@
 package io.embrace.android.embracesdk
 
 import android.app.Activity
+import io.embrace.android.embracesdk.fakes.FakeDeliveryService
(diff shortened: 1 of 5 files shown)


bidetofevil:
* **#1402** <a href="https://app.graphite.dev/github/pr/embrace-io/embrace-android-sdk/1402?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

bidetofevil:
### Merge activity

* **Sep 20, 11:48 AM EDT**: @bidetofevil merged this pull request with [Graphite](https://app.graphite.dev/github/pr/embrace-io/embrace-android-sdk/1402).

PR #1403, opened 2024-09-20
Consolidate background + scheduled workers

fractalwrench:
## Goal

Consolidates the `ScheduledWorker` class into `BackgroundWorker` as they now fulfil the same function.


bidetofevil (approved):
LGTM

PR #1404, opened 2024-09-20
Alter hierarchy of `Worker`

fractalwrench:
## Goal

Turns the `Worker` enum into a sealed class instead

## Testing


bidetofevil (approved):
LGTM

PR #1405, opened 2024-09-20
Allow priority based execution using custom comparators

fractalwrench:
## Goal

Adapts the existing `PriorityThreadPoolExecutor` so that instead of only supporting the `TaskPriority` enum, it supports arbitrary data types with a matching comparator. I've updated the implementation so that network requests are prioritised based on whether `ApiRequest` is a session, and so that the `DeliveryCacheManager` prioritises based on the supplied `TaskPriority`.

## Testing

Updated & added to unit tests.


bidetofevil (approved):
LGTM

bidetofevil (approved):
LGTM. 

Too bad we lost the default parameter. Do you think it's worthwhile to add a specialized type of `PriorityWorker` that uses the enum so we can set that as having a default NORMAL priority when submitting/scheduling? Or will we have so many different types that are widely used, most job queuing will require an explicit priority set instead of relying on a default?

fractalwrench:
@bidetofevil I think it's likely we will refactor `TaskPriority` away in favour of prioritisation of specific tasks for a given queue. Right now `TaskPriority` is only used to prioritise file caching

bidetofevil:
Ah, got.

PR #1407, opened 2024-09-20
Alter project dependency structure

fractalwrench:
## Goal

Alters the dependencies in `embrace-android-delivery` so that the core module + test fakes can be accessed. Additionally defines the serialized type that should be used for log payloads.

### DIFF
--- a/embrace-android-delivery/build.gradle.kts
+++ b/embrace-android-delivery/build.gradle.kts
@@ -12,5 +12,5 @@ apiValidation.validationDisabled = true
 
 dependencies {
     implementation(project(":embrace-android-payload"))
-    testImplementation(project(":embrace-android-payload"))
+    implementation(project(":embrace-android-core"))
 }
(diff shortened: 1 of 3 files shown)


embrace-android-delivery/build.gradle.kts
```diff
@@ -12,5 +12,6 @@ apiValidation.validationDisabled = true
 
 dependencies {
     implementation(project(":embrace-android-payload"))
+    implementation(project(":embrace-android-core"))
     testImplementation(project(":embrace-android-payload"))
```
bidetofevil, line 16: Do we still need the `testImplementation` dep now that it's also s `implementation` dep?

bidetofevil (approved):
LGTM

embrace-android-delivery/build.gradle.kts
fractalwrench: Removed this as it's unnecessary

PR #1408, opened 2024-09-20
Fix manual session control enablement

nelsitoPuglisi:
## Goal

Have a kill switch to disable ending sessions manually. There was a confused implementation with the previous key.

## Testing

Manual and Unit test.

### DIFF
--- a/embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/config/behavior/SessionBehavior.kt
+++ b/embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/config/behavior/SessionBehavior.kt
@@ -24,9 +24,9 @@ interface SessionBehavior {
     fun isGatingFeatureEnabled(): Boolean
 
     /**
-     * Whether session control is enabled, meaning that features should be gated.
+     * Whether session control is disabled, meaning that can't end sessions manually.
      */
-    fun isSessionControlEnabled(): Boolean
+    fun isSessionControlDisabled(): Boolean
 
     /**
      * Returns the maximum number of properties that can be attached to a session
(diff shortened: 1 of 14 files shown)


embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/session/orchestrator/SessionOrchestratorImpl.kt
```diff
@@ -111,7 +111,7 @@ internal class SessionOrchestratorImpl(
                 payloadFactory.startSessionWithManual(timestamp)
             },
             earlyTerminationCondition = {
-                return@transitionState shouldEndManualSession(
+                return@transitionState !shouldEndManualSession(
```
nelsitoPuglisi, line 114: `early termination condition` collects functions that prevent the execution.

embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/config/behavior/SessionBehavior.kt
```diff
@@ -24,9 +24,9 @@ interface SessionBehavior {
     fun isGatingFeatureEnabled(): Boolean
 
     /**
-     * Whether session control is enabled, meaning that features should be gated.
+     * Whether session control is disabled, meaning that can't end sessions manually.
      */
-    fun isSessionControlEnabled(): Boolean
```
bidetofevil, line 29: We can keep this as enabled at this level (app usage) so it matches the other kill switches. It's at the Config level that we should make this be identify itself as disabled

bidetofevil (commented):
Looks good other than flipping the method name in the Behavior back to `isSessionControlEnabled()` to match the other kill switches.

@fractalwrench should look at the logic flip in `OrchestratorImpl`. Maybe a method rename to `blockManualSessionEnd()` while keep the true/false returns might read better?

PR #1409, opened 2024-09-20
Alter assertion of flaky test

fractalwrench:
## Goal

The test case assumed that the job would not complete after the latch was released, but internally `BackgroundWorker` calls `shutdown` which allows completion of tasks that are already executing.

My fix is to increase the number of latches required to 2 so the `Runnable` never completes. I've added an assertion to check the latch was counted down instead.


bidetofevil (approved):
LGTM

bidetofevil:
### Merge activity

* **Sep 20, 12:30 PM EDT**: [Graphite](https://app.graphite.dev/github/pr/embrace-io/embrace-android-sdk/1409) rebased this pull request as part of a merge.

PR #1414, opened 2024-09-23
Create initial interfaces for delivery layer

fractalwrench:
## Goal

Adds the initial skeletons of the delivery layer interfaces. I've implemented a fake + empty implementation for each interface & added constructor parameters where one service depends on another.

I imagine there will be some iteration on the function signatures but hopefully having something to play around with will make the project easier to understand at a glance.


embrace-android-delivery/src/main/kotlin/io/embrace/android/embracesdk/internal/delivery/scheduling/SchedulingService.kt
```diff
@@ -0,0 +1,20 @@
...
+
+/**
+ * This service is responsible for scheduling HTTP requests to the Embrace backend.
+ */
+interface SchedulingService : CrashTeardownHandler {
+
+    /**
+     * Start scheduling requests.
+     */
+    fun startScheduling()
```
bidetofevil, line 14: I don't think we need this to be external to the service. The logic to manage any checks for retries or network state change should be internal to the service, so `onPayloadIntake()` should be sufficient for external services to notify this bad boy to do its thing.

bidetofevil (approved):
LGTM. `SchedudlingService.startScheduling` seems superfluous to me, but feel free to merge if it's easier to do so and I can remove it once I add my changes

bidetofevil:
Merged this because the CI failure is due to a flakey test that has already been fixed but not merged yet.

PR #1415, opened 2024-09-23
Implement dependency injection for delivery module

fractalwrench:
## Goal

Creates the boilerplate code required to instantiate objects in the delivery module.


bidetofevil (approved):
LGTM

PR #1416, opened 2024-09-23
Write assertions for delivery layer

fractalwrench:
## Goal

Writes some initial assertions that can be used to verify behavior in the delivery layer.


bidetofevil (approved):
LGTM

PR #1417, opened 2024-09-23
Make delivery services optional

fractalwrench:
## Goal

Makes the `DeliveryModule2` return null if the SDK is configured to only send data to OTel exporters. This avoids unnecessary work.

## Testing

Added unit test coverage.

### DIFF
--- a/embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/config/ConfigService.kt
+++ b/embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/config/ConfigService.kt
@@ -112,6 +112,13 @@ interface ConfigService {
     val appId: String?
 
     /**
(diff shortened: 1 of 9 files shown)


bidetofevil (approved):
What do you think about providing an no-op implementation of the delivery service when we aren't using it rather than make it nullable?

fractalwrench:
@bidetofevil my general preference is for null rather than no-op as it marginally reduces app size impact when rolled out across as a pattern across the entire SDK, but that's a weak preference

bidetofevil:
We should probably talk about this in a guild meeting. My preference is also marginal, but that's likely due to the Java background and wanting to avoid nulls like the plague. In Kotlin, nulls are so much more explicit, so maybe having nulls can be use to signify optionality and is totally fine?

PR #1419, opened 2024-09-24
Alter filename schema

fractalwrench:
## Goal

Alters the filename to store the version as a suffix rather than a prefix.


bidetofevil (approved):
LGTM

PR #1420, opened 2024-09-24
Initial implementation of intake service

fractalwrench:
## Goal

Creates an initial implementation of `IntakeService` which stores everything in the existing `CacheService`. This changeset attempts to fulfil the following requirements of `IntakeService`:

- On a crash the SDK will block until pending tasks have finished persisting data
- Crashes/sessions/logs/network capture are prioritised for persistence in order of importance
- Payloads are stored gzip compressed
- The scheduling service is notified after a payload is persisted
- An internal error is logged if persistence fails for whatever reason

I have not implemented cleanup if disk usage exceeds a maximum limit.

I've wrapped `CacheService` with a `PayloadStorageService` interface. My initial thought is that in future PRs we probably want to extract the functionality we want out of `CacheService`, but I'm open to discussion on that. IMO it would be good if `PayloadStorageService` also handled most of the file operations that will be triggered from `SchedulingService`, as that should simplify our synchronisation strategy.

The next task on my list is to figure out the file structure in which we want to store data on disk.

## Testing

Added unit tests to match these requirements. We will add higher level integration tests later on.


bidetofevil (approved):
LGTM

bidetofevil (commented):
Per our discussion, expanding `PayloadStorageService` to be the new CacheSevice + StorageService is a good idea. I'll update the doc - I had mixed this up with the PayloadCachingService in an earlier update, and renamed the other thing LogCachingService, but let me tweak that now

PR #1421, opened 2024-09-24
Alter stored telemetry comparator

fractalwrench:
## Goal

Refactors the comparator that will be used to prioritise telemetry to match the pattern used in `taskPriorityComparator` and others.


bidetofevil (approved):
LGTM

PR #1422, opened 2024-09-24
Add boilerplate for intake service

fractalwrench:
## Goal

Adds the boilerplate of constructor parameters & dependency injection code that `IntakeService` will require. I've also added a dependency on OTel as it'll be necessary to access the `Clock` implementation.

### DIFF
--- a/embrace-android-delivery/build.gradle.kts
+++ b/embrace-android-delivery/build.gradle.kts
@@ -13,4 +13,10 @@ apiValidation.validationDisabled = true
 dependencies {
     implementation(project(":embrace-android-payload"))
     implementation(project(":embrace-android-core"))
+    compileOnly(platform(libs.opentelemetry.bom))
+    compileOnly(libs.opentelemetry.api)
+    compileOnly(libs.opentelemetry.sdk)
+    testImplementation(platform(libs.opentelemetry.bom))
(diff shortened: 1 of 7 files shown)


embrace-android-delivery/build.gradle.kts
```diff
@@ -13,4 +13,10 @@ apiValidation.validationDisabled = true
 dependencies {
     implementation(project(":embrace-android-payload"))
     implementation(project(":embrace-android-core"))
+    compileOnly(platform(libs.opentelemetry.bom))
```
bidetofevil, line 16: So the implementation of the OTel clock that we use effectively points to the internal implementation of our own clock. I was pretty loosey goosey with which clock I reference because everything was in one module, but if it matters, we can clean that up so we don't need OTel in the module.

Perhaps one of us can do this after this is mreged?

bidetofevil (approved):
LGTM. Suggestion on how we might try to get rid of the OTel dependency at this layer, but that can be handled later.

embrace-android-delivery/build.gradle.kts
fractalwrench: I'll clean this up after merging. It looks like we only need the class because `FakeInitModule` is referencing it so it's just an issue with our tests

PR #1423, opened 2024-09-24
Refactor telemetry comparator

fractalwrench:
## Goal

Makes `storedTelemetryComparator` work on `StoredTelemetryMetadata` & makes some classes visible from the core module that will be required in future PRs.


bidetofevil (approved):
LGTM

PR #1424, opened 2024-09-24
Avoid otel dep in delivery module

fractalwrench:
## Goal

Avoids a runtime crash in our tests due to classloading issues. I've just delayed the classloading for the `openTelemetryModule` because we'll never use it from the delivery tests.


bidetofevil (approved):
LGTM

PR #1427, opened 2024-09-25
Payload storage service implementation

fractalwrench:
## Goal

Creates an initial implementation of `PayloadStorageService`. Ultimately this will fulfil a similar function that `CacheService` did in the old approach, except right now it only persists final payloads (and not cache attempts).

I added functions to store, delete, and load a payload. Payloads are stored using gzip compression & there is a total limit on the number of files stored on disk, after which less important payloads (assessed by type + age) are deleted.

I also altered the atomic file write logic somewhat compared to the `EmbraceCacheService` implementation as there isn't any possibility these files will be overwritten. 

## Discussion

1. If we detect an exception that indicates disk space has run out, should we attempt to delete old/less important payloads in favour of the new one?
2.  Is the storage limit sensible?
3. `pruneFilesIfNeeded` calls `listFiles()` every time - is it worth maintaining an in-memory list of files to avoid this?
4. How should we capture telemetry about files that are dropped in `pruneFilesIfNeeded`, if at all?
5. The `EmbraceCacheService` was falling back to the cache directory if the file directory wasn't writable. I haven't copied that logic here. Is that something that could bite us - i.e. should I include a fallback option in this PR?
6. I haven't implemented a synchronisation strategy yet - I can see value in synchronising `store()`, but I'm less certain it makes sense to add any synchronisation for other functions. These will be called from other threads that will have no way of knowing when a payload might be deleted (i.e. if the storage limit is reached)

## Testing

Added new unit tests.

### DIFF
--- a/embrace-android-delivery/src/main/kotlin/io/embrace/android/embracesdk/internal/delivery/storage/PayloadStorageServiceImpl.kt
+++ b/embrace-android-delivery/src/main/kotlin/io/embrace/android/embracesdk/internal/delivery/storage/PayloadStorageServiceImpl.kt
@@ -1,13 +1,82 @@
 package io.embrace.android.embracesdk.internal.delivery.storage
 
-import io.embrace.android.embracesdk.internal.comms.delivery.CacheService
+import io.embrace.android.embracesdk.internal.delivery.StoredTelemetryMetadata
+import io.embrace.android.embracesdk.internal.delivery.storedTelemetryComparator
 import io.embrace.android.embracesdk.internal.injection.SerializationAction
+import io.embrace.android.embracesdk.internal.telemetry.errors.InternalErrorService
+import java.io.File
+import java.io.InputStream
+import java.util.zip.GZIPOutputStream
 
 internal class PayloadStorageServiceImpl(
-    private val cacheService: CacheService
+    private val internalErrorService: InternalErrorService,
+    outputDir: Lazy<File>,
+    private val storageLimit: Int = 500 // TODO: check limit
 ) : PayloadStorageService {
 
-    override fun store(filename: String, action: SerializationAction) {
-        cacheService.cachePayload(filename, action)
+    private val payloadDir by outputDir
+
+    // TODO: synchronisation
+
+    override fun store(metadata: StoredTelemetryMetadata, action: SerializationAction) {
+        try {
+            storeImpl(metadata, action)
+        } catch (exc: Throwable) {
+            internalErrorService.handleInternalError(exc)
+        }
+    }
+
+    private fun storeImpl(
+        metadata: StoredTelemetryMetadata,
+        action: SerializationAction
+    ) {
+        pruneIfNeeded()
+
+        // write to a temporary file then rename it, to avoid sending incomplete files
+        // to the backend (i.e. where the process terminates or there isn't any disk space).
+        val tmpFile = File.createTempFile(metadata.filename, ".tmp")
+        GZIPOutputStream(tmpFile.outputStream().buffered()).use { stream ->
+            action(stream)
+        }
+
+        // move the complete file to its final location.
+        metadata.asFile().parentFile?.mkdirs()
+        tmpFile.renameTo(metadata.asFile())
+    }
+
+    override fun delete(metadata: StoredTelemetryMetadata) {
+        try {
+            metadata.asFile().delete()
+        } catch (exc: Throwable) {
+            internalErrorService.handleInternalError(exc)
+        }
+    }
+
+    override fun loadPayloadAsStream(metadata: StoredTelemetryMetadata): InputStream? {
+        return try {
+            metadata.asFile().inputStream().buffered()
+        } catch (exc: Throwable) {
+            internalErrorService.handleInternalError(exc)
+            null
+        }
+    }
+
+    private fun pruneIfNeeded() {
+        // TODO: future: avoid calling listFiles() every time by retaining an in-memory list
+        val files = payloadDir.listFiles() ?: return
+        val metadataList = files.mapNotNull {
+            StoredTelemetryMetadata.fromFilename(it.name).getOrNull()
+        }.sortedWith(storedTelemetryComparator)
+
+        if (metadataList.size >= storageLimit) {
+            val excess = metadataList.size - storageLimit + 1
+            metadataList.takeLast(excess).forEach { it.asFile().delete() }
+        }
+    }
+
+    private fun StoredTelemetryMetadata.asFile(): File = File(payloadDir, filename)
+
+    companion object {
+        internal const val OUTPUT_DIR_NAME = "embrace_payloads"
     }
 }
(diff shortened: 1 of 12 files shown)


embrace-android-delivery/src/test/kotlin/io/embrace/android/embracesdk/internal/delivery/storage/PayloadStorageServiceImplTest.kt
```diff
@@ -0,0 +1,116 @@
...
+        val files = outputDir.listFiles() ?: error("No files found")
+        val outputs = files.map { StoredTelemetryMetadata.fromFilename(it.name).getOrThrow() }
+            .sortedWith(storedTelemetryComparator)
+            .map { Pair(it.timestamp, it.envelopeType) }
+
+        val expected = listOf(
+            Pair(0L, CRASH),
+            Pair(1000L, CRASH),
+            Pair(0L, SESSION),
+            Pair(1000L, NETWORK)
```
fractalwrench, line 112: I'm not convinced this is the right behavior - we should probably prioritise the `Pair(1000L, SESSION)` right, even if it means dropping the in-memory data?

embrace-android-delivery/src/main/kotlin/io/embrace/android/embracesdk/internal/delivery/storage/PayloadStorageServiceImpl.kt
```diff
@@ -1,13 +1,82 @@
...
+
+    override fun store(metadata: StoredTelemetryMetadata, action: SerializationAction) {
+        try {
+            storeImpl(metadata, action)
+        } catch (exc: Throwable) {
+            internalErrorService.handleInternalError(exc)
+        }
+    }
+
+    private fun storeImpl(
```
bidetofevil, line 29: For files that can be overwritten, are you planning to reuse the existing EmbraceCacheService or port them over here too?

embrace-android-delivery/src/main/kotlin/io/embrace/android/embracesdk/internal/delivery/storage/PayloadStorageServiceImpl.kt
```diff
@@ -1,13 +1,82 @@
...
+import io.embrace.android.embracesdk.internal.telemetry.errors.InternalErrorService
+import java.io.File
+import java.io.InputStream
+import java.util.zip.GZIPOutputStream
 
 internal class PayloadStorageServiceImpl(
-    private val cacheService: CacheService
+    private val internalErrorService: InternalErrorService,
+    outputDir: Lazy<File>,
+    private val storageLimit: Int = 500 // TODO: check limit
```
bidetofevil, line 14: Seems fine to me as a start

embrace-android-delivery/src/main/kotlin/io/embrace/android/embracesdk/internal/delivery/storage/PayloadStorageServiceImpl.kt
```diff
@@ -1,13 +1,82 @@
...
+    }
+
+    private fun pruneIfNeeded() {
+        // TODO: future: avoid calling listFiles() every time by retaining an in-memory list
+        val files = payloadDir.listFiles() ?: return
+        val metadataList = files.mapNotNull {
+            StoredTelemetryMetadata.fromFilename(it.name).getOrNull()
+        }.sortedWith(storedTelemetryComparator)
+
+        if (metadataList.size >= storageLimit) {
```
bidetofevil, line 71: I wonder if we should trim more, especially if we are just getting rid of lower priority payloads? This avoids the constant thrashing of deleting one each time.

If that's too complicated for now, maybe just trim 20% of the max each time or something?

embrace-android-delivery/src/test/kotlin/io/embrace/android/embracesdk/internal/delivery/storage/PayloadStorageServiceImplTest.kt
bidetofevil: Yeah. If we have 500 sessions cached, there's no point trying to delivery a network capture.

That said, if a cache gets clogged up like that, we should try to notify ourselves that this is happening. If it's a lack of network, fair enough, but be should be able to side channel internal errors so they don't get lost if something bloats the cache like that.

bidetofevil (commented):
> If we detect an exception that indicates disk space has run out, should we attempt to delete old/less important payloads in favour of the new one?

I don't think so. Older sessions are more important than new logs - that's the implication of not worrying about starvation, right? The only thing we should perhaps send over high priority payloads is a message that indicates that the cache has filled up, so we will be dropping data. That is an indication of something weird going on, unless it's just offline for an extended period.

> Is the storage limit sensible?

500 seems fine. If we are talking about gzipped payloads, I can't see this being more than 20-30MB, which isn't that much in today's world, even for crap phones. That could literally be 2 unoptimized JPGs.

> pruneFilesIfNeeded calls listFiles() every time - is it worth maintaining an in-memory list of files to avoid this?

Yeah. Unless we are worried about the in memory list being out of sync with the files on disk, which I think we an protect against. We can periodically make a file system read to ensure that they aren't out of sync, but even if they are, any payloads missing from the in memory index would be delivered the next time the app starts up, and missing files will need to be dealt with anyway in case it somehow gets deleted before we get to read it.

> How should we capture telemetry about files that are dropped in pruneFilesIfNeeded, if at all?

I think we should report when we have to prune - save and report it as an internal error. Or somehow mark the session as having required pruning. I wouldn't go too far into capturing exact loss - not until we know this happens frequently enough that we want to have more data. 500 payloads is a lot, and unless apps are offline for an extended period of time, I don't see us hitting it.

And if that is the case, we'd probably want a better mechanism to deal with this - if phones are expected to be offline often, maybe we should increase the size of the cache - or allow that to be configurable somehow.

> The EmbraceCacheService was falling back to the cache directory if the file directory wasn't writable. I haven't copied that logic here. Is that something that could bite us - i.e. should I include a fallback option in this PR?

I would include a fallback option just so we don't cause any regressions, but I would want some data regarding if this ever happens. Are there different permissions to write to the files dir and the caches directory?

If anything, I'd want a fallback for when we can't cache data at all, and we have to bypass this logic entirely. Again, I don't want to do this unless we know it happens, so lets do the fallback, and report an internal error whenever we fail to write to the files directory and have to use it.

> I haven't implemented a synchronisation strategy yet - I can see value in synchronising store(), but I'm less certain it makes sense to add any synchronisation for other functions. These will be called from other threads that will have no way of knowing when a payload might be deleted (i.e. if the storage limit is reached)

As long as we don't ever return partial payloads when load, I'm ok if we don't synchronize on read - though all operations should be thread safe.

In terms of writes, I'm wondering if we can get away with not as well. If we never have two threads writing to the same file name, we should never need it? If we are storing the cached payloads separately, and we can better coordinate the writes of the periodic caching job and the synchronous session termination on the main thread, we might be able to get away with it?

bidetofevil (approved):
LGTM. I answered your discussion questions and had a thought about the pruning strategy. Lets merge this so I can build on top of it :-)

embrace-android-delivery/src/main/kotlin/io/embrace/android/embracesdk/internal/delivery/storage/PayloadStorageServiceImpl.kt
fractalwrench: I was initially planning on moving caching to the `PayloadCachingService` & have the resurrection service send them to the intake service.

PR #1428, opened 2024-09-26
Tweak payload storage implementation

fractalwrench:
## Goal

Makes some of the tweaks to the payload storage implementation that were discussed in #1427. Specifically:

- Create an in-memory list of files
- Fix prioritisation of stored telemetry deletion

## Testing

Mostly relied on existing unit tests.

### DIFF
--- a/embrace-android-delivery/src/main/kotlin/io/embrace/android/embracesdk/internal/delivery/storage/PayloadStorageServiceImpl.kt
+++ b/embrace-android-delivery/src/main/kotlin/io/embrace/android/embracesdk/internal/delivery/storage/PayloadStorageServiceImpl.kt
@@ -1,22 +1,37 @@
 package io.embrace.android.embracesdk.internal.delivery.storage
 
+import android.content.Context
 import io.embrace.android.embracesdk.internal.delivery.StoredTelemetryMetadata
 import io.embrace.android.embracesdk.internal.delivery.storedTelemetryComparator
 import io.embrace.android.embracesdk.internal.injection.SerializationAction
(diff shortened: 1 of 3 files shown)


embrace-android-delivery/src/main/kotlin/io/embrace/android/embracesdk/internal/delivery/storage/PayloadStorageServiceImpl.kt
```diff
@@ -61,22 +84,42 @@ internal class PayloadStorageServiceImpl(
...
+        val input = storedFiles.plus(metadata)
+        val removalCount = input.size - storageLimit
+        if (removalCount < 0) {
+            return false
         }
+        val removals = input
+            .sortedWith(storedTelemetryComparator)
+            .takeLast(removalCount)
+        removals.forEach(::delete)
+        internalErrorService.handleInternalError(RuntimeException("Pruned payload storage"))
```
bidetofevil, line 101: The semantics of this feel weird since it's not an error 😅 

Maybe we need to rename this later so we can relieve (my) cognitive dissonance

bidetofevil (approved):
LGTM. We can merge this as it is and sort out the pruning scheme in the slack thread

embrace-android-delivery/src/main/kotlin/io/embrace/android/embracesdk/internal/delivery/storage/PayloadStorageServiceImpl.kt
fractalwrench: Agree - it feels strange

PR #1429, opened 2024-09-26
Document synchronisation strategy for PayloadStorageService

fractalwrench:
## Goal

Adds some documentation for the `PayloadStorageService` synchronisation strategy.

@bidetofevil for review it'd be helpful to (1) let me know if you agree this is a reasonable strategy & (2) validate that there aren't any footguns in the actual implementation that I've missed. I'd also specifically be interested in your thoughts on the approach I've taken for `loadPayloadAsStream`.

### DIFF
--- a/embrace-android-delivery/src/main/kotlin/io/embrace/android/embracesdk/internal/delivery/storage/PayloadStorageService.kt
+++ b/embrace-android-delivery/src/main/kotlin/io/embrace/android/embracesdk/internal/delivery/storage/PayloadStorageService.kt
@@ -5,7 +5,18 @@ import io.embrace.android.embracesdk.internal.injection.SerializationAction
 import java.io.InputStream
 
 /**
- * Stores a completed payload to disk.
+ * Stores a completed payload to disk. This service makes several assumptions around threading
+ * that MUST be adhered to:
+ *
+ * 1. All calls to [store] are made from thread A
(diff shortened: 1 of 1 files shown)


embrace-android-delivery/src/main/kotlin/io/embrace/android/embracesdk/internal/delivery/storage/PayloadStorageService.kt
```diff
@@ -5,7 +5,17 @@ import io.embrace.android.embracesdk.internal.injection.SerializationAction
 import java.io.InputStream
 
 /**
- * Stores a completed payload to disk.
+ * Stores a completed payload to disk. This service makes several assumptions around threading
+ * that MUST be adhered to:
+ *
+ * 1. All calls to [store] are made from thread A
+ * 2. All (external) calls to [delete] are made from thread B
```
bidetofevil, line 12: What do you think about serializing all writes on Thread A, namely delete calls will be async and return right away, the method will queue a task on thread A to do the actual delete? This simplifies synchronization (there need not be any) as it will always be done on the same thread. Any issues you see with that?

fractalwrench: As discussed on our call I think this is a good idea & we should go with it

bidetofevil (approved):

PR #1430, opened 2024-09-26
Add kill switch for v2 storage

fractalwrench:
## Goal

Adds a kill switch in remote config that can be used to disable the v2 storage system if so required.

### DIFF
--- a/embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/config/behavior/AutoDataCaptureBehavior.kt
+++ b/embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/config/behavior/AutoDataCaptureBehavior.kt
@@ -56,4 +56,9 @@ interface AutoDataCaptureBehavior {
      * for apps with a lot of local files.
      */
     fun isDiskUsageCaptureEnabled(): Boolean
+
+    /**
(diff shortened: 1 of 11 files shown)


embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/config/behavior/AutoDataCaptureBehavior.kt
```diff
@@ -56,4 +56,9 @@ interface AutoDataCaptureBehavior {
      * for apps with a lot of local files.
      */
     fun isDiskUsageCaptureEnabled(): Boolean
+
+    /**
+     * Gates whether the SDK should use the v2 storage implementation or the legacy one.
+     */
+    fun isV2StorageEnabled(): Boolean
```
bidetofevil, line 63: I assume this will control the delivery layer too?

bidetofevil (approved):

embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/config/behavior/AutoDataCaptureBehavior.kt
fractalwrench: Yep

PR #1431, opened 2024-09-26
Create shim for session payload storage

fractalwrench:
## Goal

Creates a shim that will be used to switch between the v1/v2 storage systems. I've implemented this as an interface that will have 2 implementations that can be gated by the remote config added in #1430.

## Testing


bidetofevil (approved):
LGTM

PR #1433, opened 2024-09-27
Log payload storage

fractalwrench:
## Goal

Add log storage to the `PayloadStore` shim layer.

## Testing


bidetofevil (approved):
LGTM

PR #1436, opened 2024-09-27
Combine delivery modules

fractalwrench:
## Goal

Combines the 2 delivery modules given that they both live in the same module now.

## Testing


embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/injection/DeliveryModuleImpl.kt
```diff
@@ -1,26 +1,95 @@
...
                 storageModule.deliveryCacheManager,
-                apiService,
+                checkNotNull(essentialServiceModule.apiService),
                 initModule.jsonSerializer,
                 initModule.logger
             )
         }
     }
+
+    override val intakeService: IntakeService? by singleton {
```
bidetofevil, line 42: I prefer a no-op version but we can tidy that up later.

bidetofevil (approved):
LGTM

PR #1437, opened 2024-09-27
Run code through payload store

fractalwrench:
## Goal

Alters the code so that the `V1PayloadStore` is constructed within a module, and alters some assertions in the tests to run against `PayloadStore`.

## Testing


bidetofevil (approved):
LGTM

PR #1438, opened 2024-09-27
Alter integration assertions

fractalwrench:
## Goal

Improves the assertions used in the integration tests by making them wait if a payload hasn't been delivered yet. This will help when migrating these tests over to use the v2 storage layer. I attempted to make some of the assertions more specific whilst in this area.


bidetofevil (approved):
Very very nice!

PR #1443, opened 2024-09-30
Move test cases to one directory

fractalwrench:
## Goal

Moves the integration tests to under one package to make it more obvious where they all live.


bidetofevil (approved):
LGTM

PR #1444, opened 2024-09-30
[POC} alternative syntax for integration tests

fractalwrench:
## Goal

A proposal for an alternative syntax used for writing integration tests. This DSL allows 3 actions to be declared: for configuring the SDK, invoking a function on the SDK to trigger the behavior, and then asserting the results. I believe this offers a few benefits:

1. It will make longer test cases easier to read
2. When a test fails it should be more obvious at what stage of the test case it failed
3. The `Harness` & `Bootstrapper` class are directly used in lots of places where they don't need to be. This syntax opens up the door to reducing that

### DIFF
--- a/embrace-android-sdk/src/integrationTest/kotlin/io/embrace/android/embracesdk/IntegrationTestRule.kt
+++ b/embrace-android-sdk/src/integrationTest/kotlin/io/embrace/android/embracesdk/IntegrationTestRule.kt
@@ -133,6 +133,16 @@ internal class IntegrationTestRule(
         Embrace.getImpl().stop()
     }
 
+    fun runTest(
+        setupAction: Harness.() -> Unit,
+        testCaseAction: Harness.() -> Unit,
+        assertAction: Harness.() -> Unit,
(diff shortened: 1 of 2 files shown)


embrace-android-sdk/src/integrationTest/kotlin/io/embrace/android/embracesdk/IntegrationTestRule.kt
```diff
@@ -133,6 +133,16 @@ internal class IntegrationTestRule(
         Embrace.getImpl().stop()
     }
 
+    fun runTest(
+        setupAction: Harness.() -> Unit,
+        testCaseAction: Harness.() -> Unit,
+        assertAction: Harness.() -> Unit,
```
fractalwrench, line 139: Right now these all return `Harness`, but I think in future it would be better for them to return more specific types. Setup action should act on a type that alters configuration, the test case action should act on Embrace/a shim of that interface, and assert action should act on a type from the delivery layer

PR #1445, opened 2024-09-30
Update integration tests to use payload store

fractalwrench:
## Goal

Updates the integration tests to grab the payload from the `PayloadStore` rather than the `DeliveryService` (where applicable). Ultimately the tests should grab the payloads from a fake `RequestExecutionService` but that's not in place just yet, so this will act as an initial refactor.


bidetofevil (approved):
LGTM

PR #1446, opened 2024-10-01
Use noop services in delivery layer

fractalwrench:
## Goal

Uses no-op services in the delivery layer to avoid null checks everywhere.

## Testing

Added unit test coverage.


bidetofevil (approved):
If you prefer nulls we can do that too 😅 Like I said before, I think it was just old ways of thinking that is making shy away from it. Let me as Fran today what he prefers

PR #1447, opened 2024-10-01
Hookup v2 delivery layer to dependency injection

fractalwrench:
## Goal

Hooks up the v2 delivery layer via the `SchedulingServiceImpl`, so that if the flag is flipped the functionality is enabled.

## Testing

Added unit tests.


bidetofevil (approved):
LGTM

PR #1448, opened 2024-10-01
Update integration test case syntax

fractalwrench:
## Goal

Updates the syntax used in the integration test cases. This is a large diff but functionally not much has changed with the test cases - in the majority of cases this was just a change to the whitespace. In a few test cases I've untangled the actions + assertions.


bidetofevil (approved):
Skimmed it only, but we've talked about this structure and I like it.

PR #1449, opened 2024-10-01
Create action interface

fractalwrench:
## Goal

Creates an `EmbraceActionInterface` class where all actions performed on Embrace during tests should happen. I've updated the existing test cases to use the functions in the new location.


bidetofevil (approved):
Nice

PR #1450, opened 2024-10-01
Create assertion interface

fractalwrench:
## Goal

Creates a separate `EmbraceAssertionInterface` that contains functions to retrieve payloads for assertions during tests.


bidetofevil (approved):
LGTM

PR #1451, opened 2024-10-01
Reorganise integration test code

fractalwrench:
## Goal

Moves the integration test code so that it is either in the 'testcases' package or 'testframework'.


bidetofevil (approved):
LGTM

PR #1452, opened 2024-10-01
Improve payload assertions

fractalwrench:
## Goal

Continues reorganising assertions used in the integration tests & improves their readability.


bidetofevil (approved):
LGTM

PR #1453, opened 2024-10-01
Improve assertions for attributes

fractalwrench:
## Goal

Improves the assertions when verifying that attributes are included in a payload.


embrace-android-sdk/src/integrationTest/kotlin/io/embrace/android/embracesdk/testcases/SessionApiTest.kt
```diff
@@ -58,41 +57,37 @@ internal class SessionApiTest {
...
                 val attributeKeys = attrs.map { it.key }
                 validateExistenceOnly.forEach { key ->
                     attributeKeys.contains(key)
                 }
-
-                val attributesToCheck = attrs.filterNot {
-                    ignoredAttributes.contains(it.key)
-                }.toMap().toSortedMap()
-
-                val expected = mapOf(
```
bidetofevil, line 78: The original idea here is to ensure that no new attributes show up unexpectedly, so we explicitly ignore ones that may or may not be there because of race conditions. While what makes it more of a "gold file" diff, it also can get flakey. Do we want to preserve this or nah?

bidetofevil (approved):
One comment but lets merge this and we can figure out if that case is something we want to test for

embrace-android-sdk/src/integrationTest/kotlin/io/embrace/android/embracesdk/testcases/SessionApiTest.kt
fractalwrench: Hmm - I can see this still being useful. Perhaps we can add in an extra assertion after `assertMatches` that explicitly finds any new/missing attributes?

bidetofevil: What are you thinking in terms of syntax? Something like this?

```
fun assertMatches(allSpecified: Boolean, expected: MutableMap<String, Any>.() -> Unit) {
...
}
```

fractalwrench: I figured it's simplest just to leave it as-is for now: https://github.com/embrace-io/embrace-android-sdk/pull/1461

PR #1457, opened 2024-10-02
Inline test functions

fractalwrench:
## Goal

Inlines a couple of functions in test code to make the stacktraces more readable.


bidetofevil (approved):
LGTM

PR #1458, opened 2024-10-02
Reduce direct usage of bootstrapper

fractalwrench:
## Goal

Refactors the test cases to avoid directly peering into the internals of the SDK from test cases & deleted some unused assertions.


bidetofevil (approved):
Nice!

PR #1459, opened 2024-10-02
Reduce visibility of interfaces in integration tests

fractalwrench:
## Goal

Refactors the integration tests to use the new syntax & hides some properties to avoid direct access.


bidetofevil (approved):
LGTM

PR #1460, opened 2024-10-02
Delete obsolete integration tests

fractalwrench:
## Goal

Deletes integration tests that check if `null` is passed into our API. I don't believe this is a use-case that we want to support anymore given that the vast majority of Android code will be written in Kotlin, and given that a warning will show up in the IDE if a user does attempt to pass null.

This also gets rid of some of the last of our Java code (as only Java code can pass null into our API) & reduces a maintenance burden as the test is written somewhat differently to the rest of the integration tests.


bidetofevil (approved):
Cool. This level of safety is probably an overkill when it was put in, but now is definitely an overkill given our elevated expectations of SDK users.

PR #1461, opened 2024-10-03
Restore session api assertions

fractalwrench:
## Goal

Restores assertions in `SessionApiTest` that were altered in #1453.


bidetofevil (approved):
LGTM

PR #1462, opened 2024-10-03
Integration test refactor

fractalwrench:
## Goal

Refactors the integration tests in a couple of places to help make debugging easier. I made the following changes:

1. Decouple activity lifecycle simulation from starting a session
2. Rename `activityService` to `processStateService` & error if unbalanced calls are detected as this will be a problem with the test
3. Improve error messaging to state how many payloads were received if an unexpected amount were captured


bidetofevil (approved):
LGTM.

PR #1463, opened 2024-10-03
Throw if fake logger encounters internal error

fractalwrench:
## Goal

Alters `FakeEmbLogger` so that by default it throws when an internal error is encountered. A few test cases did so due to artefacts of how the test is setup so I suppressed it where necessary.


bidetofevil (approved):
LGTM

PR #1464, opened 2024-10-03
Delivery layer tweaks

fractalwrench:
## Goal

Makes the following tweaks to the delivery layer:

1. Temporarily calls into `DeliveryService` from `V2PayloadStore` when caching a payload
2. Enhances some of the test code


bidetofevil (approved):
LGTM

PR #1465, opened 2024-10-03
Allow using v2 delivery layer

fractalwrench:
## Goal

Hooks up the v2 delivery layer to the SDK behind a config field. As part of this I've allowed injecting the `RequestExecutionService` and `DeliveryService` from the integration tests, as that's what we'll use for assertions.


bidetofevil (approved):
LGTM

PR #1466, opened 2024-10-03
Use v2 delivery layer by default

fractalwrench:
## Goal

Updates the SDK to use the v2 delivery layer by default. This mostly involved flipping a switch & updating assertions in the integration tests.


bidetofevil (approved):
LGTM

PR #1469, opened 2024-10-04
Add post-setup step to integration tests that allows for almost all tests to automatically start the SDK

bidetofevil:
## Goal

Simplify integration test by adding a step to do something before setting up the test harness, and then only starting the SDK after that runs. This makes it not necessary to manually start the SDK, and also exposes a proper place to "do and verify stuff" before the SDK starts without manually needing to do so.

With this, `startImmediately` has been removed as a parameter of the harness, and `startSdk` was introduced as a parameter of `runTest` for those select tests that don't require the SDK to be started automatically.

### DIFF
--- a/embrace-android-sdk/src/integrationTest/kotlin/io/embrace/android/embracesdk/testframework/actions/EmbracePostSetupInterface.kt
+++ b/embrace-android-sdk/src/integrationTest/kotlin/io/embrace/android/embracesdk/testframework/actions/EmbracePostSetupInterface.kt
@@ -1,0 +1,13 @@
+package io.embrace.android.embracesdk.testframework.actions
+
+import io.embrace.android.embracesdk.Embrace
+import io.embrace.android.embracesdk.fakes.FakeClock
+
+internal class EmbracePostSetupInterface(
+    private val setup: EmbraceSetupInterface,
+) {
+    val embrace = Embrace.getInstance()
+
+    val clock: FakeClock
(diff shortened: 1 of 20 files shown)


embrace-android-sdk/src/integrationTest/kotlin/io/embrace/android/embracesdk/testframework/actions/EmbracePostSetupInterface.kt
```diff
@@ -0,0 +1,13 @@
+package io.embrace.android.embracesdk.testframework.actions
+
+import io.embrace.android.embracesdk.Embrace
+import io.embrace.android.embracesdk.fakes.FakeClock
+
+internal class EmbracePostSetupInterface(
```
fractalwrench, line 6: Would suggest renaming this `EmbracePreSdkInitInterface` or something similar, initially I was a bit confused about why this needed to be a separate thing to `testCaseAction`

fractalwrench (approved):
LGTM! This was on my list but I'm glad someone got to it first

embrace-android-sdk/src/integrationTest/kotlin/io/embrace/android/embracesdk/testframework/actions/EmbracePostSetupInterface.kt
bidetofevil: Sounds good

PR #1470, opened 2024-10-04
Place existing periodic caching behind interface

fractalwrench:
## Goal

Places existing periodic caching behind an interface to obscure the implementation details. I've altered `PayloadCachingService` so that the interface mostly follows the existing logic for caching right now.

## Testing


embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/delivery/caching/PayloadCachingServiceImpl.kt
```diff
@@ -0,0 +1,28 @@
...
+
+class PayloadCachingServiceImpl(
+    private val periodicSessionCacher: PeriodicSessionCacher,
+    private val periodicBackgroundActivityCacher: PeriodicBackgroundActivityCacher
+) : PayloadCachingService {
+
+    override fun shutdown() {
+    }
+
+    override fun startCaching(isInBackground: Boolean, supplier: () -> Envelope<SessionPayload>?) {
```
bidetofevil, line 16: It seems pretty imprecise for this to be a parameter passed in, and there being a chance for the payload to not match what this value is.

I understand if this is what the current implementation does, but it probably makes sense for this component to derive the background-ness of the given session and do the caching base on that?

embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/delivery/caching/PayloadCachingServiceImpl.kt
```diff
@@ -0,0 +1,28 @@
...
+        if (isInBackground) {
+            periodicBackgroundActivityCacher.scheduleSave(supplier)
+        } else {
+            periodicSessionCacher.start(supplier)
+        }
+    }
+
+    override fun stopCaching() {
```
bidetofevil, line 24: This seems pretty haphazard, not knowing what we are exactly stopping. Perhaps we should be explicit about which session we are stopping caching for?

bidetofevil (approved):
As a PR to put a shim around the existing implementation, this looks fine. But splitting out the logic like this really makes blatant that the starting/stopping of caching should be more explicit about what session is being cached or what session we are stopping the cache for.

fractalwrench:
@bidetofevil I agree - I plan on updating this in a future PR.

PR #1476, opened 2024-10-07
Consolidate session snapshotting

fractalwrench:
## Goal

Consolidates the implementation of session snapshotting so that both sessions + background activities use the `PeriodicSessionCacher` under the hood. I've avoided actually writing to disk every 2s by setting a flag when `reportBackgroundActivityStateChange` is called. If the flag is not set, no payload will be persisted for background activities.

We should probably consider whether the snapshot interval of 2s is reasonable for background activities given that we'll now be submitting executor jobs at this same interval. IMO this should be ok given it's similar to legacy behavior & that we won't take this approach with the eventual v2 caching implementation.

## Testing


embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/session/orchestrator/SessionOrchestratorImpl.kt
```diff
@@ -265,15 +277,11 @@
         endProcessState: ProcessState
     ): Envelope<SessionPayload>? {
         Systrace.traceSynchronous("on-session-cache") {
-            return if (initial.sessionId == sessionIdTracker.getActiveSessionId()) {
```
bidetofevil, line 268: Are we still at risk of caching data for a session that doesn't match the sessionId in the `initial` payload?

bidetofevil (approved):
The idea looks good overall. One question about whether we've done enough validation of overwriting a payload with the data from a later session.

embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/session/orchestrator/SessionOrchestratorImpl.kt
fractalwrench: I've moved this check to elsewhere within `SessionOrchestratorImpl`.

PR #1477, opened 2024-10-07
Extract additional caching logic to payload caching service

fractalwrench:
## Goal

Extracts some additional logic & places it in the payload caching service. 

## Testing

Added tests.


bidetofevil (approved):
LGTM. I see you've addressed the possibility of mismatched sessions in the caching :-)

PR #1478, opened 2024-10-07
Extract further payload caching

fractalwrench:
## Goal

Moves further payload caching logic out of `SessionOrchestratorImpl` so that it's isolated in a single class that will make future changes easier.

## Testing

Updated test coverage.


bidetofevil (approved):
LGTM

PR #1483, opened 2024-10-08
Cache payload in different directory to v1

fractalwrench:
## Goal

Caches session payloads in a different directory to V1. I've taken the approach of reusing the logic of the `PayloadStorageService` with a different instance. The `IntakeService` receives a request to cache a payload & then submits a job. IMO ultimately this will be the simplest way to appropriately prioritise writing payloads/caches, and to avoid repeating serialization/gzipping code in multiple places throughout the app.

## Testing

Added unit tests.


embrace-android-delivery/src/main/kotlin/io/embrace/android/embracesdk/internal/delivery/intake/IntakeService.kt
```diff
@@ -20,4 +20,9 @@ interface IntakeService : Shutdownable {
      * Stores a payload on disk as its JSON representation.
      */
     fun take(intake: Envelope<*>, metadata: StoredTelemetryMetadata)
+
+    /**
+     * Caches a payload on disk as its JSON representation.
+     */
+    fun cache(intake: Envelope<*>, metadata: StoredTelemetryMetadata)
```
bidetofevil, line 27: Do we need another method in the interface if we can tell whether a payload is complete to determine whether it's a cache or take scenario?

bidetofevil (approved):
This is fine to me if it's clearer to you at the interface level that `take` and `cache` be two separate methods. Initially I thought I preferred just one method but reading the usage, this makes the intentions more clear, so now I don't know what I prefer 😅

embrace-android-delivery/src/main/kotlin/io/embrace/android/embracesdk/internal/delivery/intake/IntakeServiceImpl.kt
```diff
@@ -23,16 +24,28 @@ class IntakeServiceImpl(
 
     override fun take(intake: Envelope<*>, metadata: StoredTelemetryMetadata) {
         worker.submit(metadata) {
-            processIntake(intake, metadata)
+            processIntake(intake, metadata, payloadStorageService)
```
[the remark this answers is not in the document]
fractalwrench: Yes. Persisting snapshots should be lower priority than capturing other data IMO.

embrace-android-delivery/src/main/kotlin/io/embrace/android/embracesdk/internal/delivery/intake/IntakeService.kt
fractalwrench: Updated to just use one method.

PR #1485, opened 2024-10-08
Improve threading model for log capture

fractalwrench:
## Goal

Improves the threading model for log capture + delivery. This makes the following changes to improve deliverability:

1. `V1PayloadStore` is now responsible for the background worker that stores log payloads, not `LogOrchestrator`. This means when enabled `V2PayloadStore` is consistently called from the same thread for all log payloads
2. Log capture now happens on the main thread which avoids the complexity of 2-3 worker threads interacting with each other. This does not affect log persistence which still happens on a background thread for v1/v2. In principle I can't see any operations that would be very expensive when capturing a log, but this is something we could profile if it needs optimizing

## Testing


bidetofevil (approved):
LGTM. I think the original intention of creating the log in the background is avoiding calling into the OTel SDK on the main thread whenever we can. But if that adds extra complications, then moving back into the main thread and asking the caller to not call from a time/place where the 1ms or so might matter.

PR #1486, opened 2024-10-08
Fix flaky test

fractalwrench:
## Goal

Addresses a flaky integration test by rolling back some changes that were made to not use a controllable executor.


embrace-android-sdk/src/integrationTest/kotlin/io/embrace/android/embracesdk/testcases/session/PeriodicSessionCacheTest.kt
```diff
@@ -26,14 +26,32 @@ internal class PeriodicSessionCacheTest {
 
     @Rule
     @JvmField
-    val testRule: IntegrationTestRule = IntegrationTestRule()
+    val testRule: IntegrationTestRule = IntegrationTestRule {
```
bidetofevil, line 29: Nice. This seems like a case that we'd eventually want to support in the test harness without needing all this boiler plate.

bidetofevil (approved):
LGTM

PR #1491, opened 2024-10-09
Refactor and improve NdkService tests

bidetofevil:
## Release Notes

**WHAT**:<br>
**WHY**:<br>

PR #1490, opened 2024-10-09
Split out logic for fetching and deleting native crash files

bidetofevil:
## Release Notes

**WHAT**:<br>
**WHY**:<br>

PR #1492, opened 2024-10-09
Resurrect all saved native crashes

bidetofevil:
## Release Notes

**WHAT**:<br>
**WHY**:<br>


bidetofevil:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/embrace-io/embrace-android-sdk/1492?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#1492** <a href="https://app.graphite.dev/github/pr/embrace-io/embrace-android-sdk/1492?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#1491** <a href="https://app.graphite.dev/github/pr/embrace-io/embrace-android-sdk/1491?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1490** <a href="https://app.graphite.dev/github/pr/embrace-io/embrace-android-sdk/1490?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`

<h2></h2>

embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/resurrection/PayloadResurrectionServiceImpl.kt
```diff
@@ -35,18 +36,25 @@ internal class PayloadResurrectionServiceImpl(
 ) : PayloadResurrectionService {
 
     override fun resurrectOldPayloads() {
+        val nativeCrashService = nativeCrashServiceProvider()
+        val nativeCrashes = nativeCrashService?.getNativeCrashes()?.associateBy { it.sessionId } ?: emptyMap()
```
fractalwrench, line 40: Perhaps a future refactor could move this to within the `NativeCrashService`, assuming nothing else needs to get a list of native crashes?

embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/resurrection/PayloadResurrectionServiceImpl.kt
```diff
@@ -35,18 +36,25 @@ internal class PayloadResurrectionServiceImpl(
...
+        nativeCrashes.values.forEach { nativeCrash ->
+            nativeCrashService?.sendNativeCrash(nativeCrash)
+        }
         payloadStorageService
             .getUndeliveredPayloads()
             .forEach { deadSessionMetadata ->
-                deadSessionMetadata.sendResurrectedPayload()
+                deadSessionMetadata.sendResurrectedPayload(nativeCrashes::get)
             }
+        // TODO: delete each native crash after its associated session is resurrected and sent
```
fractalwrench, line 49: Is this done? I assumed so given the line below.

fractalwrench (approved):
LGTM. Do we also need to avoid calling `sendCachedSessions` on the `DeliveryService` or will that come in a future PR?

embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/resurrection/PayloadResurrectionServiceImpl.kt
bidetofevil: Yeah definitely. The NdkService-related classes and interfaces seem like they could use some help. Like, some of this stuff, I think I should jus be pulling from the NDK directly and bypass this service. In fact, the service seems like it should only be sending the crash logs and not doing anything else.

All this can be consolidated once we get rid of v1. I'll add a task

bidetofevil:
> LGTM. Do we also need to avoid calling `sendCachedSessions` on the `DeliveryService` or will that come in a future PR?

Those methods are called as part of the old resurrection workflow, which looks for the old payloads. 

My original intent was to put the code that calls them behind the feature flag because I don't want two things looking at the same set of native crash files. That change would be made in the PR where I wire this up to the SDK

But thinking about this now, it may make sense to keep some of the logic around once we ship v2, as it will conceivable allow us to resurrect old session payloads without adding any awareness of them in the new resurrection service. That may be one way of getting some backwards compat in the new delivery layer without added anything there explicitly. We should think about that...

bidetofevil:
### Merge activity

* **Oct 9, 1:58 PM EDT**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/embrace-io/embrace-android-sdk/1492).

PR #1493, opened 2024-10-09
Register scheduling service as network connectivity listener

fractalwrench:
## Goal

Registers this class as a listener so that it can react to changes in network connectivity.


bidetofevil (approved):
LOL I forgot that I didn't hook that up

PR #1494, opened 2024-10-09
Handle shutdown in scheduling service

fractalwrench:
## Goal

Implements shutdown in the `SchedulingService`. I've taken the approach of just shutting the scheduling + delivery executors & not waiting, as we don't want to block on HTTP requests or I/O operations as these can just be retried on the next process, but we do want to prevent adding any further work when the process is terminating.

## Testing

Added unit test.


bidetofevil (approved):
LGTM

PR #1495, opened 2024-10-09
Record correct type when capturing logs

fractalwrench:
## Goal

Records whether the log was a log, crash, or network capture when persisting to disk.

## Testing

Added unit tests.

### DIFF
--- a/embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/session/orchestrator/V2PayloadStore.kt
+++ b/embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/session/orchestrator/V2PayloadStore.kt
@@ -1,12 +1,15 @@
 package io.embrace.android.embracesdk.internal.session.orchestrator
 
+import io.embrace.android.embracesdk.internal.arch.schema.EmbType.System
 import io.embrace.android.embracesdk.internal.clock.Clock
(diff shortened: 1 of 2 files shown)


embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/session/orchestrator/V2PayloadStore.kt
```diff
@@ -27,14 +30,30 @@ internal class V2PayloadStore(
...
         intakeService.take(envelope, createMetadata(type))
     }
 
     override fun handleCrash(crashId: String) {
         intakeService.shutdown()
     }
 
+    private fun findSupportedEnvelopeType(logs: List<Log>?): SupportedEnvelopeType {
+        // look at emb.type in all the logs. If the type is not the same for all, default to LOG.
+        val embType = logs?.mapNotNull {
```
bidetofevil, line 43: If we are assuming log payloads are homogenous, I'd rather just look at the first log and its type and use that. I can see that in the future, we get OTel logs that don't have `emb.type`, and we'll just treat them generically and not specify a type in the envelope.

fractalwrench: Updated implementation to achieve this.

bidetofevil (approved):
LGTM

PR #1496, opened 2024-10-09
Encode process id in filename

fractalwrench:
## Goal

Encodes the process ID in the filename so that it's simpler to find sessions belonging to a given process.

## Testing

Updated unit tests.


bidetofevil (approved):
LGTM. 

Reading this code, maybe we should rename this later so it doesn't get confused with the literal process ID. `app_instance_id`? We can figure that out later.

PR #1500, opened 2024-10-10
Avoid check not null in module bootstrapper

fractalwrench:
## Goal

Avoids checkNotNull to prefer returning null in the `ModuleInitBootstrapper`. Functionally this should not make much different as `apiService` will always be null if only OTel export mode is enabled, but it can't hurt to alter this in case that assumption ever breaks.


bidetofevil (approved):
LGTM

PR #1501, opened 2024-10-10
Replace scheduling worker

fractalwrench:
## Goal

Replaces the worker used for the `SchedulingService` with one that can use I/O, as I/O will be triggered the first time the service attempts to read from the `PayloadStore`. Additionally renamed some workers to more appropriate names.


bidetofevil (approved):
Oh good! I was worried about this. The nightmare scenario of a deadlock of blocking on a task that is being run on the same single threaded executor ☠️

PR #1502, opened 2024-10-10
Address backpressure in periodic caching

fractalwrench:
## Goal

Handles back pressure in periodic session caching by cancelling any pending tasks when submitting a new task request. While this won't stop any `Runnable` that is already executing it will cancel any that are in the queue, therefore helping avoid a scenario where N runnables are queued up to persist useless data.

The v1 delivery layer seems susceptible to this but I will leave that behavior as is since we plan on removing it soon.

## Testing

Added unit test.

### DIFF
--- a/embrace-android-delivery/src/main/kotlin/io/embrace/android/embracesdk/internal/delivery/intake/IntakeServiceImpl.kt
+++ b/embrace-android-delivery/src/main/kotlin/io/embrace/android/embracesdk/internal/delivery/intake/IntakeServiceImpl.kt
@@ -1,6 +1,7 @@
 package io.embrace.android.embracesdk.internal.delivery.intake
 
 import io.embrace.android.embracesdk.internal.delivery.StoredTelemetryMetadata
+import io.embrace.android.embracesdk.internal.delivery.SupportedEnvelopeType
 import io.embrace.android.embracesdk.internal.delivery.scheduling.SchedulingService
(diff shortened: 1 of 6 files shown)


embrace-android-delivery/src/main/kotlin/io/embrace/android/embracesdk/internal/delivery/intake/IntakeServiceImpl.kt
```diff
@@ -19,14 +20,23 @@ class IntakeServiceImpl(
...
     }
 
     override fun take(intake: Envelope<*>, metadata: StoredTelemetryMetadata) {
-        worker.submit(metadata) {
+        val future = worker.submit(metadata) {
             processIntake(intake, metadata)
         }
+
+        // cancel any cache attempts that are already pending to avoid unnecessary I/O.
+        if (!metadata.complete) {
```
bidetofevil, line 35: This assumes the last cache attempt was for the same thing as the current one. I THINK we can make this assumption for now since there should be a full complete session payload if we are trying to cache a new session), but in case this were to change in the future, should we add some safeguards like checking the metadata is the same before doing so? Or just that it's an envelope of the same type?

bidetofevil (approved):
LGTM - one comment about adding some safeguards to verify the thing being cached before cancelling a caching job

PR #1503, opened 2024-10-10
Review delivery threading model

fractalwrench:
## Goal

Makes some updates to the threading model for the v2 delivery layer as the old approach wasn't working. Notably:

1. The `IntakeService` and `PayloadStorageService` now share a worker for any mutative file operations
2. `PayloadStorageService` enqueues delete operations on the same thread that `IntakeService` uses to write files

## Testing


bidetofevil (approved):
LGTM. Comment about throwing IOException explicitly in the interface.

PR #1505, opened 2024-10-10
Add otel export assertions to integration tests

fractalwrench:
## Goal

Adds an initial implementation to the integration tests of assertions that check we are exporting OTel spans. In this approach the test harness always creates an exporter before each test, and the test case can optionally retrieve spans & compare them against a golden file of JSON.

This is relatively simple right now & doesn't support logs, events, or comparing OTel relationships (e.g. looking at span IDs etc). IMO this is functionality that we can add later as we add further test cases.


bidetofevil (approved):
NICE

PR #1506, opened 2024-10-10
Only write attributes for crashes if they contain valid values

bidetofevil:
## Goal

The NDK and/or the serialization layer is returning values that are effectively null/empty but with non-null string value like "null", and we weren't filtering out those cases because we assumed it would just return null when those values didn't exist.

Add in the generic ability to detect string values like that, and to skip adding those values as OTel attributes, if you explicitly say that's what you wanted.

There was also some code in the native crash log generation layer that didn't check for empty collections before serializing their values - took care of those cases too.

## Testing
Units tests all the way down

### DIFF
--- a/embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/arch/destination/LogWriterImpl.kt
+++ b/embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/arch/destination/LogWriterImpl.kt
@@ -7,6 +7,7 @@ import io.embrace.android.embracesdk.internal.session.id.SessionIdTracker
 import io.embrace.android.embracesdk.internal.session.lifecycle.EmbraceProcessStateService.Companion.BACKGROUND_STATE
 import io.embrace.android.embracesdk.internal.session.lifecycle.EmbraceProcessStateService.Companion.FOREGROUND_STATE
 import io.embrace.android.embracesdk.internal.session.lifecycle.ProcessStateService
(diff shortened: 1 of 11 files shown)


bidetofevil:
* **#1506** <a href="https://app.graphite.dev/github/pr/embrace-io/embrace-android-sdk/1506?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/arch/destination/LogWriterImpl.kt
```diff
@@ -36,7 +38,9 @@ class LogWriterImpl(
 
         var sessionState: String? = null
         sessionIdTracker.getActiveSession()?.let { session ->
-            builder.setAttribute(SessionIncubatingAttributes.SESSION_ID, session.id)
+            if (addSessionId) {
```
bidetofevil, line 41: When creating a native crash log, we don't want to attach the current sessionId as the ID, so give us a means to opt out of that.

fractalwrench (approved):
LGTM

bidetofevil:
### Merge activity

* **Oct 11, 2:42 PM EDT**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/embrace-io/embrace-android-sdk/1506).

PR #1512, opened 2024-10-11
Avoid mixing workers between different delivery systems

fractalwrench:
## Goal

Uses a separate worker for the v1/v2 delivery systems to avoid the possibility of undesired behavior if the wrong type information is submitted with a `Runnable`. This shouldn't add a thread in practice as only 1 system will be active at a time. As part of this change I also correctly gated a v1 delivery call in SDK bootstrapping behind the v2 storage feature flag.

## Testing


bidetofevil (approved):
LGTM

PR #1513, opened 2024-10-11
Added RequestExecutionServiceImpl

priettt:


embrace-android-delivery/src/main/kotlin/io/embrace/android/embracesdk/internal/delivery/execution/HttpMethodV2.kt
```diff
@@ -0,0 +1,21 @@
+package io.embrace.android.embracesdk.internal.delivery.execution
+
+import java.util.Locale
+
+enum class HttpMethodV2 {
```
bidetofevil, line 5: Why do we need this? This service is only going to POST at this point, and eventually GET. I don't think we need an abstraction over the method? And if we do, we may want to just have separate POST and GET request objects that set the method explicitly and call it a day

embrace-android-delivery/src/main/kotlin/io/embrace/android/embracesdk/internal/delivery/execution/RequestExecutionServiceImpl.kt
```diff
@@ -1,15 +1,62 @@
...
 import io.embrace.android.embracesdk.internal.delivery.SupportedEnvelopeType
+import okhttp3.Headers.Companion.toHeaders
+import okhttp3.OkHttpClient
+import okhttp3.Protocol
+import okhttp3.Request
+import okhttp3.RequestBody.Companion.toRequestBody
 import java.io.InputStream
 
-class RequestExecutionServiceImpl : RequestExecutionService {
+class RequestExecutionServiceImpl(
```
bidetofevil, line 14: lol that's it?

embrace-android-api/api/embrace-android-api.api
```diff
@@ -238,7 +238,6 @@ public final class io/embrace/android/embracesdk/network/http/HttpMethod : java/
 	public static final field POST Lio/embrace/android/embracesdk/network/http/HttpMethod;
 	public static final field PUT Lio/embrace/android/embracesdk/network/http/HttpMethod;
 	public static final field TRACE Lio/embrace/android/embracesdk/network/http/HttpMethod;
-	public static fun fromInt (Ljava/lang/Integer;)Lio/embrace/android/embracesdk/network/http/HttpMethod;
```
fractalwrench, line 241: This is a breaking change

embrace-android-delivery/build.gradle.kts
```diff
@@ -11,6 +11,8 @@ android {
 apiValidation.validationDisabled = true
 
 dependencies {
+    implementation(libs.okhttp)
+    implementation(libs.moshi)
```
fractalwrench, line 15: IMO this module doesn't need to know about Moshi specifically, we can hide that implementation detail by moving `ApiRequestV2` to the `embrace-android-payload` module.

embrace-android-delivery/src/main/kotlin/io/embrace/android/embracesdk/internal/delivery/execution/RequestExecutionServiceImpl.kt
```diff
@@ -1,15 +1,62 @@
...
 
     override fun attemptHttpRequest(
         payloadStream: () -> InputStream,
-        envelopeType: SupportedEnvelopeType
+        envelopeType: SupportedEnvelopeType,
     ): ApiResponse {
-        return ApiResponse.None
+        val apiRequest = envelopeType.endpoint.getApiRequestFromEndpoint()
+
+        val payload: ByteArray = payloadStream().use {
```
fractalwrench, line 32: Not a massive fan of loading the entire payload into a `ByteArray` as it means memory usage linearly increases with the size of the payload. Does `OkHttp` provide a mechanism that allows a stream (or Okio) equivalent to be used instead? That will have a fixed buffer so memory usage won't grow excessively.

embrace-android-api/api/embrace-android-api.api
priettt: oh this wasn't intended to be in the PR 🤦

embrace-android-delivery/src/main/kotlin/io/embrace/android/embracesdk/internal/delivery/execution/HttpMethodV2.kt
priettt: we don't! at first I thought the server expected the field, but it's just used to modify the connection

embrace-android-delivery/src/main/kotlin/io/embrace/android/embracesdk/internal/delivery/execution/RequestExecutionServiceImpl.kt
priettt: Yeah, it's what's worrying me the most, I'll try to find some alternative

fractalwrench: There's a recipe here that might be useful as a starting point, it'll probably be simplest to write to the `OutputStream` as suggested: https://square.github.io/okhttp/recipes/#post-streaming-kt-java

embrace-android-delivery/src/main/kotlin/io/embrace/android/embracesdk/internal/delivery/execution/RequestExecutionServiceImpl.kt
```diff
@@ -1,15 +1,86 @@
...
+        }
+
+        val request = Request.Builder()
+            .url(apiRequest.url)
+            .headers(apiRequest.getHeaders().toHeaders())
+            .post(requestBody)
+            .build()
+
+        okHttpClient.newCall(request).execute().use { response ->
+            if (response.isSuccessful) {
```
fractalwrench, line 52: It will be necessary to support the same responses that `ApiClientImpl` currently does

fractalwrench (commented):
This looks good - I'd like to see a few unit tests using Mockwebserver to validate the behavior before this gets merged down. It'll also be necessary to support returning different `ApiResponse` values than just `Success/Failure`

Have you checked with someone on the backend that our servers support the HTTP2 protocol specified in this changeset?

embrace-android-delivery/src/main/kotlin/io/embrace/android/embracesdk/internal/delivery/execution/RequestExecutionServiceImpl.kt
```diff
@@ -1,15 +1,86 @@
...
+        val headers = mutableMapOf(
+            "Accept" to accept,
+            "User-Agent" to userAgent,
+            "Content-Type" to contentType
+        )
+        contentEncoding?.let { headers["Content-Encoding"] = it }
+        acceptEncoding?.let { headers["Accept-Encoding"] = it }
+        appId?.let { headers["X-EM-AID"] = it }
+        deviceId?.let { headers["X-EM-DID"] = it }
+        eventId?.let { headers["X-EM-SID"] = it }
```
bidetofevil, line 81: Don't think we need eventId anymore?

bidetofevil:
I would validate the via unit tests that all the expected responses are handled and mapped back correctly. Perhaps the existing tests can be used for inspiration?

embrace-android-delivery/src/main/kotlin/io/embrace/android/embracesdk/internal/delivery/execution/HttpUrlConnectionRequestExecutionService.kt
```diff
@@ -0,0 +1,94 @@
...
+        connection.setDoOutput(true)
+
+        connection.connectTimeout = DEFAULT_TIMEOUT_SECONDS
+        connection.readTimeout = DEFAULT_TIMEOUT_SECONDS
+
+        return connection
+    }
+
+    private fun Endpoint.getApiRequestFromEndpoint(): ApiRequestV2 = ApiRequestV2(
+        url = "$coreBaseUrl/v2/${this.path}",
```
bidetofevil, line 88: Shouldn't v2 be part of either the base or the path?

embrace-android-delivery/src/main/kotlin/io/embrace/android/embracesdk/internal/delivery/execution/OkHttpRequestExecutionService.kt
```diff
@@ -0,0 +1,98 @@
...
+class OkHttpRequestExecutionService(
+    private val coreBaseUrl: String,
+    private val lazyDeviceId: Lazy<String>,
+    private val appId: String,
+    private val embraceVersionName: String,
+) : RequestExecutionService {
+
+    private val okHttpClient = OkHttpClient()
+        .newBuilder()
+        .connectTimeout(DEFAULT_TIMEOUT_SECONDS, TimeUnit.SECONDS)
```
bidetofevil, line 32: At 60s, this seems to be too long. What is the default? 15s? We should probably just use that.

If we can't establish a TCP connection in even 15s, it's unlikely to succeed unless the network connection changes.

priettt: That's what we were setting with OkHttpUrlConnection, but I agree, the default will probably be enough

embrace-android-delivery/src/main/kotlin/io/embrace/android/embracesdk/internal/delivery/execution/HttpUrlConnectionRequestExecutionService.kt
priettt: we could add it to the Endpoint class, but I didn't want to break compatibility yet, as we are using that in other places

embrace-android-delivery/src/main/kotlin/io/embrace/android/embracesdk/internal/delivery/execution/OkHttpRequestExecutionService.kt
```diff
@@ -29,8 +28,6 @@ class OkHttpRequestExecutionService(
 
     private val okHttpClient = OkHttpClient()
         .newBuilder()
-        .connectTimeout(DEFAULT_TIMEOUT_SECONDS, TimeUnit.SECONDS)
```
priettt: default is 10 seconds

embrace-android-delivery/src/main/kotlin/io/embrace/android/embracesdk/internal/delivery/execution/HttpUrlConnectionRequestExecutionService.kt
bidetofevil: Got it. No problem. We'll fix this all later.

embrace-android-delivery/src/main/kotlin/io/embrace/android/embracesdk/internal/delivery/execution/OkHttpRequestExecutionService.kt
bidetofevil: yikes. yeah, this could leave the network request hanging and increase the failure rate of requests that are queued but not sent, so ultimately not serialized and redelivered....

bidetofevil: Lets leave the old one with these bad values, but fix OkHttp to have reasonable ones. It matter less now with the new delivery layer, but the read timeout at 60 may be too long too, but we can keep that as is.

embrace-android-delivery/src/test/kotlin/io/embrace/android/embracesdk/internal/delivery/execution/HttpUrlConnectionRequestExecutionServiceTest.kt
```diff
@@ -0,0 +1,204 @@
...
+import okhttp3.mockwebserver.MockWebServer
+import okhttp3.mockwebserver.SocketPolicy
+import org.junit.After
+import org.junit.Assert.assertEquals
+import org.junit.Assert.assertTrue
+import org.junit.Before
+import org.junit.Test
+import java.net.UnknownHostException
+
+class HttpUrlConnectionRequestExecutionServiceTest {
```
bidetofevil, line 16: In theory since the APIs are the same, this test class can be used to test both implementations provided you reference the interface and not the payload, right?

embrace-android-delivery/src/main/kotlin/io/embrace/android/embracesdk/internal/delivery/execution/HttpUrlConnectionRequestExecutionService.kt
```diff
@@ -0,0 +1,95 @@
...
+import java.net.HttpURLConnection
+import java.net.URL
+
+private const val NO_HTTP_RESPONSE = -1
+private const val TOO_MANY_REQUESTS = 429
+private const val DEFAULT_TIMEOUT_MILLISECONDS = 10 * 1000
+
+class HttpUrlConnectionRequestExecutionService(
+    private val coreBaseUrl: String,
+    private val lazyDeviceId: Lazy<String>,
```
bidetofevil, line 17: Small preference to using a provider-like interface like () -> String instead of lazy since you're not initializing a singleton, but asking a for a function to give you the value.

Same comment applies to the other implementation

embrace-android-delivery/src/test/kotlin/io/embrace/android/embracesdk/internal/delivery/execution/HttpUrlConnectionRequestExecutionServiceTest.kt
bidetofevil, line 16: Might want to test timeout as well?

bidetofevil (approved):
Looks good. A couple of suggestions for additional tests but feels reasonable to merge so we can do further testing!

embrace-android-delivery/src/test/kotlin/io/embrace/android/embracesdk/internal/delivery/execution/HttpUrlConnectionRequestExecutionServiceTest.kt
priettt: I am testing timeout, but it's a bit hidden 😅 

```
    fun `return incomplete if the server returns an unexpected response code`() {
        // given a server that disconnects before sending a response
        server.enqueue(MockResponse().setSocketPolicy(SocketPolicy.DISCONNECT_AT_START))
        ...
```

As the socket disconnects when it receives a request, it will not return anything, so the actual error is a timeout. I'll add that as a comment

PR #1516, opened 2024-10-14
Fix periodic session cache test case

fractalwrench:
## Goal

Alters the periodic session cache test case so that it won't fail when run against the v2 delivery layer. I've not flipped the switch as part of this PR but have instead opted to create lots of the necessary functions etc.


embrace-android-sdk/src/integrationTest/kotlin/io/embrace/android/embracesdk/testcases/session/PeriodicSessionCacheTest.kt
```diff
@@ -24,46 +35,54 @@ import org.junit.runner.RunWith
...
                 }
+                allowPeriodicCacheExecution()
             },
             assertAction = {
-                val envelopes = getCachedSessionEnvelopes(2)
-                val endMessage = envelopes[0]
+                val endMessage = getCachedSessionEnvelopes(2).first()
                 val span = endMessage.findSpanSnapshotOfType(EmbType.Ux.Session)
-                assertNull(span.getSessionProperty("Test"))
+                assertNotNull(span.getSessionProperty("Test"))
```
priettt, line 81: why is this not null anymore?

priettt (approved):
🙌

PR #1517, opened 2024-10-14
Deletes old session payloads that have been snapshotted.

fractalwrench:
## Goal

Ensures that the periodic session caching deletes any old snapshots after persisting a new snapshot. This aims to prevent sending duplicate payloads unnecessarily.

## Testing

Added unit test.


bidetofevil (approved):
LGTM

PR #1519, opened 2024-10-15
Filter invalid ANR samples

fractalwrench:
## Goal

Filters out any invalid ANR samples (i.e. any samples that have a timestamp outside of the interval range). Once we eventually migrate this feature over to be captured via pure OTel this problem should go away entirely.

## Testing


bidetofevil (approved):
LGTM

bidetofevil (commented):
Seems weird that the ANR ended but there are still samples

PR #1520, opened 2024-10-15
Add integration tests for v1 delivery layer

fractalwrench:
## Goal

Adds integration tests that confirm the basic functions of the v1 delivery layer are still working.


bidetofevil (approved):
LGTM

PR #1521, opened 2024-10-15
Add load test for session capture

fractalwrench:
## Goal

Adds a load test that confirms 1000 sessions can be captured & delivered by the SDK without problems.


embrace-android-sdk/src/integrationTest/kotlin/io/embrace/android/embracesdk/testcases/session/SessionSpamTest.kt
```diff
@@ -0,0 +1,38 @@
...
+
+    @Rule
+    @JvmField
+    val testRule: IntegrationTestRule = IntegrationTestRule()
+
+    @Test
+    fun `session messages are recorded`() {
+        testRule.runTest(
+            testCaseAction = {
+                repeat(SESSION_COUNT) {
```
bidetofevil, line 24: Can we do check background activities as well?

bidetofevil (approved):
LGTM. One comment about also validating background activities

PR #1524, opened 2024-10-15
Use mock web server in integration tests

fractalwrench:
## Goal

Alters `StatefulSessionTest` to use mockwebserver rather than `FakeRequestExecutionService`. This exercises the whole of the SDK against a test harness & also opens the door for returning different responses (e.g. 429) to confirm the SDK can handle it properly.


bidetofevil (approved):
NICE!

PR #1528, opened 2024-10-16
Use mock web server by default

fractalwrench:
## Goal

Updates the integration tests to use mockwebserver by default. This required using the previous approach for moments + the v1 delivery tests, so after 7.0 we should be able to remove all this code & just use mock web server. There isn't a noticeable difference in execution time between these two approaches which was my main concern starting out with this approach.


bidetofevil (approved):
LGTM

PR #1529, opened 2024-10-16
Remove unnecessary json adapters

fractalwrench:
## Goal

Removes some JSON adapters that are no longer necessary. I checked by looking at each type with 'find usages' and confirmed that no serialization was done with the symbol. I'm also relying on test coverage.


bidetofevil (approved):
LGTM

PR #1530, opened 2024-10-16
Move old payload classes

fractalwrench:
## Goal

Moves some classes out of `embrace-android-payload` that no longer require JSON adapters. 

## Testing


bidetofevil (approved):
LGTM

PR #1532, opened 2024-10-17
Register connectivity listener after delivery module init

fractalwrench:
## Goal

Register a connectivity listener after delivery module init, as otherwise scheduling service isn't initialized.


bidetofevil (approved):
LGTM

bidetofevil (commented):
Ah, there's no integration test to verify this

PR #1534, opened 2024-10-17
Use version catalog where possible

fractalwrench:
## Goal

Uses a workaround in Gradle to use the version catalog in a precompiled build script. I've not been able to find a solution that would use the version catalog when setting the plugins yet, but this gets us 90% of the way there.

### DIFF
--- a/buildSrc/build.gradle.kts
+++ b/buildSrc/build.gradle.kts
@@ -1,8 +1,8 @@
 import org.jetbrains.kotlin.gradle.tasks.KotlinCompile
 
 plugins {
-    `java-gradle-plugin`
     `kotlin-dsl`
+    `kotlin-dsl-precompiled-script-plugins`
 }
 
 repositories {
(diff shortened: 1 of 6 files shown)


buildSrc/build.gradle.kts
```diff
@@ -1,24 +1,23 @@
...
 
 repositories {
     google()
     mavenCentral()
 }
 
+// NOTE: when updating any of these keep in sync with the version catalog
 dependencies {
-    compileOnly(gradleApi())
+    implementation(gradleApi())
```
bidetofevil, line 15: Can we fail the SDK build if the versions we have here isn't in sync with the very catalog?

bidetofevil (approved):
AWESOME!

buildSrc/build.gradle.kts
fractalwrench: I'm not sure of an easy way to achieve this. I've asked Daniel whether he has any suggestions on how to use version catalog with precompiled build script plugin dependencies, so we can see what he says & alter accordingly.

lacasseio: The `buildSrc` projects are a special (legacy) flavour of an included build. For historical reasons, they do not require a `settings.gradle` file, but (I believe [1]) you can add one without issue. From the settings file, you can load the catalogue from the `../gradle/libs.versions.toml`.

[1] If adding the settings.gradle file fails for some reason; just convert the `buildSrc` project into a proper include build. I don't think you need to do that.

PR #1536, opened 2024-10-17
Bump AGP to latest version

fractalwrench:
## Goal

Bumps AGP to 8.7. Also bumps the Lint dependency as this should always be kept in sync with AGP.


bidetofevil (approved):
LGTM

PR #1538, opened 2024-10-17
Add integration test for payload pruning

fractalwrench:
## Goal

Adds an integration test that checks the priority in which payloads are pruned. This also fixes a bug where the most recent session was deleted due to the reuse of a comparator that isn't correct for this specific use-case.

### DIFF
--- a/embrace-android-delivery/src/main/kotlin/io/embrace/android/embracesdk/internal/delivery/storage/PayloadStorageServiceImpl.kt
+++ b/embrace-android-delivery/src/main/kotlin/io/embrace/android/embracesdk/internal/delivery/storage/PayloadStorageServiceImpl.kt
@@ -148,9 +148,11 @@ class PayloadStorageServiceImpl(
         if (removalCount < 0) {
             return false
         }
-        val removals = input
-            .sortedWith(storedTelemetryComparator)
-            .takeLast(removalCount)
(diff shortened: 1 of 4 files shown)


embrace-android-delivery/src/main/kotlin/io/embrace/android/embracesdk/internal/delivery/storage/PayloadStorageServiceImpl.kt
```diff
@@ -148,9 +148,11 @@ class PayloadStorageServiceImpl(
         if (removalCount < 0) {
             return false
         }
-        val removals = input
-            .sortedWith(storedTelemetryComparator)
-            .takeLast(removalCount)
+        val removals = input.sortedWith(
```
bidetofevil, line 151: Oh interesting, so there's no "descending with comparator" method? That kind of sucks... it means we are maintaining a different prune sort order than delivery :/

bidetofevil (approved):
LGTM

bidetofevil:
### Merge activity

* **Oct 19, 1:19 AM EDT**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/embrace-io/embrace-android-sdk/1538).

PR #1541, opened 2024-10-18
Trigger initial delivery of stored payloads

fractalwrench:
## Goal

When the SDK loaded there was nothing that triggered the delivery of completed payloads that were stored on disk. I've taken the approach of invoking the `SchedulingService` when the SDK initializes. We could add a more specific function if that's semantically desired but for now I just called `onPayloadIntake()`.

## Testing

Added an integration test

### DIFF
--- a/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/EmbraceImpl.kt
+++ b/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/EmbraceImpl.kt
@@ -205,12 +205,12 @@ internal class EmbraceImpl @JvmOverloads constructor(
         // Send any sessions that were cached and not yet sent.
         startSynchronous("send-cached-sessions")
 
-        if (useLegacyResurrection()) {
+        if (useLegacyResurrection(configModule.configService)) {
             bootstrapper
                 .workerThreadModule
                 .priorityWorker<TaskPriority>(Worker.Priority.DeliveryCacheWorker)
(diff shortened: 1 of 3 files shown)


embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/EmbraceImpl.kt
```diff
@@ -379,7 +382,7 @@ internal class EmbraceImpl @JvmOverloads constructor(
         }
     }
 
-    private fun useLegacyResurrection(): Boolean {
-        return configService?.autoDataCaptureBehavior?.isV2StorageEnabled() != true
+    private fun useLegacyResurrection(cfgService: ConfigService?): Boolean {
+        return cfgService?.autoDataCaptureBehavior?.isV2StorageEnabled() != true
```
fractalwrench, line 386: Fixes a subtle bug where `configService` returns null because the SDK isn't initialized yet

bidetofevil: Good catch!

bidetofevil (approved):
LGTM

bidetofevil:
### Merge activity

* **Oct 19, 1:32 AM EDT**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/embrace-io/embrace-android-sdk/1541).

PR #1542, opened 2024-10-18
[DNM] Bump compile sdk version to 35

fractalwrench:
## Goal

Increases the compileSdkVersion to 35.

## Testing

## Release Notes

**WHAT**:Increase compile SDK version to 35<br>


embrace-android-features/src/main/kotlin/io/embrace/android/embracesdk/internal/capture/memory/MemoryWarningDataSource.kt
```diff
@@ -56,6 +56,7 @@ internal class MemoryWarningDataSource(
      * @param trimLevel the context of the trim, giving a hint of the amount of trimming.
      */
     override fun onTrimMemory(trimLevel: Int) {
```
bidetofevil, line 58: these don't work anymore if customers target SDK 34. This is something we should really deprecate for 7.0, if not entirely eliminated

bidetofevil (approved):
LGTM. Added comment about deprecation because it no longer works if you target SDK 34

fractalwrench:
Added deprecation notice to the config field so we remember to remove this in 7.0

PR #1545, opened 2024-10-18
[DNM] Bump lifecycle dependencies

fractalwrench:
## Goal

Bumps the lifecycle dependencies to the latest version. This involves [changing some artefacts](https://developer.android.com/jetpack/androidx/releases/lifecycle) as we no longer use Java 8 & it's possible to be more specific about what we want to include given the lifecycle libs have been further modularised since we last took a look at this.

The compose module was explicitly depending on transitive dependencies from the lifecycle lib - I've removed that as we no longer need to use a [compat class](https://developer.android.com/jetpack/androidx/releases/core#1.13.0-beta01).

## Testing


bidetofevil (approved):
Oh the compose dep removal is so nice!

PR #1552, opened 2024-10-21
[DNM] Bump profile installer dependency

fractalwrench:
## Goal

Bumps the profile installer dependency so that profiles can be installed on API 35+.

## Testing

Relied on existing test coverage & reviewed [release notes](https://developer.android.com/jetpack/androidx/releases/profileinstaller).


bidetofevil (approved):
LGTM

PR #1553, opened 2024-10-21
[DNM] Bump kotlin dependency

fractalwrench:
## Goal

Bumps Kotlin to 2.0.21. We still maintain compatibility back to 1.8.22 by setting the `apiVersion/languageVersion` attributes, so it's not anticipated this change will affect any end-user experience.

## Testing

Relied on existing test coverage & reviewed [release notes](https://kotlinlang.org/docs/whatsnew20.html).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -62,3 +62,4 @@ fastlane/readme.md
 docs/
 
 .DS_Store
+.kotlin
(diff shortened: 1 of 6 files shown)


bidetofevil (approved):
LGTM

PR #1555, opened 2024-10-21
[DNM] Move nested classes into top level classes

fractalwrench:
## Goal

Moves the `AppFramework` and `LastRunEndState` into top level classes. This is a breaking change so should be shipped as part of 7.0.


bidetofevil (approved):
LGTM

PR #1556, opened 2024-10-21
[DNM] Implement more public API in Kotlin

fractalwrench:
## Goal

Implements more of our public API in Kotlin rather than Java. For the most part this is backwards compatible apart from one case with `EmbraceNetworkRequest` and default params.

## Testing


embrace-android-api/src/main/java/io/embrace/android/embracesdk/network/http/HttpMethod.kt
```diff
@@ -0,0 +1,41 @@
+package io.embrace.android.embracesdk.network.http
+
+/**
+ * Enumeration of supported HTTP request methods.
+ *
+ *
+ * This class is part of the Embrace Public API.
+ */
+public enum class HttpMethod {
```
bidetofevil, line 9: Can we get rid of this?

bidetofevil (approved):
LGTM

embrace-android-api/src/main/java/io/embrace/android/embracesdk/network/http/HttpMethod.kt
fractalwrench: Not without breaking API changes. What would the alternative look like?

bidetofevil: OkHttp just does it with a String, and in the request logging code, we are just consuming it as a string and mapping it back to the enum with `HttpMethod.fromString` anyway. 

So it feels like that's pretty reasonable to do to get rid of this?

fractalwrench: My personal bias is towards leaving function signatures mostly the same in 7.0 and deprecating/rearranging existing classes, then addressing these types of things as part of improving the Kotlin API. I'm also generally in favour of enums in APIs over arbitrary strings unless there's good perf reasons to avoid that

PR #1566, opened 2024-10-22
Remove unnecessary null checks

fractalwrench:
## Goal

Removes unnecessary null checks from `Embrace.java`. This is a precursor refactor to converting the facade to Kotlin.

## Testing


bidetofevil (approved):
Nice! There's a bunch of tests around this too that can probably be removed as well.

PR #1567, opened 2024-10-22
[DNM] Remove deprecated apis

fractalwrench:
## Goal

Removes some deprecated APIs as part of 7.0.

## Testing

### DIFF
--- a/embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/prefs/PreferencesService.kt
+++ b/embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/prefs/PreferencesService.kt
@@ -53,17 +53,6 @@ interface PreferencesService {
     var permanentSessionProperties: Map<String, String>?
 
     /**
-     * No longer used, will be removed in a future version.
-     *
-     * Method is still present to ensure that during any upgrades to SDK3, any custom
-     * personas are merged with the user personas list.
-     *
-     * @return custom personas
-     */
-    @Deprecated("")
-    val customPersonas: Set<String>?
-
-    /**
      * Username for the user
      */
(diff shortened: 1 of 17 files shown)


embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/prefs/PreferencesService.kt
```diff
@@ -52,17 +52,6 @@ interface PreferencesService {
...
 
-    /**
-     * No longer used, will be removed in a future version.
-     *
-     * Method is still present to ensure that during any upgrades to SDK3, any custom
-     * personas are merged with the user personas list.
-     *
-     * @return custom personas
-     */
-    @Deprecated("")
```
bidetofevil, line 63: Wait, this means we can remove the entire personas feature?

bidetofevil (approved):
LGTM

embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/prefs/PreferencesService.kt
fractalwrench: Not quite. Custom personas is something that used to exist in a very old version of the SDK & was no longer settable via our API, but we kept reading it in case folks had persisted the value. I think it's safe to remove this now as enough time has passed & few people use personas anyway

bidetofevil: OK, that's different than what is done when customers call `addPersona()`? In that case, I spoke too soon..... but now that I've said it, I still want to deprecate lol

PR #1569, opened 2024-10-22
Specify mediaType for request body in OkHttpRequestExecutionService

priettt:


embrace-android-delivery/src/main/kotlin/io/embrace/android/embracesdk/internal/delivery/execution/OkHttpRequestExecutionService.kt
```diff
@@ -81,10 +82,9 @@ class OkHttpRequestExecutionService(
     }
 
     private fun generateRequestBody(payloadStream: () -> InputStream) = object : RequestBody() {
-        override fun contentType() = null
+        override fun contentType() = "application/json".toMediaType()
```
bidetofevil, line 85: Lets create a constant for this and reuse it every time so we don't create a new object every time this is called?

bidetofevil (approved):
LGTM - added a comment about creating a constant MediaType that we can always refer to to save on object creation on every invocation.

PR #1574, opened 2024-10-23
Optimize performance of v2 delivery layer

fractalwrench:
## Goal

The v2 delivery layer was setting a network connectivity listener early in the bootstrapping which led to initializing OkHttp on the main thread, which is an expensive operation. To fix this I've set the listener at the end of SDK init on a background thread as it's not essential.

## Testing

## Profiling

Screenshot 2024-10-23 at 09 37 25
Screenshot 2024-10-23 at 09 37 33


bidetofevil (approved):
LGTM

PR #1576, opened 2024-10-23
Reduce visibility of swazzler hooks

fractalwrench:
## Goal

Reduces the visibility of the swazzler hooks as they don't need to be part of our public API.


priettt (approved):

priettt (approved):
Should we do the same with FirebaseSwazzledHooks?

PR #1580, opened 2024-10-23
Default the anr monitor thread executor to non-blocking in integration tests

bidetofevil:
## Goal

Using a blockable executor for the ANR worker means a lot of jobs were stuck by default before we needed to block it to simulate ANR capture. Since we were already setting the executor to blocking mode before we capture the ANR, we can set it to default to non-blockable so we only block the thread (when we need to).

This speeds up the execution of the tests as something in the `recordSession` flow, in the background/foreground callbacks, were taking ~1s each. Something was probably waiting for the something to execute, and then timing out at the 1s mark? Anyway, didn't look further as this fixes this tests.

I suspect any long running integration tests have the same issue


fractalwrench (approved):
LGTM

bidetofevil:
### Merge activity

* **Oct 24, 12:41 PM EDT**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/embrace-io/embrace-android-sdk/1580).

PR #1584, opened 2024-10-24
Add percentage based flags for okHttp usage and v2 storage

priettt:


embrace-android-payload/src/main/kotlin/io/embrace/android/embracesdk/internal/config/remote/KillSwitchRemoteConfig.kt
```diff
@@ -15,5 +15,7 @@ data class KillSwitchRemoteConfig(
     val jetpackCompose: Boolean? = null,
 
     @Json(name = "v2_storage")
-    val v2Storage: Boolean? = null
+    val v2StoragePct: Int? = null,
+    @Json(name = "use_okhttp")
+    val useOkHttpPct: Int? = null
```
fractalwrench, line 20: We should use `Float` as that allows for non-integer percentages if we want a very small rollout for whatever reason. We probably aren't going to use it as part of this but it's the convention of how other fields do it

PR #1585, opened 2024-10-24
Precursor cleanups before converting embrace.java to kotlin

fractalwrench:
## Goal

Makes some minor cleanups that will be required before `Embrace.java` can be converted to Kotlin.


bidetofevil (approved):
LGTM

PR #1586, opened 2024-10-24
Implement embrace interface in kotlin

fractalwrench:
## Goal

Implements the embrace interface in Kotlin rather than Java.

## Testing

Relied on existing test coverage & API checks.


embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/Embrace.kt
```diff
@@ -0,0 +1,477 @@
...
+        impl.start(
+            context,
+            appFramework
+        ) { _: io.embrace.android.embracesdk.internal.payload.AppFramework? -> null }
+    }
+
+    override val isStarted: Boolean
+        get() = impl.isStarted
+
+    override fun setAppId(appId: String): Boolean {
```
bidetofevil, line 75: Is this something we want to remove or at least deprecate? We have a lot of extra complexity because we allow this to happen, and I'm not sure it's that well tested - or used

bidetofevil (approved):
LGTM. Will we get rid of the duplicate overloads in a subsequent PR?

PR #1589, opened 2024-10-25
Periodic cache bug fixes

fractalwrench:
## Goal

Makes two changes to the periodic cache logic. Firstly, the correct unit is used for the caching interval (this was altered during our delivery refactor & hadn't shipped to prod). Secondly, if `delete` fails due to `RejectedExecution` it should be run on the current thread as in this condition the worker will have shutdown.


bidetofevil (approved):
LGTM

PR #1590, opened 2024-10-25
Remove unnecessary overload calls

fractalwrench:
## Goal

Always delegate directly to `EmbraceImpl` from `Embrace`.


bidetofevil (approved):
LGTM

PR #1591, opened 2024-10-25
Hide internal APIs from end-users

fractalwrench:
## Goal

Hides internal APIs by moving them to a separate module named `embrace-internal-api`. This module has a dependency on `embrace-android-sdk` so can access the necessary internals, but crucially end-users would not have this on the compile classpath unless they explicitly add it. This is the primary motivation behind this changeset.

The hosted SDKs (and our own modules) can add this dependency if they wish to access internal APIs.


bidetofevil (approved):
Glanced at but didn't look at in detail, but agree with approach, so LGTM where L is loosely defined.

PR #1596, opened 2024-10-29
Provide no-op implementations of internal interface

fractalwrench:
## Goal

Provides no-op implementations of the internal interfaces if the SDK is not initialized yet. This should hopefully avoid NPEs on the hosted SDK side of things & make it easier to move this functionality in future.

## Testing

Added unit tests.


bidetofevil (approved):
LGTM

PR #1597, opened 2024-10-29
Create internal api module

fractalwrench:
## Goal

Creates a new module for our internal API. This is not pulled in by default - our hosted SDKs will have to explicitly add the dependency to access internal APIs. This should reduce the amount of unwanted symbols that are currently exposed to end-users.


bidetofevil (approved):
LGTM

PR #1602, opened 2024-10-30
Removes internal symbols from public api

fractalwrench:
## Goal

Removes several internal symbols from the public API.


bidetofevil (approved):
LGTM

PR #1609, opened 2024-10-31
Reformat code according to editorconfig

fractalwrench:
## Goal

Performs a mass reformat of code according to our new editorconfig.


bidetofevil (approved):

bidetofevil (commented):
Did you also get rid of the override files too so we can start from scratch on those?

PR #1610, opened 2024-10-31
Disable tap coordinate capture by default

fractalwrench:
## Goal

Disables tap coordinate capture by default

## Testing


bidetofevil (approved):
LGTM

PR #1616, opened 2024-11-01
Remove obsolete remote configs

fractalwrench:
## Goal

Removes several obsolete remote configs that either refer to experiments that are no longer used, or that will never be used in prod. I've removed code related to experiments that we won't run which should reduce complexity in our SDK.

## Testing


bidetofevil (approved):
LGTM

PR #1617, opened 2024-11-01
Remove obsolete local configs

fractalwrench:
## Goal

Removes obsolete local configs as part of 7.0. These have been removed from the swazzler instrumentation too.

## Testing


bidetofevil (approved):
LGTM

PR #1619, opened 2024-11-04
Add example app

fractalwrench:
## Goal

Adds an initial skeleton for an example app. This only has examples for adding breadcrumbs + log messages, but has been written in a way that should make it easily extensible. Here is what I think we probably want to do next:

- Port over existing samples from internal test suite + NDK app
- Archive internal apps
- Update docs site to point at this example
- Improve OTel use-case (perhaps with a worked example in a separate app under the `examples` directory)
- Build the app on CI
- Make it straightforward to use the example app with a local version of the SDK + Swazzler


examples/ExampleApp/app/build.gradle.kts
```diff
@@ -0,0 +1,60 @@
...
+        targetSdk = 35
+        versionCode = 1
+        versionName = "1.0"
+
+        testInstrumentationRunner = "androidx.test.runner.AndroidJUnitRunner"
+    }
+
+    buildTypes {
+        release {
+            isMinifyEnabled = false
```
priettt, line 24:
```suggestion
            isMinifyEnabled = true
            isShrinkResources = true
```

nit:

examples/ExampleApp/settings.gradle.kts
```diff
@@ -0,0 +1,27 @@
...
+                includeGroupByRegex("com\\.android.*")
+                includeGroupByRegex("com\\.google.*")
+                includeGroupByRegex("androidx.*")
+            }
+        }
+        mavenCentral()
+        gradlePluginPortal()
+    }
+    val swazzler_version: String by settings
+    plugins {
```
priettt, line 14: shouldn't we add embrace like the rest of the toml plugins?

priettt (approved):
looks great!

PR #1625, opened 2024-11-04
Increase number of code examples

fractalwrench:
## Goal

Increases the code examples to have parity with the existing internal test suite. Remaining examples to-do are the tracing API + network requests.


examples/ExampleApp/app/src/main/kotlin/io/embrace/android/exampleapp/ui/examples/UserExample.kt
```diff
@@ -0,0 +1,72 @@
...
+import androidx.compose.runtime.getValue
+import androidx.compose.runtime.mutableStateOf
+import androidx.compose.runtime.remember
+import androidx.compose.runtime.setValue
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.unit.dp
+import io.embrace.android.embracesdk.Embrace
+
+@Composable
+fun UserExample() {
```
bidetofevil, line 18: I guess we'll still need this as long as it hasn't been officially deprecated?

bidetofevil (approved):
LGTM. Do we have a checklist of what we want to include so we can keep track of what we still want to add when we have time?

examples/ExampleApp/app/src/main/kotlin/io/embrace/android/exampleapp/ui/examples/UserExample.kt
fractalwrench: Yep - should be trivial enough to remove, and I've avoided less used features such as personas.

fractalwrench:
The remaining features to get parity with the internal test app are network requests + the tracing API, which I'll open a separate PR for shortly. There may be some additional examples we want to create (e.g. a pure OTel exporter integration) but that could probably come later and as a separate fixture

PR #1635, opened 2024-11-06
Remove config listeners

fractalwrench:
## Goal

Removes config listeners from the SDK. Config instead will be loaded during SDK initialization & be applied for the entirety of a process lifetime. This will make it simpler to reason about how the config endpoint affects the SDK.

## Testing


bidetofevil (approved):
LGTM. We should properly document this now that it's consistent.

PR #1636, opened 2024-11-06
Avoid SharedPreferences for storing config

fractalwrench:
## Goal

Avoids using SharedPreferences to store config values & rely on the cached HTTP response instead, if one exists.

## Testing


bidetofevil (approved):
OMG yes.

PR #1637, opened 2024-11-06
Extract config loading to separate class

fractalwrench:
## Goal

Extracts the logic for loading remote config from HTTP requests/cache out of `EmbraceConfigService` and into its own class. I've lifted this pretty much as-is and put it behind the `RemoteConfigSource` interface. My intention is to refactor the behavior in a subsequent PR.

## Testing


bidetofevil (approved):
LGTM

PR #1638, opened 2024-11-07
Create interface for instrumented config

fractalwrench:
## Goal

Creates an interface for the `InstrumentedConfig` class and separates out the implementation to use `*impl` as a suffix. This approach means the bytecode can still be instrumented in a relatively straightforward way, but that fake values can also be injected during our integration tests.

## Testing

Relied on existing test coverage & performed manual tests on a few local config values to confirm they are still respected.


bidetofevil (approved):
Scanned, didn't look in detail, but agree with the idea.

PR #1639, opened 2024-11-07
Added embrace-config-schema.json for tracking and documenting local configs

priettt:


embrace-config-schema.json
```diff
@@ -0,0 +1,274 @@
...
+      "description": "Enables NDK crash capture. Defaults to false.",
+      "type": "boolean"
+    },
+    "sdk_config": {
+      "type": "object",
+      "properties": {
+        "automatic_data_capture": {
+          "type": "object",
+          "properties": {
+            "memory_info": {
```
bidetofevil, line 28: Is basically doesn't work and is deprecated (or will be). Should we not include this here so it's not suggested?

embrace-config-schema.json
```diff
@@ -0,0 +1,274 @@
...
+              "type": "boolean"
+            }
+          },
+          "minProperties": 1
+        },
+        "app": {
+          "type": "object",
+          "properties": {
+            "report_disk_usage": {
+              "description": "The SDK collects the disk usage for the app. You can disable this if your app has a large number of local files to prevent excessive resource usage, especially on lower-end devices. Defaults to true.",
```
bidetofevil, line 183: I don't understand this description...

bidetofevil (approved):
LGTM. I have lots of comments about the individual descriptions but I'll just do a PR to change them later :)

priettt:
> LGTM. I have lots of comments about the individual descriptions but I'll just do a PR to change them later :)

@bidetofevil I'm using the descriptions that were already in the schema uploaded to schemastore, but we definitely need to improve them so go ahead!

embrace-config-schema.json
priettt: can I remove it from the swazzler and SDK code?

PR #1640, opened 2024-11-07
Alter integration tests to use config service

fractalwrench:
## Goal

Alters the integration tests to use the real config service instead of a fake. This is achieved by passing in fake versions of the remote + instrumented config.


bidetofevil (approved):
YES

PR #1641, opened 2024-11-07
Config service hardening

fractalwrench:
## Goal

Integration branch for hardening the config service.


bidetofevil (approved):
Implicitly approve integration branch, when it's ready

PR #1642, opened 2024-11-07
Move remote config tests

fractalwrench:
## Goal

Reorganises the integration tests so that remote config is stored in an individual source file.


bidetofevil (approved):
LGTM

PR #1647, opened 2024-11-08
Add assertions for checking config request made

fractalwrench:
## Goal

Adds assertions to check that a config request was attempted. Additionally boosted the timeout to 10s as this may have been contributing to flakiness on CI.

It's not possible to use this right now because HttpUrlConnection doesn't play well with the tests - once OkHttp is used in the config request it should make this more testable.


bidetofevil (approved):
LGTM. 

Maybe we should start logging the runtime of these tests or even in our API calls to see how much they fluctuate in CI. It's surprising that 2 seconds isn't enough, so I wonder if there are edge cases where the execution time of telemetry capture gets ridiculous

PR #1648, opened 2024-11-08
Add workflow to clear cache

fractalwrench:
## Goal

Adds a workflow to clear the Github Actions cache. We periodically need to do this when the build cache seems to get corrupted.


bidetofevil (approved):
FWIW, I get cache weirdness issues all the time when I switch branches, so I wonder if we are just not properly constructing the cache key and invalidating caches through code changes.

PR #1650, opened 2024-11-08
Avoid inlining runTest function

fractalwrench:
## Goal

Removes `inline` from `runTest` which stops a failure in the integration tests, specifically `EmbraceInternalInterfaceTest`.

I assume this is because test code was moved in this commit, and inlining the function means some property was not always setup correctly: https://github.com/embrace-io/embrace-android-sdk/pull/1640/files#diff-3c7d753cda41b1a11be4b6fd644cc07b341483a78ea6c29980f0382211887704


bidetofevil (approved):
WHY???

PR #1651, opened 2024-11-08
Add OkHttp implementation of RemoteConfigSource interface

fractalwrench:
## Goal

Adds an OkHttp implementation of the `RemoteConfigSource` interface. Ultimately this will be used to make requests to the config endpoint - for now it's not invoked in production code. The implementation does not deal with caching/scheduling the request, but the actual HTTP request & error handling should be considered complete for the purposes of review.

## Testing

Added unit tests

### DIFF
--- a/embrace-android-core/src/test/java/io/embrace/android/embracesdk/internal/config/OkHttpRemoteConfigSourceTest.kt
+++ b/embrace-android-core/src/test/java/io/embrace/android/embracesdk/internal/config/OkHttpRemoteConfigSourceTest.kt
@@ -1,0 +1,156 @@
+package io.embrace.android.embracesdk.internal.config
+
+import androidx.test.ext.junit.runners.AndroidJUnit4
+import io.embrace.android.embracesdk.fakes.TestPlatformSerializer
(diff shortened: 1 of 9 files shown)


embrace-android-core/src/test/java/io/embrace/android/embracesdk/internal/config/OkHttpRemoteConfigSourceTest.kt
```diff
@@ -0,0 +1,155 @@
...
+import org.junit.Assert.assertEquals
+import org.junit.Assert.assertNotNull
+import org.junit.Assert.assertNull
+import org.junit.Before
+import org.junit.Test
+import org.junit.runner.RunWith
+import org.robolectric.RobolectricTestRunner
+import java.util.concurrent.TimeUnit
+
+@RunWith(RobolectricTestRunner::class)
```
bidetofevil, line 21: `AndroidJUnit4.class` doesn't work for this?

bidetofevil (approved):
LGTM. You'll be adding the caching later?

PR #1655, opened 2024-11-11
Reorganise initial bootstrapping

fractalwrench:
## Goal

Reorganises the initial part of bootstrapping around the delivery layer as this will make it easier to refactor how the config request is made. As part of this I avoided calling SharedPreferences on a background thread.

## Testing


bidetofevil (approved):
LGTM. What were we reading from there before? The cached feature flag about whether the SDK should be initialized?

PR #1656, opened 2024-11-11
Extract url construction

fractalwrench:
## Goal

Moves the construction of URLs earlier into the bootstrapping process & avoids a circular dependency on the `ConfigService` that make it impossible to construct a URL ahead of time.

## Testing


bidetofevil (approved):
Oh I see why you want to delete the old delivery layer ASAP :-)

PR #1660, opened 2024-11-12
Fix okhttp implementation

fractalwrench:
## Goal

Fixes the OkHttp implementation to decode the response using gzip & updates the tests.

## Testing

Tested against the live endpoint in the example app.


embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/config/OkHttpRemoteConfigSource.kt
```diff
@@ -41,8 +43,11 @@ internal class OkHttpRemoteConfigSource(
         if (!response.isSuccessful) {
             return null
         }
-        val config = response.body?.source()?.inputStream()?.buffered()?.use {
-            serializer.fromJson(it, RemoteConfig::class.java)
+        val config = response.body?.source()?.use { src ->
```
bidetofevil, line 46: Is the tiny payload we get back worth gzipping? Or would it be annoying and not worthwhile do do something different here?

bidetofevil (approved):
LGTM. Comment about whether we should get a compressed response at all - probably find the way it is, just wanted to raise it.

bidetofevil (approved):
LGTM

embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/config/OkHttpRemoteConfigSource.kt
fractalwrench: I'm easy either way - let's discuss internally & update in a separate PR if necessary

PR #1662, opened 2024-11-12
Create combined source of remote config

fractalwrench:
## Goal

Creates a combined source that loads remote config & schedules HTTP requests that allowing config to be persisted in the background for future process launches.

I've simplified the request retry logic so that the SDK simply attempts to request config once per hour. IMO this should be sufficient given that config will no longer be required until the next process launch, and that the config response rarely changes anyway. 

## Testing

Added unit tests.


bidetofevil (approved):
LGTM. Agree with ripping out the retry. It's a best effort anyway and will be done at startup. If the app perpetually doesn't have a network when it starts up, there's the "try next hour" to catch it.

And if we are really concerned about that edge case, we can trigger this given a successful request sent if it fails the first time - but I don't think we should do that either.

PR #1663, opened 2024-11-12
Add etag support to okhttp

fractalwrench:
## Goal

Supports etags in the OkHttp config request. The etag header value is persisted so that it can be used across processes.

## Testing

Added a unit test & removed a flaky one.

### DIFF
--- a/embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/config/source/OkHttpRemoteConfigSource.kt
+++ b/embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/config/source/OkHttpRemoteConfigSource.kt
@@ -11,6 +11,7 @@ import io.embrace.android.embracesdk.internal.serialization.PlatformSerializer
 import io.embrace.android.embracesdk.network.http.HttpMethod
 import okhttp3.OkHttpClient
 import okhttp3.Request
+import okhttp3.Response
 import okio.GzipSource
 import okio.buffer
 import java.io.IOException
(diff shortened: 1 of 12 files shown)


embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/config/source/OkHttpRemoteConfigSource.kt
```diff
@@ -27,29 +28,47 @@
...
+        val response = call.execute()
+        return processResponse(response)
+    }
+
+    private fun prepareRequest(): Request {
         val url = apiUrlBuilder.resolveUrl(Endpoint.CONFIG)
         val headers = prepareConfigRequest(url).getHeaders()
         val builder = Request.Builder().url(url)
 
+        etag?.let {
```
bidetofevil, line 49: Just to make sure, will OkHttp actually short circuit the request and return without actually pinging the server if the etag matches what it has in its local cache?

bidetofevil (approved):
LGTM. One point of clarification on my understanding of whether OkHttp will use the etag to short-circuit the request, or do we assume we only send the actual request if the cache is stale?

embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/config/source/OkHttpRemoteConfigSource.kt
fractalwrench: Yes - there's test coverage for this use-case and I've also manually tested this out by launching the example app twice with a debugger attached, and confirmed a 304 response came from the live endpoint

fractalwrench: To further clarify how the etag implementation works:

1. Client makes request to config endpoint
2. Server responds 200 with config in body
3. Client reads etag from response header & saves it for later
4. Client makes another request to the config endpoint, which includes a header with the previous etag
5. Server responds 304 with no config in body
6. Client reads status code & knows that it should rely on whatever response was stored locally

PR #1665, opened 2024-11-12
Alter remote config implementation to use okhttp

fractalwrench:
## Goal

Alters the remote config implementation to use OkHttp. This PR hooks up all the alterations in PRs made downstream of this changeset & also deletes the old config implementation.

## Testing

Updated existing tests.


embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/config/source/CombinedRemoteConfigSource.kt
```diff
@@ -5,22 +5,20 @@ import io.embrace.android.embracesdk.internal.config.store.RemoteConfigStore
...
 
-    fun getConfig(): RemoteConfig? = cfg
+    fun getConfig(): RemoteConfig? = response?.cfg
 
     fun scheduleConfigRequests() {
-        worker.submit {
-            store.retrieveEtag()?.let(httpSource::setInitialEtag)
-        }
+        response?.etag?.let(httpSource::setInitialEtag)
         worker.scheduleWithFixedDelay(
```
bidetofevil, line 22: Do we even want to schedule this with a repeated job? Why not just schedule the job for an hour later when it completes? Seems unnecessarily complicated to maintain that extra bit of state

bidetofevil (approved):
LGTM. I don't know which review I mentioned this, but how does prioritization work if we are using the same OkHttp client for the RequestExecutionService and this? I assume if you're running this from a different thread, it won't affect each other?

fractalwrench:
The config implementation uses a `BackgroundWorker` which doesn't require prioritisation, so whatever gets submitted first gets executed first.

The `OkHttpRequestExecutionService` ultimately relies on the `SchedulingService` which I believe also uses a (different) `BackgroundWorker`, but is prioritised with a comparator.

In practical terms the config request will always happen first, and I think that's the desired behavior.

PR #1666, opened 2024-11-12
Disable SDK earlier in lifecycle

fractalwrench:
## Goal

Disables the SDK much earlier on in the bootstrapping process. Previously the disable call happened in an `onForeground()` callback which is non-deterministic. Due to the fact we now load config at startup it's possible to immediately check whether the SDK should be initialized & bail out of the initialization process much earlier.

## Testing

Relied on existing tests for now. A future changeset will add more extensive integration tests.


bidetofevil (approved):
This will still ensure the config request is fired off and allowed to complete even if the SDK is disabled, so a disabled instance can be "reanimated" in the next cold start?

fractalwrench:
Yep - that's the intention, and it should lock down everything else after that point. I will add integration tests to confirm this in a separate changeset.

PR #1667, opened 2024-11-12
Remove CpuInfo

priettt:
We aren't using cpu_name or egl_info anywhere in the dashboard or the backend, so it makes sense to stop collecting them.


bidetofevil (approved):
Nice. Does this allow us to further detail the loading of the native binaries until we install the signal handler?

priettt:
> Nice. Does this allow us to further detail the loading of the native binaries until we install the signal handler?

@bidetofevil probably, but I'll wait for the NDK refactor to review the loading

PR #1669, opened 2024-11-13
Add integration tests for remote config disabling sdk

fractalwrench:
## Goal

Adds integration tests that assert the remote config response controls whether the SDK is initialized or not. There are 2 scenarios tested here:

1. If the persisted config disables the SDK, a HTTP request will always be made whose response is persisted on disk
2. If the SDK starts normally, it will always attempt a HTTP request


bidetofevil (approved):
Nice! LGTM

PR #1670, opened 2024-11-13
Store remote config response uncompressed

fractalwrench:
## Goal

Stores the remote config response on disk uncompressed as it's relatively tiny & probably not worth the hit of gzipping.


embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/config/store/RemoteConfigStoreImpl.kt
```diff
@@ -42,7 +40,7 @@ internal class RemoteConfigStoreImpl(
 
     override fun saveResponse(response: ConfigHttpResponse) {
         try {
-            GZIPOutputStream(configFile.outputStream().buffered()).use { stream ->
+            configFile.outputStream().buffered().use { stream ->
```
bidetofevil, line 43: The data coming back from the server is not compressed, or is automatically ungzipped?

bidetofevil (approved):
LGTM

embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/config/store/RemoteConfigStoreImpl.kt
fractalwrench: The data is ungzipped in `OkHttpRemoteConfigSource`

PR #1671, opened 2024-11-13
Purge config before persisting response

fractalwrench:
## Goal

Purges any previous config before attempting to persist a response, when the response etag doesn't match what is currently on disk. This is meant to mitigate against the scenario where the SDK is disabled or has a bad config persisted, and saving the new response fails for whatever reason, thus trapping the SDK in a bad state.


embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/config/source/CombinedRemoteConfigSource.kt
```diff
@@ -28,6 +28,15 @@ class CombinedRemoteConfigSource(
     }
 
     private fun attemptConfigRequest() {
-        httpSource.getConfig()?.let(store::saveResponse)
+        httpSource.getConfig()?.let {
```
bidetofevil, line 31: Any worry that we might purge but not succeed? I wonder if we can overwrite, and purge only if the write fails?

bidetofevil (approved):
LGTM. Comment about purge/save ordering and whether we can do the purge if the overwrite doesn't succeed?

fractalwrench:
@bidetofevil I've reworked this PR so that now it deletes the file if an error happened when attempting to save the config. I think this is a better approach than before as it's less complicated but achieves the same goal.

PR #1672, opened 2024-11-13
perf: systrace improvements

fractalwrench:
## Goal

Adds some extra systrace calls for config bootstrapping.


bidetofevil (approved):
LGTM

PR #1677, opened 2024-11-14
Create OkHttp on background thread during initialization

fractalwrench:
## Goal

The SDK instantiates OkHttp on the main thread which hit perf. We should load it on a background thread instead.

baseline
Screenshot 2024-11-14 at 13 06 26


priettt (approved):
good catch 🙌

PR #1679, opened 2024-11-14
Add network_security_config.xml for Proxyman testing in demo app

priettt:
I need that network config in order to use Proxyman 😬


bidetofevil (approved):
LGTM

PR #1689, opened 2024-11-18
Alter how threshold works for disabling the SDK

fractalwrench:
## Goal

Alters how the SDK calculates whether it should be disabled by checking the last 6 digits of the device ID. This matches the behavior with iOS & allows for more granular control of behavior.

## Testing


bidetofevil (approved):
LGTM

PR #1698, opened 2024-11-20
Moved cleanOldFiles to NdkServiceRepository

priettt:


fractalwrench (approved):
Happy with this as an intermediate step to reduce the responsibilities of `EmbraceNdkService` - longer-term I think we'll break up `NdkServiceRepository` between the implementations of the new interfaces

PR #1701, opened 2024-11-20
Rename JNI functions

fractalwrench:
## Goal

Renames JNI functions to follow the Kotlin code style guide & to rename them so that they reflect that they delegate for JNI calls, rather than referencing the NdkService.

## Testing

Relied on existing instrumentation tests & manual test of NDK crash in example app.


priettt (approved):
thanks I hated the underscore 😂

PR #1702, opened 2024-11-20
Remove unityCrashId, as we don't use it anymore

priettt:
- The backend isn't considering any case where NDK and JVM crashes have the same ID. It might even bring issues.
- We weren't setting the ID correctly. If installSignalHandlers gets posted to the main thread handler before we set the ID, we will always use different IDs.
- I tested crashes on an armeabi-v7a device, and we only get native crashes. The CrashDataSource handler never gets called.


fractalwrench (approved):
LGTM

PR #1703, opened 2024-11-20
Add native crash handler installer

priettt:
- Moved the native crash directory creation to EmbraceStorageService. Also handled a possible exception while this is created.
- Created EmbraceNativeCrashHandlerInstaller, with the logic related to installed the signal handlers copied from EmbraceNdkService.

Some considerations when comparing EmbraceNdkService and this new EmbraceNativeCrashHandlerInstaller:
- symbolsForCurrentArch should be moved to the native thread sampler.
- unityCrashId was removed.
- I'm passing the activeSessionId as a class dependency in form of a string, instead of a provider from SessionIdTracker. If we needed the provider to avoid resolving this too quickly, this could be rolled back.
- Listeners that handle updates to session properties, ids, or other metadata should be added in another class. We should check what happens when they receive updates and the signals are not installed, though this shouldn't happen.
- We should review the logic of cleanOldCrashFiles. Why do we only want to clean old crash files when the signal handler is installed? If installSignals fail in the main thread, this won't return false. Should we clean old files too in that case?
- initializeService was always called in a background worker from the ModuleInitBootstraper. That's why I called the background worker in install()
- Added is32bit as a boolean dependency, to not pass the whole deviceArchitecture object.
- getNativeCrashErrors and the logic associated to it should be moved to another class.

### DIFF
--- a/embrace-android-features/src/main/kotlin/io/embrace/android/embracesdk/internal/injection/NativeFeatureModuleImpl.kt
+++ b/embrace-android-features/src/main/kotlin/io/embrace/android/embracesdk/internal/injection/NativeFeatureModuleImpl.kt
@@ -8,6 +8,7 @@ import io.embrace.android.embracesdk.internal.anr.ndk.NativeThreadSamplerInstall
 import io.embrace.android.embracesdk.internal.anr.ndk.NativeThreadSamplerService
 import io.embrace.android.embracesdk.internal.config.ConfigService
 import io.embrace.android.embracesdk.internal.crash.CrashFileMarkerImpl
+import io.embrace.android.embracesdk.internal.handler.AndroidMainThreadHandler
 import io.embrace.android.embracesdk.internal.ndk.EmbraceNdkService
 import io.embrace.android.embracesdk.internal.ndk.EmbraceNdkServiceRepository
 import io.embrace.android.embracesdk.internal.ndk.NativeCrashDataSourceImpl
@@ -119,6 +120,7 @@ internal class NativeFeatureModuleImpl(
                         payloadSourceModule.deviceArchitecture.is32BitDevice
                     )
                 },
+                mainThreadHandler = AndroidMainThreadHandler()
             )
         } else {
             null
--- a/embrace-android-features/src/main/kotlin/io/embrace/android/embracesdk/internal/injection/NativeFeatureModuleImpl.kt
+++ b/embrace-android-features/src/main/kotlin/io/embrace/android/embracesdk/internal/injection/NativeFeatureModuleImpl.kt
@@ -18,7 +18,6 @@ import io.embrace.android.embracesdk.internal.ndk.NativeCrashService
 import io.embrace.android.embracesdk.internal.ndk.NativeInstallMessage
 import io.embrace.android.embracesdk.internal.ndk.NdkService
 import io.embrace.android.embracesdk.internal.ndk.jni.JniDelegateImpl
-import io.embrace.android.embracesdk.internal.storage.StorageService
 import io.embrace.android.embracesdk.internal.utils.Uuid
 import io.embrace.android.embracesdk.internal.worker.Worker
 
@@ -112,14 +111,7 @@ internal class NativeFeatureModuleImpl(
                 repository = embraceNdkServiceRepository,
                 delegate = JniDelegateImpl(),
                 backgroundWorker = workerThreadModule.backgroundWorker(Worker.Background.IoRegWorker),
-                nativeInstallMessageProvider = {
-                    createNativeInstallMessage(
-                        storageModule.storageService,
-                        essentialServiceModule.sessionIdTracker.getActiveSessionId(),
-                        essentialServiceModule.processStateService.getAppState(),
-                        payloadSourceModule.deviceArchitecture.is32BitDevice
-                    )
-                },
+                nativeInstallMessage = nativeInstallMessage ?: return@singleton null,
                 mainThreadHandler = AndroidMainThreadHandler()
             )
         } else {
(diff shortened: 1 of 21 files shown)


embrace-android-features/src/main/kotlin/io/embrace/android/embracesdk/internal/ndk/EmbraceNativeCrashHandlerInstaller.kt
```diff
@@ -0,0 +1,108 @@
...
+import io.embrace.android.embracesdk.internal.crash.CrashFileMarkerImpl
+import io.embrace.android.embracesdk.internal.logging.EmbLogger
+import io.embrace.android.embracesdk.internal.logging.InternalErrorType
+import io.embrace.android.embracesdk.internal.storage.StorageService
+import io.embrace.android.embracesdk.internal.utils.Uuid
+import io.embrace.android.embracesdk.internal.worker.BackgroundWorker
+
+private const val HANDLER_CHECK_DELAY_MS = 5000
+
+internal class EmbraceNativeCrashHandlerInstaller(
```
fractalwrench, line 18: nit: perhaps this and future classes could be called NativeCrashHandlerInstaller**Impl**, or something else appropriate to how it's implemented?

priettt: I don't like the Impl suffix, but I'd rather use the same for everything, and I don't really like the Embrace prefix either. Naming for interfaces and implementations is usually pretty bad 😅

embrace-android-features/src/main/kotlin/io/embrace/android/embracesdk/internal/ndk/EmbraceNativeCrashHandlerInstaller.kt
```diff
@@ -0,0 +1,108 @@
...
+        Systrace.traceSynchronous("native-install-handlers") {
+            delegate._installSignalHandlers(
+                reportBasePath,
+                markerFilePath,
+                activeSessionId ?: "null",
+                appState,
+                Uuid.getEmbUuid(),
+                Build.VERSION.SDK_INT,
+                is32BitDevice,
+                false
```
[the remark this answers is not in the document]
priettt: great idea 👍

embrace-android-features/src/main/kotlin/io/embrace/android/embracesdk/internal/injection/NativeFeatureModuleImpl.kt
```diff
@@ -95,6 +103,52 @@ internal class NativeFeatureModuleImpl(
...
+                    )
+                },
+                mainThreadHandler = AndroidMainThreadHandler()
+            )
+        } else {
+            null
+        }
+    }
+
+    private fun createNativeInstallMessage(
```
fractalwrench, line 130: I'd suggest converting this to a val & adding a test case that checks it is populated with the correct values

embrace-android-features/src/main/kotlin/io/embrace/android/embracesdk/internal/ndk/NativeCrashHandlerInstallerImpl.kt
```diff
@@ -0,0 +1,94 @@
...
+        Systrace.traceSynchronous("native-install-handlers") {
+            val nativeInstallMessage = nativeInstallMessageProvider() ?: run {
+                logger.trackInternalError(
+                    InternalErrorType.NATIVE_HANDLER_INSTALL_FAIL,
+                    IllegalStateException("Native install message is null")
+                )
+                return
+            }
+            delegate.installSignalHandlers(
+                nativeInstallMessage.reportPath,
```
fractalwrench, line 82: nit: using `with(nativeInstallMessage)` might reduce the amount of code required here

embrace-test-fakes/src/main/kotlin/io/embrace/android/embracesdk/fakes/FakeMainThreadHandler.kt
```diff
@@ -0,0 +1,14 @@
...
+import io.embrace.android.embracesdk.internal.handler.MainThreadHandler
+
+class FakeMainThreadHandler : MainThreadHandler {
+    override fun postAtFrontOfQueue(function: () -> Unit) {
+        function()
+    }
+
+    override fun postDelayed(runnable: Runnable, delayMillis: Long) {
+        Thread.sleep(1000L)
+        runnable.run()
```
fractalwrench, line 12: IMO you can probably remove the sleep unless it's necessary for testing, and even then, I'd prefer using a CountdownLatch or very small delay (e.g. 1ms) as anything else will unnecessarily slow down the entire test suite

embrace-android-features/src/main/kotlin/io/embrace/android/embracesdk/internal/injection/NativeFeatureModuleImpl.kt
```diff
@@ -95,6 +103,52 @@ internal class NativeFeatureModuleImpl(
...
+    override val nativeCrashHandlerInstaller: NativeCrashHandlerInstaller? by singleton {
+        if (configModule.configService.autoDataCaptureBehavior.isNativeCrashCaptureEnabled()) {
+            NativeCrashHandlerInstallerImpl(
+                configService = configModule.configService,
+                sharedObjectLoader = nativeCoreModule.sharedObjectLoader,
+                logger = initModule.logger,
+                repository = embraceNdkServiceRepository,
+                delegate = JniDelegateImpl(),
+                backgroundWorker = workerThreadModule.backgroundWorker(Worker.Background.IoRegWorker),
+                nativeInstallMessageProvider = {
```
fractalwrench, line 115: This is my bad as I originally suggested a `Provider`, but if everything we need to create the `nativeInstallMessage` is within the module, then we can probably just pass in the created object directly?

fractalwrench (approved):
LGTM

embrace-android-features/src/main/kotlin/io/embrace/android/embracesdk/internal/injection/NativeFeatureModuleImpl.kt
priettt: Yeah, that'd probably be better. I don't think we need to wait for initialization to access sessionId, appState and is32BitDevice.

embrace-test-fakes/src/main/kotlin/io/embrace/android/embracesdk/fakes/FakeMainThreadHandler.kt
priettt: yeah, I wasn't too sure about this, but I wanted postDelayed to run after postAtFrontQueue

PR #1705, opened 2024-11-21
Extract native symbol logic to different class

fractalwrench:
## Goal

Consolidates the logic for getting native symbols from the resources value and adds appropriate test coverage.

## Testing

Added new unit tests.


priettt (approved):
Nice! I think we can move this directly to the NativeThreadSamplerService, as it's the only place where we use it

PR #1708, opened 2024-11-21
Use copy on write collection for tracking stored files

fractalwrench:
## Goal

Tracing of the delivery layer from flaky tests indicates that the payload store is erroneously telling the scheduling service that 0 payloads are persisted on disk. I'm suspicious that this might be due to the implementation of `storedFiles` in `getPayloadByPriority` as it uses a `ConcurrentSkipListSet`, whose operations are async. I'd hypothesise that in our tests and also in prod this can result in a race condition where payloads aren't delivered by the scheduling service on the first attempt (although subsequent attempts do result in delivery).

Altering the implementation to use `CopyOnWriteArraySet` should fix this if the theory is correct.


bidetofevil (approved):
LGTM

PR #1710, opened 2024-11-22
Remove unused NDK code

fractalwrench:
## Goal

Removes unused features from the NDK code. Specifically we were capturing the `android.native_crash.errors` and `android.native_crash.unwind_error` errors but these were unused. Additionally the SDK had code relating to a map file that is not used.

I have deleted all code related to these fields in the JVM + NDK layer.

## Testing


bidetofevil (approved):
LGTM

PR #1712, opened 2024-11-22
Add extra debug tracing for delivery layer

fractalwrench:
## Goal

Increases the information that will be present for why an envelope got scheduled.


bidetofevil (approved):
LGTM

PR #1713, opened 2024-11-22
Attempt to reduce flakes in remote config test

fractalwrench:
## Goal

Attempts to reduce flakes in remote config test by getting the worker thread that persists config & blocking until it has completed all its tasks. The tests did not wait for the background thread so could fail intermittently.


bidetofevil (approved):
LGTM

PR #1714, opened 2024-11-22
Split ndk service

fractalwrench:
## Goal

Splits the `NativeCrashProcessor` and `SymbolService` out of the `NdkService` implementation. There are no functional changes with this PR.

## Testing

Updated existing unit tests where required.


bidetofevil (approved):
LGTM

PR #1715, opened 2024-11-22
Cache empty envelope to be used for native crash resurretion

bidetofevil:
## Goal

Add the ability to cache an empty copy of each envelope type. Extend the overwriting facility of cached session envelopes to all types. With that, we cache an envelope that a native crash can use when we background and background activity is not enabled.

## Testing
Unit tests to verify everything works, and integration tests to ensure that only one envelope is persisted at a time


bidetofevil:
* **#1715** <a href="https://app.graphite.dev/github/pr/embrace-io/embrace-android-sdk/1715?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/embrace-io/embrace-android-sdk/1715?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`

<h2></h2>

embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/session/orchestrator/V2PayloadStore.kt
```diff
@@ -41,6 +41,17 @@ internal class V2PayloadStore(
         intakeService.take(envelope, createMetadata(type, payloadType = payloadType))
     }
 
+    override fun cacheEmptyCrashEnvelope(envelope: Envelope<LogPayload>) {
+        intakeService.take(
+            intake = envelope,
+            metadata = createMetadata(
+                type = SupportedEnvelopeType.CRASH,
+                complete = false,
```
fractalwrench, line 49: Previously we didn't have a concept of incomplete logs. I haven't spotted anything in the changeset that would concern me but it'd be worth reviewing how log persistence works at a high-level to see if this change breaks any assumptions.

embrace-android-sdk/src/integrationTest/kotlin/io/embrace/android/embracesdk/testcases/features/ResurrectionFeatureTest.kt
```diff
@@ -79,6 +88,47 @@ internal class ResurrectionFeatureTest {
...
+                cacheStorageServiceProvider = { cacheStorageService }
+            },
+            testCaseAction = {
+                recordSession()
+                recordSession()
+                recordSession()
+            },
+            assertAction = {
+                getSessionEnvelopes(3)
+                assertEquals(1, cacheStorageService.getCachedCrashEnvelope().size)
```
fractalwrench, line 127: It's probably worth enhancing the assertions to check the contents of the envelope at a basic level

embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/session/orchestrator/SessionOrchestratorImpl.kt
```diff
@@ -213,6 +213,12 @@ internal class SessionOrchestratorImpl(
                 }
                 Systrace.endSynchronous()
             }
+
+            if (activeSession == null && transitionType == TransitionType.ON_BACKGROUND) {
+                // if a new session hasn't been created when we background, cache an empty envelope to be used
+                // in case a native crash needs to be sent in the future after the current process dies
+                payloadStore?.cacheEmptyCrashEnvelope(payloadFactory.createEmptyLogEnvelope())
```
fractalwrench, line 220: How does the cache get cleaned up? Is that the responsibility of the resurrection service?

embrace-android-delivery/src/main/kotlin/io/embrace/android/embracesdk/internal/delivery/intake/IntakeServiceImpl.kt
```diff
@@ -23,8 +24,8 @@ class IntakeServiceImpl(
     private val shutdownTimeoutMs: Long = 3000,
 ) : IntakeService {
 
-    private var lastSessionCacheAttempt: Future<*>? = null
-    private var lastCacheSessionRef: StoredTelemetryMetadata? = null
+    private var cachingTasks: MutableMap<SupportedEnvelopeType, Future<*>> = ConcurrentHashMap()
+    private var cacheReferences: MutableMap<SupportedEnvelopeType, StoredTelemetryMetadata> = ConcurrentHashMap()
```
fractalwrench, line 28: nit: make these val

fractalwrench: Additionally consider make a private class that has a Future property + metadata property. That will avoid the possibility of getting into an inconsistent state where a value is in one collection but not another

embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/session/orchestrator/V2PayloadStore.kt
bidetofevil: Correct. In this patch I made sure things are cleaned up at resurrection time, and there shouldn't be anything that would generically use that payload.

There's coupling in the intake service that I'm not thrilled about with respect to cleaning up the "last" cached object, but it's also the best place since it can confirm when it's actually done. Maybe a call back mechanism should be added to facilitate this, but I think I want to wait until this `PayloadStore` concept is removed before doing that.

embrace-android-delivery/src/main/kotlin/io/embrace/android/embracesdk/internal/delivery/intake/IntakeServiceImpl.kt
bidetofevil: Yeah sounds good. I had originally started along that path but abandoned it because the generic lookups got harder to read than I wanted. But let me try it again to see if I feel different given you've raised this too.

embrace-android-sdk/src/integrationTest/kotlin/io/embrace/android/embracesdk/testcases/features/ResurrectionFeatureTest.kt
bidetofevil: Sure

bidetofevil: Actually was named confusingly - we don't actually care about the envelope that is persisted given that we fake the `PayloadStorageService`, so what we want to validate is only the metadata is correct, which I'll do.

embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/session/orchestrator/PayloadStore.kt
```diff
@@ -28,6 +29,12 @@ interface PayloadStore : CrashTeardownHandler {
      */
     fun storeLogPayload(envelope: Envelope<LogPayload>, attemptImmediateRequest: Boolean)
 
+    /**
+     * Stores an empty envelope for [PayloadType.NATIVE_CRASH] for future use. One one cached version of this should
```
priettt, line 33: nit: One one

embrace-android-delivery/src/main/kotlin/io/embrace/android/embracesdk/internal/delivery/intake/IntakeServiceImpl.kt
```diff
@@ -56,24 +57,33 @@ class IntakeServiceImpl(
...
+            val lastReference = cacheReferences[metadata.envelopeType]
 
             if (metadata.complete) {
                 schedulingService.onPayloadIntake()
                 deliveryTracer?.onPayloadIntake(metadata)
-            } else if (metadata.envelopeType == SupportedEnvelopeType.SESSION) {
-                lastCacheSessionRef = metadata
+            } else {
+                cacheReferences[metadata.envelopeType] = metadata
+                if (!cacheableEnvelopeTypes.contains(metadata.envelopeType)) {
```
priettt, line 67: shouldn't we check this before storing it to cacheReferences[]?

priettt (approved):

embrace-android-delivery/src/main/kotlin/io/embrace/android/embracesdk/internal/delivery/intake/IntakeServiceImpl.kt
bidetofevil: We still want to cache it - we also want to know that we are because it's unexpected. I want to err on the side of this working and notifying us about it, rather than rejecting it.

bidetofevil:
### Merge activity

* **Nov 27, 6:59 PM EST**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/embrace-io/embrace-android-sdk/1715).

PR #1716, opened 2024-11-25
Alter native crash processing code

fractalwrench:
## Goal

Minor refactor of how the native crash payload is loaded to avoid mutable state.

## Testing

Relied on existing tests.


priettt (approved):
👍

PR #1720, opened 2024-11-25
Remove metadata/appState code from the jni layer

priettt:
Metadata and app state will either be gathered from a previous session, or cached in the JVM layer without the need of accessing C code.


embrace-android-sdk/src/main/cpp/jnibridge/emb_ndk_manager.c
```diff
@@ -99,31 +96,21 @@ JNIEXPORT void JNICALL
 Java_io_embrace_android_embracesdk_internal_ndk_jni_JniDelegateImpl_updateMetaData(JNIEnv *env,
```
fractalwrench, line 96: Would be good to remove these functions if they're unused

priettt:
> LGTM. Had some suggestions for other bits to remove but happy for that to happen in either this PR or another changeset

@fractalwrench I will to remove the commented functions in a separate PR that removes EmbraceNdkService completely! 
As for the appState comment, I mostly wanted to leave it there so we don't forget to add the appState to the payload once the JVM cache is merged 😅

PR #1721, opened 2024-11-25
Extract file storage to separate class

fractalwrench:
## Goal

Extracts file storage to a separate class so that it can be used for storing NDK payloads too.

## Testing

Mostly relied on existing test coverage as `PayloadStorageService` is well tested.


priettt (approved):

embrace-android-delivery/src/main/kotlin/io/embrace/android/embracesdk/internal/delivery/storage/FileStorageService.kt
```diff
@@ -9,7 +10,22 @@ import java.io.InputStream
 interface FileStorageService {
```
bidetofevil, line 10: We we paramaterize the key type in case other things that are stored can't be addressed with a `StoredTelemetryMetadata` object?

bidetofevil (approved):

PR #1726, opened 2024-11-26
Remove NdkService

priettt:
Remove NdkService and its uses. Use the NativeSignalHandlerInstaller to install the signal handlers.


fractalwrench (approved):
LGTM although this will need a rebase

PR #1727, opened 2024-11-26
Move native crash installer to core module

fractalwrench:
## Goal

Moves the native crash installer to the core module. This is a pre-requisite for a later refactor.

## Testing


embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/internal/injection/ModuleInitBootstrapper.kt
```diff
@@ -350,10 +352,17 @@ internal class ModuleInitBootstrapper(
                     }
 
                     postInit(NativeFeatureModule::class) {
+                        val configService = configModule.configService
                         serviceRegistry.registerServices(
                             lazy { nativeFeatureModule.nativeThreadSamplerService }
                         )
-                        nativeFeatureModule.nativeCrashHandlerInstaller?.install()
+
+                        if (configService.autoDataCaptureBehavior.isNativeCrashCaptureEnabled()) {
```
priettt, line 360: Do we need to check this here? We're already checking the ndk config in the feature module. If it's not enabled, it'll return null.

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/internal/injection/ModuleInitBootstrapper.kt
```diff
@@ -350,10 +352,17 @@ internal class ModuleInitBootstrapper(
...
+                            val worker = workerThreadModule.backgroundWorker(Worker.Background.IoRegWorker)
+                            worker.submit {
```
priettt, line 362: We are passing a worker to the nativeCrashHandlerInstaller, and `install()` happens within that worker already.

embrace-android-features/src/test/java/io/embrace/android/embracesdk/internal/injection/NativeFeatureModuleImplTest.kt
```diff
@@ -94,8 +88,6 @@ internal class NativeFeatureModuleImplTest {
         )
 
         module = createNativeFeatureModule(fakeConfigModule)
-
-        assertNotNull(module.nativeCrashHandlerInstaller)
```
priettt, line 98: could we move the native crash handler tests to the core module?

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/internal/injection/ModuleInitBootstrapper.kt
fractalwrench: Removed this redundant check

PR #1728, opened 2024-11-27
Refactor integration tests to allow testing NDK

fractalwrench:
## Goal

Refactors the integration tests to allow testing the NDK end-to-end, rather than using fake modules for most of the functionality.

A future changeset will add extra test cases.


bidetofevil (approved):
Nice. LGTM

PR #1731, opened 2024-11-28
Add extra integration test cases for native crashes

fractalwrench:
## Goal

Adds extra integration test cases that verify the native crash functionality.


priettt (approved):
it's great to have this tested 🙌

PR #1732, opened 2024-11-28
Remove concept of key spans

fractalwrench:
## Goal

Removes the concept of key spans from the SDK. This attribute was used by the backend to aggregate spans but is no longer required given the backend checks for root spans instead.

## Testing

### DIFF
--- a/embrace-android-core/src/test/java/io/embrace/android/embracesdk/internal/spans/EmbraceTracerTest.kt
+++ b/embrace-android-core/src/test/java/io/embrace/android/embracesdk/internal/spans/EmbraceTracerTest.kt
@@ -1,9 +1,7 @@
 package io.embrace.android.embracesdk.internal.spans
 
 import io.embrace.android.embracesdk.arch.assertError
-import io.embrace.android.embracesdk.arch.assertIsKeySpan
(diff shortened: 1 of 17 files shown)


embrace-android-core/src/test/java/io/embrace/android/embracesdk/internal/spans/EmbraceTracerTest.kt
```diff
@@ -347,19 +345,13 @@ internal class EmbraceTracerTest {
...
         name: String,
-        traceRoot: Boolean = true,
         errorCode: ErrorCode? = null,
     ): EmbraceSpanData {
         val currentSpans = spanSink.completedSpans()
         assertEquals(1, currentSpans.size)
         val currentSpan = currentSpans[0]
         assertEquals(name, currentSpan.name)
         currentSpan.assertIsTypePerformance()
-        if (traceRoot) {
```
bidetofevil, line 358: We should probably keep this, but verify this via checking the parent of the span

fractalwrench: Updated assertion to check parentSpanId

bidetofevil (approved):
LGTM

PR #1735, opened 2024-11-29
Add debug tracing to scheduling service

fractalwrench:
## Goal

Enhance the delivery tracing by adding some extra events in the `SchedulingService`. If an integration test fails now it should give us a decent idea of what's going wrong.


bidetofevil (approved):
LGTM

PR #1736, opened 2024-11-29
Add auto termination enum

fractalwrench:
## Goal

Adds an enum that defines the valid states for span auto termination, and puts it at the appropriate places in the Tracing API.

## Testing

Updated binary compat dump & relied on existing test coverage to assert whether pre-existing code broke.


bidetofevil (approved):
LGTM

PR #1737, opened 2024-11-29
Propagate auto termination enum to span service

fractalwrench:
## Goal

Propagates the `AutoTerminationMode` enum to the `SpanService` and `EmbraceSpan` implementations so that a future changeset can setup automatic termination of spans.

## Testing

Mainly relied on existing test coverage, and will write integration tests in future when this feature actually does something.


embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/spans/SpanService.kt
```diff
@@ -18,6 +19,7 @@ interface SpanService : Initializable {
      */
     fun createSpan(
         name: String,
+        autoTerminationMode: AutoTerminationMode,
```
bidetofevil, line 22: These should probably all default to NONE as well?

bidetofevil (approved):
LGTM - one question about setting the default in the SpanService to NONE because after this is added, usage shouldn't have to change unless they want this additional logic, just like the public API.

embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/spans/SpanService.kt
fractalwrench: I was in two minds about this but ended up thinking I'd rather have this mode required for any caller in production code so that we don't forget to set it, at the cost of making our test code more verbose. Do you think the risk of the other approach is worth it?

PR #1745, opened 2024-12-04
Remove offset support from sampling logic

pablomatiasgomez:
## Goal

Remove offset support when deciding if the sdk is enabled or not, to match the iOS, and backend logic.

## Release Notes

**WHAT**:<br>
**WHY**:<br>


embrace-android-core/src/test/java/io/embrace/android/embracesdk/internal/config/behavior/SdkModeBehaviorImplTest.kt
```diff
@@ -2,8 +2,10 @@ package io.embrace.android.embracesdk.internal.config.behavior
 
 import io.embrace.android.embracesdk.fakes.createSdkModeBehavior
 import io.embrace.android.embracesdk.internal.config.remote.RemoteConfig
+import io.mockk.core.ValueClassSupport.boxedValue
```
fractalwrench, line 5: This import looks like it's not used, assuming that's the case it can be deleted

fractalwrench (approved):
LGTM

fractalwrench (approved):

PR #1746, opened 2024-12-05
Simplify otel limits

fractalwrench:
## Goal

Simplifies the limits enforced on OTel data capture by rolling custom and system attributes + events into one. This should make it simpler for pure OTel folks to send as much data as they want by configuring the SDK appropriately.

## Testing


bidetofevil (commented):
The reason we separated system and custom attributes is so we can have a lower limit for custom attributes (e.g. max 10 attributes per span) for customers that send us data.

With this new scheme how are we enforcing those lower limits?

PR #1747, opened 2024-12-05
Support otel export only flag

fractalwrench:
## Goal

Adds a new flag to the config with the key `sdk_config.otel_export_only`. The intention is that customers who don't want to send data to Embrace can set this flag to `true`, and the SDK will avoid sending any data to Embrace. This also alters what bytecode we instrument (i.e. data limits).

## Testing

Added an integration test that confirms this workflow isn't broken.


bidetofevil (approved):
LGTM

PR #1748, opened 2024-12-06
Allow auto termination of spans

fractalwrench:
## Goal

Implements auto-termination of spans. This is achieved by checking whether a span has `AutoTerminationMode.ON_BACKGROUND` just before the session payload is constructed, and terminating it and its descendants if this is true.

## Testing

Added integration tests.

### DIFF
--- a/embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/spans/SpanRepository.kt
+++ b/embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/spans/SpanRepository.kt
@@ -2,6 +2,7 @@ package io.embrace.android.embracesdk.internal.spans
 
 import io.embrace.android.embracesdk.internal.arch.schema.EmbType
 import io.embrace.android.embracesdk.internal.utils.lockAndRun
+import io.embrace.android.embracesdk.spans.AutoTerminationMode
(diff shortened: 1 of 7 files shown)


embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/spans/SpanRepository.kt
```diff
@@ -101,4 +102,50 @@ class SpanRepository {
     }
 
     private fun notTracked(spanId: String): Boolean = activeSpans[spanId] == null && completedSpans[spanId] == null
```
bidetofevil, line 104: Will you be adding unit tests that verify this works with different edge-case combinations of descendants which might change the traversal logic? Like if a child is no longer active, but the grand child is active?

bidetofevil (approved):
LGTM. Question about adding more test cases for different combinations of span states. We need to make sure that after backgrounding, all active spans belonging to traces whose root as auto terminate configured will be terminated, irrespective of the state of its ancestors.

fractalwrench:
@bidetofevil I added some extra test cases to the integration test & made some slight tweaks to the implementation, so putting this out for review again.

bidetofevil (approved):
LGTM

PR #1749, opened 2024-12-06
Fix auto termination propagation

fractalwrench:
## Goal

Fixes a few places where the `AutoTermination` enum wasn't getting passed through. I also removed it as a parameter from `recordCompletedSpan` as it doesn't make sense in the context of that function. I debated removing it from `recordSpan` but believe it makes sense to keep it there given that the code block could potentially run for a long time.

## Testing

Relied on existing test coverage. An upcoming PR will include the integration test that originally caught these problems.


bidetofevil (approved):
LGTM

PR #1757, opened 2024-12-13
Add ability to disable data export

fractalwrench:
## Goal

Adds a `disableDataExport` function to our public API that prevents exporting data via OTel exporters or HTTP requests. This calls `stop()` which shuts down all executors internally, and deletes any persisted data/blocks OTel exporters from exporting more data.

The intention is that an app would call this function if a user opted out of tracking halfway through a session, and on the next launch they would conditionally initialize the Embrace SDK.

## Testing

Added an integration test & ran manual tests in example app.

### DIFF
--- a/embrace-android-api/src/main/kotlin/io/embrace/android/embracesdk/internal/api/EmbraceAndroidApi.kt
+++ b/embrace-android-api/src/main/kotlin/io/embrace/android/embracesdk/internal/api/EmbraceAndroidApi.kt
@@ -74,4 +74,19 @@ public interface EmbraceAndroidApi {
      * @param name the name of the view to log
      */
     public fun endView(name: String): Boolean
+
+    /**
+     * If a user wishes to opt-out of exporting data to Embrace, you should:
+     *
+     * (1) Persist this user preference somewhere that can be readily accessed between processes
+     * (2) On the next process launch, read this preference & only initialize the Embrace SDK if you wish to capture data
+     * (3) Call this function if the SDK has already been initialized.
+     *
+     * When the SDK has already been initialized this function will prevent the SDK from exporting any further data
+     * via HTTP requests or OTel exports, and will delete any persisted data that has not yet been exported.
+     *
+     * The SDK makes a best effort attempt. Some data capture/handlers may remain active until the next process launch
+     * due to technical reasons, but any captured data will not be exported.
+     */
+    public fun disable()
 }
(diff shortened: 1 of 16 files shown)


embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/EmbraceImpl.kt
```diff
@@ -269,6 +271,23 @@ internal class EmbraceImpl @JvmOverloads constructor(
...
 
+    override fun disableDataExport() {
+        if (!sdkCallChecker.started.get()) {
+            return
+        }
+        bootstrapper.openTelemetryModule.openTelemetryConfiguration.disableDataExport()
+        stop()
+
+        // delete any persisted data
+        runCatching {
```
bidetofevil, line 282: Should we do the delete even if the SDK hasn't started yet?

embrace-android-api/src/main/kotlin/io/embrace/android/embracesdk/internal/api/EmbraceAndroidApi.kt
```diff
@@ -74,4 +74,19 @@ public interface EmbraceAndroidApi {
...
+     * (2) On the next process launch, read this preference & only initialize the Embrace SDK if you wish to capture data
+     * (3) Call this function if the SDK has already been initialized.
+     *
+     * When the SDK has already been initialized this function will prevent the SDK from exporting any further data
+     * via HTTP requests or OTel exports, and will delete any persisted data that has not yet been exported.
+     *
+     * The SDK makes a best effort attempt. Some data capture/handlers may remain active until the next process launch
+     * due to technical reasons, but any captured data will not be exported.
+     */
+    public fun disableDataExport()
```
bidetofevil, line 91: Should we also disable session caching? Or perhaps just change the storage components to not actually persist any new payloads to disk? Trying to figure out if there are areas that might lead to writing any new data to disk.

fractalwrench: Calling `stop()` shuts down all the workers which effectively stops all future data persistence. I can add a comment that makes this clearer

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/EmbraceImpl.kt
fractalwrench: Yep, this sounds reasonable.

embrace-android-api/src/main/kotlin/io/embrace/android/embracesdk/internal/api/EmbraceAndroidApi.kt
bidetofevil: Ah nice!

PR #1760, opened 2024-12-16
Use cached envelope data from dead process for native crash

bidetofevil:
## Goal

For native crashes, try to use the envelope resource/metadata from the associated session to reflect the app/SDK state when the crash happened. If a session is not found, use the cached crash envelope that is written when there are no sessions to get the envelope data. Failing that, use the current SDK's envelope, but log an internal error.

This is done through using a new component, `CachedLogEnvelopeStore`, that will be be used by the `PayloadResurrectionService` to create and cache the appropriate envelope for a native crash, and then by `LogEnvelopeSource` to consume the envelope.

The workflow goes thusly:

1. Session resurrection: create and cache envelope for native crash associated with a resurrected session
2. Native crash log creation: get envelope for the given crash's `StoredTelemetryMetadata`
3. Left over native crashes (only expect 1) will use the resource and metadata off the cached crash envelope to produce the envelope for it the crash.
4. Send session-less crash
5. Clean up any left over native crashes and envelopes

## Testing

Unit and integration tests added to verify the case where a session is found, where a session isn't found but the cached crash envelope exists, as well as when neither are there.


embrace-android-core/src/test/java/io/embrace/android/embracesdk/internal/envelope/log/LogEnvelopeSourceImplTest.kt
```diff
@@ -1,57 +1,82 @@
...
+import io.embrace.android.embracesdk.internal.opentelemetry.embProcessIdentifier
 import io.embrace.android.embracesdk.internal.payload.LogPayload
+import io.embrace.android.embracesdk.internal.spans.findAttributeValue
+import io.opentelemetry.semconv.incubating.SessionIncubatingAttributes
 import org.junit.Assert.assertEquals
 import org.junit.Assert.assertTrue
 import org.junit.Before
 import org.junit.Test
 
 internal class LogEnvelopeSourceImplTest {
```
bidetofevil, line 26: These tests were quite wrong before so they were rewritten to verify the right things, plus what was added in this PR

fractalwrench (approved):
LGTM

bidetofevil:
### Merge activity

* **Dec 17, 5:21 PM EST**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/embrace-io/embrace-android-sdk/1760).

PR #1766, opened 2024-12-19
Prevent null session ID crashing ndk processing

fractalwrench:
## Goal

Prevents a null session ID from crashing in the JNI due to not checking for null. I fixed this by serializing `null` instead, which is the previous behavior before our NDK refactors.


priettt (approved):

bidetofevil (approved):
My original fix was done on the native layer but this makes more sense lol

PR #1768, opened 2024-12-19
Update swazzler test workflow

fractalwrench:
## Goal

Updates the swazzler-test workflow due to repo reorganisation.

The ref in the workflow needs to be updated before merging to main, but this allows us to test the workflow now.


bidetofevil (approved):
Nice

PR #1784, opened 2024-12-28
Avoid out of bound access to emb_exception->stacktrace

priettt:
Tested by reducing kEMBMaxSFrames to 3, and forcing a native crash. Before this fix, either the native crash is lost due to an exception, or the native crash is sent but with trash data (things like sessionId containing a stack frame or other weirdness).

Now, the native crash is sent correctly, but capped to 3.

In the future, we might want to: 
- Increase the kEMBMaxSFrames limit.
- Add some kind of warning to let the user know that the stacktrace was limited.


bidetofevil (approved):
`kEMBMaxSFrames` is defined to be 100 right? Do we know why debug/release show dramatic differences between the number of frames that are captured?

fractalwrench (approved):
LGTM

PR #1785, opened 2024-12-28
strncpy out of bounds fix

priettt:
We use strncpy in many places. In almost all places, we call it with the size of the destination array. That means something like this:

```
char destinationArray[3]
char sourceArray[5]
emb_strncpy(destinationArray, sourceArray, sizeof(destinationArray))
```

In that case, sizeof(destinationArray) will be 3. That means the while loop will execute until 4, and we will try to write `src[3]` to `dst[3]`, which is out of bounds.

While C might let you write out of bounds, it may cause memory corruption or failure. 

Changes: 
- Modified the while loop so `i` won't be equal to `destination_len`.
- Added a `source[i] != '\0'` verification in case the source is smaller than the destination. Strings passed in src should end in that termination char. If they don't, we might face another OOB. We could solve this by also passing source_len as a parameter.
- Added some test cases.
- Improved naming.

### DIFF
--- a/embrace-android-sdk/src/main/cpp/utils/string_utils.c
+++ b/embrace-android-sdk/src/main/cpp/utils/string_utils.c
@@ -3,23 +3,22 @@
 
 /**
  * Copy a string from source to destination. If the source is larger than the destination, it will be truncated.
+ * The destination buffer will always be null-terminated.
  * @param destination The destination buffer.
  * @param source The source buffer.
  * @param destination_len The length of the destination buffer.
  */
 void emb_strncpy(char *destination, const char *source, size_t destination_len) {
-    if (destination == NULL || source == NULL) {
+    if (destination == NULL || source == NULL || destination_len == 0) {
         return;
     }
-    int i = 0;
-    while (i < destination_len && source[i] != '\0') {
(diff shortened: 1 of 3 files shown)


bidetofevil (approved):
welp

embrace-android-sdk/src/main/cpp/utils/string_utils.c
```diff
@@ -1,14 +1,20 @@
...
+ */
+void emb_strncpy(char *destination, const char *source, size_t destination_len) {
+    if (destination == NULL || source == NULL) {
         return;
     }
     int i = 0;
-    while (i <= len) {
-        char current = src[i];
-        dst[i] = current;
+    while (i < destination_len && source[i] != '\0') {
```
fractalwrench, line 15: This change means the null termination character is not copied to the destination array, which could cause various problems downstream as lots of C code assumes that strings must have that character.

fractalwrench (changes requested):
Good catch on the off-by-one problem. I think we need to make one change to how the string is copied over, then this is good to go

fractalwrench (approved):

PR #1803, opened 2025-01-10
Implement business logic for file attachment

fractalwrench:
## Goal

Implements the core business logic that will be used for uploading file attachments and associating them with log records. Key concepts include:

1. There are 2 types of file attachment: those uploaded to Embrace's server, and those uploaded to a user-hosted server
2. The attributes and input from our public API will be different for these 2 types but share a lot of commonality, so I've chosen to represent them as a sealed class
3. The attachment is validated to check that its size and per-session limit has not been exceeded; an error code is set if this is the case

## Testing

Added unit tests.


embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/logs/attachments/Attachment.kt
```diff
@@ -0,0 +1,89 @@
...
+        }
+
+        override val attributes: Map<String, String> = constructAttributes(size, id, errorCode)
+    }
+
+    /**
+     * An attachment that is uploaded to a user-supplied backend.
+     */
+    class UserHosted(
+        size: Long,
```
bidetofevil, line 63: I think this may not have to be specified?

embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/logs/attachments/Attachment.kt
```diff
@@ -0,0 +1,89 @@
...
+        id: String,
+        val url: String,
+        counter: AttachmentCounter,
+    ) : Attachment(size, id, counter) {
+
+        private val errorCode: AttachmentErrorCode? = when {
+            !counter.incrementAndCheckAttachmentLimit() -> OVER_MAX_ATTACHMENTS
+            size < 0 -> UNKNOWN
+            url.isEmpty() -> UNKNOWN
+            isNotUuid() -> UNKNOWN
```
bidetofevil, line 73: Do we require UUID if customers are self hosting and validating? Seems unnecessarily strict.

embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/logs/attachments/AttachmentCounter.kt
```diff
@@ -0,0 +1,21 @@
+package io.embrace.android.embracesdk.internal.logs.attachments
+
+import io.embrace.android.embracesdk.internal.session.lifecycle.ProcessStateListener
+import java.util.concurrent.atomic.AtomicInteger
+
+/**
+ * Counts the number of attachments that should be added to log records.
+ */
+class AttachmentCounter(private val limit: Int = 5) : ProcessStateListener {
```
bidetofevil, line 9: Can we somehow tie this into session transition, or better yet use the `DataCaptureService` infra to manage this?

bidetofevil (approved):
Code looks good to me, but there are some requirement things that I have questions about that we should clarify first, especially regarding the user-hosted attachments.

Regarding tying to the counter to the session orchestrator/boundaries, feel free to ignore if there are reasons why that makes it difficult. But ideally, we won't have to worry about another thing that is slightly different from the session lifecycle but ought to be coupled to it

embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/logs/attachments/Attachment.kt
fractalwrench: The design doc states it should be supplied for user-hosted

embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/logs/attachments/Attachment.kt
fractalwrench: Currently we do require a UUID

embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/logs/attachments/AttachmentCounter.kt
fractalwrench: Yep, this is a good idea. The holiday break has expunged a lot of context from my brain, I think the session orchestrator probably is the best place to achieve this.

PR #1804, opened 2025-01-10
File attachment integration branch

fractalwrench:
## Goal

Adds an API for logging a message with a file attachment. This API is not fully functional yet and only creates a log message in the existing way, without any attachment.

## Testing

Added skeleton of integration tests.


bidetofevil (approved):
LGTM

PR #1805, opened 2025-01-10
Refactor log service implementation

fractalwrench:
## Goal

Initial refactor of the `LogService` to make it easier when we add the attachment upload feature.

## Testing

### DIFF
--- a/embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/logs/EmbraceLogService.kt
+++ b/embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/logs/EmbraceLogService.kt
@@ -32,16 +32,11 @@ class EmbraceLogService(
     private val serializer: PlatformSerializer,
 ) : LogService {
 
+    private val behavior = configService.logMessageBehavior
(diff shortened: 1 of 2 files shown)


embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/logs/EmbraceLogService.kt
```diff
@@ -32,16 +32,11 @@ class EmbraceLogService(
...
-        ),
-        Severity.ERROR to LogCounter(
-            configService.logMessageBehavior::getErrorLogLimit
-        )
+        Severity.INFO to LogCounter(behavior::getInfoLogLimit),
+        Severity.WARNING to LogCounter(behavior::getWarnLogLimit),
+        Severity.ERROR to LogCounter(behavior::getErrorLogLimit)
     )
 
     override fun log(
```
bidetofevil, line 42: This is already an improvement, but it feels like this interface is too generic and in a later PR, we may want to spilt it off so we know right off the bat what we are dealing with

bidetofevil (approved):
LGTM

PR #1814, opened 2025-01-13
Implement user hosted file attachments

fractalwrench:
## Goal

Implements user hosted file attachments. These are essentially log messages containing attributes that allow us to identify how an attachment was uploaded to a user-hosted file storage solution. User-hosted attachments can follow the regular log message flow; embrace-hosted attachments require extra logic which will be added in future PRs.

## Testing

Added integration tests.


embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/logs/attachments/Attachment.kt
```diff
@@ -12,24 +13,23 @@ import java.util.UUID
...
     val id: String,
-    val counter: AttachmentCounter,
 ) {
 
     internal companion object {
-        const val ATTR_KEY_SIZE = "emb.attachment_size"
-        const val ATTR_KEY_URL = "emb.attachment_url"
-        const val ATTR_KEY_ID = "emb.attachment_id"
-        const val ATTR_KEY_ERR_CODE = "emb.attachment_error_code"
+        val ATTR_KEY_SIZE = AttributeKey.stringKey("emb.attachment_size")
```
bidetofevil, line 19: If we're going to make them objects we can probably define them in `EmbraceAttributeKeys` like the others. It has some nice properties and make it easier for testing.

bidetofevil (approved):
LGTM. One comment about defining Embrace attribute keys

PR #1823, opened 2025-01-16
Tweak log attachment implementation after manual testing

fractalwrench:
## Goal

This changeset contains a few tweaks to the log attachment implementation that were required after carrying out manual testing:

- Created screen in example app for log messages with attachments
- Used 'attachments' instead of 'attachment' as endpoint
- Logged a message if an attachment can't be sent due to limits


embrace-android-payload/src/main/kotlin/io/embrace/android/embracesdk/internal/comms/api/Endpoint.kt
```diff
@@ -10,6 +10,6 @@ enum class Endpoint(
     LOGS("logs", "v2"),
     SESSIONS("spans", "v2"),
     CONFIG("config", "v2"),
-    ATTACHMENT("attachment", "v2"),
+    ATTACHMENT("attachments", "v2"),
```
priettt, line 13: should we also rename the enum type name? ATTACHMENT -> ATTACHMENTS

priettt (approved):

PR #1828, opened 2025-01-17
Add log message attachment example

fractalwrench:
## Goal

Adds a log message attachment sample to the example app. As this relies on a new API it won't compile without a local artefact for now, so I'll avoid merging this until the feature has shipped.


bidetofevil (approved):
LGTM

PR #1840, opened 2025-01-24
Update build-rc-workflow.yml

priettt:
Let's try it like this, and if it doesn't work, it should be without any embrace-android-sdk/


bidetofevil (approved):
LGTM

PR #1847, opened 2025-01-27
Move gradle plugin check to a different step in build-rc-workflow.yml

priettt:
If we run check along publishToSonatype, the gradle integration test plugin won't be installed, and the tests will fail


bidetofevil (approved):
LGTM

PR #1854, opened 2025-01-28
Add proguard rules for FCM capture

priettt:
Without those rules, enabling FCM capture with a proguard build resulted in a compilation error


bidetofevil (approved):
LGTM

PR #1856, opened 2025-01-29
Enable attachment example code

fractalwrench:
## Goal

Enable attachment example code after 7.1.0 is released to avoid compilation errors.


bidetofevil (approved):
LGTM

PR #1864, opened 2025-02-05
Disable daily run of publish-snapshot.yml

priettt:
It's been failing for months so I'm guessing nobody really uses it


fractalwrench (commented):
Could we simply delete this workflow given that nobody is using it?

PR #1865, opened 2025-02-05
Update build-rc-workflow.yml to use inputs instead of github.events.input

priettt:
workflow_call wasn't working, and now GitHub Actions has unified inputs, so you can use inputs for both workflow_call and workflow_dispatch


fractalwrench (approved):
LGTM

PR #1866, opened 2025-02-05
Add proguard rule

fractalwrench:
## Goal

Adds a ProGuard rule that retains config that undergoes bytecode instrumentation.


bidetofevil (approved):
LGTM

PR #1869, opened 2025-02-06
Expose OpenTelemetry span and trace ID

devgianlu:
## Goal

Expose an API to access the OpenTelemetry trace and span IDs. This allows greater flexibility when integrating with custom observability components. For example, this allows us to create `traceparent` headers for distributed tracing.

## Testing

Not much to test.

### DIFF
--- a/embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/spans/InternalTracer.kt
+++ b/embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/spans/InternalTracer.kt
@@ -97,6 +97,14 @@ class InternalTracer(
             value = value
         ) ?: false
 
+    override fun getSpanUnderlyingSpanId(spanId: String): String? {
+        return spanRepository.getSpan(spanId = spanId)?.spanId
+    }
+
+    override fun getSpanUnderlyingTraceId(spanId: String): String? {
+        return spanRepository.getSpan(spanId = spanId)?.traceId
+    }
+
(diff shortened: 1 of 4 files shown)


bidetofevil:
Hi @devgianlu - thanks for your contribution!

If you are looking for the `traceId` from the Android SDK, you can get it via an attribute on the `EmbraceSpan` object that is returned, so I suspect you are trying to use it through one of our other SDKs, like Unity or Flutter?

If that's the case, the idea is to expose it through that, as those SDKs should have access to that information already.

Or do you really need to look up a traceId based solely on the `spanId` no other references?

devgianlu:
You are correct, the work is mainly focused on getting access to trace and span IDs from the Flutter SDK. The complementary work can be found [here](https://github.com/embrace-io/embrace-flutter-sdk/compare/main...VolumioTeam:embrace-flutter-sdk:underlying-span-trace-ids) (iOS only, no Android). 

While for iOS I was able to get them directly without changing the platform specific SDK, I was not able to do so for Android thus the addition of these accessors, but I might have missed something.

bidetofevil:
The next Flutter SDK will be updated to expose the correct span and trace IDs for both iOS and Android. This should make this PR redundant?  Once it's out and you've verified that it works, I can close this. Is that OK @devgianlu?

devgianlu:
> Once it's out and you've verified that it works, I can close this. Is that OK @devgianlu?

Yes, of course!

bidetofevil:
For posterity, Embrace Flutter SDK 4.0.0 was released and addressed the issue of getting the right spanId and traceId from the object at the Flutter layer. This can be closed as it is not necessary anymore.

PR #1875, opened 2025-02-07
Refactor build plugin

fractalwrench:
## Goal

Replaces the precompiled build scripts with a build plugin. This adds a plugin extension which allows for greater configurability, which will be necessary for when we're also supporting gradle plugin modules in the project.


priettt (approved):
👍

PR #1876, opened 2025-02-07
Tweak internal plugin setup

fractalwrench:
## Goal

Makes some further tweaks to the internal plugin to make it possible to apply this to JVM modules.

### DIFF
--- a/buildSrc/build.gradle.kts
+++ b/buildSrc/build.gradle.kts
@@ -1,3 +1,4 @@
+import org.jetbrains.kotlin.gradle.dsl.JvmTarget
 import org.jetbrains.kotlin.gradle.tasks.KotlinCompile
 
 group = "io.embrace.internal"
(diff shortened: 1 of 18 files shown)


buildSrc/build.gradle.kts
```diff
@@ -41,8 +42,8 @@ project.tasks.withType(JavaCompile::class.java).configureEach {
 }
 
 // ensure the Kotlin + Java compilers both use the same language level.
-project.tasks.withType(KotlinCompile::class.java).configureEach {
-    kotlinOptions {
-        jvmTarget = JavaVersion.VERSION_1_8.toString()
+kotlin {
+    compilerOptions {
+        jvmTarget.set(JvmTarget.JVM_1_8)
```
priettt, line 47: isn't our min required version Java 11?

priettt (approved):

buildSrc/build.gradle.kts
fractalwrench: For building the plugin, yes. For the SDK we still use 1.8. I'll bump this in a separate PR

PR #1898, opened 2025-02-13
Add workflow to bump example app version

fractalwrench:
## Goal

Adds a workflow that opens a PR to bump the example app version.


bidetofevil (approved):
LGTM

PR #1900, opened 2025-02-13
Update artifact upload workflow

fractalwrench:
## Goal

Updates the artifact upload workflow to use the gradle task used for publishing the gradle plugin artifact.


bidetofevil (approved):
LGTM

PR #1908, opened 2025-02-14
Configure codecov plugin

fractalwrench:
## Goal

Configures the codecov plugin so that coverage is aggregated into one report rather than just using coverage from one module.

I wasn't able to get this working instantly with the gradle plugin module, [possibly due to this issue](https://github.com/Kotlin/kotlinx-kover/issues/728), so I'd suggest we just stick with our previous behavior & fix this when I have a bit more time to investigate what's going wrong.


bidetofevil (approved):
LGTM

PR #1926, opened 2025-02-20
Remove symbolsDir and projectType from EmbraceExtensionInternal and pass them directly to NdkTaskRegistration

priettt:
## Goal

symbolsDir and projectType are only used in NdkUploadTaskRegistration so I added them as parameters wrapped in providers to not realize them too soon.

## Testing

Relied in existing testing

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,8 +1,8 @@
 [*]
 insert_final_newline = true
-max_line_length = 120
+max_line_length = 140
 ij_continuation_indent_size = 4
-ij_visual_guides = 120
+ij_visual_guides = 140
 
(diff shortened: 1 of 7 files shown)


fractalwrench (approved):
LGTM if the editorconfig change can be removed from this changeset - that could be proposed in a separate PR if it's intentional

priettt:
> LGTM if the editorconfig change can be removed from this changeset - that could be proposed in a separate PR if it's intentional

oh I thought that was local or gitignored. I do think we should keep consistency (it's 140 everywhere else), but I'll do it in another PR

PR #1932, opened 2025-02-21
Throw an exception when building with a Gradle version lower than the minimum supported

priettt:
## Goal

Validate that the version used to build an app that uses Embrace is higher or equal than the minimum

## Testing

Added an integration test that validates this behavior


fractalwrench (approved):
LGTM. Do you plan to make a similar check against AGP's version?

PR #1941, opened 2025-02-24
Prevent activity leak in startup tracking

fractalwrench:
## Goal

Prevents a leak in startup tracing where an activity reference was held in an anonymous inner class, leading to a potential leak reported via LeakCanary.

## Testing

Tested scenario via LeakCanary and confirmed that it happened before the changeset but not after.


bidetofevil (approved):
Hang my head in shame as this was something I explicitly called out in my UI Load code, but somehow snuck in in the app startup refactor.

PR #1942, opened 2025-02-24
Throw an exception when building with a Gradle version lower than the minimum supported

priettt:
## Goal

Validate that the AGP version used to build an app that uses Embrace is higher or equal to the minimum supported.

## Testing

Validated that an exception is thrown using a custom test app. Could not add an integration test.


fractalwrench (approved):
LGTM

PR #1943, opened 2025-02-24
Update editorconfig to make it consistent with detekt and checkstyle

priettt:


fractalwrench (approved):
LGTM, would be good to get the opinion of @bidetofevil too

bidetofevil (approved):
I'm surprised this wasn't already set as that. Looks good!

PR #1950, opened 2025-02-26
Add failBuildOnUploadErrors gradle plugin property

priettt:
## Goal

If failBuildOnUploadErrors is enabled, any error during an upload task in the gradle plugin will fail the build. If it's not, build should continue as usual.

## Testing

Will add tests on specific tasks that use this property


fractalwrench (approved):
LGTM

PR #1961, opened 2025-02-27
Rollback naming logic for JvmMappingUpload tasks

priettt:
## Goal

An error was reported in 7.1.0 where multiple jvmMapping tasks are registered with the same name. This change rolls back the naming logic to the one we used in 7.0.0, though keeping the new task name.

## Testing

Relied on existing tests. Will add tests for dexguard next week.


fractalwrench (approved):
LGTM

PR #1969, opened 2025-02-28
Added CollectSharedObjectFilesTask

priettt:
## Goal

This is one of the 4 tasks NdkUploadTask will get separated into. The idea for this task is to find .so files and output them so the next task can grab them.

Some questions which I also added as TODOs in the code to resolve:
- Is it possible for a native project to not have .so files?
- Is it possible for a Unity project to not have .so files?
- Is checking `embraceConfig.ndkEnabled` at the start of the registration enough, or should we also pass it to `task.onlyIf { }`

## Testing

Will add tests in a following commit.

### DIFF
--- a/embrace-gradle-plugin/src/main/java/io/embrace/android/gradle/plugin/tasks/ndk/CollectSharedObjectFilesTask.kt
+++ b/embrace-gradle-plugin/src/main/java/io/embrace/android/gradle/plugin/tasks/ndk/CollectSharedObjectFilesTask.kt
@@ -1,0 +1,96 @@
+package io.embrace.android.gradle.plugin.tasks.ndk
+
+import io.embrace.android.gradle.plugin.config.ProjectType
+import io.embrace.android.gradle.plugin.config.UnitySymbolsDir
+import io.embrace.android.gradle.plugin.tasks.EmbraceTask
+import io.embrace.android.gradle.plugin.tasks.EmbraceTaskImpl
+import io.embrace.android.gradle.plugin.tasks.il2cpp.UnitySymbolFilesManager
+import org.gradle.api.file.ConfigurableFileCollection
+import org.gradle.api.file.DirectoryProperty
+import org.gradle.api.model.ObjectFactory
+import org.gradle.api.provider.Property
+import org.gradle.api.tasks.Input
+import org.gradle.api.tasks.InputDirectory
+import org.gradle.api.tasks.InputFiles
+import org.gradle.api.tasks.Optional
+import org.gradle.api.tasks.OutputFiles
+import org.gradle.api.tasks.SkipWhenEmpty
+import org.gradle.api.tasks.TaskAction
+import java.io.File
+import javax.inject.Inject
+
+/**
+ * Collects .so files from the project, depending on the project type.
+ */
+abstract class CollectSharedObjectFilesTask @Inject constructor(
+    objectFactory: ObjectFactory,
+) : EmbraceTask, EmbraceTaskImpl(objectFactory) {
+
+    @get:Input
+    val projectType: Property<ProjectType> = objectFactory.property(ProjectType::class.java)
+
+    @get:Input
+    val shouldFailBuildOnUploadErrors: Property<Boolean> = objectFactory.property(Boolean::class.java)
+
+    @get:Input
+    @get:Optional
+    val unitySharedObjectsDir: Property<UnitySymbolsDir?> = objectFactory.property(UnitySymbolsDir::class.java)
+
+    @get:InputDirectory
+    @get:Optional
+    val decompressedUnitySharedObjectDirectory: DirectoryProperty = objectFactory.directoryProperty()
+
+    @get:InputFiles
+    @get:SkipWhenEmpty
+    val architecturesDirectoryForNative: ConfigurableFileCollection = objectFactory.fileCollection()
+
+    @get:OutputFiles
+    val sharedObjectFiles: ConfigurableFileCollection = objectFactory.fileCollection()
+
+    @TaskAction
+    fun onRun() {
+        try {
+            collectSharedObjectFiles()
+        } catch (e: Exception) {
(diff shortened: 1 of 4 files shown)


embrace-gradle-plugin/src/main/java/io/embrace/android/gradle/plugin/config/PluginBehavior.kt
```diff
@@ -29,6 +29,11 @@ interface PluginBehavior {
      */
     val isUploadMappingFilesDisabled: Boolean
 
+    /**
+     * Whether the plugin should fail the build on upload errors, set via `embrace.failBuildOnUploadErrors`
+     */
+    val failBuildOnUploadErrors: Boolean
```
fractalwrench, line 35: Does this PR require a rebase?

embrace-gradle-plugin/src/main/java/io/embrace/android/gradle/plugin/tasks/ndk/CollectSharedObjectFilesTask.kt
```diff
@@ -0,0 +1,103 @@
...
+import org.gradle.api.tasks.OutputFiles
+import org.gradle.api.tasks.SkipWhenEmpty
+import org.gradle.api.tasks.TaskAction
+import java.io.File
+import javax.inject.Inject
+
+/**
+ * Collects .so files from the project, depending on the project type.
+ */
+abstract class CollectSharedObjectFilesTask @Inject constructor(
```
fractalwrench, line 25:
```suggestion
abstract class CopyFilesTask @Inject constructor(
```

Would suggest making this abstract by supplying the location for the files to copy as a property of the task. We could also put this in the `common` package rather than `ndk`

embrace-gradle-plugin/src/main/java/io/embrace/android/gradle/plugin/tasks/ndk/CollectSharedObjectFilesTask.kt
```diff
@@ -0,0 +1,103 @@
...
+        }
+
+        return unitySharedObjectFiles
+    }
+
+    private fun getNativeSharedObjectFiles(): Array<File> {
+        val files = architecturesDirectoryForNative.singleOrNull()?.listFiles() ?: emptyArray()
+
+        if (files.isEmpty()) {
+            // TODO: is this an error? does every native project have .so files?
```
fractalwrench, line 93: IMO we shouldn't error, we should continue execution in this unlikely scenario given we've marked the input as `SkipWhenEmpty`.

embrace-gradle-plugin/src/main/java/io/embrace/android/gradle/plugin/tasks/ndk/CollectSharedObjectFilesTask.kt
```diff
@@ -0,0 +1,103 @@
...
+    }
+
+    private fun collectSharedObjectFiles() {
+        val files = when (projectType.get()) {
+            ProjectType.UNITY -> getUnitySharedObjectFiles()
+            ProjectType.NATIVE -> getNativeSharedObjectFiles()
+            else -> error("Shared object file collection failed: unsupported project type")
+        }
+        sharedObjectFiles.from(files)
+    }
```
fractalwrench, line 70: I think this can be supplied as a provider to the task, meaning the task won't need to know about all these different concerns directly

embrace-gradle-plugin/src/main/java/io/embrace/android/gradle/plugin/tasks/ndk/CollectSharedObjectFilesTaskRegistration.kt
```diff
@@ -0,0 +1,112 @@
...
+
+        if (embraceConfig?.ndkEnabled == false) return null
+
+        val collectSharedObjectFilesTaskProvider = project.registerTask(
+            CollectSharedObjectFilesTask.NAME,
+            CollectSharedObjectFilesTask::class.java,
+            data
+        ) { task ->
+            // TODO: is the ndkEnabled check needed? We are already checking that in the execute method above
+            task.onlyIf { shouldExecuteTask(embraceConfig) }
```
fractalwrench, line 43: Can you test the behavior of this? I think this depends on whether `shouldExecuteTask` relies on not being called early in the Gradle build lifecycle, or whether the information is all there as soon as the build initializes.

embrace-gradle-plugin/src/main/java/io/embrace/android/gradle/plugin/config/PluginBehavior.kt
priettt: yup, just rebased!

embrace-gradle-plugin/src/main/java/io/embrace/android/gradle/plugin/tasks/ndk/CollectSharedObjectFilesTask.kt
priettt: good idea!

embrace-gradle-plugin/src/main/java/io/embrace/android/gradle/plugin/tasks/ndk/CollectSharedObjectFilesTask.kt
priettt: Oh you're right, I'll just remove this validation, it doesn't look like we need it

embrace-gradle-plugin/src/main/java/io/embrace/android/gradle/plugin/tasks/ndk/CollectSharedObjectFilesTaskRegistration.kt
priettt: just verified that the task won't be registrated if ndkEnabled isn't true, so this check is not necessary

PR #1974, opened 2025-03-03
Add validation for bytecode instrumentation that looks at smali files

fractalwrench:
## Goal

Adds a test that reads the smali files (generated from dex files) when disassembling the APK in a Gradle TestKit test case.

In short, this allows to verify our bytecode instrumentation at the integration test level. We can compare the expected return value against what is actually in the dex code.

Future changesets will increase the coverage of this test so that we check _all_ the config values that are instrumented, and also check the override behaviors.


bidetofevil (approved):
Ooooh so nice

PR #1988, opened 2025-03-05
Bump AGP, Gradle, Lint

fractalwrench:
## Goal

Bumps AGP, Gradle, and Lint to the latest available versions.


bidetofevil (approved):
Underlying issue fixed here: https://github.com/embrace-io/embrace-android-sdk/pull/2003

We can merge this after that is merged

PR #1992, opened 2025-03-06
Stop respecting obsolete project properties

fractalwrench:
## Goal

Stop respecting the obsolete project properties `embrace.logLevel` and `embrace.instrumentationScope`. Log level can be controlled via Gradle directly so we don't need our own logic to achieve this. Instrumentation scope has not been used in practice so is not required.

## Testing


bidetofevil (approved):
FINALLY

PR #1993, opened 2025-03-06
Add new embrace extension

fractalwrench:
## Goal

Adds the new extension that will form our plugin's public API. This is not publicly exposed as part of the plugin yet & the changeset is kept small so we can review the particulars of each property and its documentation.

### DIFF
--- a/embrace-gradle-plugin/src/main/java/io/embrace/android/gradle/plugin/api/EmbraceBytecodeInstrumentation.kt
+++ b/embrace-gradle-plugin/src/main/java/io/embrace/android/gradle/plugin/api/EmbraceBytecodeInstrumentation.kt
@@ -1,0 +1,55 @@
+package io.embrace.android.gradle.plugin.api
+
+import org.gradle.api.model.ObjectFactory
+import org.gradle.api.provider.ListProperty
+import org.gradle.api.provider.Property
+import javax.inject.Inject
+
+/**
+ * DSL for configuring the Embrace Gradle Plugin's bytecode instrumentation behavior.
+ */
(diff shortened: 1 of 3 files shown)


embrace-gradle-plugin/src/main/java/io/embrace/android/gradle/plugin/api/EmbraceBytecodeInstrumentation.kt
```diff
@@ -0,0 +1,55 @@
...
+
+    /**
+     * A list of string patterns that are used to filter classes during bytecode instrumentation. For example, 'com.example.foo.*'
+     * would avoid instrumenting any classes in the 'com.example.foo' package.
+     *
+     * This can be useful if you wish to avoid instrumenting certain parts of your codebase.
+     *
+     * Defaults to an empty list.
+     */
+    val classIgnorePatterns: ListProperty<String> =
```
bidetofevil, line 53: Do people actually use this? I hope we have tests around this because I wasn't aware folks can partially opt-out of this

bidetofevil (approved):
LGTM. A couple of non-blocking comments.

embrace-gradle-plugin/src/main/java/io/embrace/android/gradle/plugin/api/EmbraceBytecodeInstrumentation.kt
fractalwrench: Folks have certainly used it in the past & it's served as a useful escape-hatch for if instrumentation goes wrong for a specific class. There is unit test coverage for this I believe.

priettt (approved):

PR #1995, opened 2025-03-06
Ignore obsolete properties on swazzler extension

fractalwrench:
## Goal

Ignores some obsolete properties on swazzler extension and marks them as deprecated.

## Testing


bidetofevil (approved):
LGTM

PR #2004, opened 2025-03-07
Add bytecode instrumentation integration test

fractalwrench:
## Goal

Adds gradle integration test cases that assert how bytecode instrumentation functions.


bidetofevil (approved):
Nice. LGTM

PR #2009, opened 2025-03-10
Add gradle test case for dexguard

fractalwrench:
## Goal

Adds a test case that checks DexGuard mapping files are uploaded.


bidetofevil (approved):
LGTM

PR #2010, opened 2025-03-10
Tweak error message for auth error in http request

fractalwrench:
## Goal

Tweaks the error message if a HTTP request has an auth error (403).

## Testing

Before:
>Embrace HTTP request failed: /v2/store/ndk/handshake, status=403

After:
>Embrace HTTP request failed: /v2/store/ndk/handshake, status=403. Please check that src/main/embrace-config.json contains an app_id and api_token that match the settings page of your Embrace dashboard.

### DIFF
--- a/embrace-gradle-plugin/src/main/java/io/embrace/android/gradle/plugin/tasks/HttpCallResultHandler.kt
+++ b/embrace-gradle-plugin/src/main/java/io/embrace/android/gradle/plugin/tasks/HttpCallResultHandler.kt
@@ -21,11 +21,24 @@ fun handleHttpCallResult(
     failBuildOnUploadErrors: Boolean,
 ) {
     val msg = when (result) {
-        is Failure -> "failed: ${endpoint.url}, status=${result.code}"
+        is Failure -> handleHttpFailure(endpoint, result)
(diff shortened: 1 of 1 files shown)


embrace-gradle-plugin/src/main/java/io/embrace/android/gradle/plugin/tasks/HttpCallResultHandler.kt
```diff
@@ -21,11 +21,24 @@ fun handleHttpCallResult(
...
 }
+
+private fun handleHttpFailure(
+    endpoint: EmbraceEndpoint,
+    result: Failure,
+): String {
+    val msg = "failed: ${endpoint.url}, status=${result.code}"
+
+    if (result.code == 403) {
+        return "$msg. Please check that src/main/embrace-config.json contains an " +
```
bidetofevil, line 40: Sometimes a different config file is picked up depending on variant. Do we want to specify that explicitly in the error message or just loosen the error message to include the possibility of another config file being used?

bidetofevil (approved):
LGTM. Small comment that is not blocking.

PR #2011, opened 2025-03-10
Delete v1 delivery layer

fractalwrench:
## Goal

Deletes the v1 delivery layer.

## Testing


bidetofevil (approved):
kill it with fire

PR #2012, opened 2025-03-10
Tweak aei limits

fractalwrench:
## Goal

Tweaks the limits for AEI capture, so that a maximum of 64 can be processed & sent in one session, and so the trace limit is 10Mb uncompressed. The intent behind this change is to improve the deliverability of NDK crash traces.

## Testing


embrace-android-features/src/main/kotlin/io/embrace/android/embracesdk/internal/capture/aei/AeiDataSourceImpl.kt
```diff
@@ -35,7 +35,7 @@ internal class AeiDataSourceImpl(
 ) {
 
     private companion object {
-        private const val SDK_AEI_SEND_LIMIT = 32
+        private const val SDK_AEI_SEND_LIMIT = 64
```
pablomatiasgomez, line 38: Does this limit apply for all AEIs? Any chance we can make it so that native crashes don't respect this limit?

fractalwrench: This is the maximum number of AEI we request from the OS on each process launch. In practice AEI is only recorded after a process terminates so most of the time I'd expect the number of AEI sent to equal 1.

So assuming Embrace is enabled 100% of the time, the SDK will get 64 process launches where it can attempt to capture data before dropping anything. I wouldn't like to make this unbounded as there is a resource cost in capturing each AEI object

fractalwrench: And unfortunately the Android API doesn't provide a way to query _just_ for crashes/ANRs, which would make our lives much simpler...

PR #2013, opened 2025-03-10
Add ordinal for AEI native crashes

fractalwrench:
## Goal

Adds an ordinal as an OTel attribute for AEIs containing a native crash. This will be incremented whenever an AEI object contains an NDK protobuf file - assumed by checking `reason == REASON_CRASH_NATIVE` and the `trace` property not being null.

Additionally, the existing `android.crash_number` attribute will be incremented for AEI crashes.

If the AEI object does not have an NDK protobuf file, the attribute value will be explicitly set as `"null"`.

## Testing

Added test coverage.

### DIFF
--- a/embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/prefs/PreferencesService.kt
+++ b/embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/prefs/PreferencesService.kt
@@ -86,6 +86,13 @@ interface PreferencesService {
     fun incrementAndGetNativeCrashNumber(): Int
 
     /**
+     * Increments and returns the AEI crash number ordinal. This is an integer that
+     * increments on every NDK tombstone detected via an AEI trace. It allows us to check the % of AEI crashes that
+     * didn't get delivered to the backend.
+     */
+    fun incrementAndGetAeiCrashNumber(): Int
+
+    /**
      * Last javaScript bundle string url.
      */
     var javaScriptBundleURL: String?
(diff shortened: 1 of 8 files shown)


embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/prefs/PreferencesService.kt
```diff
@@ -85,6 +85,13 @@ interface PreferencesService {
      */
     fun incrementAndGetNativeCrashNumber(): Int
 
+    /**
+     * Increments and returns the AEI crash number ordinal. This is an integer that
+     * increments on every NDK tombstone detected via an AEI trace. It allows us to check the % of AEI crashes that
+     * didn't get delivered to the backend.
+     */
+    fun incrementAndGetAeiCrashNumber(): Int
```
pablomatiasgomez, line 93: should all of these be named `AeiNativeCrash` instead of `AeiCrash`? I undersatnd AEI can report both native and java crashes, right? I know we are not doing anything with the latter for now, but we may in the future, so it's good to distinguish?

fractalwrench: [AEI](https://developer.android.com/reference/android/app/ApplicationExitInfo#getTraceInputStream()) only contains a trace for ANRs and native crashes. While it does record that a JVM crash happened, it doesn't record the stacktrace. The SDK should only send an AEI with `REASON_CRASH` if an ANR trace is associated with it.

bidetofevil (approved):
LGTM

PR #2015, opened 2025-03-10
Gate AEI native crashes by ndk_enabled flag

fractalwrench:
## Goal

If `ndk_enabled` is false then AEI native crashes should not be delivered.

## Testing

Added a test case.


bidetofevil (approved):
LGTM

PR #2018, opened 2025-03-11
Alter AEI delivery priority

fractalwrench:
## Goal

Alters the priority of delivering AEI payloads. Currently these are delivered as log priority but if they contain native crashes/ANRs these deserve higher priority.

## Testing

Added unit test coverage.


bidetofevil (approved):
A 10mb file that might be blocking? I think I'd still want sessions delivered before it tbh.

Lets keep this for now and maybe talk about it with Eliab when he's back to see what he thinks?

fractalwrench:
A potentially better solution would be to avoid looking at `envelopeType.priority` in `StoredTelemetryMetadata` and use `payloadType` instead to prioritise requests.

The current implementation isn't flexible enough IMO. It's not possible to tweak the existing priorities to fit AEI in between p3-p5 as there are various assumptions made. And we want to distinguish between an ANR AEI and an AEI native crash

IMO it's reasonable to prioritise crash AEIs higher than sessions, but perhaps not ANR AEIs

bidetofevil:
If there are assumptions being made then we should take a look at refactoring how we're doing priority. I think the original intent was for the logic to be in how the priorities are determined, and the delivery layer is just a dumb thing that looks at the priority. If it has to made decisions, that defeats the purpose, so I'm OK revisiting that.

We can merge as is for now and add a task to take this on, right away even?

PR #2022, opened 2025-03-12
Move old payload classes

fractalwrench:
## Goal

Moves some classes that used to be JSON serialiazable but are no longer into the core/features modules.

## Testing


bidetofevil (approved):
LGTM

PR #2023, opened 2025-03-12
Added CreateArchToSharedObjectsMapTask.

priettt:
## Goal

This is one of the 4 tasks NdkUploadTask will get separated into. 

This task receives a directory with architectures as its input, and outputs a map of each architecture to a list with every shared object file path of that architecture.

For instance, if the input directory is this:

```
archs 
| - - - - armeabi-v7a
             | - - - - libtest1.so
             | - - - - libtest2.so
| - - - - arm64-v8a
             | - - - - libtest3.so
             | - - - - libtest4.so
```

The result will be this map: 

```
armeabi-v7a -> [path to libtest1.so, path to libtest2.so]
arm64-v8a -> [path to libtest3.so, path to libtest4.so]
```

I also moved the logic to find the architecture directories to the task registration, so it's passed as a provider to the task.

## Testing

Added an integration test for the task, still need to test the registration.


embrace-gradle-plugin/src/main/java/io/embrace/android/gradle/plugin/tasks/ndk/CompressAndHashSharedObjectFilesTask.kt
```diff
@@ -0,0 +1,81 @@
...
+import org.gradle.api.tasks.SkipWhenEmpty
+import org.gradle.api.tasks.TaskAction
+import java.io.File
+import javax.inject.Inject
+
+/**
+ * Given a directory containing architecture directories (arm64-v8a, armeabi-v7a, etc.) with shared object files (.so files),
+ * this task:
+ * 1. Compresses each shared object file using Zstd compression
+ * 2. Stores the compressed files in compressedSharedObjectFilesDirectory, preserving the architecture directory structure
```
fractalwrench, line 22: I didn't realise we were doing Zstd compression on SO files. IMO it'd be worth breaking that into a separate task that takes the uncompressed SO files as an input and the compressed files as an output. Sorry for the confusion from our discussion yesterday

embrace-gradle-plugin/src/main/java/io/embrace/android/gradle/plugin/tasks/ndk/CompressAndHashSharedObjectFilesTask.kt
```diff
@@ -0,0 +1,81 @@
...
+    @get:OutputFile
+    val architecturesToHashedSharedObjectFilesMap: RegularFileProperty = objectFactory.fileProperty()
+
+    @TaskAction
+    fun onRun() {
+        val outputMap = createOutputMap()
+
+        // Serialize the map to JSON and write it to the output file
+        val serializableMap = ArchitecturesToHashedSharedObjectFilesMap(outputMap)
+        architecturesToHashedSharedObjectFilesMap.get().asFile.outputStream().use { outputStream ->
```
fractalwrench, line 49:
```suggestion
        architecturesToHashedSharedObjectFilesMap.get().asFile.outputStream().use { outputStream ->
```

It'd be worth using a buffered stream here to improve perf

embrace-gradle-plugin/src/main/java/io/embrace/android/gradle/plugin/tasks/ndk/CompressAndHashSharedObjectFilesTaskRegistration.kt
```diff
@@ -0,0 +1,110 @@
...
+            when (it) {
+                ProjectType.UNITY -> getUnitySharedObjectFiles(project, data)
+                ProjectType.NATIVE -> getNativeSharedObjectFiles(project, variant)
+                else -> error("Project type must be either UNITY or NATIVE")
+            }
+        }
+    }
+
+    private fun getUnitySharedObjectFiles(project: Project, data: AndroidCompactedVariantData): File {
+        // TODO: Verify if errors should be thrown if the directories or SO files are not found. Improve error messaging.
```
fractalwrench, line 60: Are these todo comments new or were they present in the previous code that's been adapted into this source file?

priettt: these are new, I'm adding them so I can then think about any uncommon codepaths

PR #2028, opened 2025-03-13
Create CompressSharedObjectFilesTask, its registration class and an integration test for it

priettt:
## Goal

This is one of the tasks NdkUploadTask will get separated into.

This task receives a directory containing architecture directories (arm64-v8a, armeabi-v7a, etc.) with shared object files (.so files) inside,
compresses the .so files using Zstd and stores the compressed files in the output directory, preserving the architecture directory structure.

## Testing

Added integration test for the task, will test registration in a following PR.


embrace-gradle-plugin/src/main/java/io/embrace/android/gradle/plugin/tasks/ndk/CompressSharedObjectFilesTask.kt
```diff
@@ -0,0 +1,58 @@
...
+    val architecturesDirectory: DirectoryProperty = objectFactory.directoryProperty()
+
+    @get:OutputDirectory
+    val compressedSharedObjectFilesDirectory: DirectoryProperty = objectFactory.directoryProperty()
+
+    @TaskAction
+    fun onRun() {
+        architecturesDirectory.get().asFile
+            .listFiles()
+            ?.filter { it.isDirectory && it.listFiles()?.isNotEmpty() == true }
```
fractalwrench, line 36: nit: if you use ` ?: emptyList()` you can avoid the null coalescening operator on the rest of the functional chain

fractalwrench (approved):
LGTM

fractalwrench:
FWIW my next step would be to hook this up to the live plugin & remove the compress functionality from the NDK upload task. That will get us a bit more certainty that things haven't broken at the cost of more up-front effort

PR #2030, opened 2025-03-13
Create HashSharedObjectFilesTask, its registration class and an integration test for it

priettt:
## Goal

This is one of the tasks NdkUploadTask will get separated into.

This task receives a directory containing compressed shared object files organized by architecture, it calculates SHA1 hashes for each compressed file, and it outputs a JSON file mapping every architecture to a map of the compressed .so files names and their hashes. 

Sample output:

```
{
   "arm64-v8a": {
     "libexample1.so": "2a21dc0b99017d5db5960b80d94815a0fe0f3fc2",
     "libexample2.so": "3b32ed1c88128e6ec4b71b93a4926a1bf1f4gd3"
   },
   "armeabi-v7a": {
     "libexample1.so": "4c43fe2d77239f7fd5a71c91b85926b2g2g5he4",
     "libexample2.so": "5d54gf3e66340g8ge6b82da2c96a37c3h3h6if5"
   }
}
```

## Testing

Added integration test for the task, will test registration in a following PR.


embrace-gradle-plugin-integration-tests/src/test/java/io/embrace/android/gradle/integration/testcases/HashSharedObjectFilesTaskTest.kt
```diff
@@ -0,0 +1,72 @@
...
+                    ArchitecturesToHashedSharedObjectFilesMap::class.java
+                ).architecturesToHashedSharedObjectFiles
+
+                expectedArchitectures.forEach { architecture ->
+                    expectedSharedObjectFiles.forEach { sharedObjectFile ->
+                        val sha1Hash = deserializedOutputMap[architecture]?.get(sharedObjectFile)
+                        val compressedSoFile = projectDir.file("compressedSharedObjectFiles/$architecture/$sharedObjectFile")
+                        assertTrue(sha1Hash == calculateSha1ForFile(compressedSoFile))
+                    }
+                }
```
fractalwrench, line 68: Can you store a fixture in `test/resources` and just compare the contents, given that the test fixture should produce the same output each time?

embrace-gradle-plugin/src/main/java/io/embrace/android/gradle/plugin/tasks/ndk/HashSharedObjectFilesTask.kt
```diff
@@ -0,0 +1,64 @@
...
+        val serializableMap = ArchitecturesToHashedSharedObjectFilesMap(outputMap)
+        architecturesToHashedSharedObjectFilesMap.get().asFile.outputStream().use { outputStream ->
+            serializer.toJson(serializableMap, ArchitecturesToHashedSharedObjectFilesMap::class.java, outputStream)
+        }
+    }
+
+    private fun createOutputMap(): Map<String, Map<String, String>> =
+        compressedSharedObjectFilesDirectory.get().asFile
+            .listFiles()
+            ?.filter { it.isDirectory && it.listFiles()?.isNotEmpty() == true }
```
fractalwrench, line 48: nit: consider something like `?: emptyList()` to avoid the null coalescing operator in this chain

fractalwrench (approved):
LGTM

PR #2035, opened 2025-03-15
Remove generateHashedObjects from NdkUploadTask and replace it with the new tasks

priettt:
## Goal

Wire the result of HashSharedObjectFilesTask to NdkUploadTask. Modified test classes accordingly to account for the new tasks.

Some remarks:

- I had to add an onlyIf task to make automatic task skipping work for 7.5.1. Isn't needed for newer Gradle versions.
- getSharedObjectFilesProvider depends on two providers: projectType and merge[Variant]NativeLibs. I had to use a map inside of a flatmap to make this work (zip wasn't working for some reason).

## Testing

Adapted and relied on existing testing


fractalwrench (approved):
LGTM

PR #2046, opened 2025-03-18
Add UploadSharedObjectFilesTask and remove uploadSymbols from NdkUploadTask

priettt:
## Goal

This task handles the upload of shared object (.so) files to the Embrace backend.

It receives the directory containing the compressed .so files, the JSON mapping architectures to file hashes, and it:
- Reports the available shared object files to the backend.
- Finds the shared object files requested by the backend (if any) in the compressed files directory.
- Uploads the requested files individually to the backend.

Some observations: 
- NdkUploadTask will be renamed to InjectSymbols in a following PR.
- I had to add a specific dependency between UploadSharedObjectFiles and NdkUploadTask because the former wasn't executed, as Gradle doesn't think it's necessary (it doesn't alter any files). Those task could be independent in theory, so they could even be executed in parallel? We should check if that can be done.

## Testing

Adapted existing integration tests.


embrace-gradle-plugin/src/main/java/io/embrace/android/gradle/plugin/tasks/ndk/UploadSharedObjectFilesTask.kt
```diff
@@ -0,0 +1,150 @@
...
+    private fun getRequestedSharedObjectFiles(): Map<String, List<String>>? {
+        val params = NdkUploadHandshakeRequest(
+            requestParams.get().appId,
+            requestParams.get().apiToken,
+            variantData.get().name,
+            architecturesToHashedSharedObjectFilesMap
+        )
+
+        // TODO: we are swallowing backend errors, we should fix this.
+        return NdkUploadHandshake(okHttpNetworkService).getRequestedSymbols(params, failBuildOnUploadErrors.get())
```
fractalwrench, line 86: Is this comment still valid? Can we just propagate whatever error/exception is getting thrown and fail if that flag is enabled?

fractalwrench (approved):
LGTM

embrace-gradle-plugin/src/main/java/io/embrace/android/gradle/plugin/tasks/ndk/UploadSharedObjectFilesTask.kt
priettt: It is. If the backend returns an error, getRequestedSymbols returns null. We don't do anything in particular for that case.

priettt: I plan to review all of the TODOs when I start writing tests and reviewing edge cases tomorrow

PR #2048, opened 2025-03-20
Renamed NdkUploadTask to InjectSharedObjectFilesTask

priettt:
## Goal

This is the last task that NdkUploadFile was divided into. It injects the map of hashed object files into an ndk_symbols.xml resource. We should discuss if this is actually needed, but we want to follow what was being done in NdkUploadTask.

## Testing

Relied on existing testing.


embrace-gradle-plugin/src/main/java/io/embrace/android/gradle/plugin/tasks/ndk/InjectSharedObjectFilesTask.kt
```diff
@@ -15,13 +15,19 @@ import java.io.IOException
 import javax.inject.Inject
 
 /**
- * A task that uploads NDK symbols to Embrace.
+ * Task that injects a map of architectures to hashed shared object files into the project's resources.
+ *
+ * It reads the map from a JSON file and writes it to a new XML file. The XML file is then placed in the project's resources, to
```
fractalwrench, line 20: Non-blocking feedback: is it necessary to deserialize the file?

fractalwrench (approved):
LGTM

PR #2049, opened 2025-03-20
Update NdkUploadTasksRegistrationTest with more accurate tests

priettt:
## Goal

Adds additional testing to NdkUploadTasksRegistrationTest, testing task registrations, dependencies, and some of the inputs that they use.


embrace-gradle-plugin/src/main/java/io/embrace/android/gradle/plugin/config/ProjectType.kt
```diff
@@ -1,5 +1,9 @@
 package io.embrace.android.gradle.plugin.config
 
+/*
+*  TODO: React Native type is not used. Should we remove it?
```
fractalwrench, line 4: Yes

embrace-gradle-plugin/src/main/java/io/embrace/android/gradle/plugin/tasks/ndk/NdkUploadTasksRegistration.kt
```diff
@@ -101,12 +101,13 @@ class NdkUploadTasksRegistration(
                 task.compressedSharedObjectFilesDirectory.asFile.get().exists() &&
                     task.architecturesToHashedSharedObjectFilesMapJson.asFile.get().exists()
             }
+            // TODO: An error thrown in registration will make the build will fail. Should we use failBuildOnUploadErrors for this too?
```
fractalwrench, line 104: IMO no as it's not about failing to make a HTTP request. I think we should just default to failing the build if something goes that wrong

fractalwrench (approved):
LGTM

PR #2055, opened 2025-03-21
Added tests for HashSharedObjectFilesTask

priettt:
## Goal

- Added tests in HashSharedObjectFilesTaskTest.
- Added a expected json file so we don't need to recalculate hashes for every test.
- Added additional tests in CompressSharedObjectFilesTaskTest.
- Modified editorconfig so it doesn't add newlines to json files.


fractalwrench (approved):
LGTM

PR #2062, opened 2025-03-26
Added tests for UploadSharedObjectFilesTaskTest

priettt:
## Goal

Added additional testing for UploadSharedObjectFilesTask. 

Some things that will be added in a following PR: 
- Use failBuildOnUploadErrors to decide if we continue the build or we fail it.
- Address TODOs in the class.


fractalwrench (approved):
LGTM

PR #2063, opened 2025-03-26
Handle backend errors correctly in UploadSharedObjectFilesTask

priettt:
## Goal

When failBuildOnUploadErrors is disabled, we weren't logging the error nor returning correctly in the task. Fixed that.


fractalwrench (approved):
LGTM

PR #2066, opened 2025-03-26
Remove unneeded code from InjectSharedObjectFilesTask

priettt:
## Goal

We were mapping some unity symbol file names every time we injected the symbols map into the app resources. This mapping is not needed for Unity > 2019, so I removed it. 

This means we can stop deserializing the architecturesToHashedSharedObjectFiles json file, and just read and encode to base 64 the contents of the file directly.

For this to work, and to ensure consistency, I changed the name of the json object for every test file: from architecturesToHashedSharedObjectFiles to symbols, so it reflects what we inject. We could discuss changing the name again in the future.

Added additional testing for InjectSharedObjectFilesTaskIntegrationTest, though more tests are coming in a following PR.


fractalwrench (approved):
LGTM

PR #2070, opened 2025-03-28
Use failBuildOnUploadTask in CompressSharedObjectFilesTask

priettt:
## Goal

Use a custom property to enable and disable the task in tests. Will use this approach for the other tests too


fractalwrench (approved):
LGTM

PR #2071, opened 2025-03-28
Use failBuildOnUploadTask in HashSharedObjectFilesTask

priettt:
Use a custom property to enable and disable the task in tests.


fractalwrench (approved):
LGTM

PR #2073, opened 2025-03-31
Update instrumented-config json files to account for new OTelLimitsConfig fields

priettt:
## Goal

New fields were added to OTelLimitsConfig, so I added them to the instrumented-config json files in the order they get parsed in the tests. 

I'm not sure why the tests weren't broken in CI though.


fractalwrench (approved):
LGTM

PR #2080, opened 2025-04-08
Remove SymbolresourceInjector and move its logic to InjectSharedObjectFilesTask

priettt:
## Goal

- Removed a redundant test.
- Moved symbol injection logic to InjectSharedObjectFilesTask


fractalwrench (approved):
LGTM

PR #2081, opened 2025-04-08
Create AsmTaskRegistration to register the ASM task in the same way as the rest of the tasks

priettt:
- Removed the old AsmTasks class. 
- Altered the order in which the tasks run overall. Now ASM will be done last. This will be used to plug the output of the Injection task as an input of the ASM one.


fractalwrench (approved):
LGTM from the reorganisation POV, but I don't think this will alter the order in which the ASM task runs. It'll be necessary to use Gradle's task dependency API to achieve that.

PR #2088, opened 2025-04-10
Inject encoded shared object files map with ASM

priettt:
- Added a Base64SharedObjectFilesMap property in InstrumentedConfig, which will be used from the SDK.
- Improved ConfigClassVisitor tests.
- Make AsmTaskRegistration run even if upload tasks are disabled.


embrace-gradle-plugin/src/main/java/io/embrace/android/gradle/plugin/instrumentation/config/ConfigClassVisitorFactory.kt
```diff
@@ -17,12 +18,14 @@ object ConfigClassVisitorFactory {
...
         RedactionConfig,
-        SessionConfig;
+        SessionConfig,
+        Base64SharedObjectFilesMap;
 
         val className = "io.embrace.android.embracesdk.internal.config.instrumented.${name}Impl"
 
         fun createClassVisitor(
             cfg: VariantConfig,
+            encodedSharedObjectFilesMap: String?,
```
fractalwrench, line 28: This should be added onto `VariantConfig` like other values set via the gradle plugin and not embrace-config.json - e.g. `buildId`

fractalwrench (approved):
LGTM pending one comment

embrace-gradle-plugin/src/main/java/io/embrace/android/gradle/plugin/instrumentation/config/ConfigClassVisitorFactory.kt
priettt: I agree it would be better to have it in VariantConfig, but it might be hard to achieve that.

The encoding task depends on VariantConfig, so we can't wait for the output of the task before creating VariantConfig. I suppose we could add a field that it's initialized later, but it might end up being too complex.

I'd leave it like it is for now.

fractalwrench: Happy to go with this for now in that case - we can address this in future

PR #2091, opened 2025-04-11
Use the new EncodeMapTask output in the SDK and stop using resources

priettt:
I tried to keep this small but I had to replace many parts at once so it doesn't stop working.

- Stop looking into resources in the SDK, fetch symbols from the new instrumentedConfig field.
- Update SymbolService tests to verify behavior of the new injected symbols.
- Use project.afterEvaluate in AsmTaskRegistration to ensure that the AsmTask is found.
- Register EncodeSharedObjectFilesTask instead of InjectSharedObjectFilesTask (will remove this one in a following PR).
- Modify AndroidNdkTest: Now we're disassembling the apk and parsing the Smali to ensure the map is injected as expected. Removed resources verification.

### DIFF
--- a/embrace-android-core/src/test/java/io/embrace/android/embracesdk/internal/ndk/symbols/SymbolServiceImplTest.kt
+++ b/embrace-android-core/src/test/java/io/embrace/android/embracesdk/internal/ndk/symbols/SymbolServiceImplTest.kt
@@ -1,28 +1,21 @@
 package io.embrace.android.embracesdk.internal.ndk.symbols
 
-import android.content.Context
-import android.content.res.Resources
 import android.util.Base64
 import androidx.test.ext.junit.runners.AndroidJUnit4
 import io.embrace.android.embracesdk.fakes.FakeDeviceArchitecture
 import io.embrace.android.embracesdk.fakes.FakeEmbLogger
 import io.embrace.android.embracesdk.fakes.TestPlatformSerializer
+import io.embrace.android.embracesdk.fakes.config.FakeBase64SharedObjectFilesMap
+import io.embrace.android.embracesdk.fakes.config.FakeInstrumentedConfig
 import io.embrace.android.embracesdk.internal.payload.NativeSymbols
-import io.mockk.every
-import io.mockk.mockk
 import org.junit.Assert.assertEquals
 import org.junit.Assert.assertNull
 import org.junit.Test
 import org.junit.runner.RunWith
 
+// Needed for android.util.Base64
 @RunWith(AndroidJUnit4::class)
 class SymbolServiceImplTest {
-
-    private companion object {
-        private const val PACKAGE_NAME = "com.example.app"
-        private const val SYMBOL_RES_ID = "emb_ndk_symbols"
-        private const val RES_INT_VALUE = 1602934923
-    }
 
     private val symbolMap = mapOf(
         "armeabi-v7a" to mapOf("symbol" to "armeabi-v7a-value"),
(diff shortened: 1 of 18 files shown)


embrace-android-core/src/test/java/io/embrace/android/embracesdk/internal/ndk/symbols/SymbolServiceImplTest.kt
```diff
@@ -1,41 +1,36 @@
...
+import io.embrace.android.embracesdk.fakes.config.FakeInstrumentedConfig
 import io.embrace.android.embracesdk.internal.payload.NativeSymbols
-import io.mockk.every
-import io.mockk.mockk
 import org.junit.Assert.assertEquals
 import org.junit.Assert.assertNull
 import org.junit.Test
 import org.junit.runner.RunWith
 
+// Needed for android.util.Base64
```
fractalwrench, line 16: Could we tweak it to Kotlin/Java Base64 so the test can run on the JVM? Can be ignored if that's non-trivial

embrace-gradle-plugin/src/main/java/io/embrace/android/gradle/plugin/instrumentation/AsmTaskRegistration.kt
```diff
@@ -41,27 +41,29 @@ class AsmTaskRegistration : EmbraceTaskRegistration {
                 params.shouldInstrumentOnLongClick.set(behavior.instrumentation.onLongClickEnabled)
                 params.shouldInstrumentOnClick.set(behavior.instrumentation.onClickEnabled)
 
-                // Find the Asm transformation task by name and make it depend on encodeSharedObjectFilesTask
-                val encodeSharedObjectFilesTask = project.tryGetTaskProvider(
-                    "${EncodeSharedObjectFilesTask.NAME}${data.name.capitalizedString()}",
-                    EncodeSharedObjectFilesTask::class.java
-                ) ?: return@transformClassesWith
+                project.afterEvaluate {
```
fractalwrench, line 44: What testing have we done around this to confirm that it works?

fractalwrench (approved):
LGTM pending CI

embrace-android-core/src/test/java/io/embrace/android/embracesdk/internal/ndk/symbols/SymbolServiceImplTest.kt
priettt: We can't. We are using the Android Base64 decoder in the SDK, and we can't swap it with the Java one because it's only compatible with API > 26

priettt: We could use the Kotlin one though, if we are okay with the experimental annotation.

priettt: (we can actually use the java.util.Base64 class for encoding in the test, but we will still need the JUnit runner so the decoding  works)

embrace-gradle-plugin/src/main/java/io/embrace/android/gradle/plugin/instrumentation/AsmTaskRegistration.kt
priettt: Mostly the existing tests and debugging in different setups. Any particular test you think we should add? Tests do break without this 😅

embrace-android-core/src/test/java/io/embrace/android/embracesdk/internal/ndk/symbols/SymbolServiceImplTest.kt
fractalwrench: Let's keep this as-is for now

embrace-gradle-plugin/src/main/java/io/embrace/android/gradle/plugin/instrumentation/AsmTaskRegistration.kt
fractalwrench: I'm happy with that level of testing - the smali decompilation tests should be enough

PR #2092, opened 2025-04-11
Remove InjectSharedObjectFilesTask and associated classes

priettt:
Remove injection classes that were superseded by the new encoding task and ASM injection.


fractalwrench (approved):
LGTM

PR #2093, opened 2025-04-14
Address PR comments for the new Encode task

priettt:
- Rename encode task to make it more generic
- Use Base64.wrap


fractalwrench (approved):
LGTM

PR #2094, opened 2025-04-15
Simplify bytecode instrumentation implementation

fractalwrench:
## Goal

Begins refactoring bytecode instrumentation by creating a single `MethodVisitor` that can be configured via its parameters to instrument a specific function in a given class.

## Testing


priettt (approved):
Looks good! Why did we need a WebViewClientOverride?

fractalwrench:
The bytecode instrumentation is different depending on whether the `onPageStarted` method has been overridden or not in the webview implementation.

PR #2095, opened 2025-04-15
Rearrange logic for visiting classes

fractalwrench:
## Goal

Begins rearranging the logic that determines what classes are visited by WebObject ASM.

## Testing


bidetofevil (approved):
LGTM

PR #2098, opened 2025-04-15
Add auto start instrumentation

priettt:
## Goal
Add ASM instrumentation to start the Embrace SDK automatically. 

- Created AutoStartInstrumentationHook, which will be called from the instrumented code.
    - I decided to create this class instead of calling Embrace.start() directly, so the code goes through a different path. We could even add telemetry to log if the app was autostarted if we wanted to.
    - This class was added to the Embrace API like the rest of the hooks in the embracesdk package.
- Use ASM to insert a call to AutoStartInstrumentationHook._preOnCreate() in any onCreate method in any class that extends Application.
- Added a flag to disable auto start instrumentation, though it's turned on by default.

## Testing
Added a unit test, a InstrumentedBytecodeTestCase, and a BytecodeInstrumentationTest, similar to what we use to test other instrumentation.

### DIFF
--- a/embrace-android-sdk/api/embrace-android-sdk.api
+++ b/embrace-android-sdk/api/embrace-android-sdk.api
@@ -1,3 +1,7 @@
+public final class io/embrace/android/embracesdk/AutoStartInstrumentationHook {
+	public static fun _preOnCreate (Landroid/app/Application;)V
+}
+
 public final class io/embrace/android/embracesdk/Embrace : io/embrace/android/embracesdk/internal/api/SdkApi {
 	public static final field Companion Lio/embrace/android/embracesdk/Embrace$Companion;
 	public fun activityLoaded (Landroid/app/Activity;)V
--- a/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/AutoStartInstrumentationHook.java
+++ b/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/AutoStartInstrumentationHook.java
@@ -1,0 +1,28 @@
+package io.embrace.android.embracesdk;
+
+import androidx.annotation.NonNull;
+
+import io.embrace.android.embracesdk.annotation.InternalApi;
+import io.embrace.android.embracesdk.internal.EmbraceInternalApi;
+
+/**
+ * @hide
+ */
+@InternalApi
+public final class AutoStartInstrumentationHook {
+
+    private AutoStartInstrumentationHook() {
+    }
+
+    public static void _preOnCreate(android.app.Application application) {
+        try {
+            Embrace.getInstance().start(application);
+        } catch (Exception exception) {
+            logError(exception);
+        }
+    }
+
+    private static void logError(@NonNull Throwable throwable) {
+        EmbraceInternalApi.getInstance().getInternalInterface().logInternalError(throwable);
+    }
+}
--- a/embrace-gradle-plugin/src/main/java/io/embrace/android/gradle/plugin/api/EmbraceBytecodeInstrumentation.kt
+++ b/embrace-gradle-plugin/src/main/java/io/embrace/android/gradle/plugin/api/EmbraceBytecodeInstrumentation.kt
@@ -43,6 +43,11 @@ abstract class EmbraceBytecodeInstrumentation @Inject internal constructor(objec
         objectFactory.property(Boolean::class.java)
 
     /**
+     * Whether Embrace should automatically start in the Application class. Defaults to true.
+     */
+    val autoStartEnabled: Property<Boolean> = objectFactory.property(Boolean::class.java)
+
+    /**
      * A list of string patterns that are used to filter classes during bytecode instrumentation. For example, 'com.example.foo.*'
      * would avoid instrumenting any classes in the 'com.example.foo' package.
      *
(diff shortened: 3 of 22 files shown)


embrace-android-sdk/api/embrace-android-sdk.api
```diff
@@ -1,3 +1,7 @@
+public final class io/embrace/android/embracesdk/AutoStartInstrumentationHook {
```
fractalwrench, line 1: Ideally this would be under an `internal` package (but have public visibility) and the function would not contain an underscore

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/AutoStartInstrumentationHook.java
```diff
@@ -0,0 +1,28 @@
...
+import androidx.annotation.NonNull;
+
+import io.embrace.android.embracesdk.annotation.InternalApi;
+import io.embrace.android.embracesdk.internal.EmbraceInternalApi;
+
+/**
+ * @hide
+ */
+@InternalApi
+public final class AutoStartInstrumentationHook {
```
fractalwrench, line 12: Why Java?

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/AutoStartInstrumentationHook.java
```diff
@@ -0,0 +1,28 @@
...
+
+    public static void _preOnCreate(android.app.Application application) {
+        try {
+            Embrace.getInstance().start(application);
+        } catch (Exception exception) {
+            logError(exception);
+        }
+    }
+
+    private static void logError(@NonNull Throwable throwable) {
```
fractalwrench, line 25: This won't do anything if the SDK isn't started so IMO could be deleted, and we just swallow any unexpected exception

embrace-gradle-plugin/src/main/java/io/embrace/android/gradle/plugin/api/EmbraceBytecodeInstrumentation.kt
```diff
@@ -42,6 +42,11 @@ abstract class EmbraceBytecodeInstrumentation @Inject internal constructor(objec
     val firebasePushNotificationsEnabled: Property<Boolean> =
         objectFactory.property(Boolean::class.java)
 
+    /**
+     * Whether Embrace should automatically start in the Application class. Defaults to true.
```
fractalwrench, line 46: Defaulting this to true feels like a breaking change as folks will already have setup their own place where Embrace is initialized, which will probably be in a different location to where bytecode instrumentation assumes

embrace-gradle-plugin/src/main/java/io/embrace/android/gradle/plugin/instrumentation/visitor/ApplicationClassAdapter.kt
```diff
@@ -0,0 +1,37 @@
+package io.embrace.android.gradle.plugin.instrumentation.visitor
+
+import com.android.build.api.instrumentation.ClassContext
+import org.objectweb.asm.ClassVisitor
+import org.objectweb.asm.MethodVisitor
+
+/**
+ * Visits the Application class and returns an [ApplicationMethodAdapter] for the onCreate method.
+ */
+class ApplicationClassAdapter(
```
fractalwrench, line 10: This current implementation only works if an `Application` class exists and if the `onCreate` function has an override. I think that would be a reasonable approach but we will need to document this on the docs site (along with the new gradle DSL property)

embrace-gradle-plugin/src/main/java/io/embrace/android/gradle/plugin/instrumentation/config/model/SdkLocalConfig.kt
```diff
@@ -106,7 +106,13 @@ data class SdkLocalConfig(
...
 
     @Json(name = "app_framework")
-    val appFramework: String? = null
+    val appFramework: String? = null,
+
+    /**
+     * Whether auto-start instrumentation should be enabled or not
+     */
+    @Json(name = "auto_start")
+    val autoStart: Boolean? = null
```
fractalwrench, line 115: Why is this required as well as a property on the extension?

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/AutoStartInstrumentationHook.java
bidetofevil, line 12: This isn't really instrumentation, but starting the SDK, so maybe call it `AutoStartSdkHook` or something? And are we still using  the word "hook"

And technically this functionality is just wrapping SDK start, and the "auto" part is the result of doing the bytecode instrumentation. Maybe you can even call it `SdkInitialization` or something, as I think we'll want to add the applicationInitStart/End hooks in there too.

embrace-gradle-plugin/src/main/java/io/embrace/android/gradle/plugin/api/EmbraceBytecodeInstrumentation.kt
bidetofevil: This should probably default to false initially until we do a major version bump. But whether we ever flip it to true is probably something product should decide. Having the "no code start" opt-in-able via a gradle property already seems like a big win.

embrace-gradle-plugin/src/main/java/io/embrace/android/gradle/plugin/instrumentation/config/model/SdkLocalConfig.kt
bidetofevil: Yeah I would just go with property so there isn't two sources of truth. Going with the property also allows customers to use the SDK without a config file, which will make trying it out easier.

PR #2106, opened 2025-04-23
Rename webview instrumentation hooks

fractalwrench:
## Goal

Refactors the WebViewClient bytecode instrumentation hooks so that they are implemented in Kotlin & better reflect how we want bytecode targets to look like.

## Testing


bidetofevil (approved):
LGTM

PR #2107, opened 2025-04-23
Improve view instrumentation bytecode hooks

fractalwrench:
## Goal

Refactors the click bytecode instrumentation hooks so that they are implemented in Kotlin & better reflect how we want bytecode targets to look like.

## Testing


bidetofevil (approved):
LGTM. Do you think the existing tests are sufficient? I don't know about this specifically, but other bytecode instrumented functionality, I see to recall, don't have great coverage....

fractalwrench:
>Do you think the existing tests are sufficient

There is fairly good coverage for checking things worked at build-time (most of the files in this changeset are verifying the bytecode instructions appear as expected).

Testing the behavior at runtime is harder. IMO we should manually verify that towards the end of the project, as it's too onerous to build a test harness capable of verifying all our features end-to-end.

PR #2109, opened 2025-04-23
Alter push notification bytecode instrumentation hooks

fractalwrench:
## Goal

Refactors the push notification bytecode instrumentation hooks so that they are implemented in Kotlin & better reflect how we want bytecode targets to look like.

## Testing


bidetofevil (approved):
LGTM

PR #2114, opened 2025-04-24
Sort listed files in HashSharedObjectFilesTask to avoid flaky tests

priettt:
## Goal

listFiles() doesn't ensure order, so I added a sortedBy call to always get the same architecture and .so files order.

This shouldn't be too expensive, and it'll avoid flakiness in tests


fractalwrench (commented):
I'm not sure this will work 100% of the time as a map is usually unordered. Is the returned implementation of the map ordered, e.g. like a LinkedHashMap, or does this work for other reasons I've missed?

priettt:
> I'm not sure this will work 100% of the time as a map is usually unordered. Is the returned implementation of the map ordered, e.g. like a LinkedHashMap, or does this work for other reasons I've missed?

@fractalwrench While the function does return a map that might be unordered, I don't think it's going to be. Associate is supposed to preserve order: 

> The returned map preserves the entry iteration order of the original collection

The other part where order might be messed up is Moshi serialization, I don't know how they handle map serialization internally. But seeing that the change fixes CI and works locally, I'd say let's just give it a go.

We don't really mind if it's ordered or not outside of the tests.

fractalwrench:
TIL. If there's a guarantee in the stdlib function that the map is ordered, that's good enough for me as it's unlikely to change

fractalwrench (approved):

PR #2115, opened 2025-04-24
Alter how operand stack indices are specified

fractalwrench:
## Goal

Until now the bytecode instrumentation has assumed that we always want to mirror the function signature of the instrumented method in our own target API. This changes that assumption by allowing us to specify what variables should be loaded onto the operand stack.

Practically speaking, this means we can delete unused parameters such as `View.OnClickListener` in our bytecode API. This should simplify the code & therefore reduce surface area for bugs.

## Testing

### DIFF
--- a/embrace-gradle-plugin/src/main/java/io/embrace/android/gradle/plugin/instrumentation/visitor/BytecodeMethodInsertionParams.kt
+++ b/embrace-gradle-plugin/src/main/java/io/embrace/android/gradle/plugin/instrumentation/visitor/BytecodeMethodInsertionParams.kt
@@ -21,13 +21,14 @@ internal data class BytecodeMethodInsertionParams(
     val descriptor: String,
 
     /**
-     * The starting index of the local variable on the operand stack. We assume that by default the bytecode
-     * entrypoint will take the current object as its first parameter, then all other parameters.
+     * Declares the indices of variables on the operand stack that should be added as part of the inserted call.
+     * This MUST be entered in ascending order.
      *
-     * For example, an instrumentation entrypoint for a View.OnClickListener would have the following
-     * signature: instrumentedMethodName(android.view.View.OnClickListener thiz, android.view.View view).
+     * Virtual methods have an implicit reference as the 0th value on the stack, whereas static methods do not.
+     * After this point, the stack should contain the ordered arguments to the containing method.
      *
-     * In future we can revisit this decision, given that the current object is not usually used.
+     * As an example, listOf(1, 3) should be supplied if we wanted to pass the String and Boolean parameters of this virtual method:
+     * foo(String, Int, Boolean)
      */
-    val startVarIndex: Int = 0,
+    val operandStackIndices: List<Int>,
 )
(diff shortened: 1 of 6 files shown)


embrace-gradle-plugin/src/main/java/io/embrace/android/gradle/plugin/instrumentation/visitor/BytecodeMethodInsertionParams.kt
```diff
@@ -21,13 +21,14 @@ internal data class BytecodeMethodInsertionParams(
...
 
     /**
-     * The starting index of the local variable on the operand stack. We assume that by default the bytecode
-     * entrypoint will take the current object as its first parameter, then all other parameters.
+     * Declares the indices of variables on the operand stack that should be added as part of the inserted call.
+     * This MUST be entered in ascending order.
      *
-     * For example, an instrumentation entrypoint for a View.OnClickListener would have the following
-     * signature: instrumentedMethodName(android.view.View.OnClickListener thiz, android.view.View view).
+     * Virtual methods have an implicit reference as the 0th value on the stack, whereas static methods do not.
```
bidetofevil, line 27: The zeroth parameter is `this`?

embrace-gradle-plugin/src/main/java/io/embrace/android/gradle/plugin/instrumentation/visitor/WebViewClientClassAdapter.kt
```diff
@@ -40,7 +40,7 @@ class WebViewClientClassAdapter(
                     owner = "io/embrace/android/embracesdk/internal/instrumentation/bytecode/WebViewClientBytecodeEntrypoint",
                     name = "onPageStarted",
                     descriptor = "(Landroid/webkit/WebView;Ljava/lang/String;Landroid/graphics/Bitmap;)V",
-                    startVarIndex = 1,
+                    operandStackIndices = listOf(1, 2, 3),
                 )
```
bidetofevil, line 44: Is there a test that verifies we can skip a parameter and still be ok?

bidetofevil (approved):
LGTM. A comment about adding an additional test if we don't have it but not a big deal

embrace-gradle-plugin/src/main/java/io/embrace/android/gradle/plugin/instrumentation/visitor/BytecodeMethodInsertionParams.kt
fractalwrench: Yes, for non-static functions `this` is the 0th parameter on the operand stack.

embrace-gradle-plugin/src/main/java/io/embrace/android/gradle/plugin/instrumentation/visitor/WebViewClientClassAdapter.kt
fractalwrench: Yep - there are a bunch of test cases such as `BytecodeInstrumentationTest` and others that verify the generated bytecode matches a golden file. As these haven't changed I'm fairly confident this specific changeset is ok.

PR #2116, opened 2025-04-24
Reduce parameters in bytecode instrumentation

fractalwrench:
## Goal

Builds on #2115 by getting rid of unnecessary parameters in our bytecode API.

## Testing


bidetofevil (approved):
LGTM. 

If there's mismatch being what is declared and what that method actually is/has in terms of static-ness and/or parameter index/type, will the failure automatically surface or does every class need a bespoke test to verify that it's structurally correct? My main concern is if the underlying API we are instrumenting changes due to version bumps, I want tests or builds to blow up automatically, ideally even if we didn't add a test.

fractalwrench:
@bidetofevil I'd recommend taking a look in `embrace-bytecode-instrumentation-tests` and `BytecodeInstrumentationTest`. The first has test cases that asserts the generated bytecode look as expected when run on real-world classes. The latter does that one step further by decompiling an APK and inspecting smali, which confirms whether R8 rules were obeyed.

The aspect we are missing in our tests is asserting that the bytecode doesn't cause crashes at runtime due to signature mismatches etc. There are _some_ assertions in the existing tests that check invalid bytecode isn't generated, but it won't catch everything.

IMO manual verification feels acceptable given how infrequently bytecode instrumentation changes, and the level of effort involved in setting up a test harness for runtime changes. We can discuss that further internally though if needed

PR #2118, opened 2025-04-24
Create class visit strategy hierarchy

fractalwrench:
## Goal

Formalises the strategies that we use to decide whether a class should be visited by a feature during instrumentation. This is an intermediate step towards making it easier to define this via config.


bidetofevil (approved):
LGTM. I assume tests will be added subsequently?

PR #2119, opened 2025-04-24
Extract bytecode instrumentation features

fractalwrench:
## Goal

Extracts the information necessary to bytecode instrument the OkHttp/FCM features out to a separate data class.

## Testing


bidetofevil (approved):
Nice. LGTM

PR #2122, opened 2025-04-25
Read bytecode instrumentation from config file

fractalwrench:
## Goal

Defines the bytecode instrumentation in a bundled JSON file that is deserialized by the gradle plugin before instrumentation. This should eventually allow for bytecode instrumentation features to be added without touching Kotlin code. I've started by porting the OkHttp/FCM push notifications features.

## Testing


bidetofevil (approved):
LGTM

PR #2123, opened 2025-04-25
Add examples for bytecode instrumentation

fractalwrench:
## Goal

Adds code to the example app that can be used to manually verify most of the bytecode instrumentation features.


bidetofevil (approved):
Nice

PR #2124, opened 2025-04-25
Drive webview instrumentation via config

fractalwrench:
## Goal

Drives the webview instrumentation via config. For now the code that adds an override has been left as-is by separating it out into another visitor.

## Testing


bidetofevil (approved):
LGTM

PR #2125, opened 2025-04-25
Drive view click instrumentation via config

fractalwrench:
## Goal

Drives the instrumentation for view clicks via a config file. This covers all the previous scenarios apart from one case where a Java lambda is supplied as below:

```
view.setOnLongClickListener(lambdaView -> {
    Log.d("Embrace", "test");
    return true;
});
```

We could support this by adding an extra method visitor that specifically visits the bytecode this generates, but IMO it's likely not worth it given that few folks are writing Java, and that this will introduce complexity to our instrumentation - particularly knowing that our click instrumentation is imperfect already.

## Testing


bidetofevil (approved):
LGTM. If this worked before and doesn't work now, lets document it in case folks upgrade and stopped seeing these clicks.

PR #2126, opened 2025-04-25
Inline OkHttpNetworkService

priettt:
## Goal

Fixes error: 

```
[EmbraceGradlePlugin] [OkHttpNetworkService] Exception occurred while making network request
java.lang.NullPointerException: Cannot invoke "okhttp3.OkHttpClient.newCall(okhttp3.Request)" because the return value of "io.embrace.android.gradle.plugin.network.OkHttpNetworkService.access$getClient$p(io.embrace.android.gradle.plugin.network.OkHttpNetworkService)" is null
```

### DIFF
--- a/embrace-gradle-plugin/src/main/java/io/embrace/android/gradle/plugin/tasks/ndk/UploadSharedObjectFilesTask.kt
+++ b/embrace-gradle-plugin/src/main/java/io/embrace/android/gradle/plugin/tasks/ndk/UploadSharedObjectFilesTask.kt
@@ -50,10 +50,6 @@ abstract class UploadSharedObjectFilesTask @Inject constructor(
 
     private lateinit var architecturesToHashedSharedObjectFilesMap: Map<String, Map<String, String>>
 
-    private val okHttpNetworkService by lazy {
-        OkHttpNetworkService(requestParams.get().baseUrl)
-    }
-
     @TaskAction
     fun onRun() {
         try {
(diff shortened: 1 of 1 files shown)


embrace-gradle-plugin/src/main/java/io/embrace/android/gradle/plugin/tasks/ndk/UploadSharedObjectFilesTask.kt
```diff
@@ -90,7 +86,7 @@ abstract class UploadSharedObjectFilesTask @Inject constructor(
             variantData.get().name,
             architecturesToHashedSharedObjectFilesMap
         )
-
+        val okHttpNetworkService = OkHttpNetworkService(requestParams.get().baseUrl)
```
bidetofevil, line 89: If this is never reused, any reason why you don't just do the init in `NdkUploadHandshake`?

Or frankly, just extract that method out of the class and put in in this class? It was initially confusing what we are doing - initializing an object and then throwing it away, when it really is just a method call that internally needs an objet

bidetofevil (approved):
LGTM. Comment about why we need to create wrapper object for what is just method call where the networkService creation could just be inlined.

embrace-gradle-plugin/src/main/java/io/embrace/android/gradle/plugin/tasks/ndk/UploadSharedObjectFilesTask.kt
priettt: Mostly because I thought it looked better than inlining 😅 

I can change it if you have a strong preference, but we will probably be revisiting this later

PR #2130, opened 2025-04-28
Fix eager resolution of unitySymbolsDir in getProjectType

priettt:
## Goal

`getProjectType` was wrapped in a provider, without considering its dependency on `unitySymbolsDir`. As a result, when `projectType` was realized, it would try to realize `unitySymbolsDir` eagerly, potentially resolving it before it was available.

This change updates `getProjectType` to use `unitySymbolsDir.map`, wiring the dependency and ensuring that `unitySymbolsDir` is resolved only when needed.


fractalwrench (approved):
LGTM

PR #2132, opened 2025-04-28
Fix eager resolution of unitySymbolsDir in shared object file provider

priettt:
## Goal

`getUnitySharedObjectFiles` was accessing `unitySymbolsDir` (using `orNull`) too early, which could mess with the build or with configuration cache if the symbols directory isn't ready yet.
This change switches to using `flatMap` so `unitySymbolsDir` is resolved lazily, only when needed.

Also removed a test that didn't make sense, as the provider we receive can't be null.


fractalwrench (approved):
LGTM

PR #2144, opened 2025-04-29
Default symbol upload task registration to false if unspecified

bidetofevil:
## Goal

Change the logic to only register if the attribute is defined and true. The early return logic is difficult to read given that the object not being specified doesn't match and the logic proceeds.

## Testing
Added unit tests


priettt (approved):

bidetofevil:
### Merge activity

* **Apr 29, 6:22 PM EDT**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/embrace-io/embrace-android-sdk/2144).

PR #2146, opened 2025-04-30
Uses the Kotlin API for `StatusCode`

fractalwrench:
## Goal

Uses the Kotlin API for `StatusCode` rather than the OTel Java API.

## Testing


bidetofevil (approved):
LGTM

PR #2148, opened 2025-04-30
Use kotlin API for attribute key

fractalwrench:
## Goal

Uses the Kotlin API for `AttributeKey` rather than the OTel Java API.

## Testing


bidetofevil (approved):
LGTM

PR #2151, opened 2025-04-30
Unity projects should upload files when ndk_enabled isn't set

priettt:
## Goal

When the ndk_enabled flag is null: 
- Native projects should not upload native files
- Unity projects should. 

This is how just to make it work like before, but we will be revisiting this behavior in the future.


embrace-gradle-plugin/src/main/java/io/embrace/android/gradle/plugin/tasks/ndk/NdkUploadTasksRegistration.kt
```diff
@@ -38,109 +38,109 @@ class NdkUploadTasksRegistration(
      */
     fun RegistrationParams.execute() {
         // Only proceed if the NDK is enabled - if the config or attribute is unspecified, the default is false
```
bidetofevil, line 40:
```suggestion
        // Bail if ndk_enabled is explicitly set to false. Otherwise, let projectType specific logic decide on the default
```

bidetofevil (approved):
LGTM.

PR #2152, opened 2025-05-01
Unity projects should upload files when ndk_enabled isn't set

priettt:


bidetofevil (approved):
My comment fix didn't get in but that's ok. LGTM

PR #2154, opened 2025-05-01
Rename OTel conversion functions

fractalwrench:
## Goal

Renames some functions that convert OpenTelemetry/Embrace model classes into Embrace payload classes, so that the naming better reflects the action taken.

## Testing


bidetofevil (approved):
LGTM

PR #2155, opened 2025-05-01
Small tweak to logger definition

fractalwrench:
## Goal

Alters how and where the Kotlin API logger is accessed, and makes it more clear what property contains the Kotlin API.

## Testing


bidetofevil (approved):
LGTM

PR #2159, opened 2025-05-02
Add embrace-android-otel module

fractalwrench:
## Goal

Adds a new module that will contain OTel-specific functionality by extracting code out of `embrace-android-core`.


bidetofevil (approved):
LGTM

PR #2161, opened 2025-05-02
Refactor Embrace Attribute implementation

fractalwrench:
## Goal

Refactors Embrace's implementation of OpenTelemetry Attributes. I've achieved this by extracting extension functions out to separate files, renaming symbols to reflect their purpose, and flattening the `FixedAttribute` hierarchy. An additional benefit after these changes is that `EmbraceAttributeKey` is now an [inline value class](https://kotlinlang.org/docs/inline-classes.html) which provides a micro-optimisation whenever users create an attribute.

## Testing


bidetofevil (approved):
THANK YOU

PR #2163, opened 2025-05-02
Apply onlyIf to every task

priettt:
Even if a task is not executed because of its onlyIf, the tasks that depend on it will be executed (nice, gradle ✨).

For now, I'm applying the same onlyIf to every task. I'll check if there's a way of avoiding this in the future.


bidetofevil (approved):
LGTM

PR #2173, opened 2025-05-06
Simplify EmbType hierarchy

fractalwrench:
## Goal

Simplifies the hierarchy of `EmbType` by what seems like an unnecessary interface, given that we are always passing an `EmbType` value throughout our codebase.

## Testing


bidetofevil (approved):
LGTM. The idea is that once semantic conventions are established for the types of telemetry and their schema, they will becomes `TelemetryType` only and not tied to `EmbType` - but that seems a way off and we can always add it back later.

PR #2174, opened 2025-05-06
Remove native ANR sampling feature

fractalwrench:
## Goal

Removes the Unity ANR sampling feature as this is no longer used in the Unity SDK.


bidetofevil (approved):
hell yeah!

PR #2175, opened 2025-05-06
Begin modularising otel usage

fractalwrench:
## Goal

Tweaks the implementation to avoid referencing several OpenTelemetry concepts from within `embrace-android-features`.

## Testing


bidetofevil (approved):
LGTM

PR #2184, opened 2025-05-07
Replace bundleAndSourceMapFilesProvider with separate providers

priettt:
## Goal

Simplified bundle and sourcemap files setting in the react native gradle task.


fractalwrench (approved):
LGTM

fractalwrench (approved):

PR #2198, opened 2025-05-08
Preparatory work for kotlin api

fractalwrench:
## Goal

Creates some wrappers + fakes that are required to move over to using the Kotlin API.

## Testing

This will be covered by existing tests once the internal implementation uses these objects.


bidetofevil (approved):
LGTM. I did similar fakes in my test usage in `opentelemetry-android` so maybe we should have a repo that contains those.

PR #2199, opened 2025-05-08
Inject RN bundleId with ASM

priettt:
## Goal

Similar to what we did with NDK symbols, inject the React BundleID with ASM. This allows us to remove resource injection altogether and access the bundle ID through the project instrumented config.

## Testing
I added an integration test that verifies that the symbols are injected correctly but it's not working. I'll be trying to fix it.


embrace-gradle-plugin/src/main/java/io/embrace/android/gradle/plugin/instrumentation/config/arch/sdk/ProjectConfigInstrumentation.kt
```diff
@@ -4,10 +4,11 @@ import io.embrace.android.gradle.plugin.instrumentation.config.arch.modelSdkConf
 import io.embrace.android.gradle.plugin.instrumentation.config.arch.stringMethod
 import io.embrace.android.gradle.plugin.instrumentation.config.model.VariantConfig
 
-fun createProjectConfigInstrumentation(cfg: VariantConfig) = modelSdkConfigClass {
+fun createProjectConfigInstrumentation(cfg: VariantConfig, reactNativeBundleId: String?) = modelSdkConfigClass {
```
bidetofevil, line 7: Any reason why this is parallel to the config object rather than part of it?

priettt: it's mostly because we can't access it at the same time, but it'd be nice to aggregate the parameters in a class in the future, especially in `createClassVisitor`.

Issue #2207, opened 2025-05-14
Gradle warnings about Embrace Plugin using deprecated features / syntax


jt-gilkeson:
We use the following declaration (from the embrace documentation) in our `build.gradle`:
```
    embrace {
        buildVariantFilter {
            if (it.name.contains("debug")) {
                it.disablePluginForVariant()
                it.disableBytecodeInstrumentationForVariant()
            }
        }
    }
```

The latest gradle complains about the way the embrace plugin is build and the syntax:
`StartParameter.isConfigurationCacheRequested` property has been deprecated (used in the embrace plugin)
and
Properties should be assigned using the 'propName = value' syntax. (buildVariantFilter )

```
The StartParameter.isConfigurationCacheRequested property has been deprecated. This is scheduled to be removed in Gradle 10.0. Please use 'configurationCache.requested' property on 'BuildFeatures' service instead. Consult the upgrading guide for further information: https://docs.gradle.org/8.14/userguide/upgrading_version_8.html#deprecated_startparameter_is_configuration_cache_requested
	at org.gradle.internal.deprecation.StartParameterDeprecations.nagOnIsConfigurationCacheRequested(StartParameterDeprecations.java:28)
	at org.gradle.api.internal.StartParameterInternal.isConfigurationCacheRequested(StartParameterInternal.java:160)
	at java.base/java.lang.reflect.Method.invoke(Unknown Source)
	at io.embrace.android.gradle.plugin.buildreporter.BuildTelemetryCollector.isConfigurationCacheEnabled(BuildTelemetryCollector.kt:85)
	at io.embrace.android.gradle.plugin.buildreporter.BuildTelemetryCollector.collect(BuildTelemetryCollector.kt:38)
	at io.embrace.android.gradle.plugin.buildreporter.BuildTelemetryService$Companion$register$serviceProvider$1$1.invoke(BuildTelemetryService.kt:60)
	at io.embrace.android.gradle.plugin.buildreporter.BuildTelemetryService$Companion$register$serviceProvider$1$1.invoke(BuildTelemetryService.kt:59)
	at io.embrace.android.gradle.plugin.buildreporter.BuildTelemetryService$Companion$register$serviceProvider$1.invoke$lambda$0(BuildTelemetryService.kt:59)
	at org.gradle.api.services.internal.DefaultBuildServicesRegistry$DefaultServiceSpec.parameters(DefaultBuildServicesRegistry.java:382)
[...]

PR #2212, opened 2025-05-21
Get configuration cache telemetry from BuildFeatures

priettt:
BuildFeatures was introduced after Gradle 8.5 to fetch information about the build. Use that instead of StartParameter, which will be deprecated in Gradle 10.

Should fix issue #2207.

## Testing

In the ExampleApp, the warning following warning was showing up when running `./gradlew build --warning-mode all`:
```
The StartParameter.isConfigurationCacheRequested property has been deprecated. This is scheduled to be removed in Gradle 10.0.
```

It doesn't show up anymore after this change.


bidetofevil (approved):
LGTM

PR #2220, opened 2025-05-29
Removed afterEvaluate from AsmTaskRegistration

priettt:
## Goal

Mapped task outputs correctly to asm task parameters without using afterEvaluate. Also added a lazyTaskLookup utility function.


bidetofevil (approved):
LGTM

PR #2221, opened 2025-05-29
Remove 10 parameter limit in OpenTelemetry export

mmarczell-graphisoft:
## Goal

The Android SDK caps the number of custom properties in a log event to 10, even when external OpenTelemetry export is used. The iOS SDK doesn't. This PR lets the OpenTelemetry export contain all the properties without limitation, while the logs sent to the Embrace backend are capped to 10 properties as before.

## Testing

- PropertiesTest now verifies that PropertyUtils does not cap the property count
- EmbraceLogRecordDataTest now verifies that toEmbracePayload does cap the property count

## Release Notes

The number of log properties in the OpenTelemetry export is no longer capped to 10.


embrace-android-otel/src/main/kotlin/io/embrace/android/embracesdk/internal/otel/payload/LogMapper.kt
```diff
@@ -19,5 +19,9 @@ fun LogRecordData.toEmbracePayload(): Log {
     )
 }
 
+const val MAX_PROPERTY_SIZE = 10
+
 fun Attributes.toEmbracePayload(): List<Attribute> =
-    this.asMap().entries.map { Attribute(it.key.key, it.value.toString()) }
+    this.asMap().entries
```
bidetofevil, line 25: @priettt pointed out that you should reverse the order of `take` and `map` - feel free to fix it along with the detekt error or we can deal with it when we make the additional changes (we may change where the truncation happens so don't spend too much time on it).

bidetofevil:
Hey @mmarczell-graphisoft - looks like there are still some test failures because we were expecting the same data in the export as in the internal representation.

I did some refactoring of our internals and reimplemented what you did in #2225, which should take care of your use case. 

Can you test it when it merges?

mmarczell-graphisoft:
@bidetofevil I tested main and it worked, thanks

bidetofevil:
Excellent! Do you still need any part of the PR or can it be closed?

If `main` works as is, we will be doing a release later this week or the next so you'll be able to consume the change in an official release.

PR #2222, opened 2025-05-30
Use new GH action to upload SDK version info

fnewberg:
## Goal

- upload SDK version release data to our version metadata store

## Release Notes

**WHAT**:<br>
**WHY**:<br>


.github/workflows/release-workflow.yml
```diff
@@ -24,6 +24,28 @@ jobs:
...
+        run: |
+          if [ -z "${{ vars.SDK_VERSION_URL }}" ]; then
+            echo "Missing required variable: SDK_VERSION_URL"
+            exit 1
+          fi
+          
+          if [ -z "${{ secrets.SDK_VERSION_TOKEN }}" ]; then
+            echo "Missing required secret: SDK_VERSION_TOKEN"
+            exit 1
+          fi
```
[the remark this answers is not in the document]
fnewberg: you're right in that the other steps will fail. this just makes it really obvious if we haven't configured the repo correctly.

.github/workflows/release-workflow.yml
```diff
@@ -24,6 +24,16 @@ jobs:
     runs-on: ubuntu-latest
     timeout-minutes: 60
     steps:
+      - name: Validate access to version data service
+        uses: embrace-io/public-actions/upload-sdk-version@f4229398f257b24dbbe8b873f78e719f9af6cbbb
```
bidetofevil, line 28: What is this pinned ID for?

fnewberg: that's the sha for the version of the `upload-sdk-version` action. GH warns you if you do anything less strict than that for 3rd party actions. this may seem like a bit of overkill, but is probably a response to attacks that have occurred in the past where people changed the behavior of public actions. even though we control the action, GH sees it as a 3rd party one

PR #2224, opened 2025-06-02
Bypass instrumentation time limits on spans for non-Embrace users

bidetofevil:
## Release Notes

**WHAT**:<br>
**WHY**:<br>


embrace-android-otel/src/main/kotlin/io/embrace/android/embracesdk/internal/otel/sdk/LimitsValidator.kt
```diff
@@ -0,0 +1,52 @@
+package io.embrace.android.embracesdk.internal.otel.sdk
+
+import io.embrace.android.embracesdk.internal.config.instrumented.OtelLimitsConfigImpl
+import io.embrace.android.embracesdk.internal.config.instrumented.schema.OtelLimitsConfig
+import io.embrace.android.embracesdk.spans.EmbraceSpanEvent
+
+class LimitsValidator(
```
priettt: This class isn't only validating limits, but also other things like blank strings. Should we try to think of another name? Maybe something like TelemetryValidator, or SpanDataValidator?

bidetofevil: We should be validating log attributes as well with this, so lets go with `dataValidator`

bidetofevil:
### Merge activity

* **Jun 3, 3:58 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/embrace-io/embrace-android-sdk/2224).

PR #2225, opened 2025-06-02
Bypass limit for log properties when not sending data to Embrace

bidetofevil:
## Release Notes

**WHAT**:<br>
**WHY**:<br>


embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/logs/EmbraceLogService.kt
```diff
@@ -31,6 +31,7 @@ class EmbraceLogService(
 ) : LogService {
 
     private val behavior = configService.logMessageBehavior
+    private val bypassLimitsValidation = configService.isOnlyUsingOtelExporters()
```
mmarczell-graphisoft, line 34: @bidetofevil 

It seems quite an unexpected behaviour from the SDK, that when the Embrace backend is used, it would _also_ truncate logs exported through OTel, but not truncate them otherwise. At the very least this should be documented. But I still think it better to never truncate OTel exported logs.

bidetofevil: They will both be truncated in this implementation. We want the data to be consistent no matter what the source

bidetofevil:
### Merge activity

* **Jun 3, 4:57 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/embrace-io/embrace-android-sdk/2225).

embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/logs/EmbraceLogService.kt
bidetofevil: To explain more clearly, the idea is that when you are an Embrace customer, we will set a limit that we think is reasonable on the data logged. This will be done at instrumentation, so it will apply to all the logs collected.

But if you are not an Embrace customer and do not want our restrictions like yourself, we don't want to foist the limits on you.

The goal here is consistency - all the data exported from the SDK should be the same, regardless the destintation.

PR #2226, opened 2025-06-04
Fix a bug where disabling the sdk might cause a background exception

priettt:
## Goal

`runCatching` doesn't handle exceptions happening on another thread. This PR fixes that issue.


bidetofevil (approved):
This should fix this issue. That said, I would move this entire block after `stop()`. If the SDK isn't started when this is called, we shouldn't trigger a delete, which requires the SDK to have been started.

Either make the purge not require the SDK to have started before or don't invoke it.

PR #2239, opened 2025-06-12
Maven central portal migration

priettt:
## Goal

The legacy OSSRH publishing service will be sunset on June 30th, 2025. [[Notice here]](https://central.sonatype.org/publish/publish-guide/)

[[Here]](https://central.sonatype.org/faq/what-is-different-between-central-portal-and-legacy-ossrh/) are the differences between Legacy OSSRH and Central Portal.

I already migrated the repo through self-service migration and generated a new token, which will be set as a secret in this repo. 

This PR follows the docs on [[this page]](https://central.sonatype.org/publish/publish-portal-ossrh-staging-api/) to use the new URLs. It seems like it's a short-term solution for the migration, as they intend users to use another API for publication. I'd wait until there's an official Gradle plugin released for Central Portal publication, or we could also migrate everything to use https://github.com/vanniktech/gradle-maven-publish-plugin. 

In the meantime, this should allow us to release after June 30th.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -19,7 +19,7 @@ nexusPublishing {
         sonatype {
             username = System.getenv("SONATYPE_USERNAME")
             password = System.getenv("SONATYPE_PASSWORD")
-            nexusUrl.set(uri("https://s01.oss.sonatype.org/service/local/"))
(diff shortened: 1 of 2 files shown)


bidetofevil (approved):
LGTM

PR #2249, opened 2025-06-19
Use typealiases for opentelemetry-java

fractalwrench:
## Goal

Relies on typealiases when using symbols from opentelemetry-java as this was causing confusion now that opentelemetry-kotlin is also a thing.

## Testing


bidetofevil (approved):
LGTM

PR #2250, opened 2025-06-19
Use opentelemetry-kotlin implementation of Tracer

fractalwrench:
## Goal

Alters the SDK to use opentelemetry-kotlin's implementation of `Tracer`. This is not fully complete as the tests are failing due to context propagation not being fully ported over yet. A future PR will address this problem before merging down to the integration branch.


fractalwrench:
>Will these OptIn need to bubble up to code that uses our public APIs?

If anything does bubble up to our API then yes. Right now nothing in this changeset or the future changesets I'll be delivering should impact our public API that way, which will continue to use the opentelemetry-java symbols.

PR #2254, opened 2025-06-20
Dexguard fix: pass dependencies correctly between tasks

priettt:
## Goal

Using flatMap instead of map should carry dependencies from the compression task to the upload task in JvmMappingUploadTaskRegistration.kt.

This should fix the following issue: 

```
Task ':app:jvmMappingUploadTaskForFordexguardApkOnVariantUsProduction' uses this output of task ':app:jvmMappingCompressionTaskForFordexguardAabOnVariantUsProduction' without declaring an explicit or implicit dependency. This can lead to incorrect results being produced, depending on what order the tasks are executed.
```

Also removed old obfuscation tasks that are not present in the gradle versions we support.


priettt:
> How was this change tested?
I'm still waiting for our dexguard license to test that out, but jvm upload for non-dexguard builds is working correctly. I tested it manually.

PR #2257, opened 2025-06-20
Fix react native tests

fractalwrench:
## Goal

The React Native integration test cases were failing as they used a version of Kotlin that's incompatible with what is supplied via opentelemetry-kotlin.


bidetofevil (approved):
LGTM

PR #2258, opened 2025-06-20
Store compressed jvm outputs in different locations for every obfuscation task

priettt:
## Goal

When running `./gradlew assembleRelease bundleRelease`, users where getting an exception: 

```
Task ':app:jvmMappingUploadTaskForFordexguardApkOnVariantUsProduction' uses this output of task ':app:jvmMappingCompressionTaskForFordexguardAabOnVariantUsProduction' without declaring an explicit or implicit dependency. This can lead to incorrect results being produced, depending on what order the tasks are executed.
```

The issue was that different obfuscation tasks (APK vs AAB) were sharing the same output file path:

```
"outputs/embrace/mapping/compressed/${variant.name}/$anchorTaskWithoutVariant/$FILE_NAME_MAPPING_TXT"
```

When running both assembleRelease and bundleRelease in sequence:
- The AAB compression task creates the mapping file
- The APK compression task tries to write to the same file
- The APK upload task tries to read from a file created by a different compression task
Gradle detects this cross-task dependency without a declaration, and fails the build.

With this PR, the compression task output includes the anchor task name, so each obfuscation task type has its own output location.

## Testing

Verified locally on a dexguard test app.


bidetofevil (approved):
Is this only for a specific gradle version? It'd odd that it hasn't come up yet. Or is it that generating both a bundle and a APK for a variant doesn't happen often?

PR #2259, opened 2025-06-21
Retrieve parent from the context using the context key set by the Java SDK

bidetofevil:
## Goal

Get the parent `SpanContext` from the context key set by the Java SDK, which will always be there. Previously, it got it from the context key set by the Embrace SDK, which would not exist if the OTel tracing API implementation was used, as is in `ExternalTracerTest`.

## Testing
Fixed the tests to look for the right things. The `SpanAdapter` returned is all correct other than the fake span stored in `impl` because it was created by the FakeTracer. Also deleted an invalid test as we can really only check that the SpanContext is propagated in that test


fractalwrench (approved):
LGTM

PR #2260, opened 2025-06-23
Schedules a delayed check to stop ANR monitoring if the app is still in background after service start.

priettt:
## Goal

Stop tracking ANRs when they happen in the background, unless the ANR starts on app startup.

## Testing

Working on another PR with integration tests


fractalwrench (approved):
LGTM pending test coverage

PR #2277, opened 2025-06-24
Fix flaky EmbraceAnrServiceTimingTest.kt

priettt:
## Goal

Increase the clock time update to avoid race conditions


bidetofevil (approved):
LGTM

PR #2278, opened 2025-06-25
Use embrace clock in UI load tracing

fractalwrench:
## Goal

Uses the embrace clock in UI load tracing rather than the OTel Java clock. This required some changes to the test assertions.


embrace-android-features/src/main/kotlin/io/embrace/android/embracesdk/internal/capture/startup/AppStartupTraceEmitter.kt
```diff
@@ -44,15 +43,19 @@ import java.util.concurrent.atomic.AtomicReference
...
 internal class AppStartupTraceEmitter(
-    private val clock: OtelJavaClock,
+    private val clock: Clock,
     private val startupServiceProvider: Provider<StartupService?>,
     private val spanService: SpanService,
     private val versionChecker: VersionChecker,
     private val logger: EmbLogger,
     manualEnd: Boolean,
+    private val processCreatedMs: Long? = if (versionChecker.isAtLeast(VERSION_CODES.N)) {
+        Process.getStartElapsedRealtime()
```
fractalwrench: This returns 0 in tests and is used as the span start time. I therefore moved it to the constructor so that we can supply a fake value for our assertions.

embrace-android-features/src/main/kotlin/io/embrace/android/embracesdk/internal/capture/startup/AppStartupTraceEmitter.kt
```diff
@@ -352,15 +341,14 @@ internal class AppStartupTraceEmitter(
...
-            getStartTimeMs()?.let { traceStartTimeMs ->
-                if (applicationInitEndMs != null) {
-                    spanService.recordCompletedSpan(
-                        name = PROCESS_INIT_SPAN,
-                        startTimeMs = traceStartTimeMs,
-                        endTimeMs = applicationInitEndMs,
-                        parent = this,
-                    )
-                }
+            val startMs = applicationInitStartMs
```
fractalwrench: Assumed this was a bug, if not then the assertion in the tests is wrong.

bidetofevil: yeah this is a bug

bidetofevil: Good thing is that `applicationInitEnd()` isn't called often right now. I'll need to fix this before we merge the bytecode instrumentation change

embrace-android-features/src/main/kotlin/io/embrace/android/embracesdk/internal/capture/startup/AppStartupTraceEmitter.kt
bidetofevil: Huh. We can wrap the `Process` object and fake it as well

embrace-android-features/src/main/kotlin/io/embrace/android/embracesdk/internal/capture/startup/AppStartupTraceEmitter.kt
bidetofevil: Oh I see this logic was changed in this commit? I'll just take a look at this as a whole

embrace-android-features/src/main/kotlin/io/embrace/android/embracesdk/internal/capture/startup/AppStartupTraceEmitter.kt
bidetofevil: This is done in #2313

priettt (approved):

PR #2300, opened 2025-07-01
Migrate to vanniktech maven publish

priettt:
## Goal

Use https://github.com/vanniktech/gradle-maven-publish-plugin for publishing to Maven Central. We were using a [temporal workaround](https://central.sonatype.org/publish/publish-portal-ossrh-staging-api/), and this plugin seems better maintained and widely used in many other projects.

Changes:
- Add and use vanniktech.maven.publish.
- Add ORG_GRADLE_PROJECT_ prefix to environment variables in workflows, as the plugin looks for those variables automatically.
- Add publish-snapshot.yml. It will publish a snapshot daily, or on trigger.
- Remove release-workflow.yml, as we now only upload artifacts to maven central, and then we publish manually.
- Updated and renamed upload-artifacts-to-sonatype.yml, we now use it to publish artifacts.
- Remove uses of nexus-publish, maven-publish, and signing plugins, as they're replaced by vanniktech.
- Remove Embrace QA and Beta repos, they aren't used anymore.
- Remove java { withJavadoc... } blocks, as vanniktech adds them automatically.
- Remove gradle plugin marker creation, as vanniktech does that automatically.
- Remove signing blocks, as vanniktech reads the ORG_GRADLE_PROJECT_ env vars in our workflows, and signs releases automatically.
- Will update our internal release docs to reflect how we can test manually.

I released two test versions, one without changes, one with the vanniktech plugin. They both look the same, same artifacts are uploaded, and they both work when tested manually.


.github/workflows/publish-api-docs.yml
```diff
@@ -20,6 +20,16 @@ jobs:
     runs-on: ubuntu-latest
     timeout-minutes: 60
     steps:
+      - name: Validate access to version data service
+        uses: embrace-io/public-actions/upload-sdk-version@88167cd1a3fce3418e26c8c842026e6dfab99e41
```
bidetofevil, line 24: It is ok to pin things to this SHA?

priettt: That's how it's done everywhere, at least for this particular action. I'm not a fan, but I think it's the only way to make an action immutable. @fnewberg might have additional reasons

bidetofevil: Feels like "current version" should be defined as a variable and be referenced everywhere so we don't have to make N changes when the version revs. Not a blocker - just something to consider.

.github/workflows/publish-snapshot.yml
```diff
@@ -0,0 +1,51 @@
...
+env:
+  ORG_GRADLE_PROJECT_mavenCentralUsername: ${{ secrets.SONATYPE_TOKEN_USER }}
+  ORG_GRADLE_PROJECT_mavenCentralPassword: ${{ secrets.SONATYPE_TOKEN_USER_PASSWORD }}
+  ORG_GRADLE_PROJECT_signingInMemoryKeyId: ${{ secrets.MAVEN_ANDROID_SIGNING_KEY }}
+  ORG_GRADLE_PROJECT_signingInMemoryKey: ${{ secrets.MAVEN_ANDROID_GPG_KEY }}
+  ORG_GRADLE_PROJECT_signingInMemoryKeyPassword: ${{ secrets.MAVEN_ANDROID_SIGNING_PASSWORD }}
+
+on:
+  schedule:
+    - cron: '0 16 * * *'  # Runs every day at 16:00 UTC
```
bidetofevil: This is in the middle of a work day. Why not 5 UTC or something to ensure it's done when Jamie gets online?

.github/workflows/publish-snapshot.yml
```diff
@@ -0,0 +1,51 @@
...
+          fi
+
+      - name: Fail if version is not -SNAPSHOT
+        run: |
+          grep -q -- '-SNAPSHOT$' gradle.properties || (echo "ERROR: version must end with -SNAPSHOT" && exit 1)
+
+      - name: Setup Java
+        uses: actions/setup-java@v4
+        with:
+          distribution: 'adopt'
```
bidetofevil, line 43: Is there a way we can apply the same Java config to all of our builds, production, snapshot, and otherwise?

bidetofevil (approved):
LGTM. Some comments about logic consolidation and snapshot timing but not blocking

.github/workflows/publish-snapshot.yml
priettt: we could use a shared workflow, though I'm not sure if it'd work

PR #2302, opened 2025-07-02
Delete release-workflow.yml

priettt:
## Goal

I intended to delete this file in my previous PR but for some reason I only deleted its contents.


fractalwrench (approved):
LGTM

PR #2303, opened 2025-07-02
Update publish-snapshot.yml

priettt:


bidetofevil (approved):
Was this depending on the code already being on the machine when the workflow ran?

PR #2312, opened 2025-07-04
Read App ID from environment variable, small refactor

priettt:
## Goal

Users can now set their App ID with the EMBRACE_APP_ID environment variable. Similar to API token, if both environment variable and config file are set, the later takes precedence.

Also refactored some stuff, namely: 
- Renamed EmbraceVariantConfigurationFileBuildStrategy to BuildVariantConfigFromFile. The previous name was quite long and it didn't really mean anything.
- Removed logic from VariantConfigurationValidator that didn't really validate anything. Now this class only do validations.
- Removed configStr from VariantConfig. It wasn't used anywhere, and it was serialized every time we read the config file.

In progress:
- [x] Add tests.
- [x] Add docs. (https://github.com/embrace-io/embrace-docs/pull/375)


bidetofevil (approved):
nice to see you were able to do this refactor along with adding this functionality and new tests!

PR #2313, opened 2025-07-04
Use new API in T+ to obtain process creation request time

bidetofevil:
## Goal

Create version-aware wrapper around the Android API to get process creation time. This simplifies the app startup instrumentation, and in addition allows us to use a newer API (T+) to obtain more accurate time for when the process zygote is specialized into the app process. The previous one uses the time the process is created, which could be done well before hand to reduce the latency when the app process is created.

## Testing

Unfortunately, there's no way to fake the process start times using Roboletric, so while everything else is unit tested, we can't really differentiate between the two in a test. We just have to trust that API works as expected...


fractalwrench (approved):
LGTM

bidetofevil:
### Merge activity

* **Jul 7, 8:39 PM UTC**: @bidetofevil merged this pull request with [Graphite](https://app.graphite.dev/github/pr/embrace-io/embrace-android-sdk/2313).

PR #2314, opened 2025-07-05
Add ability to track application init time before SDK startup

bidetofevil:
## Goal

Expose the application object init start time tracking without needing the SDK to be started or do a lot of initialization that is deferred to the SDK startup time. This involves plumbing through a clock instance created in `EmbraceImpl` as well as moving some methods around so the application init start time won't result in a set of cascading inits. This time is cached and will be used when the `applicationInitEnd()` method is called.

Right now, this allows for a better fallback for Android 23 or earlier for when the app started up, but in the future, we can track how long the application object took to initialize (or even earlier).

## Testing

Added integration tests - unit tests already exist for this, as this functionality existed but just not exposed publicly.

### DIFF
--- a/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/EmbraceImpl.kt
+++ b/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/EmbraceImpl.kt
@@ -33,6 +33,8 @@ import io.embrace.android.embracesdk.internal.api.delegate.SdkStateApiDelegate
 import io.embrace.android.embracesdk.internal.api.delegate.SessionApiDelegate
 import io.embrace.android.embracesdk.internal.api.delegate.UserApiDelegate
 import io.embrace.android.embracesdk.internal.api.delegate.ViewTrackingApiDelegate
+import io.embrace.android.embracesdk.internal.clock.Clock
+import io.embrace.android.embracesdk.internal.clock.NormalizedIntervalClock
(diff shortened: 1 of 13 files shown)


embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/EmbraceImpl.kt
```diff
@@ -337,4 +341,21 @@ internal class EmbraceImpl @JvmOverloads constructor(
...
         }
+
+    override fun applicationInitStart() {
+        if (applicationInitStartMs == null) {
+            applicationInitStartMs = clock.now()
+        }
+    }
+
+    override fun applicationInitEnd() {
+        if (sdkCallChecker.check("application_init_end")) {
```
fractalwrench, line 352: What happens if the SDK hasn't been initialized at this point? Are we ok with just ignoring the consequences of this?

bidetofevil: Yeah, it just doesn't work, which is OK since this is optional.

bidetofevil:
### Merge activity

* **Jul 7, 9:05 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/embrace-io/embrace-android-sdk/2314).

PR #2316, opened 2025-07-07
Remove unused OTel java function

fractalwrench:
## Goal

Removes an unused function call that relies on OTel Java.


bidetofevil (approved):
Huh. Wonder why we needed an overload that passes the clock in....

PR #2319, opened 2025-07-07
Use kotlin API for exporters

fractalwrench:
## Goal

Refactors the span/log exporter implementation to use the Kotlin API where possible.

## Testing


bidetofevil (approved):
LGTM

PR #2324, opened 2025-07-08
Use kotlin symbol for SpanContext links

fractalwrench:
## Goal

Uses the kotlin symbol for `SpanContext` links in our implementation.


bidetofevil (approved):
LGTM

PR #2325, opened 2025-07-08
Avoid relying on compat module where not needed

fractalwrench:
## Goal

Removes a dependency on the compat module in everywhere apart from `embrace-android-otel`. In other modules we should purely rely on the public API of the Kotlin SDK.


bidetofevil (approved):
LGTM

PR #2327, opened 2025-07-08
Add InstrumentationTargetMethodEndVisitor to inject calls at the end of methods

priettt:
## Goal

With this change, we can now add `insertAtEnd` in the insert block of an instrumentation instruction in `bytecode_instrumentation_features.json`:
```
"insert": {
       ...
        "insertAtEnd": true
      },
```

With this, we'll inject the call with ASM at the end of the method, rather than the beginning. 

"The end" is going to be any return or throw instruction, so our code will be injected immediately before those lines.

### DIFF
--- a/embrace-bytecode-instrumentation-tests/src/main/java/io/embrace/test/fixtures/TargetMethodEndVisitorObj.kt
+++ b/embrace-bytecode-instrumentation-tests/src/main/java/io/embrace/test/fixtures/TargetMethodEndVisitorObj.kt
@@ -1,0 +1,29 @@
+package io.embrace.test.fixtures
+
+import kotlin.random.Random
+
+class TargetMethodEndVisitorObj {
+    fun instrumentAtEndWithReturn() {
+        print("Instrumenting at the end of a method with a return statement")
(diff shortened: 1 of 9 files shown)


priettt:
> This seems like a reasonable approach but is lacking test coverage which is a requirement for merging a change in this complex area IMO.
> 
> I'd also say that we need to try out this new visitor with a few representative functions in an actual app, to confirm that nothing breaks. The scope of our tests means things can slip through so some manual verification is required whenever we alter the building blocks of our bytecode instrumentation.

@fractalwrench do we have a way to just test the instrumentation visitor, without context? I added tests in the PR where this instrumentation is used: https://github.com/embrace-io/embrace-android-sdk/pull/2329. Do you think we should add additional testing?

fractalwrench:
@priettt `MethodReturnValueVisitorTest` is a good example of an isolated unit test for a visitor. I'd say it's worth adding an extra test case given that this forms part of the foundation of our bytecode instrumentation.

fractalwrench (approved):
LGTM

PR #2329, opened 2025-07-08
Instrument calls to applicationInitStart and applicationInitEnd

priettt:
## Goal

Added bytecode instrumentation for adding calls to applicationInitStart and applicationInitEnd on any Application.onCreate()

The resulting code should look like this: 

```
class TestApplication : Application() {
    override fun onCreate() {
        // inject call to applicationInitStart
        super.onCreate()
        // inject call to applicationInitEnd
    }
}
```
## Testing

Added bytecode instrumentation tests 

Tested manually:
- [x] Calls to applicationInitStart and applicationInitEnd are made automatically by default.
- [x] The instrumentation works correctly along `autoSdkInitializationEnabled`.
- [x] Calls are injected even when multiple return instructions are present in onCreate, or when an error is thrown.
- [x] Calls are not instrumented when `applicationInitTimingEnabled` is set to false in the `embrace.bytecodeInstrumentation` config

### DIFF
--- a/embrace-gradle-plugin/src/main/java/io/embrace/android/gradle/plugin/api/EmbraceBytecodeInstrumentation.kt
+++ b/embrace-gradle-plugin/src/main/java/io/embrace/android/gradle/plugin/api/EmbraceBytecodeInstrumentation.kt
@@ -42,14 +42,9 @@ abstract class EmbraceBytecodeInstrumentation @Inject internal constructor(objec
     val autoSdkInitializationEnabled: Property<Boolean> = objectFactory.property(Boolean::class.java)
 
     /**
-     * Whether Embrace should automatically instrument application initialization timing start. Defaults to true.
+     * Whether Embrace should automatically instrument application initialization timing. Defaults to true.
      */
-    val applicationInitTimeStartEnabled: Property<Boolean> = objectFactory.property(Boolean::class.java).convention(true)
-
-    /**
-     * Whether Embrace should automatically instrument application initialization timing end. Defaults to true.
-     */
-    val applicationInitTimeEndEnabled: Property<Boolean> = objectFactory.property(Boolean::class.java).convention(true)
+    val applicationInitTimingEnabled: Property<Boolean> = objectFactory.property(Boolean::class.java).convention(true)
 
     /**
      * Whether Embrace should automatically instrument push notifications from Firebase. Defaults to false.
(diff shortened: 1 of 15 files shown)


bidetofevil (approved):
LGTM!

embrace-gradle-plugin/src/main/java/io/embrace/android/gradle/plugin/api/EmbraceBytecodeInstrumentation.kt
```diff
@@ -41,6 +41,16 @@ abstract class EmbraceBytecodeInstrumentation @Inject internal constructor(objec
...
 
+    /**
+     * Whether Embrace should automatically instrument application initialization timing start. Defaults to true.
+     */
+    val applicationInitTimeStartEnabled: Property<Boolean> = objectFactory.property(Boolean::class.java).convention(true)
+
+    /**
+     * Whether Embrace should automatically instrument application initialization timing end. Defaults to true.
+     */
+    val applicationInitTimeEndEnabled: Property<Boolean> = objectFactory.property(Boolean::class.java).convention(true)
```
fractalwrench: IMO these should be one flag, `applicationInitTimeEnabled`. I can't think of a scenario where folks would only want to disable one.

fractalwrench: If these are true then presumably we want to set `end_startup_with_app_ready` automatically to true @bidetofevil?  https://embrace.io/docs/android/features/performance-instrumentation/#configuration

priettt: you're right, I think there's something that fails the build if you don't have flags for a certain bytecode instrumentation, so I added a flag to make that pass, but I forgot to remove it 🤦‍♂️

bidetofevil: One flag works. If they want to move one but not the other, well, just manually instrument the other one too. Ain't that hard.

bidetofevil: Answer in slack, but for posterity, that flag controls the terminal event of the entire trace while these ones simply get us the start and end of the `Application.onCreate()` method, so they can be tweaked independently.

fractalwrench (approved):
LGTM

PR #2355, opened 2025-07-14
Added tests for versions that need desugaring

priettt:
## Goal

Verify that the app doesn't build if desugaring is not enabled for apps with min sdk lower than 26, and verify that build is correct when desugaring is enabled.

### DIFF
--- a/embrace-gradle-plugin-integration-tests/src/main/java/io/embrace/android/gradle/integration/framework/IntegrationTestExtension.kt
+++ b/embrace-gradle-plugin-integration-tests/src/main/java/io/embrace/android/gradle/integration/framework/IntegrationTestExtension.kt
@@ -84,6 +84,8 @@ abstract class IntegrationTestExtension(objectFactory: ObjectFactory) {
 
         val android = checkNotNull(project.extensions.findByType(ApplicationExtension::class.java))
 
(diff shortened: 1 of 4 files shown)


embrace-gradle-plugin-integration-tests/src/main/java/io/embrace/android/gradle/integration/framework/IntegrationTestExtension.kt
```diff
@@ -110,4 +112,18 @@ abstract class IntegrationTestExtension(objectFactory: ObjectFactory) {
...
+        val android = checkNotNull(project.extensions.findByType(ApplicationExtension::class.java))
+
+        android.compileOptions {
+            // Enable core library desugaring
+            isCoreLibraryDesugaringEnabled = true
+        }
+
+        dependencies.add(
+            "coreLibraryDesugaring",
+            "com.android.tools:desugar_jdk_libs:2.1.5"
```
bidetofevil, line 135: The line to include in a gradle file is `coreLibraryDesugaring("com.android.tools:desugar_jdk_libs:<version>")`. Is this what the function will do?

bidetofevil (approved):
LGTM

PR #2359, opened 2025-07-15
Require AGP 8.3+ and special flag if minSdk < 26

bidetofevil:
## Goal

Bump required AGP version required for supporting minSdk < 26. This is needed because desugaring is now needed even when minSdk is 25 due to an OTel SDK change in `1.50.0`. As a result, we need to ensure that the AGP workaround is applied in those cases.


fractalwrench (approved):
LGTM

bidetofevil:
### Merge activity

* **Jul 15, 2:45 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/embrace-io/embrace-android-sdk/2359).

PR #2361, opened 2025-07-15
Remove dead code

fractalwrench:
## Goal

Removes some dead code that was unused.


bidetofevil (approved):
I'll revive the `ExecutionCordinator` when I have time to fix those race condition tests I had. Which is never 😭

PR #2362, opened 2025-07-15
Miscellaneous OTel tweaks

fractalwrench:
## Goal

Miscellaneous tweaks to use the new Kotlin API in the embrace-android-otel module.


bidetofevil (approved):
LGTM

PR #2367, opened 2025-07-16
Update upload-artifacts-to-maven-central.yml to accept hotfixes

priettt:
## Goal

Now we can pass 7.6.1, and it'll checkout release/7.6.0 on the checkout step.


bidetofevil (approved):
LGTM. I don't think the code injection is a problem because we've locked down who can run this. Maybe we can encode this or something later

PR #2385, opened 2025-07-18
Add entrypoint for OTel kotlin

fractalwrench:
## Goal

Adds an entrypoint for accessing OTel kotlin as an escape hatch. This is marked as an experimental API so will be subject to change without warning.


bidetofevil (approved):
Oooooooo

PR #2389, opened 2025-07-18
Add desguaring to example app

bidetofevil:
## Goal

Add desugaring to the example app and bump the version manually


fractalwrench (approved):
LGTM

PR #2390, opened 2025-07-19
Performance tweaks to various CI jobs

vitaliyf:
## Goal

Speed up builds and various other housekeeping

### DIFF
--- a/.github/workflows/android-emulator-tests.yml
+++ b/.github/workflows/android-emulator-tests.yml
@@ -1,4 +1,4 @@
-name: Android Emulator
+name: Android
 
 on:
   workflow_dispatch:
@@ -17,7 +17,7 @@ permissions:
   contents: read
 
 jobs:
-  android-emulator:
+  emulator:
     runs-on: ubuntu-latest
     timeout-minutes: 45
     strategy:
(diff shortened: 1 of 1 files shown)


.github/workflows/ci-gradle.yml
```diff
@@ -52,20 +69,27 @@ jobs:
         run: /bin/sh .github/fetch-robolectric-dependencies.sh
 
       # Build the entire project, run the tests, and run all static analysis
-      - name: Gradle Build
-        run: ./gradlew assembleRelease check -x embrace-gradle-plugin-integration-tests:test --stacktrace
+      - name: Gradle (embrace-gradle-plugin-integration-tests:test)
+        if: ${{ matrix.gradle-task == 'test' }}
+        run: ./gradlew embrace-gradle-plugin-integration-tests:test --stacktrace --parallel
```
fractalwrench: --parallel doesn't have an effect as we already have `org.gradle.parallel` in the gradle.properties

.github/workflows/ci-gradle.yml
```diff
@@ -7,20 +7,37 @@ on:
   release:
     types: [ released ]
 
-concurrency:
```
fractalwrench: This is used to cancel CI jobs for PRs that are updated at a later date with additional commits etc. Is there an alternate workflow we should follow for canceling these with Depot?

.github/workflows/ci-gradle.yml
vitaliyf: yep, i was trying to check if gradle.props is definitely taking effect - which it does, as --parallel made no difference

.github/workflows/ci-gradle.yml
vitaliyf: if you want to do that - then i can reinstate this back. we don't do this in any other repos (we let jobs continue to run) but certainly there is no harm in killing them as long as everyone knows it'll happen

fractalwrench: I think it'd be good to reinstate this if possible - everyone on the team is aware that it will happen

PR #2391, opened 2025-07-21
Enable CC parallel cache

fractalwrench:
## Goal

Enables parallel caching for Gradle's configuration cache, and tweaks a few other build options that should improve build perf.


bidetofevil (approved):
Lets go!

PR #2394, opened 2025-07-22
Add internal APIs for creating network request spans

fractalwrench:
## Goal

Adds an internal API for creating network request spans. This is required as we need to construct a span before sending a network request when we want to inject the traceparent as a HTTP header, since the traceparent should contain the same span/trace ID as the created span. Whilst we could do this internally just for Android, this functionality should also be accessible to the hosted SDKs so that they are subject to the same domain rate-limiting logic.

## Testing

Added unit + integration test cases.


bidetofevil (commented):
So I think the request where the traceparent needs to be injected into is the underlying network request that is sent to the customers's server. Have a look at `EmbraceOkHttp3NetworkInterceptor` and where `embrace.generateW3cTraceparent()` is called and injected into the actual request header.

If the backend simply wants to use the traceparent culled from the network span itself, they can just assemble it together as we have rather than pull from an attribute.

PR #2402, opened 2025-07-24
Add truncation logic go span names and attribute keys/values

bidetofevil:
## Goal

Previously, we dropped any attribute and spans whose names are too long. Now, we implementation truncation logic so that it will always succeed. There may be clashes as a result, but it should be apparent in the data as the names/keys/values will end with three dots.


bidetofevil:
* **#2402** <a href="https://app.graphite.dev/github/pr/embrace-io/embrace-android-sdk/2402?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/embrace-io/embrace-android-sdk/2402?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

embrace-android-otel/src/main/kotlin/io/embrace/android/embracesdk/internal/otel/spans/OtelSpanStartArgs.kt
```diff
@@ -24,11 +24,10 @@ class OtelSpanStartArgs(
 ) {
     var parentContext: OtelJavaContext = OtelJavaContext.root()
 
-    val spanName: String = if (internal) {
-        name.toEmbraceObjectName()
-    } else {
-        name
-    }
+    var spanName: String = name.appendEmbracePrefix(internal)
```
fractalwrench, line 27: I'm not a massive fan of making this mutable given a span name should not change after it's started. It suggests to me that we're getting the wrong abstraction around this class. Having said that, I'm refactoring how this works anyway to get context working so maybe it's fine for the purposes of this changeset?

embrace-android-otel/src/main/kotlin/io/embrace/android/embracesdk/internal/otel/sdk/DataValidator.kt
```diff
@@ -11,45 +12,87 @@ class DataValidator(
...
+        return Pair(PropertyUtils.truncate(key, maxKeyLength), truncatedValue)
     }
+
+    private fun Map<String, String>.truncate(
+        maxCount: Int,
+        maxKeyLength: Int,
+        maxValueLength: Int,
+    ): Map<String, String> =
+        mutableMapOf<String, String>().apply {
+            this@truncate.entries.take(maxCount).forEach {
```
fractalwrench, line 89: `map` might be preferable to `forEach` IMO

bidetofevil: Ah, I'll have a look

embrace-android-otel/src/main/kotlin/io/embrace/android/embracesdk/internal/otel/spans/OtelSpanStartArgs.kt
bidetofevil: For this to be immutable, we'd have to do truncation before it gets down to this layer, and there were some not-niceness about that. I'll look at it again and if it's really not doable reasonably, I'll add a comment and punt it down the road

embrace-android-otel/src/main/kotlin/io/embrace/android/embracesdk/internal/otel/sdk/DataValidator.kt
bidetofevil: Changed to directly use `associate`. Unfortunately, I put the commit on the subsequent PR, but since this is addressed, I'm not going to fuss too much about it

embrace-android-otel/src/main/kotlin/io/embrace/android/embracesdk/internal/otel/spans/OtelSpanStartArgs.kt
bidetofevil: I actually did the opposite - I made this immutable but put the truncation logic lower down by making this a pass-through parameter - it will be set on the Span when it is created, and the truncation will happen at that layer. The code that looked at this as a source of truth was removed, which was easy because we already had to support updating the name of a started span, so this is actually a good move even without this initial mutable requirement.

This change is part of #2404 because I did the work on the wrong branch lol

PR #2405, opened 2025-07-25
Combine span creator with span start args

fractalwrench:
## Goal

Combines `OtelSpanCreator` with `OtelSpanStartArgs` as they're effectively fulfilling the same functionality, but there's an unnecessary layer that makes it harder to reason about span creation.

## Testing


embrace-android-otel/src/test/kotlin/io/embrace/android/embracesdk/internal/otel/spans/OtelSpanCreatorTest.kt
priettt, line 1: should we rename this test?

PR #2413, opened 2025-07-29
Use additional otel kotlin symbols

fractalwrench:
## Goal

Use additional opentelemetry-kotlin symbols rather than opentelemetry-java.


bidetofevil (approved):
LGTM

PR #2432, opened 2025-08-07
Logging improvements in the embrace-gradle-plugin

priettt:
## Summary
- Added an EmbraceLogger class that can be created with either a Class or a String.
- Make use of loggers consistent across the plugin.
- Renamed Logger to EmbraceLogger.
- Add a log that indicates that our plugin was applied.

This PR should make our logging better. When customers saw our logs in some places like task registering, they weren't seeing any indication about who generated those logs. Now, they all use the same prefix. 

Also, while we removed a lot of logs because we were quite verbose, I think we should at least have an info log indicating that the plugin was applied.

### DIFF
--- a/embrace-gradle-plugin/src/main/java/io/embrace/android/gradle/plugin/EmbraceLogger.kt
+++ b/embrace-gradle-plugin/src/main/java/io/embrace/android/gradle/plugin/EmbraceLogger.kt
@@ -7,19 +7,11 @@ import org.gradle.api.logging.Logging
 /**
  * Embrace-specific logger that automatically adds plugin and class prefixes.
  */
-class EmbraceLogger {
-    private val gradleLogger: org.gradle.api.logging.Logger
-    private val logPrefix: String
+class EmbraceLogger(componentName: String) {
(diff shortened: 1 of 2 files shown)


embrace-gradle-plugin/src/main/java/io/embrace/android/gradle/plugin/EmbraceLogger.kt
```diff
@@ -0,0 +1,51 @@
...
+
+    constructor(clazz: Class<*>) {
+        gradleLogger = Logging.getLogger(clazz)
+        logPrefix = "[EmbraceGradlePlugin] [${clazz.simpleName}]"
+    }
+
+    constructor(componentName: String) {
+        gradleLogger = Logging.getLogger(componentName)
+        logPrefix = "[EmbraceGradlePlugin] [$componentName]"
+    }
```
fractalwrench: nit: suggest a primary constructor with a `componentName` parameter that these secondary constructors call, rather than duplicating code

fractalwrench (approved):
LGTM

PR #2434, opened 2025-08-07
Improve React Native project detection in gradle plugin

priettt:
## Summary
- Use `withPlugin("com.facebook.react")` instead of checking `node_modules/react-native` directory.
- Create separate task registration flow for React Native specific tasks.

Uses Gradle's plugin system instead of file system checks, which are failing for some customers. This should be configuration-cache safe, as it's the same thing we use for waiting for AGP. React Native tasks still register only when the React Native plugin is detected

## Changes
- Remove `isReactNativeProject` property from `PluginBehavior`
- Add `withPlugin("com.facebook.react")` detection in `EmbraceGradlePluginDelegate` 
- Create `registerReactNativeTasks()` method for RN-specific task registration
- Extract shared methods to reduce code duplication
- Update tests to remove obsolete React Native detection logic

### DIFF
--- a/embrace-gradle-plugin/src/main/java/io/embrace/android/gradle/plugin/EmbraceGradlePluginDelegate.kt
+++ b/embrace-gradle-plugin/src/main/java/io/embrace/android/gradle/plugin/EmbraceGradlePluginDelegate.kt
@@ -57,6 +57,11 @@ class EmbraceGradlePluginDelegate {
 
         taskRegistrar.registerTasks()
 
+        // Register React Native specific tasks when React Native plugin is also applied
+        project.pluginManager.withPlugin("com.facebook.react") {
+            taskRegistrar.registerReactNativeTasks()
+        }
+
         project.afterEvaluate { evaluatedProject ->
             onProjectEvaluated(evaluatedProject, agpWrapper)
         }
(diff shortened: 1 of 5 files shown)


embrace-gradle-plugin/src/main/java/io/embrace/android/gradle/plugin/EmbraceGradlePluginDelegate.kt
```diff
@@ -57,6 +57,11 @@ class EmbraceGradlePluginDelegate {
 
         taskRegistrar.registerTasks()
 
+        // Register React Native specific tasks when React Native plugin is also applied
+        project.pluginManager.withPlugin("com.facebook.react") {
```
fractalwrench, line 61: Is this available in all supported RN versions? I remember it used to be react.gradle

fractalwrench (approved):
LGTM

embrace-gradle-plugin/src/main/java/io/embrace/android/gradle/plugin/EmbraceGradlePluginDelegate.kt
priettt: yes, at least for the ones we support (0.71+)

PR #2439, opened 2025-08-08
Look for openConnection method in ancestor of wrapped URLStreamHandler

bidetofevil:
Traverse up the object hierarchy to find the `openConnection` method to be invoked in an existing URLStreamingHandler installed on an app before the Embrace SDK started. Previously, we were only looking in the wrapped class, but it could have inherited the method from an ancestor instead.


fractalwrench (approved):
LGTM

PR #2471, opened 2025-08-26
Improve AgpSupportTests

priettt:
## Goal

Add some java and kotlin files to the fixtures run in AgpSupportTest, so the compileJava and compileKotlin tasks are actually run, and not skipped for not having sources. Also set up autoAddEmbraceDependencies, so we verify that transitive dependencies don't cause any issues


fractalwrench (commented):
I think you'll need to supply an AndroidManifest.xml in the test fixture that uses this, as the default `FakeAndroidManifest.xml` sets `android:hasCode` as false

PR #2487, opened 2025-08-29
Enable Embrace SDK to use the Kotlin SDK

bidetofevil:
## Release Notes

**WHAT**:<br>
**WHY**:<br>

### DIFF
--- a/embrace-android-otel/src/main/kotlin/io/embrace/android/embracesdk/internal/otel/config/OtelSdkConfig.kt
+++ b/embrace-android-otel/src/main/kotlin/io/embrace/android/embracesdk/internal/otel/config/OtelSdkConfig.kt
@@ -24,6 +24,11 @@ import io.opentelemetry.semconv.incubating.OsIncubatingAttributes
 import io.opentelemetry.semconv.incubating.TelemetryIncubatingAttributes
 import java.util.concurrent.ConcurrentHashMap
 
+/**
+ * Globally referenceable flag that determines which OTel SDK implementation is in use for the production SDK at runtime.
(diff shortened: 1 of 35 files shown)


embrace-android-otel/src/main/kotlin/io/embrace/android/embracesdk/internal/otel/config/OtelSdkConfig.kt
```diff
@@ -24,6 +24,11 @@ import io.opentelemetry.semconv.incubating.OsIncubatingAttributes
 import io.opentelemetry.semconv.incubating.TelemetryIncubatingAttributes
 import java.util.concurrent.ConcurrentHashMap
 
+/**
+ * Globally referenceable flag that determines which OTel SDK implementation is in use for the production SDK at runtime.
+ */
+const val USE_KOTLIN_SDK: Boolean = false
```
fractalwrench, line 30: Do we have plans to make this part of RemoteConfig in a future PR, or will an end-user always have to set this value locally?

fractalwrench (approved):
LGTM

embrace-android-otel/src/main/kotlin/io/embrace/android/embracesdk/internal/otel/config/OtelSdkConfig.kt
bidetofevil: This will be a remote config, which may mean further refactoring. Right now, this is just a quick and dirty flag that is not meant to be changed, but will soon be able to be.

PR #2492, opened 2025-09-04
Update embrace-android-sdk to use new API symbols

fractalwrench:
## Goal

Some of the API surface on opentelemetry-kotlin was updated - this changeset updates embrace-android-sdk to use the correct symbols.


embrace-android-otel/src/main/kotlin/io/embrace/android/embracesdk/internal/otel/OpenTelemetryInstanceExt.kt
```diff
@@ -1,34 +1,33 @@
 package io.embrace.android.embracesdk.internal.otel
```
bidetofevil, line 1: This file should be renamed since `OpenTelemetryInstance` isn't a thing anymore?

bidetofevil (approved):
LGTM

PR #2494, opened 2025-09-05
Wrap exposed openTelemetryKotlin in EmbOpenTelemetry with EmbTracerProvider

priettt:
## Goal

The otel-kotlin instance we expose should behave the same as the otel-java instance. 

This PR modifies the instance returned by our getEmbraceOpenTelemetry method to return tracers wrapped in EmbTracer. This will set some internal properties and wire it to our internal SpanService.

Also modify the rest of the codebase to access the wrapped open telemetry instance instead of the private kotlinApi field.


fractalwrench (commented):
`kotlinApi` should be made private on `OtelSdkWrapper` IMO. We don't want to expose 2 ways of accessing the same OTel instance

fractalwrench (approved):
LGTM

PR #2496, opened 2025-09-08
Move OTel Java API to a separate opt-in module

priettt:
Introduced a new `embrace-android-otel-java` module that can be added by users who need otel-java integration. 

The following methods that were present in the Embrace class are now in the new module:
```
getOpenTelemetry(): OtelJavaOpenTelemetry
addSpanExporter(spanExporter: OtelJavaSpanExporter)
addLogRecordExporter(logRecordExporter: OtelJavaLogRecordExporter)
setResourceAttribute(key: OtelJavaAttributeKey<String>, value: String)
```

As this is a change on our API, we should release this on a new major version.

The openTelemetryKotlin instance we expose is now wrapped in an EmbTracerProvider, so it behaves as the Java method we expose.

The changes were made progressively on separate commits, so this will be best reviewed commit-by-commit.

### DIFF
--- a/embrace-android-core/src/test/java/io/embrace/android/embracesdk/internal/config/ConfigServiceImplTest.kt
+++ b/embrace-android-core/src/test/java/io/embrace/android/embracesdk/internal/config/ConfigServiceImplTest.kt
@@ -3,7 +3,7 @@ package io.embrace.android.embracesdk.internal.config
 import io.embrace.android.embracesdk.concurrency.BlockingScheduledExecutorService
 import io.embrace.android.embracesdk.core.BuildConfig
 import io.embrace.android.embracesdk.fakes.FakeClock
-import io.embrace.android.embracesdk.fakes.FakeOtelJavaLogRecordExporter
+import io.embrace.android.embracesdk.fakes.FakeLogRecordExporter
 import io.embrace.android.embracesdk.fakes.FakePreferenceService
 import io.embrace.android.embracesdk.fakes.FakeProcessStateService
 import io.embrace.android.embracesdk.fakes.config.FakeInstrumentedConfig
@@ -20,6 +20,7 @@ import io.embrace.android.embracesdk.internal.payload.AppFramework
 import io.embrace.android.embracesdk.internal.prefs.PreferencesService
 import io.embrace.android.embracesdk.internal.session.lifecycle.ProcessStateService
 import io.embrace.android.embracesdk.internal.worker.BackgroundWorker
+import io.embrace.opentelemetry.kotlin.ExperimentalApi
 import io.mockk.clearAllMocks
 import io.mockk.mockkStatic
 import io.mockk.unmockkAll
@@ -33,6 +34,7 @@ import org.junit.Before
 import org.junit.BeforeClass
 import org.junit.Test
 
+@OptIn(ExperimentalApi::class)
 internal class ConfigServiceImplTest {
 
     private lateinit var fakePreferenceService: PreferencesService
@@ -160,7 +162,7 @@ internal class ConfigServiceImplTest {
             BuildConfig.VERSION_NAME,
             SystemInfo()
         )
-        cfg.addLogExporter(FakeOtelJavaLogRecordExporter())
+        cfg.addLogExporter(FakeLogRecordExporter())
         val service = createService(config = cfg, appId = null)
         assertNotNull(service)
         assertTrue(service.isOnlyUsingOtelExporters())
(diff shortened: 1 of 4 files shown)


embrace-android-api/api/embrace-android-api.api
```diff
@@ -105,19 +105,11 @@ public abstract interface class io/embrace/android/embracesdk/internal/api/Netwo
 
 public abstract interface class io/embrace/android/embracesdk/internal/api/OTelApi {
 	public abstract fun addLogRecordExporter (Lio/embrace/opentelemetry/kotlin/logging/export/LogRecordExporter;)V
-	public abstract fun addLogRecordExporter (Lio/opentelemetry/sdk/logs/export/LogRecordExporter;)V
 	public abstract fun addSpanExporter (Lio/embrace/opentelemetry/kotlin/tracing/export/SpanExporter;)V
-	public abstract fun addSpanExporter (Lio/opentelemetry/sdk/trace/export/SpanExporter;)V
-	public abstract fun getOpenTelemetry ()Lio/opentelemetry/api/OpenTelemetry;
 	public abstract fun getOpenTelemetryKotlin ()Lio/embrace/opentelemetry/kotlin/OpenTelemetry;
```
fractalwrench, line 109: We should probably take a similar approach for any symbols from opentelemetry-kotlin, but that can be done in another changeset.

embrace-android-otel-java/src/main/kotlin/io/embrace/android/embracesdk/otel/java/EmbraceOtelJavaExtensions.kt
```diff
@@ -0,0 +1,51 @@
...
+    this.addLogRecordExporter(logRecordExporter.toOtelKotlinLogRecordExporter())
+}
+
+/**
+ * Returns an [OtelJavaOpenTelemetry] that provides working [Tracer] implementations that will record spans that fit into the Embrace data
+ * model.
+ */
+@OptIn(ExperimentalApi::class)
+fun Embrace.getJavaOpenTelemetry(): OtelJavaOpenTelemetry {
+    // TODO: Should we return noop if getOpenTelemetryKotlin is noop?
```
fractalwrench, line 35: This TODO comment should be addressed, ideally by type-checking `this.getOpenTelemetryKotlin` and returning `OtelJavaOpenTelemetry.noop()` if needed, as this avoids running through all the adapters

embrace-android-otel-java/src/main/kotlin/io/embrace/android/embracesdk/otel/java/EmbraceOtelJavaExtensions.kt
```diff
@@ -0,0 +1,51 @@
...
+/**
+ * Set an attribute on the resource used by the OTel SDK instance with the given key and String value.
+ * The value set will override any value set previously or by the Embrace SDK.
+ * This must be called before the SDK is started in order for it to take effect.
+ */
+@Deprecated(
+    "Use setResourceAttribute(key: String, value: String) instead.",
+    ReplaceWith("setResourceAttribute(key.key, value)")
+)
+fun Embrace.setJavaResourceAttribute(
```
fractalwrench, line 48: This was called `setResourceAttribute` previously - why has the symbol name changed?

priettt: To avoid having the same name for this and the `setResourceAttribute(key: String, value: String)` method. Same with `getJavaOpenTelemetry` and `addJavaSpanExporter`. Would you rather keep the names?

embrace-android-otel-java/src/main/kotlin/io/embrace/android/embracesdk/otel/java/EmbraceOtelJavaExtensions.kt
priettt: We can't typecheck against `NoopOpenTelemetry`, it's an internal object. Should we expose it in otel-kotlin?

embrace-android-otel-java/src/main/kotlin/io/embrace/android/embracesdk/otel/java/EmbraceOtelJavaExtensions.kt
fractalwrench: IMO yes, but that depends on what we call the Kotlin variants of these functions and where they live. Perhaps we can chat about it internally?

PR #2502, opened 2025-09-09
Replace OtelJavaIdGenerator with createCompatSdkFactory().tracingIdFactory and remove libs.opentelemetry.sdk dependency

priettt:


embrace-android-otel/src/main/kotlin/io/embrace/android/embracesdk/internal/otel/sdk/id/OtelIds.kt
```diff
@@ -1,24 +1,25 @@
...
 
-import io.embrace.opentelemetry.kotlin.aliases.OtelJavaIdGenerator
-import io.embrace.opentelemetry.kotlin.aliases.OtelJavaSpanId
+import io.embrace.opentelemetry.kotlin.ExperimentalApi
+import io.embrace.opentelemetry.kotlin.factory.createCompatSdkFactory
 
 object OtelIds {
-
-    private val generator = OtelJavaIdGenerator.random()
+    private val tracingIdFactory = createCompatSdkFactory().tracingIdFactory
```
fractalwrench: `createCompatSdkFactory` shouldn't really be public, I'll address this in a separate PR in the opentelemetry-kotlin repo. Basically we'll want to get an `OpenTelemetry` instance as that implements `SdkFactory`.

priettt: Yeah, this seemed off. Should we somehow expose a tracingId factory in OTel Kotlin? Or should we handle this on the sdk?

fractalwrench: I think we should pass in `openTelemetry: OpenTelemetry` to `AnrOtelMapper` and call `openTelemetry.tracingIdFactory` as that'll achieve what we need for production code.

PR #2516, opened 2025-09-12
Remove ImplicitContextKeyed from EmbSpan and EmbraceSdkSpan

priettt:
## Goal

`ImplicitContextKeyed` forces us to include `opentelemetry-java` as a dependency, and it's not really used anymore.

## Changes

- `FakeEmbraceSdkSpan` contained a Java and a Kotlin span. This didn't make sense, as the actual `EmbraceSdkSpan` only has one kotlin Span implementation.
- `EmbraceSdkSpan` and `EmbSpan` do not extend `ImplicitContextKeyed` anymore. The context operations that are actually used are handled by `opentelemetry-kotlin`. If `opentelemetry-java` is used, `opentelemetry-kotlin` will call the appropiate compat classes.


bidetofevil (approved):
LGTM. `EmbSpan` now implements the Kotlin interface, so not implementing a Java SDK interface like `ImplicitContextKeyed` makes sense. The fact that we can remove it from `EmbraceSdkSpan` and everything still works probably makes we've already made it so that those methods don't have to be called internally in order to make the inter-op work. Like you said, Fran, I think we just moved all that into the compat impl which is in a way already a wrapper, so we don't need to do it in this layer.

PR #2526, opened 2025-09-17
Add % enabled remote config for opentelemetry-kotlin SDK 

priettt:
## Goal

Added `otel_kotlin_sdk.pct_enabled` remote config to roll out the new opentelemetry-kotlin implementation, which is turned off by default.

I divided the PR in multiple commits to make it easier to review.

## Changes:
- Added otel_kotlin_sdk.pct_enabled remote config with DEFAULT_USE_KOTLIN_SDK = false for fallback.
- Moved useKotlinSdk configuration from InitModule to OtelSdkWrapper parameter to prevent early initialization issues.
- Updated OpenTelemetry module to receive configuration through `applyConfiguration` method, as configuration is not ready when the OtelModule is created.

## Testing
Updated existing ExternalTracer and ExternalOtelJavaTracer integration tests to verify the new configuration behavior works correctly.

### DIFF
--- a/embrace-android-otel/src/test/kotlin/io/embrace/android/embracesdk/internal/otel/impl/EmbTracerProviderTest.kt
+++ b/embrace-android-otel/src/test/kotlin/io/embrace/android/embracesdk/internal/otel/impl/EmbTracerProviderTest.kt
@@ -4,6 +4,7 @@ import io.embrace.android.embracesdk.fakes.FakeClock
 import io.embrace.android.embracesdk.fakes.FakeOtelKotlinClock
 import io.embrace.android.embracesdk.fakes.FakeSpanService
 import io.embrace.android.embracesdk.fakes.FakeTracerProvider
+import io.embrace.android.embracesdk.fakes.TestConstants.TESTS_DEFAULT_USE_KOTLIN_SDK
 import io.embrace.opentelemetry.kotlin.ExperimentalApi
 import io.embrace.opentelemetry.kotlin.createNoopOpenTelemetry
 import org.junit.Assert.assertNotSame
@@ -29,7 +30,7 @@ internal class EmbTracerProviderTest {
             impl = otel,
             spanService = spanService,
             clock = openTelemetryClock,
-            useKotlinSdk = true
+            useKotlinSdk = TESTS_DEFAULT_USE_KOTLIN_SDK
         )
     }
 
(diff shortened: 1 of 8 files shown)


embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/config/behavior/OtelBehaviorImpl.kt
```diff
@@ -0,0 +1,18 @@
...
+ * Provides the behavior for OpenTelemetry configuration
+ */
+class OtelBehaviorImpl(
+    override val remote: RemoteConfig?,
+) : OtelBehavior {
+
+    override val local: UnimplementedConfig = null
+
+    override fun shouldUseKotlinSdk(): Boolean {
+        return remote?.killSwitchConfig?.disableOtelKotlinSdk != true
```
fractalwrench: Won't this default the return value to `true` if `remote?.killSwitchConfig?.disableOtelKotlinSdk` is null? IMO we need to disable this behavior by default and have a % rollout due to the risk of the change.

fractalwrench (commented):
I think there are 2 requirements to follow here: firstly, the behavior should be off by default. Secondly the flag should allow a % based rollout. @bidetofevil please chime in if you've already discussed this and reached a different conclusion between you - I think this is the safest way to manage risk

priettt:
@fractalwrench I think Hanson wanted to roll it out by default and give the option to turn it off remotely if needed. Let's discuss it on Slack.

embrace-android-otel/src/main/kotlin/io/embrace/android/embracesdk/internal/otel/OtelSdkInstance.kt
```diff
@@ -15,7 +15,7 @@ import io.embrace.opentelemetry.kotlin.init.TracerProviderConfigDsl
 
 @OptIn(ExperimentalApi::class)
 internal fun createSdkOtelInstance(
-    useKotlinSdk: Boolean = USE_KOTLIN_SDK,
+    useKotlinSdk: Boolean = DEFAULT_USE_KOTLIN_SDK,
```
fractalwrench: IMO we should remove the default parameter anywhere it's used in favour of passing an instance of `OtelBehavior`, as there's a risk the default might get applied...by default

embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/injection/OpenTelemetryModuleImpl.kt
```diff
@@ -31,12 +32,7 @@ internal class OpenTelemetryModuleImpl(
     )
 ) : OpenTelemetryModule {
 
-    init {
-        if (!initModule.useKotlinSdk) {
-            // Enforce the use of default OTel Java SDK ThreadLocal ContextStorage to bypass SPI looking that violates Android strict mode
-            System.setProperty("io.opentelemetry.context.contextStorageProvider", "default")
-        }
-    }
+    private var useKotlinSdk = DEFAULT_USE_KOTLIN_SDK
```
fractalwrench: Can `DEFAULT_USE_KOTLIN_SDK` be confined to a private constant within `OtelBehavior`?

fractalwrench (approved):
LGTM pending review comments

PR #2535, opened 2025-09-19
Symbols of any kind are not uploaded in debug variants anymore

priettt:
## Goal

On debug builds, the gradle plugin should not upload anything.
We were only doing this for RN, and it should be done for NDK, Unity, Proguard files, and even our build telemetry.

This PR updates this behavior, and adds tests to verify it.

### DIFF
--- a/embrace-gradle-plugin-integration-tests/src/test/java/io/embrace/android/gradle/integration/testcases/AndroidNdkTest.kt
+++ b/embrace-gradle-plugin-integration-tests/src/test/java/io/embrace/android/gradle/integration/testcases/AndroidNdkTest.kt
@@ -17,7 +17,8 @@ class AndroidNdkTest {
     @JvmField
     val rule: PluginIntegrationTestRule = PluginIntegrationTestRule()
 
-    private val defaultExpectedVariants = listOf("debug", "release")
+    private val defaultExpectedVariants = listOf("release")
+    private val variantsSentInBuildTelemetry = listOf("debug", "release")
     private val defaultExpectedLibs = listOf("libemb-donuts.so", "libemb-crisps.so")
     private val defaultExpectedArchs = listOf("x86_64", "x86", "armeabi-v7a", "arm64-v8a")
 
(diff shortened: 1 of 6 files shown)


embrace-gradle-plugin-integration-tests/src/test/java/io/embrace/android/gradle/integration/testcases/AndroidNdkTest.kt
```diff
@@ -17,7 +17,8 @@ class AndroidNdkTest {
     @JvmField
     val rule: PluginIntegrationTestRule = PluginIntegrationTestRule()
 
-    private val defaultExpectedVariants = listOf("debug", "release")
+    private val defaultExpectedVariants = listOf("release")
+    private val variantsSentInBuildTelemetry = listOf("debug", "release")
```
fractalwrench, line 21: Do we want to send build telemetry for debug builds? I would argue not

fractalwrench (approved):
LGTM

embrace-gradle-plugin-integration-tests/src/test/java/io/embrace/android/gradle/integration/testcases/AndroidNdkTest.kt
priettt: We just send the available variant names and app IDs. We could do a lot of work to improve our telemetry, but it's probably not a priority for now 🫠

fractalwrench: If we're not uploading mapping files for debug builds then I don't see a reason to capture telemetry, as long as we do capture it for release builds & end-users can opt-out of collection

priettt: we do not send telemetry for debug builds. for not debuggable builds, we send info on what variants are configured in the project. 

if the user runs assembleRelease -> we send info on variants debug and release
if the user runs assembleDebug -> we send nothing

PR #2543, opened 2025-09-25
Bump opentelemetry-kotlin to 0.6.0

fractalwrench:
## Goal

Bumps opentelemetry-kotlin to 0.6.0


priettt (approved):
Looks good 👍 

Should we add default null arguments?

fractalwrench:
The arguments are null by default in the API's interface. As we're overriding the interface a quirk of Kotlin's syntax is that default parameters don't need to be specified in the implementation, but the default value will still be used

PR #2544, opened 2025-09-25
Bump minimum AGP and Gradle versions to 8.0.2

priettt:
## Goal

The required AGP and Gradle versions for Embrace are now 8.0.2.

Also removed Gradle 7 fixture as it won't be used anymore.


embrace-test-common/src/main/kotlin/io/embrace/android/gradle/config/TestMatrix.kt
```diff
@@ -17,33 +17,34 @@ sealed class TestMatrix(
...
      */
-    object NewerVersion : TestMatrix("8.5.2", "8.7", "2.0.0", JdkEnv.JAVA_17)
+    object NewerVersion : TestMatrix("8.5.2", "8.7", "2.0.0", JdkEnv.JAVA_17, "36")
 
     /**
      * The maximum version we currently run tests against. Newer versions may work, but are not
      * explicitly tested.
      */
-    object MaxVersion : TestMatrix("8.9.1", "8.13", "2.2.20", JdkEnv.JAVA_17)
+    object MaxVersion : TestMatrix("8.9.1", "8.13", "2.2.20", JdkEnv.JAVA_17, "36")
```
fractalwrench, line 49: It's probably worth tweaking the test matrix values in general so that there's a better spread of the versions we support - I'd like to see at least one version using Gradle 9, and the latest stable AGP version

fractalwrench (approved):
LGTM assuming you've done some internal communication on this to Ari + others

PR #2546, opened 2025-09-25
Bump min kotlin version supported to 2.0.21

priettt:
## Goal

Bump min supported kotlin version to 2.0.21

### DIFF
--- a/embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/injection/DependencyInjection.kt
+++ b/embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/injection/DependencyInjection.kt
@@ -40,7 +40,7 @@ inline fun <reified T> factory(
     noinline provider: Provider<T>,
 ): ReadOnlyProperty<Any?, T> = FactoryDelegate(provider)
 
-class FactoryDelegate<T>(private inline val provider: Provider<T>) : ReadOnlyProperty<Any?, T> {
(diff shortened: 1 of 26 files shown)


embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/injection/DependencyInjection.kt
```diff
@@ -40,7 +40,7 @@ inline fun <reified T> factory(
     noinline provider: Provider<T>,
 ): ReadOnlyProperty<Any?, T> = FactoryDelegate(provider)
 
-class FactoryDelegate<T>(private inline val provider: Provider<T>) : ReadOnlyProperty<Any?, T> {
+class FactoryDelegate<T>(private val provider: Provider<T>) : ReadOnlyProperty<Any?, T> {
```
fractalwrench, line 43: Why did this change?

priettt: > Inline property cannot have a backing field. This will become an error in Kotlin 2.1.

This is now a warning in Kotlin 2.0. Properties can't be marked as inline. Would you like to take another approach?  

I think this never worked anyway, since the fact that it's a property means that it's stored, so expanding it every time it's used doesn't make sense. That's why I just removed the inline.

bidetofevil (commented):
LFG!

PR #2547, opened 2025-09-25
Bump min java runtime supported version to 11

priettt:
## Goal

Bump source and target compatibility to Java 11


gradle/libs.versions.toml
```diff
@@ -4,7 +4,7 @@
 binaryCompatibilityValidator = "0.18.1"
 agp = "8.13.0"
 junit = "4.13.2"
-kotlinExposed = "1.9.25"
+kotlinExposed = "2.0.21"
 dokka = "1.9.20"
```
bidetofevil, line 8: Can the version dokka users be bumped as well?

PR #2576, opened 2025-10-14
Use coreLibrariesVersion instead of adding the kotlin stdlib dependency

priettt:
## Goal

There's a more "official" way of ensuring that the Kotlin version we expose is compatible with our minimum requirements: https://kotlinlang.org/api/kotlin-gradle-plugin/kotlin-gradle-plugin-api/org.jetbrains.kotlin.gradle.dsl/-kotlin-top-level-extension-config/core-libraries-version.html

> `coreLibrariesVersion` specifies the version of the core Kotlin libraries that are added to the Kotlin compile classpath, unless there is already a dependency added to this project.
> 
> The core Kotlin libraries are:
> 
> 'kotlin-stdlib'
> 'kotlin-test'
> 'kotlin-dom-api-compat'
> 'kotlin-reflect' 

This lets us avoid `Module was compiled with an incompatible version of Kotlin` errors. 

## Testing

Relied on existent testing.

### DIFF
--- a/buildSrc/src/main/kotlin/io/embrace/internal/JvmCompilerConfig.kt
+++ b/buildSrc/src/main/kotlin/io/embrace/internal/JvmCompilerConfig.kt
@@ -26,6 +26,7 @@ fun Project.configureCompilers(module: EmbraceBuildLogicExtension) {
             targetCompatibility = module.jvmTarget.get().toString()
         }
 
+        val coreLibrariesVersion = "2.0.21"
         when (val kotlin = project.extensions.getByName("kotlin")) {
             is KotlinJvmProjectExtension -> {
                 kotlin.compilerOptions {
@@ -34,6 +35,7 @@ fun Project.configureCompilers(module: EmbraceBuildLogicExtension) {
                     jvmTarget.set(target)
                     allWarningsAsErrors.set(true)
                 }
+                kotlin.coreLibrariesVersion = coreLibrariesVersion
             }
 
             is KotlinAndroidExtension -> {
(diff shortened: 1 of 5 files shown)


buildSrc/src/main/kotlin/io/embrace/internal/JvmCompilerConfig.kt
```diff
@@ -34,6 +34,7 @@ fun Project.configureCompilers(module: EmbraceBuildLogicExtension) {
                     jvmTarget.set(target)
                     allWarningsAsErrors.set(true)
                 }
+                kotlin.coreLibrariesVersion = "2.0.21"
```
bidetofevil: Isn't there a gradle properties file we define versions in that we can put this into?

priettt: I'm not sure if I like adding a property just for this. 

It would involve adding an additional line:

`val kotlinCoreLibrariesVersion = project.property("kotlinCoreLibrariesVersion") as String`

or if we use toml, two lines: 

```
val libs = project.extensions.getByType(org.gradle.api.artifacts.VersionCatalogsExtension::class.java).named("libs")
val kotlinExposedVersion = libs.findVersion("kotlinExposed").get().toString()
```

And I think I'd rather just have the version hardcoded here.

fractalwrench: As a minimum let's declare a variable within this function so that it's only one place that needs changing rather than 2 string literals

PR #2597, opened 2025-10-20
Use default parameters on APIs

fractalwrench:
## Goal

Alters our API to use default parameters rather than manually declaring function overloads.

### DIFF
--- a/embrace-android-api/api/embrace-android-api.api
+++ b/embrace-android-api/api/embrace-android-api.api
@@ -271,15 +271,11 @@ public final class io/embrace/android/embracesdk/spans/EmbraceSpan$DefaultImpls 
 
 public final class io/embrace/android/embracesdk/spans/EmbraceSpanEvent {
 	public static final field Companion Lio/embrace/android/embracesdk/spans/EmbraceSpanEvent$Companion;
-	public final fun component1 ()Ljava/lang/String;
-	public final fun component2 ()J
-	public final fun component3 ()Ljava/util/Map;
 	public fun equals (Ljava/lang/Object;)Z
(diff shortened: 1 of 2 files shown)


UPGRADING.md
```diff
@@ -30,11 +36,11 @@ get in touch if you do have a use-case that is no longer met.
...
+| `Embrace.getInstance().start(Context, AppFramework)`                    | `Embrace.start(Context)`                                                                               |
 | `Embrace.getInstance().addLogRecordExporter(LogRecordExporter)`         | Type changed to opentelemetry-kotlin API. Alternative available in `embrace-android-otel-java` module. |
 | `Embrace.getInstance().addSpanExporter(SpanExporter)`                   | Type changed to opentelemetry-kotlin API. Alternative available in `embrace-android-otel-java` module. |
-| `Embrace.getInstance().getOpenTelemetry()`                              | `Embrace.getInstance().getOpenTelemetryKotlin()` or `Embrace.getInstance().getJavaOpenTelemetry()`     |
-| `Embrace.getInstance().setResourceAttribute(AttributeKey, String)`      | `Embrace.getInstance().setResourceAttribute(String, String)`                                           |
+| `Embrace.getInstance().getOpenTelemetry()`                              | `Embrace.getOpenTelemetryKotlin()` or `Embrace.getInstance().getJavaOpenTelemetry()`                   |
```
bidetofevil: Do we still need the `getInstance()` call on before `getJavaOpenTelemetry()`?

fractalwrench: Nice catch - I'll remove the `getInstance()` call

PR #2599, opened 2025-10-20
Modularise schema for instrumentation

fractalwrench:
## Goal

Modularises the schema that our SDK uses to capture instrumentation. This changeset adds the pure JVM `embrace-android-instrumentation-schema` module that contains `EmbType`, `SchemaType`, and friends. I've made a few tweaks to avoid requiring any inter-module dependencies.

I imagine this module will eventually be used as part of an instrumentation API when the features module is broken up.

## Testing


embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/network/logging/NetworkCaptureDataSourceImpl.kt
```diff
@@ -25,7 +25,28 @@ internal class NetworkCaptureDataSourceImpl(
      */
     override fun logNetworkCapturedCall(networkCapturedCall: NetworkCapturedCall) {
         return logWriter.addLog(
-            SchemaType.NetworkCapturedRequest(networkCapturedCall),
+            SchemaType.NetworkCapturedRequest(
```
bidetofevil, line 28: Not blocking, but we may want to add parameter names when there are that many parameters of primitive types

bidetofevil (approved):
LGTM. One non-blocking comment about using parameter names when invoking functions with a tonne of primitives as paramters

PR #2604, opened 2025-10-21
Modularise config

fractalwrench:
## Goal

This changeset moves the `ConfigService` interface and `*Behavior` classes into a separate module named `embrace-android-config` that is JVM-only. Its only dependency is on the `embrace-android-payload` module (required to read the remote config response). I've also extracted the fakes into a separate 'fakes' module that can be used for test, as `embrace-test-fakes` is beginning to have too many interdependencies.

I have _not_ moved the HTTP client that fetches or stores the remote config values as that is still fairly entangled.

## Testing


bidetofevil (approved):
Oh I like this

PR #2607, opened 2025-10-21
Extract HUC instrumentation into new module

bidetofevil:
## Goal

Extract HUC instrumentation to another module and initialize it via reflection, similar to how the Compose Tap instrumentation is done

## Testing

- Best effort unit tests, while integration test will test when the class is not found (but the SDK still starts). More thorough testing will be added in the next PR to disable this by default

### DIFF
--- a/settings.gradle.kts
+++ b/settings.gradle.kts
@@ -15,7 +15,6 @@ include(
     ":embrace-android-payload",
     ":embrace-android-delivery",
     ":embrace-android-okhttp3",
-    ":embrace-android-instrumentation-huc",
     ":embrace-android-fcm",
     ":embrace-android-compose",
     ":embrace-android-otel",


bidetofevil:
* **#2607** <a href="https://app.graphite.dev/github/pr/embrace-io/embrace-android-sdk/2607?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/embrace-io/embrace-android-sdk/2607?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/internal/instrumentation/bytecode/InstrumentationInitializer.kt
```diff
@@ -0,0 +1,11 @@
+package io.embrace.android.embracesdk.internal.instrumentation.bytecode
+
+import androidx.annotation.Keep
+
+@Keep
+internal class InstrumentationInitializer {
```
bidetofevil: The original plan was to put this in a place where I can write an integration test, but there doesn't seem like a module I can drop this into which would allow for it. At this point, does it make sense to just have an `object`?

embrace-android-huc/build.gradle.kts
```diff
@@ -0,0 +1,21 @@
+plugins {
+    id("com.android.library")
+    kotlin("android")
+    id("io.embrace.internal.build-logic")
+}
+
+description = "Embrace Android SDK: HUC"
+
+android {
+    namespace = "io.embrace.android.embracesdk.huc"
```
fractalwrench: nit: I personally prefer `embrace-android-instrumentation-http-url-connection` (or perhaps `embrace-android-instrumentation-huc` for short) to follow the new convention for module names

embrace-android-huc/build.gradle.kts
```diff
@@ -0,0 +1,21 @@
+plugins {
+    id("com.android.library")
+    kotlin("android")
```
fractalwrench: Can you check whether this could use `kotlin("jvm")` instead? I assume it will probably need to be an android module as it's relying on core for now

embrace-android-huc/src/main/java/io/embrace/android/embracesdk/instrumentation/huc/InternalNetworkApi.kt
```diff
@@ -0,0 +1,23 @@
...
+    fun isNetworkSpanForwardingEnabled(): Boolean
+    fun recordNetworkRequest(embraceNetworkRequest: EmbraceNetworkRequest)
+    fun shouldCaptureNetworkBody(url: String, method: String): Boolean
+    fun logInternalError(error: Throwable)
+    fun getSdkCurrentTimeMs(): Long
+    fun isStarted(): Boolean
+    fun generateW3cTraceparent(): String?
+}
+
+object NoopInternalNetworkApi : InternalNetworkApi {
```
fractalwrench: can this be internal too?

embrace-android-huc/src/main/java/io/embrace/android/embracesdk/internal/instrumentation/bytecode/HucBytecodeEntrypoint.kt
```diff
@@ -0,0 +1,22 @@
...
+import io.embrace.android.embracesdk.internal.EmbraceInternalApi
+
+/**
+ * @hide
+ */
+@Keep
+object HucBytecodeEntrypoint {
+    @Keep
+    @JvmStatic
+    fun init() {
```
fractalwrench: Would suggest another name here, as `init()` could be very easily confused with `<init>()`

embrace-gradle-plugin-integration-tests/src/test/resources/bytecode-instrumentation-enabled.json
```diff
@@ -45,6 +45,15 @@
           "embraceHook": "invoke-static {p1}, Lio/embrace/android/embracesdk/internal/instrumentation/bytecode/FcmBytecodeEntrypoint;->onMessageReceived(Lcom/google/firebase/messaging/RemoteMessage;)V"
         }
       ]
+    },
```
fractalwrench: `bytecode-instrumentation-disabled.json` could have an equivalent test fixture for when it's disabled

embrace-gradle-plugin/src/main/java/io/embrace/android/gradle/plugin/api/EmbraceBytecodeInstrumentation.kt
```diff
@@ -52,6 +52,11 @@ abstract class EmbraceBytecodeInstrumentation @Inject internal constructor(objec
     val firebasePushNotificationsEnabled: Property<Boolean> =
         objectFactory.property(Boolean::class.java)
 
+    /**
+     * Whether HUC instrumentation should be enabled. Defaults to true.
+     */
+    val hucEnabled: Property<Boolean> = objectFactory.property(Boolean::class.java).convention(true)
```
fractalwrench: `httpUrlConnectionEnabled` is a more descriptive name for an API IMO

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/internal/instrumentation/bytecode/InstrumentationInitializer.kt
fractalwrench: I'm no longer 100% convinced that bytecode instrumentation is the correct answer here. The main point is how can we write integration tests for any feature that uses bytecode instrumentation to initialize itself, given that the gradle plugin doesn't run in our integration tests?

I think the answer might be either to just retain the pattern that we were previously using (initializing directly within `ModuleInitBootstrapper`). That seems undesirable given that it leads to a really large initialization process, but is easy in the short-term.

Alternatively, we need some sort of plugin/post-SDK-init callback that allows instrumentation to be registered. That's a bit more complicated to design but feels like the longer-term approach

bidetofevil: Fair enough. Let me change this in the short term to use reflection and go from there.

embrace-android-huc/build.gradle.kts
bidetofevil: Yeah, appears that way

embrace-android-huc/build.gradle.kts
bidetofevil: Renamed the module to `embrace-android-instrumentation-huc`. The namespace should also be `io.embrace.android.embracesdk.instrumentation.huc`

embrace-android-instrumentation-huc/src/main/java/io/embrace/android/embracesdk/instrumentation/huc/InternalNetworkApi.kt
```diff
@@ -0,0 +1,23 @@
+package io.embrace.android.embracesdk.instrumentation.huc
+
+import io.embrace.android.embracesdk.network.EmbraceNetworkRequest
+
+internal interface InternalNetworkApi {
```
bidetofevil, line 5: This is a hacky construct and I wish I could get rid of it, but alas, I'll wait until we refactor this into the new instrumentation API before doing that. If I do that.

embrace-android-instrumentation-huc/src/test/kotlin/io/embrace/android/embracesdk/instrumentation/huc/HttpUrlConnectionTrackerTest.kt
```diff
@@ -0,0 +1,57 @@
...
+import org.junit.Assert.assertEquals
+import org.junit.Assert.assertFalse
+import org.junit.Assert.assertNotEquals
+import org.junit.Assert.assertTrue
+import org.junit.Before
+import org.junit.Test
+import org.junit.runner.RunWith
+
+@RunWith(AndroidJUnit4::class)
+internal class HttpUrlConnectionTrackerTest {
```
bidetofevil, line 21: There's only so many things we can test on a regular JVM with no access to the internet, so this is a best-effort that was better than the previous best-effort.

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/EmbraceImpl.kt
```diff
@@ -176,6 +177,38 @@ internal class EmbraceImpl(
         val endTimeMs = clock.now()
         sdkCallChecker.started.set(true)
         end()
+
+        if (configModule.configService.networkBehavior.isHttpUrlConnectionCaptureEnabled()) {
```
bidetofevil, line 181: Integration tests run with this enabled, but it will fail on the class lookup.

I will test this more thoroughly in the PR where this instrumentation is turned off, as I can localize the checks in the singular integration test with this enabled instead of having to change the defaults now and change it back again.

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/EmbraceImpl.kt
```diff
@@ -176,6 +177,38 @@ internal class EmbraceImpl(
...
+                        val trackerObject = objectField.get(null)
+                        val initMethod = trackerClass.getDeclaredMethod(
+                            "registerUrlStreamHandlerFactory",
+                            Boolean::class.java,
+                            SdkStateApi::class.java,
+                            InstrumentationApi::class.java,
+                            NetworkRequestApi::class.java,
+                            EmbraceInternalInterface::class.java,
+                        )
+                        initMethod.invoke(
```
bidetofevil, line 197: The init is also slightly delayed, so if there's a network request that was initiated during SDK startup and after where this was originally initialized, it won't be caught.

But because we're turning this off, I don't think it's a big deal.

settings.gradle.kts
```diff
@@ -9,11 +9,13 @@ include(
     ":embrace-android-utils",
     ":embrace-android-features",
     ":embrace-android-instrumentation-api",
+    ":embrace-android-instrumentation-huc",
     ":embrace-android-instrumentation-taps",
     ":embrace-android-instrumentation-schema",
     ":embrace-android-payload",
     ":embrace-android-delivery",
     ":embrace-android-okhttp3",
+    ":embrace-android-instrumentation-huc",
```
fractalwrench: This is a dupe and can be removed

bidetofevil:
### Merge activity

* **Oct 23, 2:26 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/embrace-io/embrace-android-sdk/2607).
* **Oct 23, 2:27 PM UTC**: [Graphite](https://app.graphite.dev/github/pr/embrace-io/embrace-android-sdk/2607) rebased this pull request as part of a merge.
* **Oct 23, 2:28 PM UTC**: @bidetofevil merged this pull request with [Graphite](https://app.graphite.dev/github/pr/embrace-io/embrace-android-sdk/2607).

PR #2611, opened 2025-10-23
Use service loader for tap instrumentation

fractalwrench:
## Goal

Uses a `ServiceLoader` to load the tap instrumentation. This alters `embrace-android-sdk` so that it has no direct knowledge of the `embrace-android-instrumentation-taps` module, and simply loads its provided service during SDK initialization.

There are a few important pieces:

- `InstrumentationProvider`: interface that returns a `DataSourceState`
- `InstrumentationContext`: arguments that are used when creating `DataSourceState` (e.g. config service)
- `TapsInstrumentationProvider`: implementation that defines how the tap instrumentation data capture should be controlled

And a few other things to note:

- `EmbraceImpl` has been updated to load everything via SPI after the module bootstrapping has finished
- Most out-of-the-box instrumentation should be updated to use this pattern
- I've got around bytecode instrumentation by retaining an internal reference to the `DataSource` within the instrumentation module, when it's created. This might be good enough as a solution, or we might need something with a more explicit API
- I considered an approach with KSP instead of SPI. In that approach we could use codegen instead of service loaders. If we run into perf problems that option is still open, but it felt more complicated
- I included `embrace-android-instrumentation-taps` in `embrace-android-sdk` as an anchor module. This isn't a requirement for future instrumentation.

## Testing

Relied on existing test coverage & also carried out manual testing to confirm things show up as expected in the dash.


embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/EmbraceImpl.kt
```diff
@@ -201,6 +206,23 @@ internal class EmbraceImpl(
         }
     }
 
+    /**
+     * Loads instrumentation via SPI and registers it with the SDK.
+     */
+    private fun registerInstrumentation() {
+        val loader = ServiceLoader.load(InstrumentationProvider::class.java)
```
priettt, line 246: should we add a try catch here?

embrace-android-instrumentation-api/src/main/kotlin/io/embrace/android/embracesdk/internal/arch/InstrumentationInstallArgs.kt
```diff
@@ -0,0 +1,16 @@
...
+import io.embrace.android.embracesdk.internal.clock.Clock
+import io.embrace.android.embracesdk.internal.config.ConfigService
+import io.embrace.android.embracesdk.internal.logging.EmbLogger
+
+/**
+ * Provides references to essential functionality that can be used when registering instrumentation via SPI.
+ */
+interface InstrumentationContext {
+    val configService: ConfigService
+    val sessionSpanWriter: SessionSpanWriter
```
bidetofevil, line 13: I'd like to see a higher level abstraction than this eventually. That it's a span that's being written to is too low-level of a detail for a good API with the right abstractions

embrace-android-instrumentation-api/src/main/kotlin/io/embrace/android/embracesdk/internal/arch/InstrumentationContext.kt
```diff
@@ -0,0 +1,16 @@
...
+
+import io.embrace.android.embracesdk.internal.arch.destination.SessionSpanWriter
+import io.embrace.android.embracesdk.internal.clock.Clock
+import io.embrace.android.embracesdk.internal.config.ConfigService
+import io.embrace.android.embracesdk.internal.logging.EmbLogger
+
+/**
+ * Provides references to essential functionality that can be used when registering instrumentation via SPI.
+ */
+interface InstrumentationContext {
```
bidetofevil: Do we want another `*Context`? Once everything is in place, we should probably rethink our naming

embrace-android-instrumentation-api/src/main/kotlin/io/embrace/android/embracesdk/internal/arch/InstrumentationInstallArgs.kt
```diff
@@ -0,0 +1,16 @@
...
+import io.embrace.android.embracesdk.internal.arch.destination.SessionSpanWriter
+import io.embrace.android.embracesdk.internal.clock.Clock
+import io.embrace.android.embracesdk.internal.config.ConfigService
+import io.embrace.android.embracesdk.internal.logging.EmbLogger
+
+/**
+ * Provides references to essential functionality that can be used when registering instrumentation via SPI.
+ */
+interface InstrumentationContext {
+    val configService: ConfigService
```
bidetofevil, line 12: The `ConfigService` interface can probably be genericized in the future and be detached from Embrace-y

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/EmbraceImpl.kt
bidetofevil: I would be nice to have a detailed accounting of what failed during this loading process and log internal errors when appropriate.

embrace-android-sdk/embrace-proguard.cfg
```diff
@@ -24,6 +24,10 @@
 -keep class io.embrace.android.embracesdk.internal.anr.ndk.NativeThreadSamplerNdkDelegate { *; }
 -keep class io.embrace.android.embracesdk.internal.ndk.jni.JniDelegateImpl { *; }
 
+## Keep internal instrumentation API
```
bidetofevil, line 30: Do we still need to `-keep` the listener class that we look for via reflection, or is that no longer needed?

embrace-android-instrumentation-api/src/main/kotlin/io/embrace/android/embracesdk/internal/arch/InstrumentationContext.kt
fractalwrench: Renamed to `InstrumentationInstallArgs`

embrace-android-sdk/embrace-proguard.cfg
fractalwrench: The compose class is still initialized via reflection. We can change that in a future update & get rid of the proguard entry.

PR #2614, opened 2025-10-23
Support trace instrumentation API

fractalwrench:
## Goal

Moves existing code to different modules to make it possible to capture a trace via the Instrumentation API. This effectively involved wrapping the `SpanService/EmbraceSpan` APIs to avoid depending on the public module.

## Testing

### DIFF
--- a/embrace-android-instrumentation-api/src/main/kotlin/io/embrace/android/embracesdk/internal/arch/destination/SpanToken.kt
+++ b/embrace-android-instrumentation-api/src/main/kotlin/io/embrace/android/embracesdk/internal/arch/destination/SpanToken.kt
@@ -1,0 +1,5 @@
+package io.embrace.android.embracesdk.internal.arch.destination
+
+interface SpanToken {
+    fun stop(endTimeMs: Long? = null)
+}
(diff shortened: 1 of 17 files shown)


embrace-android-instrumentation-api/src/main/kotlin/io/embrace/android/embracesdk/internal/arch/destination/SpanToken.kt
```diff
@@ -0,0 +1,5 @@
+package io.embrace.android.embracesdk.internal.arch.destination
+
+interface SpanToken {
```
bidetofevil, line 3: Is the goal of this interface just to provide an object that can be stopped anywhere without the whole interface being exposed?

bidetofevil (commented):
LGTM. I have a wilder idea which I'll mention in slack, but this is a good start to start putting boundaries around things.

bidetofevil (approved):

embrace-android-instrumentation-api/src/main/kotlin/io/embrace/android/embracesdk/internal/arch/destination/SpanToken.kt
fractalwrench: Exactly that.

PR #2615, opened 2025-10-23
Improve project build scripts

fractalwrench:
## Goal

Improves the build scripts used in the projects. I've made a few changes:

- Removed `afterEvaluate` by replacing a plugin extension with gradle.properties files on individual modules
- Avoided applying plugins where they aren't required
- Only applied publish plugin to modules that require publishing
- Avoided adding dependencies where they weren't required by a module


buildSrc/src/main/kotlin/io/embrace/internal/JvmCompilerConfig.kt
```diff
@@ -14,41 +14,36 @@ fun Project.configureJvmWarningsAsErrors() {
...
-        val target = when (module.jvmTarget.get()) {
-            JavaVersion.VERSION_11 -> JvmTarget.JVM_11
-            else -> error("Unsupported jvm target: ${module.jvmTarget.get()}")
-        }
-        // ensure the Kotlin + Java compilers both use the same language level.
-        project.tasks.withType(JavaCompile::class.java).configureEach {
-            sourceCompatibility = module.jvmTarget.get().toString()
-            targetCompatibility = module.jvmTarget.get().toString()
-        }
+fun Project.configureCompilers() {
```
bidetofevil, line 17: Is there anyway of synchronizing our various version definitions so they live in one place? That is, either have it in code in a single place, or defined somewhere like a properties file so there's one source of truth. As this stuff gets more complex, I worry about modules being out of alignment

Or am I looking at the former already without realizing it? 😅

bidetofevil (commented):
Looks good for me, but I'll let @priettt give the final approval

buildSrc/src/main/kotlin/io/embrace/internal/JvmCompilerConfig.kt
fractalwrench: The versions in this plugin get applied to all modules in the library. I think it's out-of-scope for this PR given that this mainly aims to improve build performance/remove unnecessary dependencies from some modules, but it should be possible to lookup versions exclusively from the version catalog with a bit of investigation.

PR #2620, opened 2025-10-26
Add URLStreamHandler and HttpsURLConnection wrappers for HUC lite instrumentation

bidetofevil:
## Goal

Add wrappers for classes that construct `HttpsUrlConnection`​ into which instrumentation can be added later

### DIFF
--- a/embrace-android-instrumentation-huc-lite/build.gradle.kts
+++ b/embrace-android-instrumentation-huc-lite/build.gradle.kts
@@ -2,6 +2,7 @@ plugins {
     kotlin("android")
     id("com.android.library")
     id("io.embrace.internal.build-logic")
+    id("com.vanniktech.maven.publish")
 }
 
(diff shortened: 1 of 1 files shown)


embrace-android-instrumentation-huc-lite/lint-baseline.xml
```diff
@@ -0,0 +1,4 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<issues format="6" by="lint 8.9.2" type="baseline" client="gradle" dependencies="false" name="AGP (8.9.2)" variant="all" version="8.9.2">
```
fractalwrench, line 2: Does this need a rebase or something? I'm wondering how the lint file was generated with AGP 8.9.2 given we've not used that version for a while?

bidetofevil: It was derived from the OkHttp one. I figure we can regenerate it later if we actually add anything here, which I don't want to lol.

PR #2629, opened 2025-10-27
Split up preference service implementation

fractalwrench:
## Goal

Creates `KeyValueStore`, an interface that abstracts away `PreferencesService` and will ultimately replace it. Currently `PreferencesService` has too many concerns and grows linearly with each bit of instrumentation we add. Replacing it allows individual instrumentation to create their own interfaces that don't know anything about other instrumentation, but have all the values show up in the same place.

## Testing

### DIFF
--- a/embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/envelope/resource/DeviceImpl.kt
+++ b/embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/envelope/resource/DeviceImpl.kt
@@ -20,7 +20,7 @@ internal class DeviceImpl(
     override val systemInfo: SystemInfo,
     private val logger: EmbLogger,
 ) : Device {
-    override var isJailbroken: Boolean? = null
+    override var isJailbroken: Boolean? = false
     override var screenResolution: String = ""
     private val jailbreakLocations: List<String> = listOf(
         "/sbin/",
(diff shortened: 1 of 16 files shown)


embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/prefs/EmbracePreferencesService.kt
```diff
@@ -253,18 +146,16 @@ internal class EmbracePreferencesService(
     }
 
     override fun decreaseNetworkCaptureRuleRemainingCount(id: String, maxCount: Int) {
-        prefs.setIntegerPreference(
-            NETWORK_CAPTURE_RULE_PREFIX_KEY + id,
-            getNetworkCaptureRuleRemainingCount(id, maxCount) - 1
-        )
+        impl.getInt(NETWORK_CAPTURE_RULE_PREFIX_KEY + id)
```
priettt: shouldn't we be writing the value back, minus one?

embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/envelope/resource/DeviceImpl.kt
```diff
@@ -75,18 +75,9 @@ internal class DeviceImpl(
...
         backgroundWorker.submit {
-            val storedIsJailbroken = preferencesService.jailbroken
-            // load value from shared preferences
-            if (storedIsJailbroken != null) {
-                isJailbroken = storedIsJailbroken
-            } else {
-                isJailbroken = checkIfIsJailbroken()
-                preferencesService.jailbroken = isJailbroken
-            }
+            isJailbroken = checkIfIsJailbroken()
```
priettt, line 79: do we want to always recompute, without checking if there's already a value in sharedPreferences?

fractalwrench: IMO yes, because users can enable/disable root masking conditionally, so this value might be different on the same device between different process launches (or even within the same process).

embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/prefs/EmbracePreferencesService.kt
fractalwrench: Yep - good catch!

PR #2630, opened 2025-10-28
Improve instrumentation API

fractalwrench:
## Goal

Attempts to improve the Instrumentation API by merging a few redundant concepts & removing a few unnecessary layers. These changes include:

- `TelemetryDestination` is the single interface that comprises of the functions that were defined separately on `LogWriter`, `SessionSpanWriter`, and `LogWriter`
- `DataSource` no longer requires generic parameters and doesn't expose the implementation detail of whether it records spans/logs/session span events
- `SpanAttributeData` was removed in favour of just supplying key/value parameters as strings
- `CurrentSessionSpan` no longer has functions that add events/attributes; this is done via `TelemetryDestination` instead
- `DataSource` had various alterations including:
  - `inputValidation` parameter is now optional as in practice it isn't used frequently
  - `TelemetryDestination` is an implementation detail that isn't exposed
  - Folks _must_ call `captureTelemetry` to capture telemetry, which runs data capture in a try-catch after checking data limits
  - `countsTowardsLimits` parameter was removed, as this effectively just required a try-catch around some action that didn't actually capture telemetry

Additionally I've updated the existing implementation so that it can be use the altered interface.

## Testing

Updated existing test coverage & wrote some new unit tests.

### DIFF
--- a/embrace-bytecode-instrumentation-tests/src/test/resources/TargetMethodEndVisitorObj_expected.txt
+++ b/embrace-bytecode-instrumentation-tests/src/test/resources/TargetMethodEndVisitorObj_expected.txt
@@ -23,13 +23,13 @@ public final class io/embrace/test/fixtures/TargetMethodEndVisitorObj {
   // access flags 0x11
   public final instrumentAtEndWithReturn()V
    L0
-    LINENUMBER 7 L0
+    LINENUMBER 8 L0
     LDC "Instrumenting at the end of a method with a return statement"
     GETSTATIC java/lang/System.out : Ljava/io/PrintStream;
     SWAP
     INVOKEVIRTUAL java/io/PrintStream.print (Ljava/lang/Object;)V
    L1
-    LINENUMBER 8 L1
+    LINENUMBER 9 L1
     INVOKESTATIC io/embrace/gradle/plugin/instrumentation/InjectedClass.injectedMethod ()V
     RETURN
    L2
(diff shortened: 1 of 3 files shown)


embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/injection/EssentialServiceModuleImpl.kt
```diff
@@ -75,17 +75,20 @@ class EssentialServiceModuleImpl(
...
-                writer = openTelemetryModule.currentSessionSpan
+                destination = telemetryDestination
             )
         }
     }
 
-    override val logWriter: LogWriter by singleton {
-        LogWriterImpl(
+    override val telemetryDestination: TelemetryDestination by singleton {
+        openTelemetryModule.currentSessionSpan
```
priettt: why do we need this? (the call to         openTelemetryModule.currentSessionSpan)

priettt (approved):

embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/injection/EssentialServiceModuleImpl.kt
fractalwrench: We don't, this was left in error. I'll remove it.

PR #2633, opened 2025-10-28
Reduce use of preferences service

fractalwrench:
## Goal

Replaces some uses of `PreferencesService` with `KeyValueStore` where the code should ideally be moved to a separate module.


bidetofevil (approved):
LGTM

PR #2635, opened 2025-10-30
Add instrumentation to HUC wrapper

bidetofevil:
## Goal

Create a more limited version of HUC instrumentation that doesn't support the advanced features of the old one (`embrace-android-instrumentation-huc`​), but is less invasive in how it does the instrumentation because it doesn't need to wrap streams, count bytes, and read and inject  headers. Therefore, it's more appropriate to be used to track requests to third-party APIs that use this network stack as a fallback, so it will be the new default if no configurations are set. As it doesn't support HTTP/2, it's not recommended to be used for first-party APIs because of the inherent perf penalties, though if someone were to use it, they can use the old instrumentation to track everything as before.

This new lite HUC instrumentation will be enabled by default.

### DIFF
--- a/embrace-android-instrumentation-huc-lite/src/main/kotlin/io/embrace/android/embracesdk/instrumentation/huclite/InstrumentedHttpsURLConnection.kt
+++ b/embrace-android-instrumentation-huc-lite/src/main/kotlin/io/embrace/android/embracesdk/instrumentation/huclite/InstrumentedHttpsURLConnection.kt
@@ -36,13 +36,17 @@ internal class InstrumentedHttpsURLConnection(
     private val cachedResponseStatusCode: AtomicInteger = AtomicInteger(0)
     private val requestData: RequestData? =
         if (sdkStateApi.isStarted) {
-            RequestData(
-                connection = wrappedConnection,
-                now = instrumentationApi::getSdkCurrentTimeMs,
-                trackRequest = sdkStateApi::isStarted,
-                networkRequestApi = networkRequestApi,
-                errorHandler = internalInterface::logInternalError
-            )
+            runCatching {
+                RequestData(
+                    connection = wrappedConnection,
+                    now = instrumentationApi::getSdkCurrentTimeMs,
+                    trackRequest = sdkStateApi::isStarted,
+                    networkRequestApi = networkRequestApi,
+                    errorHandler = internalInterface::logInternalError
+                )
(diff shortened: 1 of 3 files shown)


embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/EmbraceImpl.kt
fractalwrench, line 261: It would be preferable to initialize the instrumentation by depending on the `embrace-android-instrumentation-api` module and using SPI.

embrace-android-instrumentation-huc-lite/src/main/kotlin/io/embrace/android/embracesdk/instrumentation/huclite/InstrumentedHttpsURLConnection.kt
```diff
@@ -17,9 +26,162 @@ import javax.net.ssl.SSLSocketFactory
...
+        try {
+            wrappedConnection.connect()
+        } catch (t: Throwable) {
+            requestData?.clientError(t)
+            throw t
+        }
+    }
+
+    override fun getOutputStream(): OutputStream {
+        requestData?.startRequest()
```
fractalwrench, line 67: These should be in a try-catch unless there's a good reason?

bidetofevil: Do you mean for `requestData.startRequest()`? You mean so we don't throw? Yeah, I can add that within the call.

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/EmbraceImpl.kt
bidetofevil: Yup, going to add that in the next PR

bidetofevil:
### Merge activity

* **Oct 30, 7:30 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/embrace-io/embrace-android-sdk/2635).
* **Oct 30, 7:30 PM UTC**: @bidetofevil merged this pull request with [Graphite](https://app.graphite.dev/github/pr/embrace-io/embrace-android-sdk/2635).

PR #2637, opened 2025-10-30
Modularise compose tap instrumentation

fractalwrench:
## Goal

Creates a separate module for the Jetpack Compose tap instrumentation.

## Testing

Manually confirmed that Compose taps are still captured


UPGRADING.md
```diff
@@ -27,6 +27,13 @@ such as adding exporters.
 `Embrace.getInstance()` is deprecated in favour of `Embrace`. For example, you can now call `Embrace.start(Context)` instead
 of `Embrace.getInstance().start(Context)`.
 
+## Internal modularisation
+
+The Embrace SDK is going through a process of modularisation. If you rely on `embrace-android-sdk` then you should notice
+very little change as most alterations are internal. However, you should be aware that:
+
+- `embrace-android-compose` has been renamed as `embrace-android-instrumentation-compose`
```
bidetofevil: This should be `compose-tap` or something less generic?

gradle/libs.versions.toml
```diff
@@ -17,7 +17,7 @@ protobuf = "4.33.0"
 profileinstaller = "1.3.1" # version pinned to 1.3.1 because of AGP bug
 okhttp = "4.12.0"
 firebase = "25.0.1"
-compose = "1.0.5"
+compose = "1.9.3"
```
bidetofevil: I remember there was a reason we couldn't upgrade. Maybe it was Kotlin related and no longer applies. Do you remember,  @priettt ?

bidetofevil (approved):

UPGRADING.md
fractalwrench: Renamed to `compose-tap`

gradle/libs.versions.toml
fractalwrench: I've left this as-is for the purposes of this PR. When I tested this out in the example app it worked ok, and it's a `compileOnly` dependency, but I'd be intrigued to know what the original blocker for upgrading was

PR #2641, opened 2025-10-30
Add Network request logging as a telemetry destination

bidetofevil:
## Goal

Add the ability to log network requests into the `TelemetryDestination`​ API. Note that the network capture isn't part of the API that was moved over as it's not needed for the HUC Lite instrumentation.


bidetofevil:
Going to do it different. Will abandon

PR #2644, opened 2025-10-31
Use precompiled build script

fractalwrench:
## Goal

Refactors the build logic to use precompiled build script plugins. This gets IDE syntax highlighting and should also be faster to compile the build scripts themselves as Gradle can cache them.


embrace-android-api/build.gradle.kts
```diff
@@ -1,8 +1,5 @@
 plugins {
-    id("com.android.library")
-    kotlin("android")
-    id("io.embrace.internal.build-logic")
-    id("com.vanniktech.maven.publish")
+    id("embrace-prod-android-conventions")
```
priettt: should we also use the embrace-public-api-conventions here?

fractalwrench: Yep good catch.

PR #2647, opened 2025-10-31
Network status instrumentation modularisation

fractalwrench:
## Goal

Creates a separate module for the network status change instrumentation.

## Testing


bidetofevil (approved):
LGTM

PR #2648, opened 2025-10-31
Modularise breadcrumb features

fractalwrench:
## Goal

Moves `BreadcrumbDataSource`, `RnActionDataSource`, and `InternalErrorDataSource` into the core module as they are all called by API invocations and therefore must be present. In future we _might_ move `RnActionDataSource` to some internal API module, or obviate it entirely by having RN pass the correct schema type down.


bidetofevil (approved):
LGTM

PR #2651, opened 2025-10-31
Add isolated projects telemetry

priettt:
## Goal
Track whether Gradle's isolated projects feature is enabled in build telemetry to understand customer adoption.

## Changes
- Add `isIsolatedProjectsEnabled()` method to `BuildTelemetryCollector` that checks both system property (`-Dorg.gradle.unsafe.isolated-projects`) and gradle property (`gradle.properties`) for the isolated projects setting


embrace-gradle-plugin/src/main/java/io/embrace/android/gradle/plugin/buildreporter/BuildTelemetryCollector.kt
```diff
@@ -93,6 +94,15 @@ class BuildTelemetryCollector {
     private fun Project.isParallelExecutionEnabled() =
         this.gradle.startParameter.isParallelProjectExecutionEnabled
 
+    private fun Project.isIsolatedProjectsEnabled(): Boolean {
+        return try {
+            System.getProperty("org.gradle.unsafe.isolated-projects") == "true"
```
fractalwrench: `systemWrapper.getProperty` can be used here

PR #2653, opened 2025-10-31
Refactor HUC Lite instrumentation to use TelemetryDestination

bidetofevil:
## Goal

Convert HUC Lite instrumentation to a `DataSource`​ and initialize via SPI. Some clean up and optimization still need to be done, but this should be functional, so I want to get it reviewed first.  
  
Note that we still need to enforce limits - this has to be shared with all the other network requests, so it will likely have to be plumbed through the args that come through during registration. But I thought I'd get this in review


embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/arch/destination/TelemetryDestinationImpl.kt
```diff
@@ -129,6 +129,7 @@ internal class TelemetryDestinationImpl(
             type = type,
             attributes = attributes,
         )
+        processStateService.sessionUpdated()
```
fractalwrench, line 134: Ideally every function in `TelemetryDestination` should call this. Perhaps that's a good candidate to break off into a separate PR?

embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/session/lifecycle/EmbraceProcessStateService.kt
```diff
@@ -125,6 +125,12 @@ internal class EmbraceProcessStateService(
         return sessionOrchestrator != null
     }
 
+    override fun sessionUpdated() {
```
fractalwrench, line 128: Why route this through `ProcessStateService`? `SessionOrchestrator` already has a functionally equivalent check on whether the app is in the background. IMO we can just use that (and possibly rename it from `reportBackgroundActivityStateChange` to something like `onDataCaptured`)

embrace-android-instrumentation-huc-lite/src/main/kotlin/io/embrace/android/embracesdk/instrumentation/huclite/InstrumentedUrlStreamHandlerFactory.kt
```diff
@@ -12,19 +9,19 @@ import java.net.URLStreamHandlerFactory
...
 internal class InstrumentedUrlStreamHandlerFactory(
     private val httpsHandler: URLStreamHandler,
-    private val sdkStateApi: SdkStateApi,
-    private val instrumentationApi: InstrumentationApi,
-    private val networkRequestApi: NetworkRequestApi,
-    private val internalInterface: EmbraceInternalInterface,
+    private val sdkStarted: () -> Boolean,
+    private val currentTimeMs: () -> Long,
+    private val hucLiteDataSource: HucLiteDataSource,
+    private val errorHandler: (Throwable) -> Unit,
```
fractalwrench: Could we supply Clock and possibly EmbLogger here like the other instrumentation does? I think a few other classes could be updated in the same way.

embrace-android-instrumentation-huc-lite/src/main/kotlin/io/embrace/android/embracesdk/instrumentation/huclite/InstrumentedHttpsURLConnection.kt
```diff
@@ -26,27 +20,22 @@ import javax.net.ssl.SSLSocketFactory
...
-                    connection = wrappedConnection,
-                    now = instrumentationApi::getSdkCurrentTimeMs,
-                    trackRequest = sdkStateApi::isStarted,
-                    networkRequestApi = networkRequestApi,
-                    errorHandler = internalInterface::logInternalError
-                )
-            }.onFailure {
-                internalInterface.logInternalError(InstrumentationException(it))
-            }.getOrNull()
+        if (sdkStarted()) {
```
fractalwrench: Fine to keep this for this PR, but if we're using a `DataSource` hooked up via SPI then by definition the SDK will have been initialized when this executes. So ultimately we should be able to get rid of this check

embrace-android-instrumentation-huc-lite/build.gradle.kts
```diff
@@ -12,12 +12,11 @@ android {
 }
 
 dependencies {
-    compileOnly(project(":embrace-android-core"))
-    compileOnly(project(":embrace-android-sdk"))
-    compileOnly(project(":embrace-internal-api"))
-    testImplementation(project(":embrace-android-core"))
-    testImplementation(project(":embrace-android-sdk"))
-    testImplementation(project(":embrace-internal-api"))
```
fractalwrench, line 17: 🔥

embrace-android-instrumentation-huc-lite/src/main/kotlin/io/embrace/android/embracesdk/internal/instrumentation/HucLiteDataSource.kt
```diff
@@ -0,0 +1,193 @@
...
+
+    @Volatile
+    var enabled = false
+
+    override fun onDataCaptureEnabled() {
+        InstrumentationInitializer().installURLStreamHandlerFactory(
+            sdkStarted = { enabled },
+            currentTimeMs = args.clock::now,
+            hucLiteDataSource = this,
+            errorHandler = fun(t: Throwable) {
```
fractalwrench: Out of curiosity, why `fun(t: Throwable)` rather than `{ t ->`?

embrace-android-instrumentation-huc-lite/src/main/kotlin/io/embrace/android/embracesdk/internal/instrumentation/HucLiteDataSource.kt
```diff
@@ -0,0 +1,193 @@
...
+                args.logger.trackInternalError(InternalErrorType.DATA_SOURCE_DATA_CAPTURE_FAIL, t)
```
fractalwrench: If you're running code within `captureTelemetry`, you can just throw an exception and it will automatically get logged with this reason ✨ 

Hopefully this should allow quite a few try-catches to be removed

embrace-android-instrumentation-huc-lite/src/main/kotlin/io/embrace/android/embracesdk/internal/instrumentation/HucLiteDataSource.kt
```diff
@@ -0,0 +1,193 @@
...
+class HucLiteDataSource(
+    private val args: InstrumentationInstallArgs,
+) : DataSourceImpl(
+    args = args,
+    limitStrategy = NoopLimitStrategy
+) {
+    val telemetryDestination = args.destination
+
+    @Volatile
+    var enabled = false
```
fractalwrench: This is discouraged in the contract for `DataSource` as `onDataCaptureEnabled` can be called multiple times. In practice this should just get called once as long as data capture is enabled for both session + background activities.

embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/arch/destination/TelemetryDestinationImpl.kt
bidetofevil: Yeah that makes sense. I was trying to minimize the functional impact of this but might as well do it right in another PR.

embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/session/lifecycle/EmbraceProcessStateService.kt
bidetofevil: Yeah this was to minimize change, but that would definitely be better overall and make sense if I broke it out into another PR.

embrace-android-instrumentation-huc-lite/src/main/kotlin/io/embrace/android/embracesdk/instrumentation/huclite/InstrumentedHttpsURLConnection.kt
bidetofevil: Yeah, I was trying to think of an edge case where the DS would be initialized but then we stopped wanting to do this. But threading this through makes things more complicated, so I'll see if I truly need it at this layer

embrace-android-instrumentation-huc-lite/src/main/kotlin/io/embrace/android/embracesdk/instrumentation/huclite/InstrumentedUrlStreamHandlerFactory.kt
bidetofevil: I think it had something to do with the interfaces not being available in a previous iteration/location of the instrumentation. I would definitely prefer that - and ideally have it provided by the data source instead of factory install time, so let me see if that's possible.

embrace-android-instrumentation-huc-lite/src/main/kotlin/io/embrace/android/embracesdk/internal/instrumentation/HucLiteDataSource.kt
bidetofevil: I wouldn't need this if I got rid of the `sdkStarted()` method provision. Fingers crossed I can do that.

embrace-android-instrumentation-huc-lite/src/main/kotlin/io/embrace/android/embracesdk/internal/instrumentation/HucLiteDataSource.kt
bidetofevil: Oh probably because my grasp of Kotlin syntax can still be improved 😅

embrace-android-instrumentation-huc-lite/src/main/kotlin/io/embrace/android/embracesdk/internal/instrumentation/HucLiteDataSource.kt
bidetofevil: ooh right!

bidetofevil: I'm not going through `captureTelemetry` because I need different arguments for errors and completed requests, and I wanted to consolidate that logic within `RequestData`, but I was still able to reduce the use of this

fractalwrench (approved):
LGTM

PR #2660, opened 2025-11-03
Remove unused code from AnrService

fractalwrench:
## Goal

Removes some code from the `AnrService` that doesn't fulfil any useful function.

### DIFF
--- a/embrace-android-features/src/main/kotlin/io/embrace/android/embracesdk/internal/instrumentation/anr/EmbraceAnrService.kt
+++ b/embrace-android-features/src/main/kotlin/io/embrace/android/embracesdk/internal/instrumentation/anr/EmbraceAnrService.kt
@@ -78,15 +78,6 @@ internal class EmbraceAnrService(
         }
     }
 
-    override fun handleCrash(crashId: String) {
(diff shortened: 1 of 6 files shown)


embrace-android-features/src/main/kotlin/io/embrace/android/embracesdk/internal/instrumentation/anr/EmbraceAnrService.kt
```diff
@@ -78,15 +78,6 @@ internal class EmbraceAnrService(
         }
     }
 
-    override fun handleCrash(crashId: String) {
-        this.anrMonitorWorker.submit {
-            livenessCheckScheduler.stopMonitoringThread()
```
bidetofevil, line 83: Is this handled somewhere along the crash handling process? I believe this was needed so there isn't a race condition during crash handling when things are being shut down and we still try to track ANRs.

fractalwrench: I wasn't aware of that condition - I'll close this PR

PR #2664, opened 2025-11-03
Improve crash data source

fractalwrench:
## Goal

Makes miscellaneous tweaks to `CrashDataSource` to allow moving it to its own instrumentation module.

## Testing


bidetofevil (commented):
LGTM, assuming there's another way to handle the stopping of the ANR monitoring thread

PR #2673, opened 2025-11-03
Expose session update API through TelemetryDestintationImpl

bidetofevil:
## Goal

Allow `TelemetryDestination`​ to update the SDK via `ProcessStateService`​ when session data has been updated. With this, any explicit updates in workflows that use `TelemetryDestination`​ can be removed because it will be done implicitly.

### DIFF
--- a/embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/session/lifecycle/EmbraceProcessStateService.kt
+++ b/embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/session/lifecycle/EmbraceProcessStateService.kt
@@ -126,9 +126,7 @@ internal class EmbraceProcessStateService(
     }
 
     override fun sessionUpdated() {
-        if (isInBackground) {
-            sessionOrchestrator?.onSessionDataUpdate()
-        }
+        sessionOrchestrator?.onSessionDataUpdate()
     }
 
     companion object {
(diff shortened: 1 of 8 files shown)


embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/session/lifecycle/ProcessStateService.kt
```diff
@@ -46,4 +46,9 @@ interface ProcessStateService : LifecycleEventObserver, Closeable {
      * Returns true if this is fully initialized
      */
     fun isInitialized(): Boolean
+
+    /**
+     * Notify the SDK that the current session has updated telemetry or metadata
+     */
+    fun sessionUpdated()
```
fractalwrench, line 53: Same comment from previous PR - I don't think this really needs to be on `ProcessStateService`. `SessionOrchestrator` and friends feel like a better fit.

bidetofevil: This means exposing SessionOrchestrator to `TelemetryDestinationImpl` directly, and the latter is current a dependency of `SessionOrchestratorImpl`, which is added to `EmbraceProcessService` as a listener. I'm not really sure about that.

Any other ideas?

bidetofevil: I've updated the implementation so that this is essentially a pass-through and the filtering out of non-background updates is done in the `SessionOrchestratorImpl`, but this still needs to threaded through somehow.

fractalwrench: Hmm. Let's just go with this for now & we can unpick it later.

PR #2680, opened 2025-11-04
Bump test matrix for gradle tests

fractalwrench:
## Goal

Bumps the test matrix for our gradle plugin tests as AGP/Gradle versions have moved on quite a bit since we last did it. The RN fixture didn't pass on Gradle 9.2 but I believe this is down to the fixture itself being old.


bidetofevil (approved):
Good times

PR #2681, opened 2025-11-04
Split storage of ordinals from preferences service

fractalwrench:
## Goal

Splits the storage of ordinals such as crash/session number from the `PreferencesService` into a separate interface, `OrdinalStore`.

## Testing

Updated unit tests.


embrace-android-infra/src/main/kotlin/io/embrace/android/embracesdk/internal/store/OrdinalStore.kt
```diff
@@ -0,0 +1,32 @@
+package io.embrace.android.embracesdk.internal.store
+
+interface OrdinalStore {
+
+    /**
+     * Increments and returns the crash number ordinal. This is an integer that
+     * increments on every crash. It allows us to check the % of crashes that
+     * didn't get delivered to the backend.
+     */
+    fun incrementAndGetCrashNumber(): Int
```
bidetofevil: Any thought about simplifying this by exposing an enum directly and reducing this down to one method? 4 is right in the grey area of whether we should have dedicated methods.

bidetofevil (approved):
LGTM. One comment about whether we want to use/expose an enum for the thing being tracked but fine either way.

embrace-android-infra/src/main/kotlin/io/embrace/android/embracesdk/internal/store/OrdinalStore.kt
fractalwrench: I'll convert this PR to draft as I like this idea better than the current implementation.

fractalwrench:
Numbers will remain the same. We are still writing to the same location, we're just changing the abstraction

PR #2687, opened 2025-11-04
Modularise OkHttp instrumentation

fractalwrench:
## Goal

Modularises the OkHttp instrumentation. `OkHttpDataSource` is called by an OkHttp interceptor, performs OkHttp-specific logic, then calls into `NetworkRequestDataSource` and `NetworkCaptureDataSource` with the captured request.

## Testing


embrace-android-instrumentation-okhttp/src/main/java/io/embrace/android/embracesdk/internal/instrumentation/okhttp/OkHttpDataSource.kt
```diff
@@ -25,39 +27,46 @@ import kotlin.math.abs
  * Captures OkHttp requests as telemetry.
  */
 internal class OkHttpDataSource(
-    private val embrace: Embrace = Embrace,
-    private val embraceInternalApi: EmbraceInternalApi = EmbraceInternalApi,
+    args: InstrumentationArgs,
+    private val networkRequestDataSourceProvider: () -> NetworkRequestDataSource?,
```
bidetofevil, line 31: Interesting. How much are we actually leveraging the "DataSource-ness" of `NetworkRequestDataSource` and `NetworkCaptureDataSource`? Is it more the lifecycle stuff? Feels like these aren't quite like the Provider-loaded DataSources that I think of when I think of what our DataSources are. Does this feel like a slightly different construct at this point?

fractalwrench: IMO there's a reasonable amount of `DataSource` that those two use. I think `OkHttpDataSource` might need a different construct longer term.

bidetofevil: Yeah, the provider-loaded stuff may actually be less data-source-y. It's more to do with initialization of the singleton more than anything. Some of the lifecycle stuff it depends on overlap with the the actual data capturing stuff, but it's more like an instrumentation bootstrapping piece.

OK cool. Good to think about but this ultimately is totally fine. We can dive in later if we feel a need to.

PR #2689, opened 2025-11-04
Use ErrorCodeAttribute in TelemetryDestination

bidetofevil:
## Goal

Use existing internal sealed class ErrorCodeAttribute to represent the public API concept `ErrorCode`​ in `TelemetryDestination`


fractalwrench (approved):
LGTM

PR #2695, opened 2025-11-05
[DNM] POC for state

bidetofevil:
## Goal

POC for network connection status as a State

### DIFF
--- a/embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/arch/destination/TelemetryDestinationImpl.kt
+++ b/embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/arch/destination/TelemetryDestinationImpl.kt
@@ -222,6 +222,14 @@ class TelemetryDestinationImpl(
         override fun getStartTimeMs(): Long? = span.getStartTimeMs()
 
         override fun setSystemAttribute(key: String, value: String) = span.setSystemAttribute(key, value)
+
+        override fun addEvent(name: String, eventTimeMs: Long, attributes: Map<String, String>) {
+            span.addEvent(
+                name = name,
+                timestampMs = eventTimeMs,
+                attributes = attributes
+            )
+        }
     }
 }
 
--- a/embrace-android-instrumentation-network-status/src/main/kotlin/io/embrace/android/embracesdk/internal/instrumentation/network/NetworkStatusDataSource.kt
+++ b/embrace-android-instrumentation-network-status/src/main/kotlin/io/embrace/android/embracesdk/internal/instrumentation/network/NetworkStatusDataSource.kt
@@ -19,6 +19,9 @@ class NetworkStatusDataSource(
     }
 
     private var span: SpanToken? = null
+    private var state: SpanToken? = null
+    private var currentState: NetworkStatus = NetworkStatus.UNKNOWN
+    private var transitionCount = 0
 
     override fun onNetworkConnectivityStatusChanged(status: NetworkStatus) {
         // close previous span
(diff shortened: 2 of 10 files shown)


embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/arch/destination/TelemetryDestinationImpl.kt
```diff
@@ -184,5 +184,13 @@ internal class TelemetryDestinationImpl(
             span.stop(endTimeMs = endTimeMs)
             processStateService.sessionUpdated()
         }
+
+        override fun addEvent(name: String, eventTimeMs: Long, attributes: Map<String, String>) {
```
bidetofevil, line 226: The real implementation would add a `changeState` method to `TelemetryDestination` and abstract the details away, but adding this and using what's there is just convenient for the POC

embrace-android-instrumentation-network-status/src/main/kotlin/io/embrace/android/embracesdk/internal/instrumentation/network/NetworkStatusDataSource.kt
```diff
@@ -27,5 +30,31 @@ class NetworkStatusDataSource(
         captureTelemetry {
             span = startSpanCapture(SchemaType.NetworkStatus(status.value), timestamp)
         }
+        stateChange(status, timestamp)
```
bidetofevil, line 34: The real implementation would create a new `DataSource` to capture each new state, probably with some shared infra that does the basics.

embrace-android-instrumentation-schema/src/main/kotlin/io/embrace/android/embracesdk/internal/arch/schema/EmbType.kt
```diff
@@ -28,6 +28,10 @@ sealed class EmbType(type: String, subtype: String?) : EmbraceAttribute {
         object ThermalState : Performance("thermal_state")
 
         object UiLoad : Performance("ui_load")
+
+        object State : System("state") {
```
bidetofevil, line 32: I thought I had to put the type to be `Performance` to have it show up in the dashboard, but apparently I didn't do that properly and it still showed up lol. The proper `emb.type` would be `sys.state`.

embrace-android-instrumentation-schema/src/main/kotlin/io/embrace/android/embracesdk/internal/arch/schema/SchemaType.kt
```diff
@@ -282,6 +282,16 @@ sealed class SchemaType(
             ExceptionAttributes.EXCEPTION_MESSAGE to (throwable.message ?: "")
         )
     }
+
+    class NetworkState(initialState: String) : SchemaType(
```
bidetofevil, line 286: Each State should have a new `SchemaType` while sharing a `TelemetryType`. Initial state should be derived from the telemetry type in the real impl,

embrace-android-instrumentation-schema/src/main/kotlin/io/embrace/android/embracesdk/internal/arch/schema/SchemaType.kt
```diff
@@ -282,6 +282,16 @@ sealed class SchemaType(
...
+        telemetryType = EmbType.Performance.State,
+        fixedObjectName = "state-network-connectivity"
+    ) {
+        override val schemaAttributes: Map<String, String> = mapOf(
+            embInitialState.name to initialState,
+            "initial_state" to initialState
```
bidetofevil, line 292: This non-emb-prefixed attribute is just so it shows up on the dashboard - `emb.` ones, we hide by default, which we may want to change if we want to expose `emb-` attributes in the future.

bidetofevil:
POC. Don't need it.

PR #2698, opened 2025-11-05
Sort JVM mapping requests by app ID to fix flaky test

priettt:
## Goal
Fix the flaky `AndroidBuildTypeProductFlavorTest > bundle` test that was failing intermittently in CI due to non-deterministic request ordering when Gradle builds variants in parallel.

## Changes
- Modified `verifyJvmMappingRequestsSent()` in `AssertionInterface.kt` to sort both actual requests and expected values by app ID before comparison
- Added size assertion when build IDs are provided to ensure app IDs and build IDs lists match
- Updated function documentation to reflect the sorting behavior


bidetofevil (commented):
LGTM

fractalwrench (approved):
LGTM

PR #2703, opened 2025-11-06
Deprecate autoAddDependencies

fractalwrench:
## Goal

Deprecates `embrace.autoAddDependencies` as it only adds the `embrace-android-sdk` and `embrace-android-instrumentation-okhttp` modules. In a future release (likely v9) we can remove this functionality entirely.


fractalwrench:
Personally I'd like to avoid adding anything else to this part of the gradle plugin as it's low value & has historically caused issues with how it adds dependencies

PR #2706, opened 2025-11-06
Refactor okhttp interceptors

fractalwrench:
## Goal

Refactors the OkHttp interceptors. This extracts most of the code out of the actual interceptors and moves it into a new class named `OkHttpDataSource`, which will eventually be where most of the logic around data capture happens.

`EmbraceOkHttpInterceptor` is retained as a simple interceptor. This is kept so we can always attempt to proceed the chain if telemetry capture isn't initialised yet (its function will become more apparent in future PRs).


embrace-android-instrumentation-okhttp/src/main/java/io/embrace/android/embracesdk/internal/instrumentation/okhttp/OkHttpDataSource.kt
```diff
@@ -19,33 +19,76 @@ import okio.Buffer
...
+        InterceptorType.NETWORK -> interceptNetworkRequest(chain)
+    }
+
+    private fun interceptApplicationRequest(chain: Interceptor.Chain): Response {
+        val startTime = embrace.getSdkCurrentTimeMs()
+        val request: Request = chain.request()
+        return try {
+            // we are not interested in response, just proceed
+            chain.proceed(request)
+        } catch (e: EmbraceCustomPathException) {
```
bidetofevil, line 46: Funny enough, I can't find where this is thrown anymore, so we might not even need this block. Not a big problem to keep it in for now though.

fractalwrench: It's used for GraphQL requests so we should keep it until we can write more specific instrumentation for that: https://embrace.io/docs/best-practices/graphql/?_highlight=embracecustompath#override-paths-of-requests

PR #2707, opened 2025-11-06
Refactor network capture to use KeyValueStore

fractalwrench:
## Goal

Refactors network capture to use `KeyValueStore` instead of `PreferencesService`.

## Testing


bidetofevil (approved):
LGTM

PR #2708, opened 2025-11-06
Add HttpNetworkRequest model class

fractalwrench:
## Goal

Adds `HttpNetworkRequest` as a class used to hold network request data internally. This avoids using the public API at the telemetry capture layer which is an obstacle to removing the use of `Embrace/EmbraceInternalApi` in our network request instrumentation.

As part of this change I also separated `NetworkLoggingService` and `NetworkCaptureService` so that they are called individually from each other & therefore have no knowledge of each other.

## Testing


bidetofevil (approved):
LGTM. I think we can leverage this piece to record the source of the network request/instrumentation (e.g. okhttp vs huc vs huc-lite), but that's for later

PR #2710, opened 2025-11-06
Combine network capture service and data source

fractalwrench:
## Goal

Combine the network capture service and data source, which has allowed a lot of the internal implementation details to be hidden from the interface and in future will allow further modularisation.

## Testing

Updates existing test coverage.


bidetofevil (approved):
LGTM. This kind of DataSource is almost like an extension of `TelemetryDestination` to record something of a similar shaped that goes on top of the `DataSource` interface/impl. I can see State being implemented similarly, which will keep `TelemetryDestination` agnostic to data types.

I'd be curious how you plan to use this in the actual OkHttp module This kind of DataSource template probably belongs in the instrumentation-api layer for now, but I wonder how many of these value-add we plan to add, and whether that deserves another layer...

Anyway, a problem for later. Looks great!

fractalwrench:
My current thoughts were to have one main place that routes all the network requests, and then have okhttp/huc/manual sources have the minimal amount of logic to record something. Effectively that'd mean the code would mainly be interacting with specific concepts like `HttpUrlConnection` and interceptors.

There's definitely a bit of experimentation to be done to strike the right balance IMO. It's a bit of an unusual case

PR #2730, opened 2025-11-10
Complete modularisation of JVM crash instrumentation

fractalwrench:
## Goal

Completes the modularisation of our JVM crash instrumentation. This creates a new module named `embrace-android-instrumentation-crash-jvm` that contains `JvmCrashDataSource`.

Other modules look up this type to register crash handlers for use in teardown, and to augment JS exception telemetry.

## Testing

Updated unit tests.


embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/internal/api/delegate/ReactNativeInternalInterfaceImpl.kt
```diff
@@ -27,7 +32,19 @@ internal class ReactNativeInternalInterfaceImpl(
         stacktrace: String?,
     ) {
         if (embrace.isStarted) {
-            jsCrashService?.logUnhandledJsException(name, message, type, stacktrace)
+            val registry = bootstrapper.instrumentationModule.instrumentationRegistry
```
bidetofevil, line 35: Not sure how I feel about random thinks having access to data source. But I guess all these hosted SDK APIs are special....

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/internal/injection/InternalInterfaceModuleImpl.kt
```diff
@@ -19,7 +18,7 @@ internal class InternalInterfaceModuleImpl(
     payloadSourceModule: PayloadSourceModule,
     instrumentationModule: InstrumentationModule,
     embrace: EmbraceImpl,
-    crashModule: CrashModule,
+    bootstrapper: ModuleInitBootstrapper,
```
bidetofevil, line 21: Is it possible to export just the interface for the registry for lookup rather than the whole thing?

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/internal/injection/ModuleInitBootstrapper.kt
```diff
@@ -442,6 +420,27 @@ internal class ModuleInitBootstrapper(
         }
     }
 
+    fun loadInstrumentation() {
+        val registry = instrumentationModule.instrumentationRegistry
+        val instrumentationProviders = ServiceLoader.load(InstrumentationProvider::class.java)
+        registry.loadInstrumentations(instrumentationProviders, instrumentationModule.instrumentationArgs)
+    }
+
+    fun postLoadInstrumentation() {
```
bidetofevil, line 429: Installing listeners/handlers to data sources feels like a proper abstraction eh?

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/internal/api/delegate/ReactNativeInternalInterfaceImpl.kt
fractalwrench: I agree this isn't ideal & is something I'd like to avoid in future. For our current API, I believe this is the best way to achieve our goal. Longer-term I think we should split out these APIs to the specific instrumentation modules rather than requiring the `Embrace` class to obtain a reference to a given data source.

bidetofevil: Our internal stuff shouldn't depend directly on our external API. That creates a conflict of interest internally and probably some dependency stuff we'd rather not have to deal with.

PR #2734, opened 2025-11-11
Add warning to prevent Compose version bump

priettt:
## Goal
Added a warning comment in libs.versions.toml to prevent bumps to Compose dependency version.

## More info
The `getAllSemanticsNodes` signature changed in Compose 1.6+, adding a second parameter (`skipDeactivatedNodes`). If we compile the SDK against Compose 1.6+, the bytecode will reference the two-parameter method. Apps using older Compose versions (like 1.0.5) at runtime will crash with `NoSuchMethodError` since they only have the single-parameter version.

Compiling against 1.0.5 works for both old and new runtime versions since newer Compose versions still include the deprecated single-parameter method.


fractalwrench (approved):
LGTM

PR #2737, opened 2025-11-11
Create separate module for ANR instrumentation

fractalwrench:
## Goal

Creates a separate module that contains the ANR instrumentation. Unlike the other instrumentation modules this is somewhat unusual in that it does not use the `DataSource` interface or SPI directly. This is because the existing instrumentation is something of a compromise where ANR intervals are not recorded as OTel signals until the session ends. IMO we should address this as part of a future refactor, at which point we can lean more into the `DataSource` API (or its State API equivalent).

## Testing

Relied on existing test coverage & manually verified E2E.


bidetofevil (approved):
So one feature we will do in the relatively near future is to make it so we can add call stack samples to any span, which essentially will require decoupling that aspect of span annotation from the ANR instrumentation. Some details still need to be worked out as to what the format is like, but the idea is that the addition of these attributes will be a service accessible outside of the ANR stuff.   
  
This seems like a step towards that, and if there's anything that would make that harder to do, now is probably a good time to address it if there's one of those "six of one, half dozen the other" decisions to be made. (I don't think I see it?)

Anyway, LGTM

PR #2740, opened 2025-11-11
Reduce mock usage

fractalwrench:
## Goal

Avoids some unnecessary mocks where `ApplicationProvider` will suffice.


bidetofevil (approved):
Oh I didn't know that test fixture exists. Nice!

PR #2751, opened 2025-11-13
Reduce use of modules in instrumentation

fractalwrench:
## Goal

Replaces the use of modules with `InstrumentationArgs` for ANR/NDK crash detection instrumentation code. This will make it easier to extract instrumentation to its own module in the long-term.


bidetofevil (approved):
LGTM

PR #2752, opened 2025-11-13
Reorganise some utils classes

fractalwrench:
## Goal

Reorganises some utils classes by moving them to more specific modules or inlining them where they are only used in one place or are superseded by Kotlin language features.


bidetofevil (approved):
Excellent. LGTM

PR #2754, opened 2025-11-13
Simplify AppStateService

fractalwrench:
## Goal

Simplifies `AppStateService` by removing implementation details from the interface.

## Testing


embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/session/lifecycle/AppStateServiceImpl.kt
```diff
@@ -120,28 +122,9 @@ internal class AppStateServiceImpl(
         }
     }
 
-    override fun close() {
```
bidetofevil, line 123: Nice. I'm not sure if there are any system services that we need to be "closeable" since the SDK can't be disabled in process once started, and we'll never turn these off with new feature flags

bidetofevil (approved):
LGTM

PR #2760, opened 2025-11-14
Avoid tracing ID factory use

fractalwrench:
## Goal

Refactors `AnrOtelMapper` to create its own span ID/trace ID rather than requiring a dependency on `embrace-android-otel`. This is the only bit of instrumentation that requires this & longer-term I'd like to avoid the necessity of doing it this way.

### DIFF
--- a/embrace-android-features/src/main/kotlin/io/embrace/android/embracesdk/internal/instrumentation/anr/AnrOtelMapper.kt
+++ b/embrace-android-features/src/main/kotlin/io/embrace/android/embracesdk/internal/instrumentation/anr/AnrOtelMapper.kt
@@ -6,36 +6,37 @@ import io.embrace.android.embracesdk.internal.clock.millisToNanos
 import io.embrace.android.embracesdk.internal.instrumentation.anr.payload.AnrInterval
 import io.embrace.android.embracesdk.internal.instrumentation.anr.payload.AnrSample
 import io.embrace.android.embracesdk.internal.otel.payload.toEmbracePayload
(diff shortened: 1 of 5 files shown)


embrace-android-features/src/main/kotlin/io/embrace/android/embracesdk/internal/instrumentation/anr/AnrOtelMapper.kt
```diff
@@ -6,36 +6,37 @@ import io.embrace.android.embracesdk.internal.clock.millisToNanos
...
 class AnrOtelMapper(
     private val anrService: AnrService,
     private val clock: Clock,
     private val spanService: SpanService,
-    private val tracingIdFactory: TracingIdFactory,
+    private val random: Random = Random.Default,
 ) {
+
+    private companion object {
+        const val INVALID_SPAN_ID: String = "0000000000000000"
```
bidetofevil, line 29: Don't we have this defined somewhere else so you don't need to redefine it here?

fractalwrench: This is defined in `embrace-android-otel`. It felt problematic to add a dependency on the entire module given that we only need this and a function that generates a span/trace ID

bidetofevil: Yeah definitely. It conceptually feels like it doesn't belong here, so I think a future move can move this logic internal to where they would have access to that module.

PR #2765, opened 2025-11-14
Remove session type concept

fractalwrench:
## Goal

Removes the concept of `SessionType`. We don't have any telemetry that is using this API to determine whether to capture telemetry or not, and `AppStateTracker` is now accessible from the instrumentation API so it doesn't make sense to have two concepts.

I have refactored the `DataSourceState` implementation so that they are notified when a new session occurs as this does require resetting limits for data capture.

## Testing

Added an integration test to assert that data limits are actually reset.


bidetofevil (approved):
So DataSources are now supposed to manage differentiation in type by themselves? Is it done through a combination of registration of handlers and just checking background/foreground when it is enabled?

PR #2778, opened 2025-11-17
Tweak upgrade doc

bidetofevil:
## Goal

Added details and tweaked for Americanized spellings per our style guide.

Also removed the upgrading to `7.3`​ instructions as they are redundant.


fractalwrench (approved):
LGTM

PR #2782, opened 2025-11-18
Reorganise modules further

fractalwrench:
## Goal

Organises the dependency modules by moving `SystemServiceModule` to `CoreModule`. Additionally this removes registration of listeners towards the end of the bootstrapping process.

## Testing


bidetofevil (approved):
System belongs to Core, This makes sense thematically too. LGTM

PR #2784, opened 2025-11-18
Extract module dependency graph

fractalwrench:
## Goal

Creates `ModuleGraph`, which holds all the modules the SDK requires to be initialized. This has two implementations: uninitialized and initialized. I've extracted this out from `ModuleInitBootstrapper` as this avoids the use of `lateinit var` and will allow additional actions on the modules to be extracted out in future PRs.

## Testing

Relied on existing test coverage. I also audited every access of each module to confirm they're behind a check that the SDK is started.

### DIFF
--- a/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/internal/injection/InitializedModuleGraph.kt
+++ b/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/internal/injection/InitializedModuleGraph.kt
@@ -1,0 +1,311 @@
+package io.embrace.android.embracesdk.internal.injection
+
+import android.content.Context
+import io.embrace.android.embracesdk.internal.instrumentation.anr.AnrModule
+import io.embrace.android.embracesdk.internal.instrumentation.anr.AnrModuleSupplier
+import io.embrace.android.embracesdk.internal.instrumentation.crash.ndk.NativeCoreModule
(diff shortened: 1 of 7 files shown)


embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/internal/injection/InitializedModuleGraph.kt
```diff
@@ -0,0 +1,311 @@
...
+            EmbTrace.trace("sdk-disable-check") {
+                // kick off config HTTP request first so the SDK can't get in a permanently disabled state
+                EmbTrace.trace("load-config-response") {
+                    module.combinedRemoteConfigSource?.scheduleConfigRequests()
+                }
+
+                EmbTrace.trace("behavior-check") {
+                    if (module.configService.sdkModeBehavior.isSdkDisabled()) {
+                        // bail out early. Caught at a higher-level that relies on this specific type
+                        throw SdkDisabledException()
```
bidetofevil, line 87: I understand why we're using an exception here in terms of the flow of the code, but this feel slight odd given that this is just an case we expect to happen.

fractalwrench: I did find it a bit odd. I can't think of an alternative approach that still allows us to make all the module properties `val`, so I figured it was a reasonable tradeoff

PR #2796, opened 2025-11-20
Move sdk init code

fractalwrench:
## Goal

Moves SDK init code out of `ModuleInitBootstrapper` and `EmbraceImpl` into a separate file.

### DIFF
--- a/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/EmbraceImpl.kt
+++ b/embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/EmbraceImpl.kt
@@ -3,7 +3,6 @@ package io.embrace.android.embracesdk
 import android.annotation.SuppressLint
 import android.content.Context
 import android.util.Log
-import io.embrace.android.embracesdk.core.BuildConfig
 import io.embrace.android.embracesdk.internal.EmbraceInternalApi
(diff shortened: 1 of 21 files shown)


embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/EmbraceImpl.kt
```diff
@@ -200,23 +156,11 @@ internal class EmbraceImpl(
...
-    private fun registerDeliveryNetworkListener() {
-        bootstrapper.deliveryModule.schedulingService?.let(
-            bootstrapper.essentialServiceModule.networkConnectivityService::addNetworkConnectivityListener
-        )
-    }
-
     /**
      * Shuts down the Embrace SDK.
      */
     fun stop() {
```
bidetofevil, line 162: If this is only done during SDK startup, we might be able to refator this out entire so it doesn't appear like this can be called at any time

bidetofevil (approved):
LGTM. Keep going!

embrace-android-sdk/src/main/java/io/embrace/android/embracesdk/EmbraceImpl.kt
fractalwrench: It's called from other places unfortunately so we need to keep this or something like it

PR #2798, opened 2025-11-20
Tweaks to span capture on TelemetryDestination

fractalwrench:
## Goal

Adds the ability to specify parent/name/type and access other information when capturing spans from `TelemetryDestination`. In future I think we'll want to refactor this API to consolidate all these different functions, but that should probably happen as part of the data modelling work.

### DIFF
--- a/embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/arch/destination/TelemetryDestinationImpl.kt
+++ b/embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/arch/destination/TelemetryDestinationImpl.kt
@@ -15,6 +15,7 @@ import io.embrace.android.embracesdk.internal.arch.state.AppStateTracker
 import io.embrace.android.embracesdk.internal.clock.Clock
 import io.embrace.android.embracesdk.internal.clock.nanosToMillis
 import io.embrace.android.embracesdk.internal.otel.sdk.toEmbraceObjectName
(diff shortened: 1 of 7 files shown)


embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/arch/destination/TelemetryDestinationImpl.kt
```diff
@@ -176,22 +202,32 @@ class TelemetryDestinationImpl(
...
+            span.addAttribute(key, value)
+        }
+
+        override fun getStartTimeMs(): Long? = span.getStartTimeMs()
+
+        override fun setSystemAttribute(key: String, value: String) = span.setSystemAttribute(key, value)
+    }
+}
+
+internal fun ErrorCodeAttribute?.toErrorCode(): ErrorCode? {
```
bidetofevil, line 226: Does this need to be exposed to the rest of the module or can it live inside this file/class and be shared by both classes?

bidetofevil (approved):
LGTM other than one non-blocking comment

embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/arch/destination/TelemetryDestinationImpl.kt
fractalwrench: This needs to be internal to avoid generating a synthetic accessor.

PR #2805, opened 2025-11-24
Move startup tracing to separate module

fractalwrench:
## Goal

Moves startup tracing into a separate module.


embrace-android-instrumentation-startup-trace/src/test/kotlin/io/embrace/android/embracesdk/fakes/FakeMainThreadHandler.kt
```diff
@@ -0,0 +1,34 @@
+package io.embrace.android.embracesdk.fakes
+
+import android.os.Handler
+import android.os.Looper
+import android.os.Message
+import io.embrace.android.embracesdk.internal.handler.MainThreadHandler
+import java.util.LinkedList
+import java.util.Queue
+
+class FakeMainThreadHandler(
```
bidetofevil, line 10: Where is this being used? I can't seem to find a reference to it in this review

fractalwrench: This is required by `HandlerMessageDrawDetectorTest`

PR #2814, opened 2025-11-25
Create module for telemetry persistence

fractalwrench:
## Goal

Creates a module that has extracted the code for persisting telemetry to disk. This is used by the NDK crash instrumentation _and_ the delivery layer so requires separating to enable further modularisation.

## Testing


embrace-android-telemetry-persistence/src/main/kotlin/io/embrace/android/embracesdk/internal/delivery/storage/FileStorageServiceImpl.kt
```diff
@@ -10,6 +9,7 @@ import java.io.FileNotFoundException
 import java.io.InputStream
 import java.util.concurrent.CopyOnWriteArraySet
 import java.util.concurrent.RejectedExecutionException
+import kotlin.collections.forEach
```
priettt: why do we need this import?

fractalwrench: IntelliJ keeps messing with my imports when moving code for whatever reason. I must have missed this one making it into the changeset & can remove it

PR #2819, opened 2025-11-25
Move StorageLocationExt

fractalwrench:
## Goal

Moves `StorageLocationExt` to `embrace-android-telemetry-persistence` to facilitate NDK modularisation. This is a JVM-only module so can't rely on the `Context` parameter that was passed in previously.


embrace-android-sdk/src/integrationTest/kotlin/io/embrace/android/embracesdk/testcases/features/DisableSdkFeatureTest.kt
```diff
@@ -39,7 +39,11 @@ internal class DisableSdkFeatureTest {
     @Before
     fun setUp() {
         val ctx = ApplicationProvider.getApplicationContext<Context>()
-        embraceDirs = StorageLocation.entries.map { it.asFile(ctx, FakeEmbLogger()).value }
+        embraceDirs = StorageLocation.entries.map { it.asFile(
+            logger = FakeEmbLogger(),
+            rootDirSupplier = { ctx.filesDir },
+            fallbackDirSupplier = { ctx.cacheDir }
+        ) .value }
```
priettt: will probably get caught by lint, but just in case: 

that `) .value` looks weird

PR #2824, opened 2025-11-25
Use SPI to load ndk crash instrumentation

fractalwrench:
## Goal

Use service loading to initialize the NDK crash instrumentation.


embrace-android-instrumentation-crash-ndk/src/main/kotlin/io/embrace/android/embracesdk/internal/instrumentation/crash/ndk/NdkCrashInstrumentationProvider.kt
```diff
@@ -0,0 +1,60 @@
...
+                )
+
+                val nativeCrashHandlerInstaller = NativeCrashHandlerInstallerImpl(
+                    args,
+                    sharedObjectLoader = sharedObjectLoader,
+                    delegate = delegate,
+                    mainThreadHandler = AndroidMainThreadHandler(),
+                    outputDir = nativeOutputDir,
+                )
+                nativeCrashHandlerInstaller.install()
```
bidetofevil, line 49: Isn't it a bit weird to kick off an async load of the native binary here but then create an instance of the data source later? Are there any cases you'd want to invoke both of these separately?

fractalwrench: It's kind of weird but does match existing behavior. I think this is definitely something to simplify further in future & consolidate into the same place.

PR #2830, opened 2025-11-27
Remove unused proguard lines

fractalwrench:
## Goal

Removes ProGuard lines that are no longer relevant and fixes a few wrong symbols where packages have changed.


fractalwrench:
@bidetofevil it is possible to specify package name keep rules. The problem is that it keeps everything from the package. There's also the `@Keep` annotation that's an alternative to providing proguard rules.

PR #2835, opened 2025-11-27
Rename build telemetry JSON keys

priettt:
## Goal
Improve readability of build telemetry JSON API by replacing abbreviated keys with descriptive names.


embrace-gradle-plugin/src/main/java/io/embrace/android/gradle/plugin/buildreporter/BuildTelemetryCollector.kt
```diff
@@ -116,6 +118,19 @@ class BuildTelemetryCollector {
             null
         }
     }
+
+    private fun getKotlinJvmTarget(project: Project): String? {
+        return if (project.pluginManager.hasPlugin("org.jetbrains.kotlin.android")) {
```
fractalwrench, line 123: This is probably good enough for our purposes. `jvm` or `multiplatform` could also be in play but should be much less common.

priettt: This is definitely something we should revisit in the future.

If we end up refactoring the build telemetry process, one of the things we could do is call `withPlugin` with the three options, jvm, android and multiplatform, get the data from the first one we get, and ignore the rest.

PR #2839, opened 2025-11-28
Alter OTel payload mapper

fractalwrench:
## Goal

Makes `AnrService` implement `OtelPayloadMapper`. This avoids exposing several implementation details such as `AnrInterval` from the module.


bidetofevil (approved):
Nice. LGTM.

PR #2841, opened 2025-11-30
Remove incorrect comment from proguard config

priettt:
## Goal
Fix failing tests by removing an inline comment that was causing issues in proguard configuration.

## Changes
- Removed incorrect `// payload` comment from proguard rule in embrace-proguard.cfg
- Commented out ExampleApp task in CI workflow until 8.0 release


fractalwrench (commented):
Let's keep the example app CI workflow - I'll fix things up so that it builds as part of the same project. The proguard fix looks good.

PR #2847, opened 2025-12-02
Simplify ANR detection code

fractalwrench:
## Goal

Begins simplifying the ANR detection code.

## Testing


embrace-android-instrumentation-anr/src/test/kotlin/io/embrace/android/embracesdk/internal/instrumentation/anr/detection/TargetThreadHandlerTest.kt
```diff
@@ -82,13 +78,11 @@ internal class TargetThreadHandlerTest {
         handler.handleMessage(msg)
         assertEquals(2, executorService.submitCount)
         executorService.runCurrentlyBlocked()
```
priettt, line 80: should we keep this test? it's pretty much the same as the one above now

fractalwrench: I'll open a follow-up PR to delete it.

PR #2850, opened 2025-12-02
Increase default log limits for warnings and errors

nelsitoPuglisi:
## Goal

Increase default log limits for warnings and errors as expected by our backend and match iOS

## Testing

Unit tests updated


pablomatiasgomez (approved):
I'm glad this was lower and not higher :)

PR #2864, opened 2025-12-04
Inline stacktrace sampler functions

fractalwrench:
## Goal

Inlines a few functions within `AnrStacktraceSampler`


bidetofevil (approved):
LGTM

PR #2865, opened 2025-12-04
Alter ANR capture to avoid using ThreadInfo class

fractalwrench:
## Goal

Uses a class that doesn't have JSON annotations & also avoids creating an unnecessary list when only one thread is captured for a given sample at a time.

## Testing


embrace-android-instrumentation-api/src/main/kotlin/io/embrace/android/embracesdk/internal/arch/stacktrace/ThreadTrace.kt
```diff
@@ -0,0 +1,10 @@
+package io.embrace.android.embracesdk.internal.arch.stacktrace
+
+class ThreadTrace(
```
bidetofevil: Is this a Trace or more like a callstack sample? What if we just called it ThreadSample or something like that?

bidetofevil (approved):
LGTM. Question about whehter we should call the thing with the call stack a `ThreadTrace` or something else that doesn't use the word "trace", which means something else.

PR #2871, opened 2025-12-05
Simplify ANR threading model

fractalwrench:
## Goal

Simplifies the threading model used for ANRs as now that concurrent collections + synchronisation are used by `AnrStacktraceSampler` there's no need to run on a specific worker thread.


bidetofevil (approved):
LGTM.   
  
So `AnrStacktraceSampler` methods are now being run on a background thread, or was the worker thread always only used for synchronization and not for offloading work?

PR #2873, opened 2025-12-05
Move ANR code package

fractalwrench:
## Goal

Moves some source files to a separate package as they'll eventually be extracted to a different module.


bidetofevil (approved):
LGTM

PR #2875, opened 2025-12-05
Infra for capturing State using a base DataSource

bidetofevil:
## Goal

Add a generic interface in `TelemetryDestination` to create, modify, and terminate the a state span for the given session. It will be used by a `DataSource` to cut a span that represents the state and its changes throughout a session. 

Further tests are forthcoming

### DIFF
--- a/embrace-android-instrumentation-api/src/main/kotlin/io/embrace/android/embracesdk/internal/arch/datasource/StateDataSource.kt
+++ b/embrace-android-instrumentation-api/src/main/kotlin/io/embrace/android/embracesdk/internal/arch/datasource/StateDataSource.kt
@@ -119,4 +119,4 @@ abstract class StateDataSource<T>(
  * Maximum transitions per state span that the backend can expect. This could be overridable in the future, at which point the max value
  * should be encoded in the data.
  */
-private const val DEFAULT_MAX_TRANSITIONS = 1000
+private const val DEFAULT_MAX_TRANSITIONS = 100


embrace-android-instrumentation-api/src/main/kotlin/io/embrace/android/embracesdk/internal/arch/datasource/TelemetryDestination.kt
```diff
@@ -77,6 +77,11 @@ interface TelemetryDestination {
         events: List<SpanEvent> = emptyList(),
     )
 
+    /**
+     * Start a recording the given [SchemaType.State] for the current session
+     */
+    fun createSessionState(state: SchemaType.State): StateToken?
```
fractalwrench: More tactical, but can we consider making the return type non-null and returning a no-op instance of `StateToken` if it's not possible to create something for whatever reason? This is a pet peeve with our current Tracing API (that we should probably go fix separately)

embrace-android-instrumentation-schema/src/main/kotlin/io/embrace/android/embracesdk/internal/arch/schema/SchemaType.kt
```diff
@@ -282,6 +281,27 @@ sealed class SchemaType(
...
+    abstract class State(stateSpanName: String) :
+        SchemaType(
+            telemetryType = EmbType.State,
+            fixedObjectName = stateSpanName
+        )
+
+    class NetworkState(
+        initialStatus: Status,
+    ) : State("state-network") {
+        enum class Status(val value: String) {
```
fractalwrench:
```suggestion
        enum class Status(private val value: String) {
```

embrace-android-instrumentation-schema/src/main/kotlin/io/embrace/android/embracesdk/internal/arch/schema/SchemaType.kt
```diff
@@ -282,6 +281,27 @@ sealed class SchemaType(
...
+
+    abstract class State(stateSpanName: String) :
+        SchemaType(
+            telemetryType = EmbType.State,
+            fixedObjectName = stateSpanName
+        )
+
+    class NetworkState(
+        initialStatus: Status,
+    ) : State("state-network") {
```
fractalwrench: `EmbType` and `SchemaType` are currently separate and we've talked about combining these in the past. This might be a good time to rationalise concepts if we're adding another concept of state to this hierarchy.

embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/arch/destination/TelemetryDestinationImpl.kt
```diff
@@ -222,6 +238,35 @@ class TelemetryDestinationImpl(
...
+            )
+        }
+    }
+
+    private class StateTokenImpl(
+        val spanToken: SpanToken,
+    ) : StateToken {
+        private val transitionCount = AtomicInteger(0)
+
+        override fun update(timestampMs: Long, newValue: String) {
```
fractalwrench: This enforces that every value should be a non-null string. Are we happy with that trade-off? How should null values be represented, and how would complex values be coerced to strings?

embrace-android-instrumentation-api/src/main/kotlin/io/embrace/android/embracesdk/internal/arch/datasource/StateToken.kt
```diff
@@ -0,0 +1,7 @@
+package io.embrace.android.embracesdk.internal.arch.datasource
+
+interface StateToken {
+    fun update(timestampMs: Long, newValue: String)
```
fractalwrench: How would multiple state values be recorded at the same time? Is it necessary to create a span for every single value? I imagine we could bump into span limits quite quickly if we track several dozen state values and there are various transitions.

embrace-android-instrumentation-api/src/main/kotlin/io/embrace/android/embracesdk/internal/arch/datasource/StateToken.kt
```diff
@@ -0,0 +1,7 @@
+package io.embrace.android.embracesdk.internal.arch.datasource
+
+interface StateToken {
```
fractalwrench: Nit: feels like it could be nice to include functions that provide a default time via a clock implementation, rather than explicitly passing the timestamp in

embrace-android-instrumentation-api/src/main/kotlin/io/embrace/android/embracesdk/internal/arch/datasource/StateToken.kt
fractalwrench: How do we handle the case where `newValue` is the same as the current state value?

embrace-android-instrumentation-network-status/src/main/kotlin/io/embrace/android/embracesdk/internal/instrumentation/network/NetworkStateDataSource.kt
```diff
@@ -0,0 +1,61 @@
...
+            stateToken = newStateSpan(newState)
+        } else if (newState != currentState) {
+            captureTelemetry {
+                currentStateToken.update(
+                    timestampMs = timestamp,
+                    newValue = newState.value
+                )
+            }
+            currentState = newState
+        }
```
fractalwrench: This feels quite complicated & I'm convinced that we'll forget to do all these checks in at least some instrumentation modules. I think it'd be beneficial to create some abstraction for tracking & updating the state that handles these sorts of checks.

embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/arch/destination/TelemetryDestinationImpl.kt
bidetofevil: If null means the default state, it should be explicitly stated and defined in the state definition. For example, for Network state, it's `UNKNOWN`.  Supporting null seems like giving data sources the ability to "set to default" without knowing the value, which I don't think we support without a use case - it seems too undefined, no pun intended lol.

embrace-android-instrumentation-api/src/main/kotlin/io/embrace/android/embracesdk/internal/arch/datasource/StateToken.kt
bidetofevil: This is deliberate - the transition time isn't the time that `TelemetryDestination` receives the update.   
  
I should make that more explicit from the interface standpoint. I'll tweak.

embrace-android-instrumentation-api/src/main/kotlin/io/embrace/android/embracesdk/internal/arch/datasource/StateToken.kt
bidetofevil: Good question. In the Network state implementation, that is filtered out. I think there's utility in recording a transition in case this is the result of debouncing during a period of high oscillation, e.g. a data source will not send an update more than once a second, but within that second, the state goes from foo to bar back to foo.  
  
Of course, we then need to ask the question: do we need to keep track of debouncing data? I feel like adding that to the design is a good idea.   
  
Let me tweak the interface docs and propose a way to amend the design to handle it more explicitly.

bidetofevil: For every state there will only be one span, and we assume right now there could be a couple dozen or so max. For every session, adding that many spans (that will be discarded on the server side) doesn't seem so bad, but that is a consideration.   
  
An alternative design is to put all the span events on the session span, but that was discarded because it feels too messy and may cause synchronization overhead. But maybe 1 span for all states is also an option? Let me put this out to the group.

embrace-android-instrumentation-api/src/main/kotlin/io/embrace/android/embracesdk/internal/arch/datasource/TelemetryDestination.kt
bidetofevil: Good idea

embrace-android-instrumentation-network-status/src/main/kotlin/io/embrace/android/embracesdk/internal/instrumentation/network/NetworkStateDataSource.kt
bidetofevil: Oh yeah. I think a new base data source for State would be beneficial. I haven't done that yet but I can do that after - I might as well as a POC impl of Where You Are based on activity navigation and what we currently do with the View breadcrumb to validate

embrace-android-instrumentation-schema/src/main/kotlin/io/embrace/android/embracesdk/internal/arch/schema/SchemaType.kt
bidetofevil: Will do when I update

embrace-android-instrumentation-schema/src/main/kotlin/io/embrace/android/embracesdk/internal/arch/schema/SchemaType.kt
bidetofevil: I agree that we should revisit these related concepts, but I actually think `State` sort of justifies splitting these out, as different States will share an `EmbType`but may have a different `SchemaType`, e.g. `NetworkState`  
  
At the same time, you're right that something is not quite right given we have to add this concept of a abstract `SchemaType` to unify what should've been unified by a shared `EmbType`.   
  
Let me take a look at this to see if it's something I can do along side it or after.

embrace-android-instrumentation-api/src/main/kotlin/io/embrace/android/embracesdk/internal/arch/datasource/StateToken.kt
bidetofevil: Actually, I'm going to take the time out of the the `end()` because the token-holder shouldn't have to know deal with that. But I'm still keeping the timestamp in the update method but make it more clear why.

embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/arch/destination/TelemetryDestinationImpl.kt
bidetofevil: Thinking more about it, I think the value type could be specified, e.g. Int, enum, etc.   
  
I still think it should be non-null though.

embrace-android-instrumentation-schema/src/main/kotlin/io/embrace/android/embracesdk/internal/arch/schema/SchemaType.kt
bidetofevil: This is accessed by the data source, which seems bad and another reason why a StateToken should be typed and there's a generic way to turn the type to an attribute value when it's recorded.

embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/arch/destination/TelemetryDestinationImpl.kt
bidetofevil: I've made the state token typed, and utilized toString() to get it written into a string attribute. A state that is bound to a value. I had a value type interface initially but that seems too heavy handed

fractalwrench: It seems sensible to enforce that a state always needs some value, even if it's 'unknown'. Encoding the type could be useful. I do think it would be good to be explicit about what types are supported as it's only a matter of time before someone feels the need to encode an arbitrary map or JSON object...

embrace-android-instrumentation-api/src/main/kotlin/io/embrace/android/embracesdk/internal/arch/datasource/StateToken.kt
fractalwrench: Is the State API intended to replace all of our instrumentation? If not, what is its primary use-case?

embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/arch/destination/TelemetryDestinationImpl.kt
bidetofevil: Right now I have it as anything that can be `toString()`'d, which... is everything. 

I think if as long as the blob folks want to `toString()` meets the criteria we've laid out, it's fine? The low-ish cardinality requirement we have in the backend will enforce this anyway, and erroneous use will be ferreted out.

embrace-android-instrumentation-api/src/main/kotlin/io/embrace/android/embracesdk/internal/arch/datasource/StateToken.kt
bidetofevil: The State API is meant to replace the modelling of metadata that can be effectively distilled into a single, mutable attribute value for some duration of time that globally has a relatively low cardinality (100s-ish). The metadata is persisted uniformly per session along with the transition events, which can be considered both telemetry and metadata. 

Examples are things like network connection status, low power mode, and where in the app you are

Examples of things that are NOT state is the number of API calls for a session, heap size, user name, and how long it took the SDK to startup

Something like a user journey can be seen as having characteristics of a State, but won't be modelled as one because it is something more (what specific user journey a user is on currently CAN be a State if we want it to be though).

Does that make sense? I'll copy and paste some of this into the design doc so folks don't have to go to the other docs to get the basics.

embrace-android-instrumentation-schema/src/main/kotlin/io/embrace/android/embracesdk/internal/arch/schema/SchemaType.kt
bidetofevil: I want to unify this with `NetworkStatus` but it's a bit trickier than I though, so I'll do it after

embrace-android-instrumentation-api/src/main/kotlin/io/embrace/android/embracesdk/internal/arch/datasource/StateToken.kt
fractalwrench: Yep, that definition was helpful

embrace-android-sdk/src/main/kotlin/io/embrace/android/embracesdk/internal/injection/SdkInitActions.kt
```diff
@@ -105,6 +106,10 @@ internal fun ModuleGraph.postLoadInstrumentation() {
     registry.findByType(NetworkStatusDataSource::class)?.let {
         essentialServiceModule.networkConnectivityService.addNetworkConnectivityListener(it)
     }
+    registry.findByType(NetworkStateDataSource::class)?.let {
+        essentialServiceModule.networkConnectivityService.addNetworkConnectivityListener(it)
```
fractalwrench: Why are two network connectivity listeners being added?

embrace-android-instrumentation-network-status/src/main/kotlin/io/embrace/android/embracesdk/internal/instrumentation/network/NetworkStateDataSource.kt
```diff
@@ -0,0 +1,61 @@
...
+                )
+            }
+        }
+    }
+
+    override fun resetDataCaptureLimits() {
+        setNewStateSpan(currentState)
+    }
+
+    override fun cleanCollections() {
```
fractalwrench: Not required for this PR, just stating my intention that we should combine these two functions on DataSource in future as they effectively do the same thing

embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/arch/destination/TelemetryDestinationImpl.kt
```diff
@@ -222,6 +236,40 @@ class TelemetryDestinationImpl(
...
+            span.addEvent(
+                name = name,
+                timestampMs = eventTimeMs,
+                attributes = attributes
+            )
+        }
+    }
+
+    private class SessionStateTokenImpl<T>(
+        val spanToken: SpanToken,
```
fractalwrench: Can this be private?

embrace-android-instrumentation-api/src/main/kotlin/io/embrace/android/embracesdk/internal/arch/datasource/SessionStateToken.kt
```diff
@@ -0,0 +1,18 @@
+package io.embrace.android.embracesdk.internal.arch.datasource
+
+/**
+ * An object representing a State for a particular session. A valid instance can be used to update or end state tracking for a particular
+ * session
+ */
+interface SessionStateToken<T> {
+    /**
+     * Update the state to the current value. There is no guarantee that the new value is different than current one due to debouncing,
+     * so implementations need to handle those cases as they see fit.
```
fractalwrench: It's unclear to me what the last sentence means

embrace-android-instrumentation-api/src/main/kotlin/io/embrace/android/embracesdk/internal/arch/datasource/SessionStateToken.kt
```diff
@@ -0,0 +1,18 @@
...
+     */
+    fun update(updateDetectedTimeMs: Long, newValue: T)
```
fractalwrench: Did we want an overload without the explicit timestamp?

embrace-android-instrumentation-network-status/src/main/kotlin/io/embrace/android/embracesdk/internal/instrumentation/network/NetworkStateDataSource.kt
fractalwrench: It'd be nice to have that in this PR but I won't leave it as blocking as long as it gets done, as this definitely offloads too much state management to the instrumentation IMO

bidetofevil: I'll do this as part of the infra vs network state split

embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/arch/destination/TelemetryDestinationImpl.kt
bidetofevil: oh yeah definitely

embrace-android-instrumentation-api/src/main/kotlin/io/embrace/android/embracesdk/internal/arch/datasource/SessionStateToken.kt
bidetofevil: I don't want it at this point. I want the data source to explicitly give me the time of the update, which it must have to know. I think the delay is material, and given the clock is always available to the data source, I don't mind asking it to do it bit more work.

embrace-android-instrumentation-api/src/main/kotlin/io/embrace/android/embracesdk/internal/arch/datasource/SessionStateToken.kt
bidetofevil: So the idea is that not `update` calls will necessarily modify the underlying data if, say, too many updates come within a short interval. The last sentence give the implementation permission to drop any updates it doesn't want to deal with, or do something like count the number of updates it has dropped (which is likely what's going to happen in the real implementation), but it is stated like the toekn exposes the current state value, which it doesn't.   
  
Hmm. Let me tweak this then. It is a bit confusing now that I look at it again.

embrace-android-sdk/src/main/kotlin/io/embrace/android/embracesdk/internal/injection/SdkInitActions.kt
bidetofevil: The other ones is the existing span-producing NetworkSTATUS data source which is different than this new NetworkSTATE one. The Status one I hope we can deprecate and delete in the next major version one this one is stable.

embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/arch/destination/TelemetryDestinationImpl.kt
fractalwrench: That's fine by me. I'd like us to document it in the SDK code.

embrace-android-instrumentation-api/src/main/kotlin/io/embrace/android/embracesdk/internal/arch/datasource/StateDataSource.kt
```diff
@@ -0,0 +1,57 @@
...
+            sessionStateToken = args.destination.startSessionStateCapture(
+                state = schemaTypeFactory(initialValue)
+            )
+        } catch (t: Throwable) {
+            logger.trackInternalError(InternalErrorType.SESSION_STATE_CREATION_FAIL, t)
+        }
+    }
+}
+
+private const val MAX_TRANSITIONS = 1000
```
fractalwrench: This seems potentially excessive given there could be dozens of spans. Have we checked with the backend what sort of hard limit they want?

embrace-android-instrumentation-api/src/main/kotlin/io/embrace/android/embracesdk/internal/arch/datasource/StateDataSource.kt
```diff
@@ -0,0 +1,57 @@
...
+abstract class StateDataSource<T>(
+    private val args: InstrumentationArgs,
+    defaultValue: T,
+    private val schemaTypeFactory: (initialValue: T) -> SchemaType.State<T>,
+) : MemoryCleanerListener, DataSourceImpl(
+    args = args,
+    limitStrategy = UpToLimitStrategy { MAX_TRANSITIONS }
+) {
+    private var currentState: T = defaultValue
+    private var sessionStateToken: SessionStateToken<T>? = null
```
fractalwrench: Consider `AtomicReference` as that will ensure data liveness if `onStateChange` is called from multiple threads

embrace-android-instrumentation-api/src/main/kotlin/io/embrace/android/embracesdk/internal/arch/datasource/SessionStateToken.kt
fractalwrench: I think this approach does make it marginally more likely that a span event could have a timestamp before the span start time (or that we could somehow screw up the timestamp capture). Will the backend process state transitions ok if the timestamps are not within the boundaries of a span?

embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/arch/destination/TelemetryDestinationImpl.kt
bidetofevil: Yup, good call and will do.

embrace-android-instrumentation-api/src/main/kotlin/io/embrace/android/embracesdk/internal/arch/datasource/SessionStateToken.kt
bidetofevil: Yes. The fuzziness is understood, and the transition events are noted for not always landing within the bounds of the session in which it was delivered. The backend and/or UI, should they wish to, can hide or "reparent" if it really matters since the state span will be consumed and discarded, and the UI can just query for transitions based on timestamps and have the delayed one show up after the session it was in is received, if it ever is.  
  
The async nature of this stuff means that even the instrumentation that picks up the timestamp could be delayed, so even if we lock everything and get the transitions right, there is still going to be a mismatch with reality. In aggregate, this will hopefully wash out. What we do is just a good approximation and on a mobile device, transitions just ain't going to be very crisp, so we just have to do our best and deal with the fuzziness.

embrace-android-instrumentation-api/src/main/kotlin/io/embrace/android/embracesdk/internal/arch/datasource/StateDataSource.kt
bidetofevil: The original idea is that this wasn't but you're right - it could. Let me fix that.

embrace-android-instrumentation-api/src/main/kotlin/io/embrace/android/embracesdk/internal/arch/datasource/StateDataSource.kt
bidetofevil: That's an open question I need to validate (it's in the doc). I figure this is only in place at the data level to prevent the degenerative case, and even that high would only make sense if we are dealing with super long sessions. (There's also a proposal to add in like a "dropped updates" attribute in the event so we can record details of debouncing - it's not in the implementation yet though)  
  
It probably makes sense for the UI to be smart about this and not just show every transition event if they are clustered, but we shouldn't prevent that to be recorded at the data level.  
  
And the state implementations themselves should probably do some debouncing on their own. I could even build that into network state implementation if we want, or at least create an example that does so.

embrace-android-instrumentation-api/src/main/kotlin/io/embrace/android/embracesdk/internal/arch/datasource/SessionStateToken.kt
fractalwrench: Cool. Another explanation to maybe condense and add to KDocs?

embrace-android-instrumentation-api/src/main/kotlin/io/embrace/android/embracesdk/internal/arch/datasource/StateDataSource.kt
fractalwrench: IMO we should set this lower if it's an open question. E.g. 200 or something around that ballpark?

embrace-android-instrumentation-api/src/main/kotlin/io/embrace/android/embracesdk/internal/arch/datasource/SessionStateToken.kt
bidetofevil: Yeah for sure

embrace-android-instrumentation-api/src/main/kotlin/io/embrace/android/embracesdk/internal/arch/datasource/StateDataSource.kt
bidetofevil: This value is only there to put a ceiling on things to avoid disaster, and it should be overridable by each state if it calls for.

embrace-android-instrumentation-schema/src/main/kotlin/io/embrace/android/embracesdk/internal/arch/schema/SchemaType.kt
bidetofevil: I'll be entering a separate task to track this.

embrace-android-config/src/main/kotlin/io/embrace/android/embracesdk/internal/config/behavior/AutoDataCaptureBehavior.kt
```diff
@@ -64,4 +64,9 @@ interface AutoDataCaptureBehavior {
      * Whether the app startup trace will be waiting for a call to Embrace.appReady() to signal completion
      */
     fun isEndStartupWithAppReadyEnabled(): Boolean
+
+    /**
+     * Whether the state-as-span feature is enabled
+     */
+    fun isStateCaptureEnabled(): Boolean
 }
```
bidetofevil, line 72: This feature flag is expected to be always off until we are ready to ship the feature. Having a flag allows us to test this without actually putting it in production.

embrace-android-instrumentation-api/src/main/kotlin/io/embrace/android/embracesdk/internal/arch/datasource/StateDataSource.kt
fractalwrench: @bidetofevil I'd like to set this limit initially lower unless the data model guild is happy with setting it at this threshold & doesn't have any concerns

bidetofevil: This has been reviewed by the data model guild, but we should look at it at a state-by-state level to see if we want further throttling. The backend folks are aware that they shouldn't just show all the change events on the timeline - but capping it to too low of a ceiling might make us drop valuable data that should be dropped further town the line, if the size of these state spans aren't a concern.

fractalwrench: I don't think it needs formal review, but I do think it could use a quick discussion (sync or async) so that everybody is on the same page about what a sensible limit is. This isn't isolated to Android either as all the other SDKs will need to pick the same limit

bidetofevil: Sure, Let talk about this on Thursday?

PR #2877, opened 2025-12-08
Extract thread sampling

fractalwrench:
## Goal

Extracts the actual sampling to `ThreadStacktraceSampler` and confines all the logic around setting interval codes etc to the ANR-specific class.

## Testing

### DIFF
--- a/embrace-android-instrumentation-anr/src/main/kotlin/io/embrace/android/embracesdk/internal/instrumentation/anr/ThreadBlockageSampler.kt
+++ b/embrace-android-instrumentation-anr/src/main/kotlin/io/embrace/android/embracesdk/internal/instrumentation/anr/ThreadBlockageSampler.kt
@@ -1,13 +1,15 @@
 package io.embrace.android.embracesdk.internal.instrumentation.anr
 
 import io.embrace.android.embracesdk.internal.arch.stacktrace.ThreadSample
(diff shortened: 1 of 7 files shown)


embrace-android-instrumentation-anr/src/main/kotlin/io/embrace/android/embracesdk/internal/instrumentation/anr/ThreadBlockageSampler.kt
```diff
@@ -122,11 +95,39 @@ internal class ThreadBlockageSampler(
...
+
+            // if a stacktrace is repeated in consecutive elements, omit it from the payload. This
+            // is done for legacy reasons with the aim of saving payload size.
+            val dedupe = when (lastSample) {
+                current -> null
+                else -> current
+            }
+            lastSample = current
+
+            ThreadBlockageSample(
```
bidetofevil, line 116: So we are just supposed to know that a blank sample means the stack is what in the previous non-null stack?  
  
I sure hope it's documented somewhere other than in ppl's heads lol

fractalwrench: I'll add an explicit test case for this in a separate PR. I think we have implicit coverage but explicit would be better.

PR #2878, opened 2025-12-08
Simplify interval clearing logic

fractalwrench:
## Goal

Simplifies the ANR interval capture logic by creating a new collection rather than clearing all the samples from the previous one. This avoids the need for a mutex. I've also moved the interval clearing logic to `getThreadBlockageIntervals` as that's the only place it's required.

## Testing


embrace-android-instrumentation-anr/src/main/kotlin/io/embrace/android/embracesdk/internal/instrumentation/anr/ThreadBlockageSampler.kt
```diff
@@ -48,52 +50,48 @@ internal class ThreadBlockageSampler(
...
-                    ThreadBlockageInterval(
-                        startTime = lastUnblockedMs.get(),
-                        lastKnownTime = clock.now(),
-                        samples = sampler.get().getThreadBlockageSamples(),
-                    )
+        val intervals = intervalSink.get()
+        val results = intervals.toMutableList()
+
+        // add any in-progress intervals
+        if (blocked.get()) {
```
bidetofevil, line 57: This could be done without a lock because we can never get new intervals while this is running?

fractalwrench: It can be done without a lock because `results` is a local copy of the original collection.

PR #2879, opened 2025-12-08
Rename ANR to thread blockage

fractalwrench:
## Goal

Renames 'ANR' to 'thread blockage' as that more accurately describes what we capture.


priettt (approved):
Looks good! 

Three questions:
- Should we call the feature MainThreadBlockage instead? Or are we tracking any thread blockage?
- Should we rename the module to embrace-android-instrumentation-thread-blockage?
- Should we update user facing docs and dashboard in the future?

fractalwrench:
Eventually we'll be using this code to track general thread blockages (I think). I agree on renaming/reworking docs but suggest we do that as part of 9.0.

PR #2881, opened 2025-12-08
Remove functions from internal interface

fractalwrench:
## Goal

Removes functions that are unused from Embrace's internal interfaces (used by the hybrid SDKs). This reduces the maintenance burden on the Android SDK and removes scope for bugs etc. I've determined whether something is unused or not by checking these source files:

https://github.com/embrace-io/embrace-unity-sdk/blob/main/io.embrace.sdk/Scripts/Native/Embrace_Android.cs
https://github.com/embrace-io/embrace-react-native-sdk/blob/main/packages/core/android/src/main/java/io/embrace/rnembracecore/EmbraceManagerModule.java
https://github.com/embrace-io/embrace-flutter-sdk/blob/main/embrace_android/android/src/main/kotlin/io/embrace/flutter/EmbracePlugin.kt

## Removed functions

The following functions are removed from `EmbraceInternalApi`:

- logInfo
- logWarning
- logError
- recordCompletedNetworkRequest
- recordIncompleteNetworkRequest
- isAnrCaptureEnabled
- isNdkEnabled
- stopSdk
- logHandledJsException
- setCacheableJavaScriptBundleUrl
- logRnAction
- logRnView


embrace-android-sdk/src/main/kotlin/io/embrace/android/embracesdk/internal/api/ViewTrackingApi.kt
```diff
@@ -23,6 +23,7 @@ internal interface ViewTrackingApi {
     /**
      * Logs a React Native Redux Action.
      */
+    @Deprecated("This function has no effect.")
     fun logRnAction(
```
bidetofevil, line 27: I thought these were removed? Why did we effectively un-deprecate these instead?

fractalwrench: `ViewTrackingApi` is part of our public API. This is technically something to delete in the next major version bump (even if it's unlikely anybody is using it).

PR #2887, opened 2025-12-09
Create session change listener

fractalwrench:
## Goal

Creates a `SessionChangeListener` interface as that's effectively what `MemoryCleanerListener` is doing already, and we have an existing interface already in the `SessionIdTracker`. In future PRs I think we should move some other objects over to use this interface such as `UserService/SessionProperties/InstrumentationRegistry`

## Testing


embrace-android-instrumentation-api/src/main/kotlin/io/embrace/android/embracesdk/internal/arch/SessionChangeListener.kt
```diff
@@ -0,0 +1,10 @@
+package io.embrace.android.embracesdk.internal.arch
+
+fun interface SessionChangeListener {
+
+    /**
+     * Called when the 'session' has changed. This is called _after_ the previous session has
+     * ended.
+     */
+    fun onSessionChange()
```
bidetofevil: Can we put the timing into the method call, e.g. `onPostSessionChange()` or something to that effect? 

I can see us eventually doing pre/post for session ending and starting tbh, and having the interface pass information like sessionIds so that the code in the hook won't have to keep track or find it in other ways.

PR #2888, opened 2025-12-09
Add try-catch for sdk initialization

fractalwrench:
## Goal

Adds a try-catch for the SDK's initialization. This seems to have been removed in a refactor at some point.


bidetofevil (approved):
I think this was a wrapper that called startImpl or something which is then try-caught.  
  
Should we put something out in logcat to let folks know the SDK startup was interrupted by an unexpected error?

PR #2893, opened 2025-12-10
[DO NOT MERGE] Read versionName, versionCode and packageName on build time

priettt:
## Goal

Inject versionName, versionCode and packageName on build time, instead of accessing them through the packageManager

## Changes

Removed PackageVersionInfo in favor of said fields injected at buildtime in ProjectConfig, and added them to BuildInfo. 
Tested on an app with many flavors and buildTypes, and the collected data is the same.


embrace-gradle-plugin/src/main/java/io/embrace/android/gradle/plugin/model/AndroidCompactedVariantData.kt
```diff
@@ -13,24 +15,32 @@ data class AndroidCompactedVariantData(
...
 ) : Serializable {
 
     companion object {
         @Suppress("ConstPropertyName")
         private const val serialVersionUID = 1L
 
         fun from(variant: Variant): AndroidCompactedVariantData {
-            val buildtype = variant.buildType?.lowercase()
-            val debuggable = buildtype?.contains("debug") ?: false
+            val appVariant = variant as ApplicationVariant
```
fractalwrench, line 30: Is this the only place we check for `ApplicationVariant` right now? I'm curious about whether this would throw an error in libraries if a plugin is applied there for whatever reason.

embrace-gradle-plugin/src/main/java/io/embrace/android/gradle/plugin/model/AndroidCompactedVariantData.kt
```diff
@@ -13,24 +15,32 @@ data class AndroidCompactedVariantData(
...
+            val mainOutput = appVariant.outputs.single {
```
fractalwrench, line 31: This seems risky - can an app variant have more than one output?

priettt: This predicate `it.outputType == VariantOutputConfiguration.OutputType.SINGLE` should mean that we only bring one output. 

It does bring the question though, what happens with split APK support? I'll have a look and check what we can do

fractalwrench: Looking at the enum, I assume that we should attempt to get an output from `UNIVERSAL` too: https://developer.android.com/reference/tools/gradle-api/8.3/null/com/android/build/api/variant/VariantOutputConfiguration.OutputType

I do think a safe cast for `ApplicationVariant` and `singleOrNull` would be preferable as it would avoid breaking the build & allow the plugin to degrade gracefully

priettt: agreed, working on it!

PR #2896, opened 2025-12-11
Tweak config assertion

fractalwrench:
## Goal

Tweaks an assertion in the remote config tests so that the test won't fail if the remote config is present on disk but doesn't contain valid JSON. I believe sporadic failures on CI and locally have been down to a race between the test assertion and the HTTP request where the config file is present on disk but empty. I've fixed this by returning null if valid JSON isn't present and waiting for another iteration in `returnIfConditionMet`.


bidetofevil (approved):
LGTM. The race condition is just in the test, where we can maybe wait for the file to be present, but more difficult to parse and validate?

PR #2897, opened 2025-12-11
Reduce internal interface

fractalwrench:
## Goal

Removes some functions from the Android SDK's internal interface that the Flutter SDK (and others) are no longer using. The relevant changes from a Flutter POV can be found on `EmbraceInternalInterface` and everything else is just a consequence of removing those functions.


bidetofevil (approved):
LGTM

PR #2898, opened 2025-12-11
Begin refactoring log delegate implementation

fractalwrench:
## Goal

Begins refactoring the log delegate implementation to make it easier to pass custom schema types and attributes down.

## Testing


fractalwrench:
@bidetofevil I hadn't thought quite that far ahead. I want `LogService` to just accept a generic OTel log/supply a custom schema. Making it more like `SpanService` or rationalizing with `TelemetryDestination` might be good next steps after that.

PR #2900, opened 2025-12-11
Begin consolidating log attributes

fractalwrench:
## Goal

Begins rationalizing how log attributes are constructed.

### DIFF
--- a/embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/logs/LogService.kt
+++ b/embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/logs/LogService.kt
@@ -1,9 +1,8 @@
 package io.embrace.android.embracesdk.internal.logs
 
 import io.embrace.android.embracesdk.LogExceptionType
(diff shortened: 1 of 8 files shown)


embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/logs/LogService.kt
```diff
@@ -15,15 +14,12 @@ interface LogService : SessionChangeListener {
      */
     fun log(
```
bidetofevil, line 15: Feels like that log term, there should be a generic `log` method has Embrace-y concepts, similar to `SpanService.createSpan()`,  and all the details specific to the Embrace Log will sit on top in a DataSource or something that that calls into this via the `TelemetryDestination`.  Is that the vision?

fractalwrench: Answered here: https://github.com/embrace-io/embrace-android-sdk/pull/2898#issuecomment-3643192891

PR #2902, opened 2025-12-11
Move stacktrace attribute in logs

fractalwrench:
## Goal

Moves the location stacktraces are added as a log attribute


bidetofevil (approved):
LGTM

PR #2908, opened 2025-12-11
Remove unused variantVersion field from VariantConfig

priettt:
## Goal
Clean up dead code by removing the `variantVersion` field that was being collected but never used anywhere in the codebase.

## Changes
- Removed `variantVersion` parameter from `VariantConfig` data class
- Removed `versionName` parameter from `AndroidCompactedVariantData` 
- Updated all test files to remove the unused field
- Removed empty string placeholder that was marked as "not used for now"


fractalwrench (approved):
LGTM

PR #2911, opened 2025-12-12
Do session transition inside of SessionTracker

bidetofevil:
## Goal

The first of step of rationalizing session bookkeeping and create a better defined session lifecycle where instrumentation can hook into its various parts is to consolidate the logic of starting, ending, and transitioning between sessions in one place: `SessionTracker`

This PR restructures the existing transitioning by largely keeping the code path the same. `SessionOrchestrator` and `SessionTracker` has also been refactored so that the latter contains the source of truth when it comes to what the active session is.

A future PR will add a hook so instrumentation can "do something" before a session ends so that state spans can be finalized within them. More refactoring can be done after that, but lets get that in first.

## Testing

Existing tests should cover this

### DIFF
--- a/embrace-android-core/src/test/kotlin/io/embrace/android/embracesdk/internal/session/id/SessionTrackerImplTest.kt
+++ b/embrace-android-core/src/test/kotlin/io/embrace/android/embracesdk/internal/session/id/SessionTrackerImplTest.kt
@@ -1,5 +1,7 @@
 package io.embrace.android.embracesdk.internal.session.id
 
+import io.embrace.android.embracesdk.fakes.fakeSessionZygote
+import io.embrace.android.embracesdk.internal.arch.state.AppState
 import io.embrace.android.embracesdk.internal.logging.EmbLoggerImpl
 import org.junit.Assert.assertEquals
 import org.junit.Assert.assertNull
@@ -16,7 +18,7 @@ internal class SessionTrackerImplTest {
     }
 
     @Test
-    fun `test set session id`() {
+    fun `verify tracker implementation`() {
         var count = 0
         tracker.addListener {
             count++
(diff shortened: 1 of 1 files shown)


embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/session/id/SessionTracker.kt
```diff
@@ -2,26 +2,28 @@ package io.embrace.android.embracesdk.internal.session.id
...
 import io.embrace.android.embracesdk.internal.arch.state.AppState
+import io.embrace.android.embracesdk.internal.session.SessionZygote
 
 interface SessionTracker {
 
     /**
      * Gets the currently active session, if present.
      */
-    fun getActiveSession(): SessionData?
+    fun getActiveSession(): SessionZygote?
```
fractalwrench, line 12: The class name is weird if we're going to use it like this. I'd suggest `SessionToken` or something other than `SessionZygote`

embrace-android-core/src/test/kotlin/io/embrace/android/embracesdk/internal/session/id/SessionTrackerImplTest.kt
```diff
@@ -26,15 +25,15 @@ internal class SessionTrackerImplTest {
...
+//        tracker.setActiveSession("123", AppState.FOREGROUND)
+//        assertEquals(SessionData("123", AppState.FOREGROUND), tracker.getActiveSession())
+//        assertEquals("123", tracker.getActiveSessionId())
+//        assertEquals(1, count)
+//
+//        tracker.setActiveSession("456", AppState.FOREGROUND)
+//        assertEquals(2, count)
+//
+//        tracker.setActiveSession(null, AppState.BACKGROUND)
+//        assertEquals(3, count)
```
fractalwrench: This should be removed

embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/session/id/SessionTracker.kt
bidetofevil: Agree. I'll rename after these merge.

embrace-android-core/src/test/kotlin/io/embrace/android/embracesdk/internal/session/id/SessionTrackerImplTest.kt
bidetofevil: Haha, oops! I should write proper tests for them actually. Let me do that first.

PR #2913, opened 2025-12-12
Alter log function signatures

fractalwrench:
## Goal

Simplified the function signature required for `logMessageImpl` by creating an optional `LogExceptionType` parameter that holds data about an error/exception. 

## Testing


bidetofevil (approved):
LGTM

PR #2914, opened 2025-12-12
Alter unity/flutter logging

fractalwrench:
## Goal

Alters the implementation of the Unity/Flutter internal logging APIs so that they use Embrace's public Log API only. For backwards compatibility I've altered the existing functions to use the public API for now - see `FlutterInternalInterfaceImpl`. This demonstrates what the invocation should look like from the relevant hybrid SDK.

I've achieved this by creating a `Custom` class on `SchemaType` and inspecting the `attributes` that are passed in. If `emb.type` is set we construct the schema appropriately. All other attributes such as stacktrace etc are just treated as a normal key-value-pair - it's only necessary to know the semantic convention.

## Testing

Updated unit/integration tests.

### DIFF
--- a/embrace-android-sdk/src/main/kotlin/io/embrace/android/embracesdk/internal/api/delegate/FlutterInternalInterfaceImpl.kt
+++ b/embrace-android-sdk/src/main/kotlin/io/embrace/android/embracesdk/internal/api/delegate/FlutterInternalInterfaceImpl.kt
@@ -4,8 +4,6 @@ import io.embrace.android.embracesdk.EmbraceImpl
 import io.embrace.android.embracesdk.Severity
 import io.embrace.android.embracesdk.internal.EmbraceInternalInterface
 import io.embrace.android.embracesdk.internal.FlutterInternalInterface
-import io.embrace.android.embracesdk.internal.arch.schema.EmbType.System.FlutterException.embFlutterExceptionContext
-import io.embrace.android.embracesdk.internal.arch.schema.EmbType.System.FlutterException.embFlutterExceptionLibrary
 import io.embrace.android.embracesdk.internal.envelope.metadata.HostedSdkVersionInfo
 import io.embrace.android.embracesdk.internal.logging.EmbLogger
 import io.embrace.android.embracesdk.internal.logs.LogExceptionType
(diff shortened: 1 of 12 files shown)


embrace-android-sdk/src/main/kotlin/io/embrace/android/embracesdk/internal/api/delegate/FlutterInternalInterfaceImpl.kt
```diff
@@ -66,20 +64,26 @@ internal class FlutterInternalInterfaceImpl(
         exceptionType: LogExceptionType,
     ) {
         if (embrace.isStarted) {
-            val attrs = mutableMapOf<String, String>()
-            context?.let { attrs[embFlutterExceptionContext.name] = it }
-            library?.let { attrs[embFlutterExceptionLibrary.name] = it }
+            val attrs = mutableMapOf(
+                "emb.type" to "sys.flutter_exception",
```
bidetofevil, line 68: Prefer to use the key `EmbType.Custom.key.name` as the key name

embrace-android-sdk/src/main/kotlin/io/embrace/android/embracesdk/internal/api/delegate/FlutterInternalInterfaceImpl.kt
```diff
@@ -66,20 +64,26 @@ internal class FlutterInternalInterfaceImpl(
...
+                "emb.private.send_mode" to "immediate",
```
bidetofevil, line 69: Prefer to use `embSendMode.name to SendMode.IMMEDIATE.name`

embrace-android-sdk/src/main/kotlin/io/embrace/android/embracesdk/internal/api/delegate/FlutterInternalInterfaceImpl.kt
```diff
@@ -66,20 +64,26 @@ internal class FlutterInternalInterfaceImpl(
...
+            )
+            context?.let { attrs["emb.exception.context"] = it }
```
bidetofevil: Similarly prefer used `embFlutterExceptionContext.name` - ditto for all other Embrace attributes that are defined in a constant

bidetofevil (approved):
LGTM. Just a comment about preferring to use names derived from attributes that are defined as constants rather than hardcoding the string.

PR #2918, opened 2025-12-12
Add session end listener for instrumentation

bidetofevil:
## Goal

Add a hook to invoke code just before a session ends through the `SessionTracker`. Listeners can be through `ServiceRegistry`, `InstrumentationRegistry`, and `InstrumentationArgs`. Further consolidation will be done in subsequent PRs.

### DIFF
--- a/embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/session/id/SessionTrackerImpl.kt
+++ b/embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/session/id/SessionTrackerImpl.kt
@@ -42,10 +42,10 @@ internal class SessionTrackerImpl(
             }
             endingSession.endSessionCallback()
         }
-        activeSession = startSessionCallback()?.apply {
-            runCatching {
-                sessionChangeListeners.forEach(SessionChangeListener::onPostSessionChange)
-            }
+
+        activeSession = startSessionCallback()
+        runCatching {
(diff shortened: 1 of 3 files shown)


embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/session/id/SessionTrackerImpl.kt
```diff
@@ -14,20 +15,18 @@ internal class SessionTrackerImpl(
...
-    private val listeners = CopyOnWriteArraySet<SessionChangeListener>()
+    private val sessionChangeListeners = CopyOnWriteArraySet<SessionChangeListener>()
+    private val sessionEndListeners = CopyOnWriteArraySet<SessionEndListener>()
 
     @Volatile
     private var activeSession: SessionToken? = null
-        set(value) {
-            field = value
-            try {
-                listeners.forEach(SessionChangeListener::onPostSessionChange)
```
fractalwrench, line 24: Removing this will mean the session ID isn't updated appropriately at startup for the cached NDK payload.

fractalwrench (approved):
LGTM

embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/session/id/SessionTrackerImpl.kt
bidetofevil: Oh? I'll fix. And add a test for it

PR #2921, opened 2025-12-12
Use app version and package name for OTel service attributes

### LINKED ISSUE #2223: OTel service version field should contain the version of the client app
When exporting to OTel, every telemetry item contains a field called telemetry.distro.version and one called service.version. Currently both are set to the version of the OpenTelemetry SDK. If I understand correctly, service.version was intended to contain the version of the "service" (= web server application) that the telemetry describes. Therefore the Embrace SDK should set this field to the version of the mobile app.

See https://github.com/embrace-io/embrace-android-sdk/blob/d2581cf323fdad3e11f380db8c02c757b7455d5f/embrace-android-otel/src/main/kotlin/io/embrace/android/embracesdk/internal/otel/config/OtelSdkConfig.kt#L32-L34

I would send a PR but I need some guidance regarding where to obtain the application version, because it seems OtelSdkConfig is initialized before CoreModule.

priettt:
## Goal

Fix issue #2223 by using the mobile app's version and package name for OTel service attributes instead of the SDK version.

## Changes

- Updated `OtelSdkConfig` to accept `appVersion` and `packageName` parameters
- Modified resource attributes mapping:
  - `service.name` now uses the app package name (e.g., "com.example.app")
  - `service.version` now uses the app version name (e.g., "1.0.0")
  - `telemetry.distro.name` and `telemetry.distro.version` remain unchanged (SDK name/version)
- Wired `getPackageName()` and `getVersionName()` from build-time injected config through `OpenTelemetryModuleImpl`
- Updated all tests to provide the new parameters

Closes #2223

### DIFF
--- a/embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/injection/OpenTelemetryModuleImpl.kt
+++ b/embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/injection/OpenTelemetryModuleImpl.kt
@@ -50,6 +50,8 @@ class OpenTelemetryModuleImpl(
             logSink = logSink,
             sdkName = BuildConfig.LIBRARY_PACKAGE_NAME,
             sdkVersion = BuildConfig.VERSION_NAME,
+            appVersion = initModule.instrumentedConfig.project.getVersionName() ?: "UNKNOWN",
+            packageName = initModule.instrumentedConfig.project.getPackageName() ?: "UNKNOWN",
             systemInfo = initModule.systemInfo,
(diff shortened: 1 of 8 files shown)


priettt:
> LGTM. Do you know if it's still possible to override this default behavior by calling `setResourceAttribute` and whether we have a test for that?

@fractalwrench Using `Embrace.setResourceAttribute("service.name", "custom.name")`, users can override the attributes on exported spans. I verified that by modifying the `LogcatSpanExporter` on the ExampleApp, so it also prints `span.resource.attributes`. Also it's tested in OtelExportTest.

That won't change the value sent to our backend in `resource.app_ecosystem_id` though.

PR #2922, opened 2025-12-13
Auto invoke session listeners for instrumentation

bidetofevil:
## Goal

Make `InstrumentationRegistry` implement session listeners and auto register datasources that implement those interfaces.


fractalwrench (approved):
LGTM

PR #2924, opened 2025-12-15
Add internal API to add custom resource values

fractalwrench:
## Goal

Adds an internal API to add custom values to the 'resource' object in the `Envelope` type. This requires a custom Moshi serializer that delegates to the default `EnvelopeResourceJsonAdapter` implementation generated by Moshi - additional fields are appended at the end.

This is only intended for internal use and only adds values to the HTTP payload, not OTLP.

## Testing

Added tests.


bidetofevil (approved):
LGTM. I assume the additional resource attributes are immutable after the SDK starts, right?

PR #2939, opened 2025-12-17
[EMBR-9564] Feature: Included the pr-title-formatter job

ArielDemarco:
## Goal
Included the `pr-title-formatter` action to be used as part of Android development workflow

### DIFF
--- a/.github/workflows/set-pr-title.yml
+++ b/.github/workflows/set-pr-title.yml
@@ -1,0 +1,18 @@
+name: PR Title Formatter
+
+on:
+  pull_request:
+    types: [opened, edited, synchronize]
+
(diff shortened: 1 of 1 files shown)


fractalwrench (commented):
I tend to use the `style:`, `refactor:`, and `api:` prefixes reasonably often. @bidetofevil and @priettt might also have their own prefixes that they use that I've missed.

Would adding this action mean that _every_ PR mandated a ticket, or would it remain optional? I would have reservations about the overhead introduced from requiring every PR to have a ticket given that we tend to create stacks of PRs to split things up into small review (unless there's some automation that could automagically create an appropriate ticket)

PR #2940, opened 2025-12-17
Make state value type non-null and formalize attribute names

bidetofevil:
## Goal

Define and use the correct Embrace attributes when setting state attributes. Add a bunch of syntactic sugar for testing states as well. As part of that, make the type for the state values to be explicitly non-null, which I should've done earlier anyway.

### DIFF
--- a/embrace-android-instrumentation-schema/src/main/kotlin/io/embrace/android/embracesdk/internal/arch/attrs/EmbraceAttributeKeys.kt
+++ b/embrace-android-instrumentation-schema/src/main/kotlin/io/embrace/android/embracesdk/internal/arch/attrs/EmbraceAttributeKeys.kt
@@ -119,3 +119,30 @@ val embAttachmentErrorCode: EmbraceAttributeKey = EmbraceAttributeKey.create("at
  * The name of the Activity that app startup completed in
  */
 val embStartupActivityName: EmbraceAttributeKey = EmbraceAttributeKey.create("startup_activity")
+
+/**
+ * The initial value a state began with
+ */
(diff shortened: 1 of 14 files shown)


embrace-android-instrumentation-schema/src/main/kotlin/io/embrace/android/embracesdk/internal/arch/attrs/EmbraceAttributeKeys.kt
```diff
@@ -119,3 +119,13 @@ val embAttachmentErrorCode: EmbraceAttributeKey = EmbraceAttributeKey.create("at
  * The name of the Activity that app startup completed in
  */
 val embStartupActivityName: EmbraceAttributeKey = EmbraceAttributeKey.create("startup_activity")
+
+private fun createStateKey(id: String) = EmbraceAttributeKey.create("state.$id")
+
+val embStateNewValue: EmbraceAttributeKey = createStateKey("new_value")
```
fractalwrench, line 131: These are missing Kdoc like the other attributes in this file

fractalwrench (approved):
LGTM

embrace-android-instrumentation-schema/src/main/kotlin/io/embrace/android/embracesdk/internal/arch/attrs/EmbraceAttributeKeys.kt
bidetofevil: Ah yeah, will add

PR #2948, opened 2025-12-18
Add test case where no data is exported to embrace

fractalwrench:
## Goal

Adds an integration test case where data export to Embrace via our HTTP API is disabled, but OTLP export is enabled.


bidetofevil (approved):
LGTM

PR #2950, opened 2025-12-19
Add applied limits tracking to telemetry service

priettt:
## Goal
Add observability for when the SDK applies limits to telemetry data (truncations or drops). 

I will use the new methods on a following PR.

## Changes
- Added `AppliedLimitType` enum to represent limit types (TRUNCATE_ATTRIBUTES, TRUNCATE_STRING, DROP)
- Added `logAppliedLimit()` method to `TelemetryService` interface
- Implemented limit tracking in `EmbraceTelemetryService` with counter aggregation
- Limits tracked as private attributes: `emb.private.applied_limit.{telemetry_type}.{limit_type}`
- Added comprehensive test coverage for the new functionality
- Updated `FakeTelemetryService` to support the new interface method


embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/telemetry/AppliedLimitType.kt
```diff
@@ -0,0 +1,26 @@
...
+
+    /**
+     * A telemetry item was completely dropped due to rate limits or other constraints.
+     */
+    DROP;
+
+    /**
+     * Converts the enum to its attribute name format (lowercase with underscores).
+     */
+    fun toAttributeName(): String = name.lowercase()
```
fractalwrench: I'd suggest using a string literal rather than the object name due to obfuscation. Same for other enums in this changeset

embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/telemetry/TelemetryService.kt
```diff
@@ -14,6 +14,14 @@ interface TelemetryService {
      */
     fun logStorageTelemetry(storageTelemetry: Map<String, String>)
 
+    /**
+     * Tracks when a limit is applied to telemetry (truncation or drop).
+     *
+     * @param telemetryType The type of telemetry that had the limit applied
+     * @param limitType The type of limit that was applied
+     */
+    fun logAppliedLimit(telemetryType: LimitedTelemetryType, limitType: AppliedLimitType)
```
fractalwrench: Would suggest `record` or `track` rather than `log`

embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/telemetry/LimitedTelemetryType.kt
```diff
@@ -0,0 +1,26 @@
+package io.embrace.android.embracesdk.internal.telemetry
+
+/**
+ * Represents the type of telemetry that had a limit applied to it.
+ */
+enum class LimitedTelemetryType {
```
fractalwrench: Is this restricted to values from our public API right now? What's our plan for other instrumentation, and is there any way we could make this more 'automatic' so that it's not a requirement to create a new entry for each bit of telemetry?

embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/telemetry/EmbraceTelemetryService.kt
```diff
@@ -28,9 +29,18 @@ internal class EmbraceTelemetryService(
         this.storageTelemetryMap.putAll(storageTelemetry)
     }
 
+    override fun logAppliedLimit(telemetryType: LimitedTelemetryType, limitType: AppliedLimitType) {
+        val id = "applied_limit.${telemetryType.toAttributeName()}.${limitType.toAttributeName()}"
+        val key = EmbraceAttributeKey.create(id, isPrivate = true).name
+        synchronized(appliedLimitCountMap) {
```
fractalwrench: This function is a potential hot path if limits are being applied, so it'd be good to optimize its performance so that it doesn't block unnecessarily. Rather than synchronising on the map, I'd suggest `Map<String, AtomicInteger>` as that would allow you to get or create the value, then synchronise on that value.

This allows concurrent access for most calls but prevents it when the key is the same.

embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/telemetry/LimitedTelemetryType.kt
priettt: hmmm would you prefer passing a string? we could also use reflection but it won't be performant.

fractalwrench: Definitely not reflection. I suppose an enum is ok. I think it'd be good to consider the use-case of how we capture data limits being exceeded from data sources, for example.

embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/telemetry/EmbraceTelemetryService.kt
```diff
@@ -50,6 +59,16 @@ internal class EmbraceTelemetryService(
         return result
     }
 
+    private fun getAndClearAppliedLimitTelemetry(): Map<String, String> {
+        synchronized(appliedLimitCountMap) {
```
fractalwrench: `synchronized` isn't really doing much here as this seems to be the only place where the synchronisation block occurs. I'd favour the approach used in `getAndClearStorageTelemetry` instead as it copies the map without requiring a mutex.

embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/telemetry/LimitedTelemetryType.kt
priettt: For data sources, I don't see any way other than adding a field in every data source that stores the name of what's being tracked. This could either be an enum member, or even a string: 

```
 class TapDataSource(args: InstrumentationArgs) : DataSourceImpl(
      args = args,
      limitStrategy = UpToLimitStrategy(...),
      telemetryType = LimitedTelemetryType.TAP_DATASOURCE      // add this
  )
```

It's a bit ugly, but it doesn't involve reflection or accessing the class name at runtime, which could add overhead and not work on obfuscated builds

PR #2952, opened 2025-12-20
Add current state values on all logs

bidetofevil:
## Goal

Add current state values to all logs

### DIFF
--- a/embrace-android-instrumentation-api/src/main/kotlin/io/embrace/android/embracesdk/internal/arch/datasource/StateDataSource.kt
+++ b/embrace-android-instrumentation-api/src/main/kotlin/io/embrace/android/embracesdk/internal/arch/datasource/StateDataSource.kt
@@ -29,7 +29,7 @@ abstract class StateDataSource<T : Any>(
     private val unrecordedTransitions = AtomicReference(noUnrecordedTransitions)
 
     /**
-     * Notify that the state has changed at the given time to teh given value, with the given number of transitions dropped by the
+     * Notify that the state has changed at the given time to the given value, with the given number of transitions dropped by the
      * since the last time this data source was notified of a state change.
      */
     fun onStateChange(updateDetectedTimeMs: Long, newState: T, droppedTransitions: Int = 0) {


embrace-android-instrumentation-api/src/main/kotlin/io/embrace/android/embracesdk/internal/arch/datasource/StateDataSource.kt
```diff
@@ -15,17 +16,22 @@ import java.util.concurrent.atomic.AtomicReference
...
     args = args,
     limitStrategy = UpToLimitStrategy { maxTransitions }
 ) {
+    val stateAttributeKey = createStateKey(stateTypeFactory(defaultValue).stateName)
     private val currentState: AtomicReference<T> = AtomicReference(defaultValue)
     private val sessionStateToken: AtomicReference<SessionStateToken<T>?> = AtomicReference()
     private val unrecordedTransitions = AtomicReference(noUnrecordedTransitions)
 
+    /**
+     * Notify that the state has changed at the given time to teh given value, with the given number of transitions dropped by the
```
fractalwrench:
```suggestion
     * Notify that the state has changed at the given time to the given value, with the given number of transitions dropped by the
```

embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/arch/destination/TelemetryDestinationImpl.kt
```diff
@@ -99,6 +101,11 @@ class TelemetryDestinationImpl(
                     setStringAttribute(it.key, it.value)
                 }
             }
+
+            currentStatesProvider().forEach {
+                setStringAttribute(it.key.name, it.value.toString())
+            }
```
fractalwrench, line 107: I don't understand how the `foreach` could ever get executed if `currentStatesProvider` is always `emptyMap`

bidetofevil: This is set asynchronously in `SdkInitActions`. After SDK init, this should provide a key-value map of all stat values implemented in `InstrumentationRegistryImpl`

PR #2953, opened 2025-12-22
Avoid obfuscating fragment class names

fractalwrench:
## Goal

Adds a keep rule that avoids obfuscating Fragment class names but allows for shrinking of code.


bidetofevil (approved):
LGTM. I thought Fragments were like Activities in that they weren't obfuscated by default? Doesn't hurt I guess

PR #2958, opened 2025-12-22
Move config code

fractalwrench:
## Goal

Moves config code to its own module out of core.


bidetofevil (approved):
LGTM. Now that we only rely on OkHttp for networking, it makes sense to put this higher than the payload delivery layer.

PR #2961, opened 2025-12-23
Refactor how config endpoint is resolved

fractalwrench:
## Goal

Simplifies how the config endpoint and base data URL are calculated.


embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/config/source/OkHttpRemoteConfigSource.kt
```diff
@@ -72,12 +65,18 @@ internal class OkHttpRemoteConfigSource(
...
+        val headers = mutableMapOf(
+            "Accept" to "application/json",
+            "User-Agent" to "Embrace/a/" + configEndpoint.sdkVersion,
+            "Content-Type" to "application/json",
+            "X-EM-AID" to configEndpoint.appId,
+            "X-EM-DID" to configEndpoint.deviceId,
+            "Accept-Encoding" to "gzip",
+        )
+        etag?.let {
+            headers["If-None-Match"] = it
```
priettt, line 75: I think we're setting this header twice

embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/config/source/ConfigEndpoint.kt
```diff
@@ -0,0 +1,25 @@
...
+    instrumentedConfig: InstrumentedConfig,
+    val sdkVersion: String,
+    apiLevel: Int,
+) {
+
+    val appId: String = checkNotNull(instrumentedConfig.project.getAppId())
+    private val baseUrl = instrumentedConfig.baseUrls.getConfig() ?: "https://a-$appId.config.emb-api.com"
+    private val operatingSystemCode = "$apiLevel.0.0"
+
+    val url: String by lazy {
```
priettt: I'm curious about this lazy: all fields we use here should be available on construction, are string templates that expensive?

embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/injection/DeliveryModuleImpl.kt
```diff
@@ -156,11 +157,14 @@ class DeliveryModuleImpl(
             null
         } else {
             val appId = configModule.configService.appId ?: return@singleton null
-            val coreBaseUrl = configModule.urlBuilder?.baseDataUrl ?: return@singleton null
+            val coreBaseUrl = initModule.instrumentedConfig.baseUrls.getData() ?: "https://a-$appId.data.emb-api.com"
+            val endpoint = Endpoint.SESSIONS
+            val url = "$coreBaseUrl/${endpoint.version}/${endpoint.path}".split(endpoint.path).first()
```
priettt: Can't we just do this: `val url = "$coreBaseUrl/${Endpoint.SESSIONS.version}/"`

embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/config/source/ConfigEndpoint.kt
fractalwrench: Nope they're not. I'll refactor this to use `run`.

embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/injection/DeliveryModuleImpl.kt
fractalwrench: Yep, that's far better than what was there before.

PR #2966, opened 2025-12-23
Remove config module

fractalwrench:
## Goal

Removes the extra layer of the `ConfigModule` in favor of using `ConfigService` directly.

## Testing


bidetofevil (approved):
Very nice!

PR #2968, opened 2025-12-29
Track applied limits for data sources

priettt:
## Goal
Track when data source limits are exceeded. Data sources now automatically report when they drop telemetry due to rate limits.

## Changes
- Moved telemetry classes (`TelemetryService`, `AppliedLimitType`) from `embrace-android-core` to `embrace-android-infra` to make them easily available.
- Modified `DataSourceImpl` to accept optional `telemetryType` parameter
- Updated `DataSourceImpl.captureTelemetry()` to automatically track when limits are exceeded (`AppliedLimitType.DROP`)

### DIFF
--- a/embrace-android-instrumentation-api/src/main/kotlin/io/embrace/android/embracesdk/internal/arch/datasource/DataSource.kt
+++ b/embrace-android-instrumentation-api/src/main/kotlin/io/embrace/android/embracesdk/internal/arch/datasource/DataSource.kt
@@ -12,7 +12,7 @@ interface DataSource {
 
     /**
      * The name used to identify this data source for telemetry and tracking purposes.
-     * Should be in snake_case format (e.g., "tap_data_source").
+     * Should be in snake_case format and be unique (e.g., "tap_data_source").
      */
     val instrumentationName: String
 


embrace-android-instrumentation-api/src/main/kotlin/io/embrace/android/embracesdk/internal/arch/datasource/DataSourceImpl.kt
```diff
@@ -47,6 +51,11 @@ abstract class DataSourceImpl(
                 invalidInputCallback()
             } else if (limitStrategy.shouldCapture()) {
                 return destination.action()
+            } else {
+                // Track that a limit was exceeded if telemetry type was provided
+                telemetryType?.let {
+                    telemetryService.trackAppliedLimit(it, AppliedLimitType.DROP)
+                }
```
fractalwrench: An alternative approach would be to have `val instrumentationName: String` on `DataSource` and use it to track the applied limit here. That would avoid needing to maintain a separate enum hierarchy that gets passed into data sources & would probably come in use for other reasons (e.g. looking up data sources by key, rather than type)

fractalwrench (commented):
I think it'd be preferable to add a name/key attribute to the `DataSource` interface as that would guarantee the telemetry type is available, and could also be used for other purposes.

embrace-android-instrumentation-api/src/main/kotlin/io/embrace/android/embracesdk/internal/arch/datasource/DataSource.kt
```diff
@@ -10,6 +10,12 @@ package io.embrace.android.embracesdk.internal.arch.datasource
  */
 interface DataSource {
 
+    /**
+     * The name used to identify this data source for telemetry and tracking purposes.
+     * Should be in snake_case format (e.g., "tap_data_source").
```
fractalwrench:
```suggestion
     * Should be in snake_case format and be unique (e.g., "tap_data_source").
```

fractalwrench (approved):
LGTM

PR #2970, opened 2025-12-30
Create OTel-agnostic service to create OTel logs

bidetofevil:
## Goal

Create a new entry point, `EventService`, in the `embrace-android-otel` module responsible for converting all parameters into OTel and using its API to create a LogRecord. `TelemetryDestination` did that previously for logs, but like `SpanService` for Spans, it should be consolidated within that module. Note that `LogService` is not the analogue for `SpanService`, as it is closer to a DataSource for Embrace Logs rather than a general purpose creator of OTel LogRecords which is what's needed.

It's called `EventService` because OTel is moving in that direction in calling its logging API an Events API. Also we already have a LogService...

## Testing

Moved testing of capabilities previously done at `TelemetryDestination` into new tests at the lower level, `EventService`

### DIFF
--- a/embrace-android-otel/src/main/kotlin/io/embrace/android/embracesdk/internal/otel/logs/EventServiceImpl.kt
+++ b/embrace-android-otel/src/main/kotlin/io/embrace/android/embracesdk/internal/otel/logs/EventServiceImpl.kt
@@ -1,0 +1,66 @@
+package io.embrace.android.embracesdk.internal.otel.logs
+
+import io.embrace.android.embracesdk.Severity
+import io.embrace.android.embracesdk.internal.arch.schema.PrivateSpan
+import io.embrace.android.embracesdk.internal.arch.schema.SchemaType
+import io.embrace.android.embracesdk.internal.utils.Provider
+import io.embrace.android.embracesdk.internal.utils.Uuid
+import io.embrace.opentelemetry.kotlin.ExperimentalApi
+import io.embrace.opentelemetry.kotlin.logging.Logger
+import io.embrace.opentelemetry.kotlin.logging.model.SeverityNumber
+import io.embrace.opentelemetry.kotlin.semconv.IncubatingApi
+import io.embrace.opentelemetry.kotlin.semconv.LogAttributes
+import java.util.concurrent.TimeUnit
+
+@OptIn(ExperimentalApi::class, IncubatingApi::class)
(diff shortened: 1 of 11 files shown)


embrace-android-otel/src/main/kotlin/io/embrace/android/embracesdk/internal/otel/logs/EventServiceImpl.kt
```diff
@@ -0,0 +1,66 @@
...
+            }
+
+            with(schemaType) {
+                setStringAttribute(telemetryType.key.name, telemetryType.value)
+                attributes().forEach {
+                    setStringAttribute(it.key, it.value)
+                }
+            }
+
+            embraceAttributes.forEach {
```
fractalwrench, line 56: Consider a defensive copy here to avoid concurrent modification exceptions

fractalwrench (approved):
LGTM

embrace-android-otel/src/main/kotlin/io/embrace/android/embracesdk/internal/otel/logs/EventServiceImpl.kt
bidetofevil: Good call. I was assuming immutability but of course the underlyign impl doesn't have to be

bidetofevil:
### Merge activity

* **Jan 5, 4:05 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/embrace-io/embrace-android-sdk/2970).
* **Jan 5, 4:05 PM UTC**: @bidetofevil merged this pull request with [Graphite](https://app.graphite.com/github/pr/embrace-io/embrace-android-sdk/2970).

embrace-android-otel/src/main/kotlin/io/embrace/android/embracesdk/internal/otel/logs/EventServiceImpl.kt
bidetofevil: Oh crap I didn't commit my fix for this. Will do that right now

bidetofevil: Done in #2984

PR #2974, opened 2025-12-30
Set allWarningsAsErrors to true

priettt:
## Goal

Set allWarningsAsErrors = true, and add a compiler flag `("-Xsuppress-version-warnings")` to avoid the "Language version 2.0 is deprecated" warning.


priettt:
> LGTM. Do you have docs for where you found this flag?

@fractalwrench nope, some back and forth with chatgpt. 

There should be a brief description if you run `kotlinc -X`, but -X flags are experimental, and have no guarantees, so we should check again in the future in case they gives us a way to actually suppress the version error.

Apollo uses this same suppresion, and some other libraries too, so I think this should be fine for now

PR #2977, opened 2025-12-31
Only grab session properties off dead sessions for native crashes

bidetofevil:
## Goal

`Span.getSessionProperties()` was grabbing all attributes off the session span and assuming they were that session's session properties. This was pointed out in the original code review way back when but I missed it. This resulted in a bunch of junk data on native crash logs that might've been erroneously interpreted as attributes.  
  
This addresses that by making that helper function look for only session properties - it also means we don't have to stick the prefix on there, as the original filtered attributes should already have it.


bidetofevil:
### Merge activity

* **Jan 5, 4:05 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/embrace-io/embrace-android-sdk/2977).

PR #2978, opened 2026-01-02
Track applied limits for log operations

priettt:
## Goal
Enable tracking of applied limits for log operations to provide visibility into when logs are dropped, truncated, or otherwise limited by the SDK.

## Changes
  - Add `TelemetryService` parameter to `LogServiceImpl` to track applied limits
  - Track when logs are dropped due to rate limiting
  - Track when log messages exceed maximum length and are truncated
  - Track when log attributes exceed count limit
  - Track when log attribute keys or values are truncated
  - Add tests for limit tracking scenarios

### DIFF
--- a/embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/injection/LogModuleImpl.kt
+++ b/embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/injection/LogModuleImpl.kt
@@ -30,7 +30,8 @@ class LogModuleImpl(
         LogServiceImpl(
             essentialServiceModule.telemetryDestination,
             configService,
-            logLimitingService
+            logLimitingService,
+            initModule.telemetryService
         )
     }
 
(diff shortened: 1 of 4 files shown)


embrace-android-instrumentation-api/src/main/kotlin/io/embrace/android/embracesdk/internal/arch/datasource/DataSourceImpl.kt
```diff
@@ -47,6 +51,11 @@ abstract class DataSourceImpl(
                 invalidInputCallback()
             } else if (limitStrategy.shouldCapture()) {
                 return destination.action()
+            } else {
+                // Track that a limit was exceeded if telemetry type was provided
+                telemetryType?.let {
```
fractalwrench: Mentioned on the other PR, but I think that requiring `val name: String` on `DataSource` would be a better way of achieving this

embrace-android-infra/src/main/kotlin/io/embrace/android/embracesdk/internal/telemetry/LimitedTelemetryType.kt
```diff
@@ -0,0 +1,25 @@
+package io.embrace.android.embracesdk.internal.telemetry
+
+/**
+ * Represents the type of telemetry that had a limit applied to it.
+ */
+enum class LimitedTelemetryType(val attributeName: String) {
```
fractalwrench: Will we still need this enum if all the data sources are identified via other means?

PR #2979, opened 2026-01-02
Reset main thread crash handler on integration test iteration

bidetofevil:
## Goal

When initializing the JVM crash data source, we were assuming that if there was a main thread exception handler already, it was set by the app. This affected our integration tests because we just reset the SDK, so we were setting wrapping our own handler. 

So first, lets ensure that we don't double-set the main thread exception handler in production (it won't happen anyway, but just in case), and we do a reset on every integration test run.

### DIFF
--- a/embrace-android-sdk/src/integrationTest/kotlin/io/embrace/android/embracesdk/testframework/SdkIntegrationTestRule.kt
+++ b/embrace-android-sdk/src/integrationTest/kotlin/io/embrace/android/embracesdk/testframework/SdkIntegrationTestRule.kt
@@ -34,11 +34,11 @@ import io.embrace.android.embracesdk.testframework.server.FakeApiServer
 import io.embrace.opentelemetry.kotlin.ExperimentalApi
 import io.embrace.opentelemetry.kotlin.logging.export.toOtelKotlinLogRecordExporter
 import io.embrace.opentelemetry.kotlin.tracing.export.toOtelKotlinSpanExporter
-import java.io.File
 import okhttp3.Protocol
 import okhttp3.mockwebserver.MockWebServer
 import org.junit.Assert.assertEquals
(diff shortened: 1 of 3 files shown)


embrace-android-sdk/src/integrationTest/kotlin/io/embrace/android/embracesdk/testframework/SdkIntegrationTestRule.kt
```diff
@@ -201,7 +201,7 @@ internal class SdkIntegrationTestRule(
      * Setup the Embrace SDK so it's ready for testing.
      */
     override fun before() {
-
+        Thread.setDefaultUncaughtExceptionHandler(null)
```
fractalwrench, line 204: Does this interfere with Junit's default uncaught exception handler at all?

fractalwrench (approved):
LGTM

embrace-android-sdk/src/integrationTest/kotlin/io/embrace/android/embracesdk/testframework/SdkIntegrationTestRule.kt
bidetofevil: It shouldn't. Or rather, this is the default state fo the first test run anyway, and we already to this in tests where we explicitly need this.

bidetofevil:
### Merge activity

* **Jan 5, 3:54 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/embrace-io/embrace-android-sdk/2979).
* **Jan 5, 3:54 PM UTC**: @bidetofevil merged this pull request with [Graphite](https://app.graphite.com/github/pr/embrace-io/embrace-android-sdk/2979).

PR #2987, opened 2026-01-06
Remove session orchestration module

fractalwrench:
## Goal

Replaces the `SessionOrchestrationModule` implementation with a function.

## Testing


bidetofevil (approved):
LGTM. "This module could be an object" is a thing :-)

PR #2988, opened 2026-01-06
Verify app name is used as service name in resource

bidetofevil:
## Goal

Make the package name picked up during at app build overridable in tests. To thread this through, we also need to thread through the overridden instrumentation config via the fully overridden InitModule in order for that to be picked up. It should've been like that anyway.


fractalwrench (approved):
LGTM

PR #2994, opened 2026-01-07
Add OTel resource attributes to the embrace payload envelopes

bidetofevil:
## Goal

Add OTel resource attributes to the Embrace payload envelope `resource` object

### DIFF
--- a/embrace-android-sdk/src/integrationTest/kotlin/io/embrace/android/embracesdk/testcases/session/SessionPayloadTest.kt
+++ b/embrace-android-sdk/src/integrationTest/kotlin/io/embrace/android/embracesdk/testcases/session/SessionPayloadTest.kt
@@ -5,6 +5,7 @@ import io.embrace.android.embracesdk.assertions.getSessionId
 import io.embrace.android.embracesdk.internal.payload.AppFramework
 import io.embrace.android.embracesdk.testframework.SdkIntegrationTestRule
 import org.junit.Assert.assertEquals
+import org.junit.Assert.assertFalse
 import org.junit.Assert.assertTrue
 import org.junit.Rule
 import org.junit.Test
@@ -23,6 +24,7 @@ internal class SessionPayloadTest {
                 embrace.setResourceAttribute("resource-attr", "foo")
             },
             testCaseAction = {
+                embrace.setResourceAttribute("bad-resource-attr", "foo")
                 recordSession()
 
             },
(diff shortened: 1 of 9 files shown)


embrace-android-sdk/src/integrationTest/kotlin/io/embrace/android/embracesdk/testcases/session/SessionPayloadTest.kt
```diff
@@ -17,8 +17,11 @@ internal class SessionPayloadTest {
     val testRule: SdkIntegrationTestRule = SdkIntegrationTestRule()
 
     @Test
-    fun `device and app attributes are present in session envelope`() {
+    fun `device, app, and otel resource attributes are present in session envelope`() {
         testRule.runTest(
+            preSdkStartAction = {
```
fractalwrench, line 23: Can we add a test case/assertion that verifies that nothing happens if someone calls this API after embrace has initialized?

fractalwrench (approved):
LGTM assuming payload schema has been agreed with backend folks

embrace-android-sdk/src/integrationTest/kotlin/io/embrace/android/embracesdk/testcases/session/SessionPayloadTest.kt
bidetofevil: I think there's a test for that already but i'll add it here if there isn't.

bidetofevil: This is done in `OTelApiDelegateTest` but I'll add it to this test anyway

embrace-android-sdk/src/integrationTest/kotlin/io/embrace/android/embracesdk/testframework/assertions/JsonComparator.kt
```diff
@@ -30,6 +31,7 @@ internal object JsonComparator {
                 else -> "$path.$key"
             }
             if (expectedValue == IGNORE_VALUE) {
+                findMissingKeys = false
```
bidetofevil, line 34: I got the logic wrong and this breaks the missing key diffing mechanism. I'll need to fix the test some other way if we keep the current serialization form.

bidetofevil:
The `.` keys are fine. I will find a way to test this better and get this reviewed again

PR #3041, opened 2026-01-28
Upgrade to AGP 9

priettt:
## Goal

Upgrade Android Gradle Plugin to version 9.0.0 along with related dependency updates.

## Changes

- Bump AGP from 8.13.2 to 9.0.0
- Bump Dokka from 2.1.0 to 2.2.0-Beta
- Bump AndroidX Benchmark from 1.4.1 to 1.5.0-alpha01
- Set Robolectric SDK 35 globally via `GlobalConfigProvider` SPI in `embrace-test-common`
- Update Dokka config for AGP 9 source set naming (`release` instead of `main`)
- Regenerate lint baselines for AGP 9
- Update build conventions (remove explicit kotlin-android plugin)
- Update bytecode instrumentation test expectations (module name suffix change)
- Add OpenTelemetry BOM platform dependency where needed
- Update ExampleApp to AGP 9.0.0
- Fix microbenchmark tests (remove incorrect RobolectricTest inheritance)

## Considerations

**Robolectric SDK version:** AGP 9 changes the default `targetSdk` for unit tests from `minSdk` to `compileSdk`. Previously, all Robolectric tests ran against SDK 21 (our minSdk). With AGP 9, they would run against SDK 35 (compileSdk) by default.

To handle this transition, a `GlobalConfigProvider` SPI implementation in `embrace-test-common` sets SDK 35 as the default for all Robolectric tests. This replaces the previous `RobolectricTest` base class approach, removing the need for every test class to extend it. Method-level `@Config` overrides still work as expected.

**Bytecode test expectations:** AGP 9 changes how Kotlin metadata encodes module names. The `_release` suffix is no longer appended to module names in the compiled bytecode, so the expected test output files were updated to reflect this change.

**Kotlin Android plugin:** AGP 9 now automatically applies the Kotlin Android plugin when using `com.android.library` or `com.android.application`, so the explicit `kotlin("android")` declarations were removed to avoid conflicts. This applies to both the SDK modules and the ExampleApp.

**Dokka source sets:** AGP 9 renames the default Android library source set from `main` to `release`. The Dokka configuration was updated to use `configureEach` with a filter for `name == "release"` to accommodate this change.

**Lint baselines:** AGP 9 ships with a newer version of Android Lint that detects additional issues. The baseline files were regenerated to accommodate these new lint checks. The new issues are all from Robolectric test dependencies (InvalidPackage warnings), so they don't affect production code.

**OpenTelemetry BOM:** Added `platform(libs.opentelemetry.bom)` dependency in modules using OpenTelemetry. AGP 9 has stricter dependency resolution, and the BOM is now required to ensure consistent versioning across OpenTelemetry artifacts.


vitaliyf (approved):

buildSrc/src/main/kotlin/embrace-android-conventions.gradle.kts
```diff
@@ -37,6 +36,7 @@ android {
                     this.exceptionFormat = TestExceptionFormat.FULL
                 }
                 test.maxParallelForks = (Runtime.getRuntime().availableProcessors() / 3) + 1
+                test.maxHeapSize = "2g"
```
fractalwrench, line 39: Why is this change necessary?

embrace-android-sdk/src/integrationTest/kotlin/io/embrace/android/embracesdk/testcases/features/HucInstrumentationEnabledTest.kt
```diff
@@ -13,13 +15,15 @@ import org.junit.Assert.assertTrue
...
+import org.robolectric.annotation.Config
 
 @RunWith(AndroidJUnit4::class)
-internal class HucInstrumentationEnabledTest {
+internal class HucInstrumentationEnabledTest : RobolectricTest() {
     @Rule
     @JvmField
     val testRule: SdkIntegrationTestRule = SdkIntegrationTestRule()
 
+    @Config(sdk = [21])
```
fractalwrench, line 24: Why is 21 required?

embrace-android-sdk/src/integrationTest/kotlin/io/embrace/android/embracesdk/testcases/features/HucLiteInstrumentationEnabledTest.kt
```diff
@@ -15,13 +16,16 @@ import org.junit.Assert.assertTrue
...
+import kotlin.intArrayOf
 
 @RunWith(AndroidJUnit4::class)
-internal class HucLiteInstrumentationEnabledTest {
+internal class HucLiteInstrumentationEnabledTest: RobolectricTest() {
     @Rule
     @JvmField
     val testRule: SdkIntegrationTestRule = SdkIntegrationTestRule()
 
+    @Config(sdk = [21])
```
fractalwrench, line 26: Similar question on why 21 is required here

embrace-android-sdk/src/integrationTest/kotlin/io/embrace/android/embracesdk/testcases/session/BackgroundActivityDisabledTest.kt
```diff
@@ -142,6 +144,7 @@ internal class BackgroundActivityDisabledTest {
         )
     }
 
+    @Config(sdk = [21])
```
fractalwrench, line 146: Another query on why 21 is required. This API level is deprecated in Robolectric so ideally we'd move away from it, _unless_ it's required to run the test case.

embrace-test-common/src/main/kotlin/io/embrace/android/embracesdk/RobolectricTest.kt
```diff
@@ -0,0 +1,6 @@
+package io.embrace.android.embracesdk
+
+import org.robolectric.annotation.Config
+
+@Config(sdk = [35])
+abstract class RobolectricTest
```
fractalwrench: Would suggest using global configuration via a properties file or similar rather than creating an inheritance hierarchy: https://robolectric.org/configuring/#robolectricproperties-file

.github/workflows/ci-gradle.yml
```diff
@@ -93,11 +93,11 @@ jobs:
 
       - name: Run Kover Code Coverage
         if: ${{ matrix.gradle-task == 'assembleRelease' }}
-        run: ./gradlew koverXmlReportRelease
+        run: ./gradlew koverXmlReport
```
fractalwrench: Why does this task need to change? The codecov report seems inaccurate on the PR now and I wonder whether it's related.

priettt: https://github.com/Kotlin/kotlinx-kover/issues/785

kover doesn't support AGP 9 yet, I'll try to migrate everything to koverXmlReport and exclude the test folders

buildSrc/src/main/kotlin/embrace-android-conventions.gradle.kts
priettt: My guess is that using more SDK targets in Robolectric + its beta version was causing out of memory errors. This only affects tests so it shouldn't be that bad

embrace-android-sdk/src/integrationTest/kotlin/io/embrace/android/embracesdk/testcases/features/HucInstrumentationEnabledTest.kt
priettt: Some of the tests where failing when running on SDK 35 with an error with Robolectric's ShadowStatsManager, which doesn't affect SDK 21. When we stop supporting SDK 21, we should look into it.

PR #3064, opened 2026-02-04
Refactor CI workflows to invoke bash scripts

fractalwrench:
## Goal

Refactors the CI workflows so that instead of specifying scripts in the yaml, they invoke standalone scripts. This should make it much easier to test out changes locally when CI is failing as it'll be possible to execute the same logic locally without relying on Github Actions.


.github/workflows/create-release-branch.yml
```diff
@@ -21,27 +21,20 @@ jobs:
     timeout-minutes: 60
     runs-on: ubuntu-latest
     steps:
-      - name: Configure git
-        run: |
-          git config --global user.name 'embrace-ci[bot]'
-          git config --global user.email 'embrace-ci@users.noreply.github.com'
```
vitaliyf, line 27: you should keep this here since this should only run in CI and nobody should ever, even accidentally, fake who they are

.github/workflows/generate-baseline-profile.yml
```diff
@@ -17,17 +17,14 @@ jobs:
     timeout-minutes: 30
 
     steps:
-      - name: Enable KVM group perms
-        run: |
-          echo 'KERNEL=="kvm", GROUP="kvm", MODE="0666", OPTIONS+="static_node=kvm"' | sudo tee /etc/udev/rules.d/99-kvm4all.rules
-          sudo udevadm control --reload-rules
-          sudo udevadm trigger --name-match=kvm
```
vitaliyf, line 24: Would anyone ever need to run this locally? I thought this was something specific to GitHub runners

.github/workflows/generate-baseline-profile.yml
```diff
@@ -61,46 +54,23 @@ jobs:
...
           SDK_VERSION: ${{ steps.sdk_version.outputs.sdk_version }}
-        run: |
-          # Update the embrace_version in gradle.properties
-          sed -i "s/embrace_version=.*/embrace_version=${SDK_VERSION}/" gradle.properties
+        run: "$GITHUB_WORKSPACE/.github/scripts/update-benchmark-version.sh" "$SDK_VERSION"
 
       - name: Clean Managed Devices
         working-directory: android-sdk-benchmark
-        run: ./gradlew cleanManagedDevices --unused-only
+        run: "$GITHUB_WORKSPACE/.github/scripts/gradle-clean-managed-devices.sh"
```
vitaliyf, line 61: There are a couple that have $GITHUB_WORKSPACE/.github/scripts/name.sh and all others do not - you should get rid of $GITHub_WORKSPACE everywhere

.github/workflows/publish-api-docs.yml
```diff
@@ -30,21 +30,23 @@ jobs:
         env:
```
vitaliyf, line 30: are we even using this workflow? it's using deprecated publishing via branches so if we are in fact using it then we should refactor it, and not move this code into scripts

.github/scripts/choose-runner.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
```
vitaliyf, line 1: keep this in yanml file. this logic is my responsibility and i don't want to hunt for it in .sh script in one repo, and in .yaml files in 30 others. also, it doesn't make sense to run this locally.

.github/scripts/clear-ci-cache.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+set -euo pipefail
+
+# Clear all GitHub Actions cache for the repository
+
+gh cache delete --all --repo embrace-io/embrace-android-sdk
```
vitaliyf, line 6: why would someone run this locally?

.github/scripts/configure-git.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+set -euo pipefail
+
+# Configure git with embrace-ci bot credentials
+
+git config --global user.name 'embrace-ci[bot]'
+git config --global user.email 'embrace-ci@users.noreply.github.com'
```
vitaliyf, line 7: delete this file, it should not exist

.github/workflows/ci-gradle.yml
```diff
@@ -21,13 +21,16 @@ jobs:
...
+          sparse-checkout: .github/scripts
+          persist-credentials: false
+
       - id: choose
-        run: |
-          if [ ${GITHUB_ACTOR} != 'dependabot[bot]' ]; then
-              echo "where=${DEPOT_RUNNER}" >> $GITHUB_OUTPUT
-          else
-              echo "where=ubuntu-latest" >> $GITHUB_OUTPUT
-          fi
```
vitaliyf, line 30: keep this in .yaml

.github/workflows/ci-gradle.yml
```diff
@@ -21,13 +21,16 @@ jobs:
...
+        env:
+          ACTOR: ${{ github.actor }}
+        run: .github/scripts/choose-runner.sh "$ACTOR" "$DEPOT_RUNNER" >> $GITHUB_OUTPUT
```
vitaliyf, line 33: i would keep anything that has to do with GITHUB_OUTPUT variables in yaml files. otherwise, it's a pain in the ass to ever follow what it does

vitaliyf (changes requested):
For the record, I'm 👎 on this kind of structure in general because it's a nightmare to follow what each workflow does since you have to refer back and forth between multiple files. I agree it's easier to run individual parts individually, but it's hard to run the entire thing. There is https://github.com/nektos/act to run workflows locally, or I sometimes create a throw-away repo and test workflow changes in it by skipping slow steps.

I know because Unity repo is in part like this, and it's been super annoying to debug.

But ultimately, you are responsible for 90% of this logic, so you can do it the way you want to.

I left comments about remaining 10%, as someone who watches similar functionality in dozens of repos and frequently does bulk search/replace across all of them to keep them in sync :)

fractalwrench:
If you're strongly against this approach then I'd defer to you. I'd heard that extracting scripts makes it easier to debug things locally and also makes it easier to migrate to other CI providers, but if it's caused documented problems on other repos I'm happy to take your lead

PR #3210, opened 2026-04-10
Use ordinal for user session number

fractalwrench:
## Goal

Supply the user session number with a persisted ordinal.

## Testing

Added a unit test case.


bidetofevil (approved):
LGTM

PR #3211, opened 2026-04-10
Add callbacks for user sessions

fractalwrench:
## Goal

Adds the capability to register callbacks for when a user session starts. This will be used to replace the existing `endSession(clearUserInfo: Boolean)` public API. A subsequent PR will add the necessary public APIs to complete this change.

## Testing

Added unit tests.


bidetofevil (approved):
LGTM

PR #3212, opened 2026-04-10
Implement storage of user session metadata

fractalwrench:
## Goal

Implements a persistence solution for metadata around the user session so that session stitching can be maintained even if the first process dies.

## Testing

Added unit tests.


bidetofevil (approved):
LGTM

PR #3213, opened 2026-04-10
Rename public API for sessions

fractalwrench:
## Goal

Renames the public API and documentation for sessions to either refer to 'user sessions' or 'session parts'. This does require breaking changes to our interface, primarily around user session properties, and will therefore be merged into an integration branch until we're ready to ship the next major version.

## Testing


bidetofevil (approved):
Are we using the term "UserSession" in public interfaces now? I thought everything external was still being called "Session", and we'll get around this in the API by just changing the method name to like `getSeessionId()` or something just to force a breaking change?  
  
Feel free to merge (when you can) if that's not the case

fractalwrench:
Yes, that's the plan

PR #3214, opened 2026-04-10
Alter user session properties API

fractalwrench:
## Goal

Alters the user session properties API to allow specifying the scope of added properties.


bidetofevil (approved):
LGTM.   
  
Reading the description of the scopes, I think we need some really good docs and explainers when we ship this, including examples. I think some folks will have difficulty parsing these.

PR #3215, opened 2026-04-10
Load user session information on initialization

fractalwrench:
## Goal

Loads persisted metadata about the previous user session so that it can be restored between different processes.

## Testing

Added unit tests.


embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/session/UserSessionOrchestratorImpl.kt
```diff
@@ -16,25 +13,32 @@ internal class UserSessionOrchestratorImpl(
...
     override fun onNewSessionPart() {
         synchronized(lock) {
-            if (state == USER_SESSION_ACTIVE) {
-                val current = checkNotNull(metadata)
-                if (clock.now() - current.startTimeMs >= maxDurationMs) {
+            val current = state
+            if (current is UserSessionState.Active) {
+                if (isOverMaxDurationLimit(current.metadata)) {
                     terminateSession()
                     startNewSession()
```
bidetofevil: Is this and the call below `startNewUserSession()`?

bidetofevil (approved):
LGTM

embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/session/UserSessionOrchestratorImpl.kt
fractalwrench: Yes. Renamed

PR #3220, opened 2026-04-13
Combine session orchestration

fractalwrench:
## Goal

Combines `SessionPartOrchestratorImpl` and `UserSessionOrchestratorImpl` so the two components are easier to reason about, given their logic is tightly interwoven anyway.

## Testing

Relied on existing test cases.


embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/session/orchestrator/SessionOrchestratorImpl.kt
```diff
@@ -40,14 +51,30 @@ internal class SessionPartOrchestratorImpl(
...
 
     private var state = appStateTracker.getAppState()
+    private val maxDurationMs: Long = configService.sessionBehavior.getMaxSessionDurationMs()
+    private val inactivityTimeoutMs: Long = configService.sessionBehavior.getSessionInactivityTimeoutMs()
+
+    @Volatile
+    private var userSessionState: UserSessionState = UserSessionState.Initializing
 
     init {
+        loadPersistedUserSession()
```
bidetofevil, line 61: Can the load ever fail?

bidetofevil (approved):
LGTM. One question I had is what if the user session init when the metadata loading fails, will the error cascade up or will the metadata layer pass throw the error/exception?

embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/session/orchestrator/SessionOrchestratorImpl.kt
fractalwrench: I'll add a fallback so that `NoActiveSession` is set as the default in the scenario where this does fail.

PR #3221, opened 2026-04-13
Alter config fields for user session

fractalwrench:
## Goal

Alters the config fields that control user session as the requirements changed.


bidetofevil (approved):
LGTM

PR #3222, opened 2026-04-13
Alter how user session config applied

fractalwrench:
## Goal

Ensures that the max duration + inactivity timeout for user sessions is persisted and _used_ when restoring a user session, rather than the value available from the `ConfigService`.


bidetofevil (approved):
LGTM

PR #3223, opened 2026-04-13
Detect anomalies in clock for persisted user session

fractalwrench:
## Goal

Detects a backwards shift in time for persisted user session. If this occurs, log an internal error and immediately start a new user session.

## Testing

Added unit tests.


bidetofevil (approved):
LGTM

PR #3231, opened 2026-04-14
Fix CI workflow

fractalwrench:
## Goal

Fixes the CI workflow as it wasn't running the gradle plugin integration tests. The step is present in CI but a refactor removed 'test' from the matrix, meaning it never ran.

I spotted this when the gradle plugin integration tests started failing locally due to `androidx.navigation` complaining about needing AGP 8.1.1 or greater. Consequently I've downgraded this to 2.8.X as that version is not affected.


vitaliyf (approved):

bidetofevil (approved):
LGTM

PR #3232, opened 2026-04-14
Prevent incorrect build id generation in gradle plugin

fractalwrench:
## Goal

The gradle plugin uses a build ID to disambiguate mapping file requests. This was generated in a data class as a default parameter which means if configuration cache was used this would not be regenerated. In practice this should happen rarely on CI as configuration cache tends to be disabled, but altering the implementation to avoid this using a [ValueSource](https://docs.gradle.org/current/kotlin-dsl/gradle/org.gradle.api.provider/-value-source/index.html) seems like a good idea.

## Testing

Added a unit test & relied on existing gradle plugin test cases for regression testing.


bidetofevil (approved):
LGTM.   
  
Couple of clarification questions: does this break any of the existing caches since this attribute is different every time, or is this effectively bypassed when doing caching staleness calculations?  
  
Do we have a test to verify that the integrity of the cache so if this were to break it, it woudl be caught?

bidetofevil (approved):

fractalwrench:
>Couple of clarification questions: does this break any of the existing caches since this attribute is different every time

We want to invalidate cached inputs (for our tasks only) each build as that guarantees the mapping file/build ID will be distinct, and that any changes in the mapping file are picked up.

>Do we have a test to verify that the integrity of the cache so if this were to break it, it would be caught?

`BuildIdValueSource` is the closest test. IMO it would be fairly tricky to write a test that checks the integrity of the cache as it would require running a build at least twice with configuration cache enabled.

PR #3234, opened 2026-04-14
Alter transition logic for user session

fractalwrench:
## Goal

Alters the transition logic for the user session so that it's always ready before a new session part.


bidetofevil (approved):
LGTM

PR #3235, opened 2026-04-14
Support inactivity timeout for user sessions

fractalwrench:
## Goal

Support inactivity timeout for user sessions. In summary, once the process has been in the background for X ms, the background session part should be split and a new user session created. On reentering the foreground a new user session must always be created.

## Testing

Added unit tests.


embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/session/orchestrator/SessionOrchestratorImpl.kt
```diff
@@ -106,9 +111,16 @@ internal class SessionOrchestratorImpl(
     }
 
     override fun onForeground() {
+        inactivityTimerState?.cancel()
+        val exceeded = inactivityTimerState?.exceeded ?: false
```
bidetofevil: Does this need to hold the lock so that it won't race with the background thread terminating the session due to inactivity? You can probably hold a reference to the value of `inactivityTimerState` before nulling it out, then just evaluate this within `transitionState()`. Or something.

fractalwrench: I've added a lock around this code.

bidetofevil (approved):
LGTM

PR #3236, opened 2026-04-14
Add callbacks for user sessions

fractalwrench:
## Goal

Adds an API for registering callbacks for user sessions. This allows a library consumer to register a callback and react to changes in the session lifecycle. These states are currently available:

1. Creation of an active user session
2. End of a user session
3. A user session restoring itself in a new process

This attempts to fill parity with the existing functionality on iOS. This is also required because we plan on deprecating `endSession(clearUserInfo: Boolean)` and asking users who want to clear user data to do so in a callback.

## Testing

Added unit tests.

### DIFF
--- a/embrace-android-api/src/main/kotlin/io/embrace/android/embracesdk/SessionStateEvent.kt
+++ b/embrace-android-api/src/main/kotlin/io/embrace/android/embracesdk/SessionStateEvent.kt
@@ -17,6 +17,11 @@ public sealed class SessionStateEvent(
      * 1. When a new user session starts
      * 2. When an existing user session is restored in a new process
      * 3. When a listener is registered, if a user session is already active
+     *
+     * It's important to note that the same userSessionId value may be present in multiple, distinct events of
+     * [io.embrace.android.embracesdk.SessionStateEvent.UserSessionActive]. This API is primarily intended so that
+     * library consumers can perform operations when a user session is brought into memory, such as setting
+     * a user ID or user session properties.
      */
     public class UserSessionActive(userSessionId: String) : SessionStateEvent(userSessionId)
 
--- a/embrace-android-api/src/main/kotlin/io/embrace/android/embracesdk/SessionStateEvent.kt
+++ b/embrace-android-api/src/main/kotlin/io/embrace/android/embracesdk/SessionStateEvent.kt
@@ -1,0 +1,27 @@
+package io.embrace.android.embracesdk
+
+/**
+ * Represents an event in the user session lifecycle.
+ */
+public sealed class SessionStateEvent(
+
+    /**
+     * ID representing the user session.
+     */
+    public val userSessionId: String
+) {
+
+    /**
+     * A user session is active. This event is fired in the following conditions:
+     *
+     * 1. When a new user session starts
+     * 2. When an existing user session is restored in a new process
+     * 3. When a listener is registered, if a user session is already active
+     */
+    public class UserSessionActive(userSessionId: String) : SessionStateEvent(userSessionId)
+
+    /**
+     * The current user session has ended.
+     */
+    public class UserSessionEnded(userSessionId: String) : SessionStateEvent(userSessionId)
+}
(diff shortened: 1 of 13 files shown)


embrace-android-api/src/main/kotlin/io/embrace/android/embracesdk/SessionStateEvent.kt
```diff
@@ -0,0 +1,23 @@
+package io.embrace.android.embracesdk
+
+/**
+ * Represents an event in the user session lifecycle.
+ */
+public enum class SessionStateEvent {
```
bidetofevil: It might be good to include an ID here so folks can know which session ended/started/restored?

embrace-android-api/src/main/kotlin/io/embrace/android/embracesdk/SessionStateEvent.kt
```diff
@@ -0,0 +1,23 @@
...
+
+    /**
+     * The current user session has ended.
+     */
+    USER_SESSION_ENDED,
+
+    /**
+     * An existing user session, started in a previous process, has been restored.
+     */
+    USER_SESSION_RESTORED,
```
bidetofevil: Your comment on slack is right - this is more like a subset of of USER_SESSION_ACTIVE. Perhaps a boolean that says whether it was restored would be good info to include?

embrace-android-api/src/main/kotlin/io/embrace/android/embracesdk/SessionStateEvent.kt
```diff
@@ -0,0 +1,23 @@
+package io.embrace.android.embracesdk
+
+/**
+ * Represents an event in the user session lifecycle.
+ */
+public enum class SessionStateEvent {
+
+    /**
+     * A user session is active. Fired when a new session starts, and immediately when a listener
```
bidetofevil: The description will need amending if we are using it to denote when an existing user session becomes active after a new app launch as well.

embrace-android-api/src/main/kotlin/io/embrace/android/embracesdk/SessionStateEvent.kt
fractalwrench: I'll add the user session ID.

embrace-android-api/src/main/kotlin/io/embrace/android/embracesdk/SessionStateEvent.kt
fractalwrench: So my current thinking is that it doesn't really make sense to disambiguate these events, as the intended use-case for this API is simply to set values at the start/end of when a user session is created for a given process. As such, I think it makes sense to invoke this when the listener is first set, when a new user session is created, and when a process restores an existing user session.

If folks can think of other use-cases being useful then I suppose it would be possible to create subclasses of `UserSessionActive` now that we're using a sealed class hierarchy?

bidetofevil: This seems ok. But we should make clear in the API docs that the session value fired may not be unique, e.g. when a user session is restored, the same ID will be passed along that event. So if customers are doing bookkeeping on their side about how many sessions are created, they need to be aware of this.

embrace-android-api/src/main/kotlin/io/embrace/android/embracesdk/SessionStateEvent.kt
```diff
@@ -0,0 +1,27 @@
+package io.embrace.android.embracesdk
+
+/**
+ * Represents an event in the user session lifecycle.
+ */
+public sealed class SessionStateEvent(
```
bidetofevil, line 6: Are instance of these events equal if they have the same concrete class and the same userSessionId?  If so, we need to implement isEquals and hashCode. But there is an argument to be made that they are NOT equal, given starting a new session and restoring it are two different events with the same sessionId, though right now at least, they shouldn't ever be in the same app instance. So..... :shrug:  
  
Either way, we should state the equality criteria in the comments, whichever way we choose.

bidetofevil (approved):
LGTM. Non-blocking comment about equality that may require a code change, depending which side of a coin you land on.

PR #3237, opened 2026-04-14
Remove clearUserInfo parameter

fractalwrench:
## Goal

Removes the `clearUserInfo` parameter. The preferred API for v9 will be the user session callbacks added in #3236.


bidetofevil (commented):
The old callback is blocking while the one won't be. If we don't depend on this ourselves, it'll just be a change we need to communicate.  
  
But is this something we can do before 9.0 as it is technically a breaking change?

fractalwrench:
@bidetofevil this change is being targeted to an integration branch so won't ship until v9.

I don't understand the comment about the 'old callback', can you clarify? We don't depend on the `clearUserInfo` parameter ourselves and I've checked that the original rationale to build it was a workaround, so it feels like this API is a better fit...

bidetofevil:
> The old callback is blocking while the one won't be. If we don't depend on this ourselves, it'll just be a change we need to communicate.
>
> But is this something we can do before 9.0 as it is technically a breaking change?

​Ignore this comment. This was made when I thought this was meant to replace the session orchestrator transition boundary callbacks. I should've deleted it lol.

bidetofevil (approved):
LGTM. Ignore my other comment.

PR #3238, opened 2026-04-14
Harmonise start timestamp for session part and user session

fractalwrench:
## Goal

Harmonises the start timestamp used for the session part and user session.

## Testing

Added unit tests.


bidetofevil (approved):
Ah, perfect. LGTM.

PR #3242, opened 2026-04-15
Tweak inactivity timeout minimum for user sessions

fractalwrench:
## Goal

Tweaks the minimum inactivity timeout for user sessions, as this should now be 30s.


bidetofevil (approved):
LGTM

PR #3243, opened 2026-04-15
Support max duration timer

fractalwrench:
## Goal

Adds a max duration timer for foreground user sessions that terminates the active user session and starts a new one if the threshold is exceeded.

## Testing

Added unit tests.

### DIFF
--- a/embrace-android-core/config/detekt/baseline.xml
+++ b/embrace-android-core/config/detekt/baseline.xml
@@ -1,5 +1,7 @@
 <?xml version='1.0' encoding='UTF-8'?>
 <SmellBaseline>
   <ManuallySuppressedIssues/>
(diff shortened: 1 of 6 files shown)


bidetofevil (approved):
LGTM.   
  
Non-blocking question: are there sequencing of how the two timeouts are fired such that they can interact (epecially if you throw foregrounding/backgrounding into the mix) where one timer kills the session, and the other timer, detecting the condition that it shouldn't kill the session, skips it but doesn't reschedule the timer? Or will the ending by the first time cause a reschedule which makes how the second timer behaves irrelevant?

fractalwrench:
I can't spot any problems. The actual Runnable for both timeouts invokes functions that synchronise on the same lock.

PR #3244, opened 2026-04-15
Support tracking session part number correctly

fractalwrench:
## Goal

Track the session part number correctly and persist it so that the information is available between processes.

## Testing

Added unit tests.


bidetofevil (approved):
LGTM

PR #3247, opened 2026-04-16
Set emb.is_final_session_part attribute

fractalwrench:
## Goal

Sets the `emb.is_final_session_part` attribute to 1 if the session part is known to be the last in the user session. Other session parts don't need to set the value.

## Testing

Added unit tests.


bidetofevil (approved):
LGTM

PR #3248, opened 2026-04-16
Add termination reason to session part

fractalwrench:
## Goal

Adds the termination reason to session part attributes.

## Testing

Added unit tests.


bidetofevil (approved):
LGTM

PR #3249, opened 2026-04-16
Rename symbols to session part

fractalwrench:
## Goal

Renames various symbols from session to session part.


bidetofevil (approved):
LGTM

PR #3250, opened 2026-04-16
Supply attributes to session span

fractalwrench:
## Goal

Supplies several user session attributes to the session span. This includes:

- `emb.user_session_start_ts`
- `emb.user_session_id`
- `emb.session_part_id`
- `emb.user_session_number`
- `emb.user_session_max_duration_seconds`
- `emb.user_session_inactivity_timeout_seconds`
- `emb.user_session_part_number`
- `session.id`

## Testing

Added unit tests.

### DIFF
--- a/embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/session/orchestrator/SessionPartSpanAttrPopulatorImpl.kt
+++ b/embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/session/orchestrator/SessionPartSpanAttrPopulatorImpl.kt
@@ -17,15 +17,15 @@ internal class SessionPartSpanAttrPopulatorImpl(
     private val metadataService: MetadataService,
 ) : SessionPartSpanAttrPopulator {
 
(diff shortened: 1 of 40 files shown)


embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/session/orchestrator/SessionPartSpanAttrPopulatorImpl.kt
```diff
@@ -16,7 +17,7 @@ internal class SessionPartSpanAttrPopulatorImpl(
     private val metadataService: MetadataService,
 ) : SessionPartSpanAttrPopulator {
 
-    override fun populateSessionSpanStartAttrs(session: SessionPartToken) {
+    override fun populateSessionSpanStartAttrs(session: SessionPartToken, userSession: UserSessionMetadata) {
```
bidetofevil, line 20: we may want to rename this variable to part because reading the body of the function it confused me a bit :sweat_smile:

bidetofevil (approved):
LGTM. One nitty non-blocking comment.

PR #3251, opened 2026-04-16
Create SessionIdProvider

fractalwrench:
## Goal

Create `SessionIdProvider` which is responsible for providing the user session ID + session part ID without exposing unnecessary details to other components.


bidetofevil (approved):
LGTM

PR #3252, opened 2026-04-16
Delete telemetry older than 7 days

fractalwrench:
## Goal

Deletes telemetry that is older than 7 days. This is intended to avoid the scenario where in-order delivery gets indefinitely stuck for a given payload type.

## Testing

Added unit tests.

### DIFF
--- a/embrace-android-telemetry-persistence/src/main/kotlin/io/embrace/android/embracesdk/internal/delivery/storage/FileStorageServiceImpl.kt
+++ b/embrace-android-telemetry-persistence/src/main/kotlin/io/embrace/android/embracesdk/internal/delivery/storage/FileStorageServiceImpl.kt
@@ -1,5 +1,6 @@
 package io.embrace.android.embracesdk.internal.delivery.storage
 
+import io.embrace.android.embracesdk.internal.clock.Clock
 import io.embrace.android.embracesdk.internal.delivery.StoredTelemetryMetadata
 import io.embrace.android.embracesdk.internal.logging.InternalErrorType
 import io.embrace.android.embracesdk.internal.logging.InternalLogger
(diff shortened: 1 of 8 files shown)


embrace-android-telemetry-persistence/src/main/kotlin/io/embrace/android/embracesdk/internal/delivery/storage/FileStorageServiceImpl.kt
```diff
@@ -120,7 +145,7 @@ class FileStorageServiceImpl(
         logger.trackInternalError(InternalErrorType.PAYLOAD_STORAGE_FAIL, RuntimeException("Pruned payload storage"))
```
bidetofevil, line 145: Do we want to log a different error (not really an error, but...) for cases where we prune due to timing reasons vs when the cache is just getting too full? I think the former will happen periodically just because apps and devices go offline and don't come back for a while, but the latter does seem like a pretty rare case where if it goes up dramatically, it might indicate an issue with the delivery layer.

bidetofevil (approved):
One comment about logging a different internal error if we truncate by time vs capacity, but LGTM otherwise

embrace-android-telemetry-persistence/src/main/kotlin/io/embrace/android/embracesdk/internal/delivery/storage/FileStorageServiceImpl.kt
fractalwrench: I don't want to log internal errors for that because I don't think it's actionable data and it's likely to be noisy (even if it's a low frequency, we're dealing with large numbers of data). Fredric does have plans to add an 'audit trail' for when data is dropped that feels like a better fit

PR #3254, opened 2026-04-16
Set session IDs appropriately on native crashes

fractalwrench:
## Goal

Stores and loads the user session + session part IDs on native crashes.

## Testing

Added unit tests.


bidetofevil (approved):
I wonder if it's slightly risky storing two different identifying attributes in case one gets corrupted or goes missing because of a race condition. I think I'm more conscientious about this because it's on the native layer...

Or is it better to have that risk instead of adding complexity to store this as a compound userSessionId-partId tuple? You would never have one without the other, I imagine, so they do kind of belong together.

Anyway, I'll approve this if this is the direction you want to go. I can see the merits of both, but just wanted to bring it up.

fractalwrench:
I favour storing them as separate fields as it's easier to serialize/deserialize. I feel that encoding strings is going to introduce complexity to the C layer I'd rather avoid.

PR #3255, opened 2026-04-16
Attempt to fix flaky AeiFeatureTest

fractalwrench:
## Goal

Attempts to fix the flaky `AeiFeatureTest` by setting the timestamps of AEI objects so that they always exceed or are equal to the `FakeClock`.


bidetofevil (approved):
LGTM

PR #3259, opened 2026-04-17
Rate limit manual end of user session

fractalwrench:
## Goal

Rate limits the manual ending of user sessions via the user session start timestamp, not the session part timestamp.

## Testing

Added a unit test.


bidetofevil (approved):
LGTM

PR #3260, opened 2026-04-17
Persist inactivity timeout for user sessions

fractalwrench:
## Goal

Persists the inactivity timeout for user sessions so that it's handled correctly when a user session spans multiple processes.

## Testing

Added unit tests.

### DIFF
--- a/embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/session/UserSessionMetadataStore.kt
+++ b/embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/session/UserSessionMetadataStore.kt
@@ -10,6 +10,7 @@ internal class UserSessionMetadataStore(private val store: KeyValueStore) {
 
     private companion object {
         const val KEY_SESSION = "embrace.user_session"
+        const val KEY_LAST_ACTIVITY_TS = "emb.user_session_last_activity_ts"
     }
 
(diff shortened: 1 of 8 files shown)


embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/session/UserSessionMetadataStore.kt
```diff
@@ -10,6 +10,7 @@ internal class UserSessionMetadataStore(private val store: KeyValueStore) {
 
     private companion object {
         const val KEY_SESSION = "embrace.user_session"
+        const val KEY_LAST_ACTIVITY_TS = "emb.user_session_last_activity_ts"
```
bidetofevil, line 13: Can we just build this right into the new semconv or is the session stuff not moved over yet?

bidetofevil (approved):
LGTM

embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/session/UserSessionMetadataStore.kt
fractalwrench: This is an internal implementation detail (a key used for persistence) so doesn't belong in semconv IMO

PR #3261, opened 2026-04-17
Return correct ID in SdkStateApiDelegate

fractalwrench:
## Goal

Returns the correct ID in `SdkStateApiDelegate`. The user session ID should be returned rather than the session part ID.

## Testing

Updated unit tests.


bidetofevil (approved):
LGTM. This is going into the feature branch?

PR #3262, opened 2026-04-17
Alter syntax for timers

fractalwrench:
## Goal

Tweaks the syntax for checking whether the max duration/inactivity timeout are reached.

### DIFF
--- a/embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/session/UserSessionMetadata.kt
+++ b/embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/session/UserSessionMetadata.kt
@@ -1,5 +1,6 @@
 package io.embrace.android.embracesdk.internal.session
 
+import io.embrace.android.embracesdk.internal.clock.Clock
 import io.embrace.android.embracesdk.semconv.EmbSessionAttributes
 import io.opentelemetry.kotlin.semconv.SessionAttributes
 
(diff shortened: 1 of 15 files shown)


embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/session/UserSessionMetadata.kt
```diff
@@ -15,6 +16,12 @@ data class UserSessionMetadata(
     val partNumber: Int,
     val lastActivityMs: Long,
 ) {
+    fun isOverMaxDuration(clock: Clock): Boolean =
+        clock.now() - startTimeMs >= maxDurationSecs * 1_000L
```
bidetofevil: Technically this is equals to or over, but semantically meeting the max is when we should kill the session already.  Maybe I'm being too pedantic so feel free to ignore, but should the method chains be about that, e.g. shouldEndUserSession() or hasUserSessionReachedMax() or something like that?

bidetofevil (approved):
LGTM. Some pendantioc thoughts that you could ignore about the naming of the methods that says the user session is too long

embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/session/UserSessionMetadata.kt
fractalwrench: This feels like an off-by-one error so I've updated the implementation.

PR #3263, opened 2026-04-17
Refactor inactivity timeout

fractalwrench:
## Goal

Refactors the inactivity timeout so that it doesn't rely on a `boolean` flag.


bidetofevil (approved):
Very nice. LGTM

PR #3264, opened 2026-04-17
Update semantic conventions

fractalwrench:
## Goal

Removes `emb.is_final_session ` (renamed to `emb. is_final_session_part`) and renames `emb.user_session_termination_reason`. Both of these are due to spec changes.


bidetofevil (approved):
LGTM

PR #3265, opened 2026-04-17
Enforce manual session end requirements correctly

fractalwrench:
## Goal

Retain a timestamp of the last manual session end so that it can never be >=5s between manual session ends.


bidetofevil (approved):
LGTM

PR #3267, opened 2026-04-17
Handle backwards time shift

fractalwrench:
## Goal

Handles a backwards shift in time. Although this is unlikely with our `Clock` implementation, the agreed way of handlin handling this use-case is by logging an internal error and immediately starting a new user session.

## Testing

Added a unit test.


bidetofevil (approved):
Nice. LGTM

PR #3269, opened 2026-04-17
Thread through correct session ordinals

fractalwrench:
## Goal

Thread through the correct values for the user session + session part numbers.


bidetofevil (approved):
LGTM

PR #3270, opened 2026-04-17
Wire through session ids

fractalwrench:
## Goal

Wires through the user session/session part IDs to the required places in the SDK.

## Testing

Updated unit tests.


bidetofevil (approved):
LGTM

PR #3274, opened 2026-04-20
Prevent setting part id without a user session id

fractalwrench:
## Goal

Prevents setting the Session Part ID if the User Session ID is not set.

### DIFF
--- a/.github/workflows/android-emulator-tests.yml
+++ b/.github/workflows/android-emulator-tests.yml
@@ -48,7 +48,7 @@ jobs:
 
       # Cache the emulator
       - name: AVD cache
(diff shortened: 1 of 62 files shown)


bidetofevil:
The code seems to be asserting the opposite, i.e. a sessionId can only be set if there is a partId? Or is there something I'm missing?

fractalwrench:
@bidetofevil the expected behavior is that if there's a session part ID, the user session ID must always be set.

bidetofevil:
Oh I guess this isn't asserting anything, but rather expecting that there is ever only a sessionId IF there is a sessionPartId. Sorry, had to get my head around that first.

bidetofevil (approved):
LGTM

PR #3275, opened 2026-04-20
Wire in correct session ids for native crashes

fractalwrench:
## Goal

Correctly wires in the user session ID + session part ID for native crashes.


bidetofevil (approved):
LGTM

PR #3281, opened 2026-04-21
Always set session id attributes even when empty

fractalwrench:
## Goal

A change in the backend's requirements mean that session ID attributes should be populated with an empty string when no user session is active.

## Testing

Updated existing tests.


bidetofevil (approved):
LGTM.

PR #3283, opened 2026-04-21
Add initial user session integration tests

fractalwrench:
## Goal

Adds initial integration tests for the user session functionality.


bidetofevil (approved):
LGTM

PR #3284, opened 2026-04-21
Add integration tests for user session ID propagation

fractalwrench:
## Goal

Adds integration test that assert user session ID + session part IDs are propagated correctly.


bidetofevil (approved):
LGTM

PR #3295, opened 2026-04-22
Add integration tests for user session lifecycle

fractalwrench:
## Goal

Adds some initial integration tests for the user session lifecycle.


bidetofevil (approved):
LGTM

PR #3296, opened 2026-04-22
Add integration tests for session sequence number

fractalwrench:
## Goal

Adds integration tests that check the sequence of the session part number is correct between user sessions, and corrects the existing implementation that made it a continuously incrementing number.


bidetofevil (approved):
LGTM

PR #3305, opened 2026-04-23
Create UuidSource class

fractalwrench:
## Goal

Creates a `UuidSource` class that is injected into classes that require it rather than relying on the static `Uuid.getEmbUuid()`. This allows greater testability as `Random` can be supplied to ensure deterministic IDs.


bidetofevil (approved):
LGTM

PR #3306, opened 2026-04-23
Add integration tests for user session

fractalwrench:
## Goal

Adds integration tests for the user session that compare a known golden file output.


bidetofevil (approved):
LGTM

PR #3307, opened 2026-04-23
Add golden file tests for user session id association

fractalwrench:
## Goal

Adds golden file tests that check the user session ID and session part ID allow correct association of payloads in various scenarios.


bidetofevil (approved):
LGTM

PR #3317, opened 2026-04-24
Additional basic session golden file tests

fractalwrench:
## Goal

Adds some extra user session golden file tests that assert sessions can have multiple parts and the max duration/inactivity timeout are serialized as attributes.


bidetofevil (approved):
LGTM

PR #3318, opened 2026-04-24
Add integration test that asserts user session ends after max duration

fractalwrench:
## Goal

Adds an integration test that asserts a user sessions that has exceeded the maximum duration will create a new user session.


bidetofevil (approved):
LGTM

PR #3319, opened 2026-04-24
Fix ndk tests

fractalwrench:
## Goal

The instrumented tests were failing to compile on the integration branch CI. This fixes them and adds the new user session ID attribute.


bidetofevil (approved):
LGTM

PR #3320, opened 2026-04-24
Add manual user session end integration test

fractalwrench:
## Goal

Adds an integration tests that asserts the behavior when a user session is manually ended.


bidetofevil (approved):
LGTM

PR #3321, opened 2026-04-24
Add golden file test for inactivity timeout with background sessions

fractalwrench:
## Goal

Adds a golden file test for inactivity timeout with background sessions enabled.


bidetofevil (approved):
Nice. LGTM

PR #3322, opened 2026-04-24
Create a golden file test for a log from the background

fractalwrench:
## Goal

Create a golden file test for a log from the background without any user session. This fixes a bug where the initial session part created a short-lived user session.

### DIFF
--- a/embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/session/orchestrator/SessionPartSpanAttrPopulatorImpl.kt
+++ b/embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/session/orchestrator/SessionPartSpanAttrPopulatorImpl.kt
@@ -8,6 +8,7 @@ import io.embrace.android.embracesdk.internal.session.LifeEventType
 import io.embrace.android.embracesdk.internal.session.SessionPartToken
 import io.embrace.android.embracesdk.internal.session.UserSessionMetadata
 import io.embrace.android.embracesdk.semconv.EmbSessionAttributes
+import io.opentelemetry.kotlin.semconv.SessionAttributes
 import java.util.Locale
 
 internal class SessionPartSpanAttrPopulatorImpl(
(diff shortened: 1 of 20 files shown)


embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/session/orchestrator/SessionPartSpanAttrPopulatorImpl.kt
```diff
@@ -29,12 +29,17 @@ internal class SessionPartSpanAttrPopulatorImpl(
                 addSessionPartAttribute(EmbSessionAttributes.EMB_SESSION_START_TYPE, it)
             }
 
-            userSession.attributes.forEach { (key, value) ->
-                addSessionPartAttribute(key, value.toString())
+            if (userSession != null) {
+                addSessionPartAttribute(EmbSessionAttributes.EMB_SESSION_PART_ID, sessionPart.sessionPartId)
+                addSessionPartAttribute(EmbSessionAttributes.EMB_USER_SESSION_PART_NUMBER, sessionPart.sessionPartNumber.toString())
+                userSession.attributes.forEach { (key, value) ->
```
bidetofevil, line 35: Don't we need to add appropraite properties even if we don't have user sesion?

fractalwrench: That's covered by the `else` branch. If no user session is active, the only required user session attributes are the session part ID, user session ID, and `session.id`, which are all set as empty strings. The other attributes supplied from `UserSessionMetadata` such as part number don't make sense in that context.

bidetofevil: Got it. Missed that it was specifically userSession attributes that were being added

bidetofevil (approved):
LGTM

PR #3342, opened 2026-04-29
fix console log when no app ID is set

fnewberg:
## Goal

If not app ID was set, the startup console log emitted was

```
Embrace SDK version 8.4.0-SNAPSHOT startednull
```

This change emits it as

```
Embrace SDK version 8.4.0-SNAPSHOT started without an app ID
```


bidetofevil (approved):
LGTM

PR #3350, opened 2026-05-05
Create and use interface to atomically get congruent pairs of userSession/sessionPart IDs

bidetofevil:
When session and part transitions are happening, there could be telemetry being logged in different threads. When that happens, there could be raise conditions such that user session IDs are out of sync with the session part IDs because they are separate, unsynchronized reads.

This PR spins of that functionality into a new interface so that these ID pairs can be obtained atomically. This PR just adds this interface and changes some tests and infra around - the consumption will happen in subsequent PRs.


embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/injection/EssentialServiceModule.kt
```diff
@@ -20,6 +21,7 @@ interface EssentialServiceModule {
     val networkConnectivityService: NetworkConnectivityService
     val sessionPartTracker: SessionPartTracker
     val sessionIdProvider: SessionIdProvider
+    val activeSessionIdsProvider: ActiveSessionIdsProvider
```
fractalwrench: These concepts should be combined IMO if we're going down this route (or at least `sessionIdProvider` and `activeSessionIdsProvider`)

embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/injection/UserSessionOrchestrationModule.kt
```diff
@@ -9,4 +10,5 @@ import io.embrace.android.embracesdk.internal.session.orchestrator.SessionOrches
 interface UserSessionOrchestrationModule {
     val sessionOrchestrator: SessionOrchestrator
     val sessionIdProvider: SessionIdProvider
+    val activeSessionIdsProvider: ActiveSessionIdsProvider
```
fractalwrench: Can be combined

embrace-android-sdk/src/integrationTest/kotlin/io/embrace/android/embracesdk/testcases/features/UserSessionPropertiesTest.kt
```diff
@@ -98,33 +99,19 @@ internal class UserSessionPropertiesTest {
         )
     }
 
+    @Ignore("user session scoped properties don't persist across app instances")
```
fractalwrench: These changes don't seem relevant to the rest of the PR

fractalwrench (commented):
This feels like the right approach. The changeset seems to have some unrelated test code changes bundled in, and IMO it would also be worth combining the `SessionIdProvider` and `ActiveSessionIdsProvider` interfaces as part of this.

embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/injection/EssentialServiceModule.kt
bidetofevil: I initially had them combined, but the caller then has to decided whether they should use the non-atomic calls or the atomic call depending on the use case. 

In fact ripping them apart is most responsible for blowing up the size of the PR, so let me combine them again so the PR is easier to review, and we can discuss further and rip them up again if we decide that is a cleaner approach.

embrace-android-sdk/src/integrationTest/kotlin/io/embrace/android/embracesdk/testcases/features/UserSessionPropertiesTest.kt
bidetofevil: Yeah I put them on top of the test PR in case there were conflicting files and the rebase/merge didn't go well. But because they don't really, I can separate the two

fractalwrench (approved):
LGTM

PR #3358, opened 2026-05-06
Use threadsafe UuidSource that generates UUIDs deterministically

bidetofevil:
Create a threadsafe UuidSource to use in tests so that we can write gold files with specific UUIDs


fractalwrench (approved):
I don't think this necessarily addresses the problem behavior you were seeing as fundamentally there can still be a race between two threads to determine which one gets a UUID in a given sequence. However, I think this is a good change overall for readability and would be happy for it to be merged if the `@Synchronized` was removed from `TestUuidSource`.

Addressing the flake will be a case of either faking a specific worker in the test (likely the periodic cacher) so that its execution is predictable, or loosening the test assertion.

bidetofevil:
Ugh, it's the ordering of the UUID-generation calls and not the `random.nextLong()` calls, eh? OK. Attempt number #2 forthcoming.

bidetofevil:
Not sure I like loosening the assertion as that partially reduces the effectiveness of having a golden file test like this. Let me take a look to see if there's a tweak without adding additional complexity....

bidetofevil:
Changed the golden file test assertions and removed the synchronization in the UUID generation method in the test UuidSource

fractalwrench (commented):
This approach looks ok but I think `__EMBRACE_TEST_IGNORE__` needs to be removed from the golden file fixtures in favour of what was there before (or new UUIDs if the ID generated was different). There's a lot of implicit test coverage (e.g. asserting the UUIDs match for log/session envelopes associated with the same user session)

bidetofevil:
Ok. Let me modify the golden file assertion methods to take in the expected ID that we retrieve dynamically.

fractalwrench (approved):
LGTM

PR #3360, opened 2026-05-06
Allow manual ending of user sessions in background

fractalwrench:
## Goal

Allows the manual ending of user sessions in the background, which was previously disallowed.

## Testing

Updated existing tests.


bidetofevil (approved):
LGTM. I can remove this one from the ignored tests too!

PR #3361, opened 2026-05-06
Include user session ID and session part ID in span links

fractalwrench:
## Goal

Includes the user session ID and session part ID in span links.

## Testing

Added unit test coverage.


bidetofevil (approved):
LGTM. I can remove my ignored test for this

PR #3365, opened 2026-05-07
[DNM] Update symbols on example app

fractalwrench:
## Goal

Updates symbols on example app for changes in this integration branch. This can only be merged once an artifact is shipped that can be used by the example app.


bidetofevil (approved):
LGTM

PR #3366, opened 2026-05-07
Update release workflow

fractalwrench:
## Goal

Updates the CI workflows so that the maven central publishing [is automatic](https://vanniktech.github.io/gradle-maven-publish-plugin/central/#publishing-releases).


bidetofevil (approved):
LGTM

PR #3372, opened 2026-05-08
Report the user agent for instrumented HTTP requests

lemnik:
## Goal
Capture the `User-Agent` header (when present) as the `user_agent.original` [span attribute](https://opentelemetry.io/docs/specs/semconv/registry/attributes/user-agent/) to assist debugging different clients.

This PR also adds `userAgent` as an optional property to the `EmbraceNetworkRequest` and appropriate bridge functions to maintain binary compatibility, so that the `userAgent` can be manually reported. `User-Agent` headers can also be extracted from `NetworkCaptureData` when manually reporting a `EmbraceNetworkRequest`.

## Testing
Updated the existing network instrumentation tests to either capture or ignore the `userAgent` and assert its appropriate values on the generated spans.

### DIFF
--- a/embrace-android-api/src/main/kotlin/io/embrace/android/embracesdk/network/EmbraceNetworkRequest.kt
+++ b/embrace-android-api/src/main/kotlin/io/embrace/android/embracesdk/network/EmbraceNetworkRequest.kt
@@ -65,11 +65,6 @@ public class EmbraceNetworkRequest private constructor(
     public val w3cTraceparent: String?,
 
     /**
-     * Optional user agent to report for the request
-     */
-    public val userAgent: String?,
-
-    /**
      * Network capture data for the request.
      */
     public val networkCaptureData: NetworkCaptureData?,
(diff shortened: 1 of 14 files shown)


embrace-android-api/src/main/kotlin/io/embrace/android/embracesdk/network/EmbraceNetworkRequest.kt
```diff
@@ -166,10 +222,58 @@ public class EmbraceNetworkRequest private constructor(
...
+                errorMessage = errorMessage,
+                traceId = traceId,
+                w3cTraceparent = w3cTraceparent,
+                userAgent = networkCaptureData?.userAgent,
+                networkCaptureData = networkCaptureData
+            )
+        }
+
         internal fun HttpMethod.asString(): String = name.uppercase(Locale.getDefault())
+        internal val NetworkCaptureData.userAgent: String?
```
bidetofevil: So `NetworkCaptureData` is rarely used in practice, so I think I prefer to avoid to complexity of using the request header from there to populate the field. Allowing folks to optionally specifying it should be enough

bidetofevil:
The overall structure looks good. I like how this has satisfied the original requirements and gives us more by using the `user-agent` request header. Some points about the implementation that I want to discuss:

- We want attributes to have low cardinality so we can better filter and group by them. `OkHttp` seems to satisfy that, but does Http(s)UrlConnection? Like, does it contain extraneous information like the OS, OS version, and other things?
- Seems like using `user_agent.name` and `user_agent.version` would be better for keeping the cardinality low (and more nuanced filtering) - `user_agent.original` seems like it can contain a bunch of things that are just what the network stack is
- We can implement this for our internal instrumentation without exposing this in the custom API, which I don't think are that widely used anyway. As such, I think we should do that first given we don't know how we are going to use the data beyond internal understanding of our customers usage at this point, so exposing an API might give customers the wrong impression of what we are going to do with it. Maybe after we've solidified our own usage of it, we can do the exposure?

fractalwrench (approved):
LGTM

bidetofevil (approved):
LGTM

PR #3379, opened 2026-05-11
Tweak codecov threshold

fractalwrench:
## Goal

Tweaks the threshold used for codecov so that the minimum target for the project is 75%, but fluctuations of <1% are allowed on PRs. This should reduce the number of PRs that have CI check failures related to codecov.

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -2,7 +2,10 @@ coverage:
   status:
     project:
       default:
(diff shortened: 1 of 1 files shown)


bidetofevil (approved):
LGTM. I think this is non-blocking anyway but good to not have it flag stuff needlessly.

PR #3380, opened 2026-05-11
Add session index

fractalwrench:
## Goal

Renames `emb.user_session_part_number` to `emb.user_session_part_index` and additionally adds `emb.session_part_number` which increments monotonically for session parts.

## Testing


bidetofevil (approved):
LGTM. Does this have coverage of the index increasing for background activities?

fractalwrench:
Yes, that is asserted in the golden files.

PR #3381, opened 2026-05-11
Fix max duration timeout

fractalwrench:
## Goal

Fixes the max duration timeout so that it's accurately applied to background activities.

The behavior of background session parts now looks as expected. In this example inactivity timeout was set at 1m, and max duration at 2m. The first user session is exactly 2m due to the max duration timeout. The second user session exceeds the max duration because it's in the background. The third user session looks normal.

Screenshot 2026-05-11 at 12 51 42

### DIFF
--- a/embrace-android-core/src/test/kotlin/io/embrace/android/embracesdk/internal/session/orchestrator/SessionOrchestratorTest.kt
+++ b/embrace-android-core/src/test/kotlin/io/embrace/android/embracesdk/internal/session/orchestrator/SessionOrchestratorTest.kt
@@ -1370,46 +1370,6 @@ internal class SessionOrchestratorTest {
     }
 
     @Test
-    fun `stale inactivity timer does not terminate user session rotated by max duration in BG`() {
-        // Inactivity > max duration ensures the max-duration timer fires first in BG. The
-        // inactivity timer scheduled by the previous user session's onBackground is then a stale
-        // anchor on a terminated session — if it later fires, it would incorrectly terminate the
-        // new session created by the max-duration rotation.
-        configService = FakeConfigService(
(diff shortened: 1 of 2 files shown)


embrace-android-core/src/test/kotlin/io/embrace/android/embracesdk/internal/session/orchestrator/SessionOrchestratorTest.kt
```diff
@@ -1235,6 +1235,223 @@ internal class SessionOrchestratorTest {
...
+    fun `stale inactivity timer does not terminate user session rotated by max duration in BG`() {
+        // Inactivity > max duration ensures the max-duration timer fires first in BG. The
+        // inactivity timer scheduled by the previous user session's onBackground is then a stale
+        // anchor on a terminated session — if it later fires, it would incorrectly terminate the
+        // new session created by the max-duration rotation.
+        configService = FakeConfigService(
+            backgroundActivityBehavior = createBackgroundActivityBehavior(
+                remoteCfg = RemoteConfig(backgroundActivityConfig = BackgroundActivityRemoteConfig(threshold = 100f))
+            ),
+            sessionBehavior = FakeUserSessionBehavior(
```
bidetofevil: I don't think the inactivity timer can exceed the max duration time in production (see `UserSessionBehaviorImpl`), so i don't think this codepath would be possible. Can you trigger this in a way that is possible in production?  
  
And given there is logic like this in the real UserSessionBehavior, perhaps we should use that for testing this?

fractalwrench: Removed this test case

bidetofevil (approved):
LGTM

PR #3382, opened 2026-05-11
Populate span termination attributes correctly

fractalwrench:
## Goal

Alters the `Span.toFailedSpan` extension function so that it sets the termination reason. `hasEmbraceAttribute` had overloads for both `Span` and `Map<String, Any>`, which meant in the context of `toFailedSpan` the `mutableMapOf` was checked for attributes. 

## Testing

Enhanced unit test coverage.


bidetofevil (approved):
LGTM

PR #3383, opened 2026-05-11
Prevent duplicate crash requests

fractalwrench:
## Goal

Prevents duplicate crash requests from being sent by shutting down the `SchedulingService` when a crash is received.

## Testing

Added unit tests.


embrace-android-delivery/src/main/kotlin/io/embrace/android/embracesdk/internal/delivery/intake/IntakeServiceImpl.kt
```diff
@@ -111,6 +124,9 @@ class IntakeServiceImpl(
         }
     }
 
+    private fun StoredTelemetryMetadata.isCrashPayload(): Boolean =
```
bidetofevil: Specifically, this is a payload after which the process is expected to terminate. Like, an native crash is a crash payload but wouldn't and shouldn't trigger this. Should this be renamed so it's clearer when this is read?

bidetofevil (approved):
This looks good to me in terms of fixing the current behaviour, so I think we should merge it as is (except make a non-blocking rename suggestion).   
  
While talking to Fredric about this, we said he wanted to hold the process open to try to send the crash with a small timeout so we can potentially get this earlier and without requiring the app to startup again.  
  
Do you see any complications if we did this, i.e. can we confirm the delivery request finished, either the HTTP request is persisted or the initial send was successful? We can discuss this offline

fractalwrench:
@bidetofevil I think we need to pick one approach. If we care more about duplicates than deliverability then we should not attempt to make a HTTP request when crashing. If it's the inverse, we can attempt to make a HTTP request.

This is also a numbers game from my POV. We're very unlikely to stop duplicate crash reports entirely but we can reduce their frequency through different delivery strategies. I'll leave this open until there's consensus on the approach we take.

bidetofevil:
For the record, I also think we should persist the payload and not send until the next app startup. The kind of edge cases attempting delivery will reduce (but not prevent entirely) doesn't seem worth the extra complexity and potential complications down stream caused by the indeterminism. 

The vast majority of crash apps wills start again, so this is mostly just a delay. And if this crash is so unique that all cases of it are from an app never starts up again, it's probably too obscure for most ppl to care about. And if there's something structural that prevents the app or SDK from ever starting up again... well that is even less likely IMO,.

PR #3384, opened 2026-05-11
Add lazy-init support for state data sources

bidetofevil:
Add functionality to `StateDataSource` to delay state capture. Everything is encapsulated in the state infra and doesn't touch the core DataSource interfaces.

This will be used in a subsequent PR for the manual screen tracking API.

### DIFF
--- a/embrace-android-instrumentation-api/src/main/kotlin/io/embrace/android/embracesdk/internal/arch/datasource/StateDataSource.kt
+++ b/embrace-android-instrumentation-api/src/main/kotlin/io/embrace/android/embracesdk/internal/arch/datasource/StateDataSource.kt
@@ -7,6 +7,7 @@ import io.embrace.android.embracesdk.internal.arch.SessionPartEndListener
 import io.embrace.android.embracesdk.internal.arch.limits.UpToLimitStrategy
 import io.embrace.android.embracesdk.internal.arch.schema.SchemaType
 import io.embrace.android.embracesdk.internal.logging.InternalErrorType
+import java.util.concurrent.atomic.AtomicBoolean
 import java.util.concurrent.atomic.AtomicReference
 
 /**
@@ -27,6 +28,7 @@ abstract class StateDataSource<T : Any>(
     private val currentState: AtomicReference<T> = AtomicReference(defaultValue)
     private val partStateToken: AtomicReference<SessionPartStateToken<T>?> = AtomicReference()
     private val unrecordedTransitions = AtomicReference(noUnrecordedTransitions)
+    private val initialized = AtomicBoolean(initializeOnCreation)
 
     /**
      * Notify that the state has changed at the given time to the given value, with the given number of transitions dropped by the
@@ -38,6 +40,10 @@ abstract class StateDataSource<T : Any>(
         transitionAttributes: Map<String, String> = emptyMap(),
         droppedTransitions: Int = 0,
     ) {
+        if (!initialized.getAndSet(true)) {
+            onDataCaptureEnabled()
+        }
+
         val oldState = currentState.getAndSet(newState)
         val currentStateToken = partStateToken.get()
         // Track the number of transitions dropped by instrumentation that didn't cause this to be invoked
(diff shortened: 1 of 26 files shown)


bidetofevil:
* **#3385** <a href="https://app.graphite.com/github/pr/embrace-io/embrace-android-sdk/3385?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#3384** <a href="https://app.graphite.com/github/pr/embrace-io/embrace-android-sdk/3384?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/embrace-io/embrace-android-sdk/3384?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

fractalwrench (commented):
This feels like the approach could be the wrong way round to me. I think there's a distinction to be made between whether the `DataCaptureOrchestrator` allows a `DataSource` to capture data, and whether that `DataSource` is ready to capture data (an internal implementation detail).

Is it feasible to have the data source internally maintain state on whether it's ready to capture data, or are there reasons that's not possible?

bidetofevil:
Yeah, the lifecycle might be a bit unclear. Before this, a data source is basically ready to receive or passively capture data after creation, which is broken up into the object instantiation, and the calling of the `onDataCaptureEnabled()` callback, which is effectively a creation callback.

This PR decouples the callback invocation optionally, which creates an implicit init step that is post creation. But as seen from the changes to `StateDataSource`, it has created an implicit state of "created but not ready to capture data", which isn't expressed in the `DataSource` interface and only in `StateDataSource`. You're right to point out the disconnect.  
  
I think I can push whether or not a datasource is ready to capture data up to the DS level, but I'll have to see if I can truly guard the capturing of data (e.g. have some logic in `captureTelemetry).`Let me spend a bit of time on this today.

bidetofevil:
I'm not completely satisfied with this, but it is a good point to take stock to see if it's good enough for now.  
  
To summarize the change, I've basically tightly coupled `DataSourceState` with`DataSourceImpl` whereas before it was coupled with `DataSource` only. The reason I made this change was so `DataSourceState` can selectively enable a datasource upon creation, which is a new concept being introduced, as it is primarily responsible for doing that in production codepaths. 

The reason this concept cannot be put on the `DataSource` interface is because the implementation will need to guarantee that data sources not enabled at creation need to be enabled some other time - and whether or not something is enabled has to be congruent with this, and the only way to guarantee that is embedding everything within an implementation, which is what this does.  
  
Alternatively, I could have put the concept on the interface and made the non-DataSourceImpl-based implementations fend for themselves, which I don't like. I could've also extracted this into another interface and provided some state management implementation, but that seems over the top in terms of complexity. As such this is what I've settled on.

Another wrinkle that makes this trickier is that `DataSourceState` and `DataSource` are tightly coupled, but you can also create a datasource on its own without a DataSourceState, which enforces some lifecycle stuff. This is probably something we can address later, as this change exposes that issue a bit more.

fractalwrench:
I think this one could benefit from sync discussion (either in the guild or later on) given that it affects a core internal API and I think it'd be beneficial to have everyone on the same page with the approach before reviewing a changeset

bidetofevil:
Pushed all this stuff to the `StateDataSource`

fractalwrench (approved):
LGTM

embrace-android-instrumentation-api/src/main/kotlin/io/embrace/android/embracesdk/internal/arch/datasource/StateDataSource.kt
```diff
@@ -97,15 +113,23 @@ abstract class StateDataSource<T : Any>(
         createSessionStateSpan(currentState.get())
     }
 
+    private fun startStateCapture() {
+        if (!stateCaptureActive.getAndSet(true)) {
```
lemnik: nit: any reason not to use a single strong CAS instead of `getAndSet`?

```suggestion
        if (stateCaptureActive.compareAndSet(false, true)) {
```

lemnik (approved):

embrace-android-instrumentation-api/src/main/kotlin/io/embrace/android/embracesdk/internal/arch/datasource/StateDataSource.kt
bidetofevil: No good reason at all. More like a force of habit, but a compareAndSet makes more sense, so lets do that.

bidetofevil:
### Merge activity

* **May 20, 3:04 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/embrace-io/embrace-android-sdk/3384).
* **May 20, 3:05 PM UTC**: [Graphite](https://app.graphite.com/github/pr/embrace-io/embrace-android-sdk/3384) rebased this pull request as part of a merge.
* **May 20, 3:05 PM UTC**: @bidetofevil merged this pull request with [Graphite](https://app.graphite.com/github/pr/embrace-io/embrace-android-sdk/3384).

PR #3394, opened 2026-05-14
Fix AEI user session ID population

fractalwrench:
## Goal

Fixes how the user session IDs are populated for AEI logs so that the necessary attributes are present for the backend.

## Testing

Updated test coverage.


embrace-android-sdk/src/integrationTest/resources/user_session_aei_log.json
```diff
@@ -32,35 +32,15 @@
...
-          "value": "__EMBRACE_TEST_IGNORE__"
+          "value": "1a2b3c4d5e6f7a8b9c0d1e2f3a4b5c6d"
         },
         {
           "key": "emb.type",
           "value": "sys.exit"
         },
         {
           "key": "emb.user_session_id",
-          "value": "__EMBRACE_TEST_USER_SESSION_ID__"
```
bidetofevil, line 63: This placeholder should probably be restored?

bidetofevil (approved):
I assume if we want the sending session info, we have to fix the issue where the session metadata supplier is set after the instrumentation is loaded?

PR #3398, opened 2026-05-15
Improve Compose-Tap Instrumentation Performance

lemnik:
## Goal
Reduce the overhead of the Compose tap instrumentation by only walking `View` and `SemanticNode`s that have bounds enclosing the `[x, y]` location of the `MotionEvent`. This avoids walking `ViewGroup`s and composables that are off-screen, invisible or otherwise cannot be the tapped element.

This PR also adds `Text` semantics as a fallback label, which will pick up the text labels of elements (such as buttons) that don't have a `SemanticNode` with a content description or `OnClick` label.

### DIFF
--- a/embrace-android-instrumentation-compose-tap/src/main/kotlin/io/embrace/android/embracesdk/internal/instrumentation/compose/tap/EmbraceGestureListener.kt
+++ b/embrace-android-instrumentation-compose-tap/src/main/kotlin/io/embrace/android/embracesdk/internal/instrumentation/compose/tap/EmbraceGestureListener.kt
@@ -22,11 +22,9 @@ internal class EmbraceGestureListener(
 
     override fun onSingleTapUp(event: MotionEvent): Boolean {
         try {
-            if (event.actionMasked == MotionEvent.ACTION_UP) {
-                activityRef.get()?.let { activity ->
-                    activity.window?.let {
-                        logTapUp(it.decorView, event)
-                    }
+            activityRef.get()?.let { activity ->
+                activity.window?.let {
+                    logTapUp(it.decorView, event)
                 }
             }
         } catch (e: Throwable) {
--- a/embrace-android-instrumentation-compose-tap/build.gradle.kts
+++ b/embrace-android-instrumentation-compose-tap/build.gradle.kts
@@ -12,4 +12,8 @@ dependencies {
     implementation(project(":embrace-android-instrumentation-api"))
     implementation(project(":embrace-android-instrumentation-taps"))
     compileOnly(libs.compose)
+
+    testImplementation(project(":embrace-android-instrumentation-api-fakes"))
+    testImplementation(libs.compose)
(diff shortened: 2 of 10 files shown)


fractalwrench (commented):
This approach looks much better than the previous implementation. My main question is what do we think is an appropriate way to test this change (and broadly the feature as a whole). If we did have automated test coverage that'd make me feel a lot better about making changes in this area, even if it was just the basic case of asserting that click events are actually captured. However, I'm not sure of the effort level that involves...@bidetofevil thoughts?

bidetofevil:
Are there Compose test infrastructure we can use with Roboletric or something?

embrace-android-instrumentation-compose-tap/src/main/kotlin/io/embrace/android/embracesdk/internal/instrumentation/compose/tap/EmbraceGestureListener.kt
```diff
@@ -22,11 +22,9 @@ internal class EmbraceGestureListener(
 
     override fun onSingleTapUp(event: MotionEvent): Boolean {
         try {
-            if (event.actionMasked == MotionEvent.ACTION_UP) {
```
bidetofevil, line 25: Do we not need this to dedupe up/down/whatever events for the same action?

bidetofevil:
Will `androidx.compose.ui.test` be helpful here or are they meant to be used to unit test custom Composables, whereas we need to test view trees that have gone through bytecode instrumentation. I suppose we can simulate that too....

bidetofevil:
Seems like `AndroidComposeTestRule` can give us the platform but we have to manually write up the instrumentation that otherwise would be done in the build.   
  
We can also create macrobenchmark tests for both performance and integration verifiction

embrace-android-instrumentation-compose-tap/src/main/kotlin/io/embrace/android/embracesdk/internal/instrumentation/compose/tap/EmbraceGestureListener.kt
lemnik: The [only possible path to get here](https://cs.android.com/android/platform/superproject/+/android-latest-release:frameworks/base/core/java/android/view/GestureDetector.java;l=781-798) from `GestureDetector` means that the event is `ACTION_UP`. We could leave it in for safety, but it's effectively duplicate logic right now.

bidetofevil: Nah, remove if it's not necessary

bidetofevil (approved):
LGTM pending the addition of more test

fractalwrench (approved):
LGTM

PR #3399, opened 2026-05-15
Tighten intake of envelopes when a crash happens

fractalwrench:
## Goal

Tightens up what envelopes are accepted by the `IntakeService` when a crash happens to avoid persisting unnecessary cached session payloads that show as duplicates. The `PeriodicSessionPartCacher` now has a condition where it terminates at the start of an operation if a crash has happened.

Additionally, `IntakeService` now detects whether an envelope is a JVM crash, and if so prevents taking any envelope other than a JVM crash and a complete session payload. This is achieved with an enum state machine.

## Testing

Added unit tests.

### DIFF
--- a/embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/session/caching/PeriodicSessionPartCacher.kt
+++ b/embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/session/caching/PeriodicSessionPartCacher.kt
@@ -18,10 +18,14 @@ class PeriodicSessionPartCacher(
 
     private var scheduledFuture: ScheduledFuture<*>? = null
 
+    @Volatile
+    private var stopped = false
+
     /**
      * It starts a background job that will schedule a callback to do periodic caching.
      */
     fun start(provider: Provider<Envelope<SessionPartPayload>?>) {
+        stopped = false
         scheduledFuture = this.worker.scheduleWithFixedDelay(
             onPeriodicCache(provider),
             0,
@@ -31,6 +35,9 @@ class PeriodicSessionPartCacher(
     }
 
     private fun onPeriodicCache(provider: Provider<Envelope<SessionPartPayload>?>) = Runnable {
+        if (stopped) {
+            return@Runnable
+        }
         EmbTrace.trace("snapshot-session") {
             try {
                 provider()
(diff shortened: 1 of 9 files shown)


embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/session/caching/PeriodicSessionPartCacher.kt
```diff
@@ -18,10 +18,14 @@ class PeriodicSessionPartCacher(
 
     private var scheduledFuture: ScheduledFuture<*>? = null
 
+    @Volatile
+    private var stopped = false
```
bidetofevil, line 22: Do we need to synchronize around this or do we handle race conditions fine?

embrace-android-delivery/src/main/kotlin/io/embrace/android/embracesdk/internal/delivery/intake/IntakeServiceImpl.kt
```diff
@@ -40,13 +51,27 @@ class IntakeServiceImpl(
...
+        if (metadata.isCrashTerminatingProcess() && metadata.complete) {
+            if (state.compareAndSet(State.ACTIVE, State.CRASH_RECEIVED)) {
+                schedulingService.shutdown()
+                // non-blocking shutdown: reject subsequent submissions but defer the drain to
+                // payloadStore.handleCrash's later intakeService.shutdown() call
+                worker.shutdownAndWait(0)
+                processIntake(intake, metadata, staleEntry)
+                return immediateFuture()
             }
+            throw RejectedExecutionException("Rejecting as crash already accepted")
```
bidetofevil: I assume this and the later throw isn't expected to happen and if it does, we will catch this and log an internal error?

bidetofevil (approved):
LGTM.  A couple of non-blocking questions. Also, can we test this in the integration test as well?

embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/session/caching/PeriodicSessionPartCacher.kt
fractalwrench: Volatile should be sufficient here

embrace-android-delivery/src/main/kotlin/io/embrace/android/embracesdk/internal/delivery/intake/IntakeServiceImpl.kt
fractalwrench: I've altered this to return `immediateFuture()` instead.

PR #3400, opened 2026-05-15
Alter what stored telemetry metadata is persisted

fractalwrench:
## Goal

Alters `StoredTelemetryMetadata` so that the active user session ID and session part ID is encoded in the filename, or 'none' if none is present. This will be required to dedupe cached/complete session payloads, but is likely to have a variety of use-cases.

## Testing

Added unit tests.


bidetofevil (approved):
LGTM

PR #3401, opened 2026-05-15
Dedupe session payloads at resurrection

fractalwrench:
## Goal

Alters the payload resurrection logic so that session payloads are deduped if a payload with a similar session ID exists in `payloadStorageService`, or if multiple payloads exist in `cacheStorageService` with the same ID.

This should help mitigate the frequency at which session payload HTTP requests have duplicates, although some baseline level will always be expected as it's impossible to atomically make a HTTP request and delete the file in the context of the environment we run in.

## Testing

Added unit tests.

### DIFF
--- a/embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/resurrection/PayloadResurrectionServiceImpl.kt
+++ b/embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/resurrection/PayloadResurrectionServiceImpl.kt
@@ -126,7 +126,8 @@ internal class PayloadResurrectionServiceImpl(
                     if (resource != null && metadata != null) {
                         cachedLogEnvelopeStore.create(
                             storedTelemetryMetadata = createNativeCrashEnvelopeMetadata(
-                                sessionPartId = nativeCrash.sessionPartId
+                                sessionPartId = nativeCrash.sessionPartId,
+                                userSessionId = nativeCrash.userSessionId,
                             ),
                             resource = resource,
                             metadata = metadata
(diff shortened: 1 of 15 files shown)


embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/resurrection/PayloadResurrectionServiceImpl.kt
```diff
@@ -164,6 +169,63 @@ internal class PayloadResurrectionServiceImpl(
...
+     * Splits payloads into two collections: one that should be resurrected, and another that should be
+     * discarded.
+     *
+     * Cached session payloads can be redundant either when another cached payload has the same session
+     * ID, or when the session ID is matched by a complete payload in [payloadStorageService].
+     *
+     * Non-SESSION payloads and SESSION payloads with no encoded session IDs (legacy v1 filenames)
+     * bypass the dedupe checks.
+     */
+    private fun dedupeSessionPayloads(
```
bidetofevil, line 182: I like having this as a fallback, but are are conditions for which is not not preventable other than a poorly timed crash that happened before the incomplete session cache payload is deleted but after the complete payload is finalized?

bidetofevil (approved):
LGTM. A non-blocking comment about what conditions actually cause these dupes and if its worthwhile to reduce them or just let the deduping do the job

embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/resurrection/PayloadResurrectionServiceImpl.kt
fractalwrench: This is in place as a fallback for any conditions we might not have thought about. I've attempted to tighten up the intake layer so that the incomplete payload is always deleted.

The main ones I can think of are process termination before the cached payload can be deleted, intermittent I/O errors, and upgrades from old versions of the SDK.

PR #3402, opened 2026-05-15
Revert "Prevent duplicate crash requests"

fractalwrench:
Reverts embrace-io/embrace-android-sdk#3383


bidetofevil (approved):
LGTM. Are there any changes that remained to reduce the duplicates caused by the race conditions, or is this a reset to before any changes were made?

PR #3405, opened 2026-05-18
Reduce possible races in CurrentSessionPartSpanImpl

lemnik:
## Goal
Encapsulate the current session span with its trace counters so that there cannot be races between the modification and checking of the counters and the transition between session spans.

This PR also replaces the mix of `AtomicReference`, `AtomicBoolean` and `synchronized` calls with `@Volatile` properties (for the correct memory-effects) and `synchronized` calls on a clearly named `sessionTransitionLock`.


fractalwrench (approved):
LGTM

PR #3406, opened 2026-05-18
Alter internal logger so that severity is specified with `InternalErrorType`

fractalwrench:
## Goal

Alters `InternalErrorType` to a sealed class hierarchy that contains an associated severity. This can be a first step towards triaging existing internal errors so that internal telemetry is less noisy. By default I set every error type as `Severity.ERROR`. A subsequent PR will change these for individual error types and prevent sending them as internal errors. The idea here is that we can route errors below `ERROR` severity to a different location in future.

## Testing

Added unit tests.


bidetofevil (approved):
Nice. I have a feeling we can down-shift most of these to warnings.

PR #3407, opened 2026-05-18
Set metadata provider earlier

fractalwrench:
## Goal

Configures the metadata provider before instrumentation is loaded so that the collected attributes always contain session information such as `emb.user_session_id`.

## Testing

Added a unit test.


bidetofevil (approved):
LGTM

PR #3408, opened 2026-05-18
Fix test assertion

fractalwrench:
## Goal

Fixes a test assertion that assumed only 2 sessions would be created, when in fact 3 will be created. This led to flaky behavior on CI as the call often succeeded as the test framework polls until N envelopes have been received.

Additionally, I've made some `emb.state` attributes less specific in golden files as they were causing flaky behavior, and removed a flaky test case that is covered by other test cases in the same file.


bidetofevil (approved):
LGTM

PR #3409, opened 2026-05-19
Resolve internal NoSuchElementExceptions

lemnik:
## Goal
Replace the use of `CopyOnWriteArraySet.toList()` with a new `threadSafeToList()` extension function that avoids the race between the `size` check and unconditional `iterator().next()` in the stdlib `toList()` implementation.

The `threadSafeToList()` extension function was added to our existing `CollectionExtensions.kt`.


fractalwrench (approved):
LGTM

PR #3411, opened 2026-05-19
Reduce noisy internal errors

fractalwrench:
## Goal

Alters the reporting of internal errors so that some are no longer reported by reducing their importance to `WARNING`.


bidetofevil (approved):
LGTM

PR #3412, opened 2026-05-19
Avoid NPEs from GZIPInputStream construction

lemnik:
## Goal
Check that the `InputStream`s returned from `cacheStorageService.loadPayloadAsStream` is non-null before passing them to `GZIPInputStream` (which would result in a `NullPointerException` from the `GZIPInputStream` constructor).

These can happen due to thread races, cache clears, or I/O errors (which are already reported by the `FileStorageServiceImpl`).

## Testing
Added new unit tests to `PayloadResurrectionServiceImplTest` where the payload is returned as `null` and the test can ensure that no errors are raised or logged in these cases.


fractalwrench (approved):
LGTM

PR #3413, opened 2026-05-19
Handle cases where payloads are not valid GZIP

lemnik:
## Goal
When a payload is not valid GZIP the `GZIPInputStream` constructor throws a `ZIPException`, these should be ignored and the payload be deleted.

This PR only adds the `ZIPException` handling and relies of the existing code path to ignore the `null` `InputStream` and delete the payload.

## Testing
Added another unit test, and assert the `deleteCount`s


fractalwrench (approved):
LGTM

bidetofevil (approved):
LGTM

PR #3426, opened 2026-05-21
Reduce internal error noise

lemnik:
## Goal
Reduce the amount of noise produced by `trackInternalError`s for errors that are expected:

 - App startup may not be recording due to user interactions, leading to errors that cannot be actioned
 - `pruneStorage` should not try to report internal errors when the disk is full to avoid making the situation worse
 - `loadPayloadAsStream` should not report `FileNotFoundException`s as these are expected

## Testing
Updated test to not expect internal errors that have been removed

### DIFF
--- a/embrace-android-delivery/src/test/kotlin/io/embrace/android/embracesdk/internal/delivery/storage/PayloadStorageServiceImplTest.kt
+++ b/embrace-android-delivery/src/test/kotlin/io/embrace/android/embracesdk/internal/delivery/storage/PayloadStorageServiceImplTest.kt
@@ -80,7 +80,7 @@ class PayloadStorageServiceImplTest {
     @Test
     fun `load non existent file`() {
         assertNull(service.loadPayloadAsStream(metadata))
-        assertNotNull(logger.internalErrorMessages.single().throwable)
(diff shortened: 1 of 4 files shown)


embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/resurrection/PayloadResurrectionServiceImpl.kt
```diff
@@ -30,6 +30,7 @@ import io.embrace.android.embracesdk.internal.session.getUserSessionProperties
 import io.embrace.android.embracesdk.internal.utils.Provider
 import io.embrace.android.embracesdk.semconv.EmbSessionAttributes
 import io.opentelemetry.kotlin.semconv.SessionAttributes
+import java.io.FileNotFoundException
```
fractalwrench: This import doesn't seem necessary

fractalwrench (commented):
This looks good once CI is passing (there's a failing assertion in one of the delivery tests that looks related to this)

fractalwrench (approved):

PR #3433, opened 2026-05-26
Added clock drift attributes for GNSS and Network time to the session

lemnik:
## Goal
Calculate and report the differences between the anchored wall time and GNSS time && Network time (when available). These time deltas are added as attributes to the session span on-end.

## Testing
Added new test cases to cover the new attributes.

### DIFF
--- a/embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/capture/metadata/ClockDrift.kt
+++ b/embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/capture/metadata/ClockDrift.kt
@@ -4,12 +4,37 @@ package io.embrace.android.embracesdk.internal.capture.metadata
  * Reported difference / drift between our internal clock and the system auxiliary clocks (GNSS and Network Time). Positive values
  * indicate that the specified clock is behind our clock, negative values indicate it is ahead.
  */
-data class ClockDrift(
+class ClockDrift private constructor(
     val networkDriftMillis: Long?,
     val gnssDriftMillis: Long?,
 ) {
-    constructor(wallTimeMillis: Long, networkTimeMillis: Long?, gnssTimeMillis: Long?) : this(
-        networkDriftMillis = networkTimeMillis?.let { wallTimeMillis - it },
-        gnssDriftMillis = gnssTimeMillis?.let { wallTimeMillis - it },
-    )
+    override fun equals(other: Any?): Boolean {
+        if (this === other) return true
+        if (javaClass != other?.javaClass) return false
+
+        other as ClockDrift
+
+        if (networkDriftMillis != other.networkDriftMillis) return false
+        if (gnssDriftMillis != other.gnssDriftMillis) return false
+
+        return true
+    }
+
+    override fun hashCode(): Int {
+        var result = networkDriftMillis?.hashCode() ?: 0
+        result = 31 * result + (gnssDriftMillis?.hashCode() ?: 0)
+        return result
+    }
+
+    override fun toString(): String {
+        return "ClockDrift(networkDriftMillis=$networkDriftMillis, gnssDriftMillis=$gnssDriftMillis)"
+    }
+
+    companion object {
+        fun fromWallDrift(wallTimeMillis: Long, networkTimeMillis: Long?, gnssTimeMillis: Long?) =
+            ClockDrift(
+                networkDriftMillis = networkTimeMillis?.let { wallTimeMillis - it },
+                gnssDriftMillis = gnssTimeMillis?.let { wallTimeMillis - it },
+            )
+    }
 }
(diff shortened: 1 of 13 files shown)


embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/capture/metadata/ClockDrift.kt
```diff
@@ -0,0 +1,15 @@
...
+
+/**
+ * Reported difference / drift between our internal clock and the system auxiliary clocks (GNSS and Network Time). Positive values
+ * indicate that the specified clock is behind our clock, negative values indicate it is ahead.
+ */
+data class ClockDrift(
+    val networkDriftMillis: Long?,
+    val gnssDriftMillis: Long?,
+) {
+    constructor(wallTimeMillis: Long, networkTimeMillis: Long?, gnssTimeMillis: Long?) : this(
```
bidetofevil: We have a coding convention (found in [AGENT.md](http://AGENT.md)) that states we shouldn't have business logic in data classes. Perhaps we can pull this out into a factory method that takes in these parameters and constructs an instance instead? We can lock down the constructor so folks don't create these on their own and instead always go through the factory method.

embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/capture/metadata/EmbraceMetadataService.kt
```diff
@@ -90,6 +93,32 @@ internal class EmbraceMetadataService(
 
     override fun getDiskUsage(): DiskUsage? = diskUsage
 
+    override fun getClockDrift(): ClockDrift? {
```
bidetofevil: Can we test this via roboletric to ensure we get the expected values in the relevant Android versions?

embrace-android-semconv/src/main/kotlin/io/embrace/android/embracesdk/semconv/EmbSessionAttributes.kt
```diff
@@ -16,6 +16,18 @@ object EmbSessionAttributes {
     @ExperimentalSemconv
     const val EMB_CLEAN_EXIT: String = "emb.clean_exit"
 
+    /**
+     * The time delta between wall-time and the GNSS reported time in milliseconds (Wall Time - GNSS Time)
+     */
+    @ExperimentalSemconv
+    const val EMB_CLOCK_GNSS_DRIFT_MILLIS: String = "emb.clock_gnss_drift_millis"
```
bidetofevil: For general semantic conventions, the unit should be defined in the YAML so it doesn't have to be defined in the constant or attribute name

embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/capture/metadata/ClockDrift.kt
bidetofevil: We should probably add some tests at this level for sanity's sakes too.

bidetofevil (approved):
LGTM in general. Some non-blocking suggestions you can do as part of this, in a separate PR, or not at all if you feels strongly about it.

embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/capture/metadata/EmbraceMetadataService.kt
```diff
@@ -90,6 +93,32 @@ internal class EmbraceMetadataService(
...
+        if (Build.VERSION.SDK_INT < Build.VERSION_CODES.Q) {
+            return null
+        }
+
+        val wallTime = clock.now()
+
+        val gnssTime = runCatching { SystemClock.currentGnssTimeClock().millis() }.getOrNull()
```
bidetofevil: I assume these calls are super fast and won't block on a binder call or something?

embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/capture/metadata/ClockDrift.kt
lemnik: Thanks. I've moved the calculation logic out into the `companion object` rather than extracting a full factory method, since I've also adjusted how the clocks are sampled (based on your IPC question: we want to sample the wall time for each AUX clock).

embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/capture/metadata/EmbraceMetadataService.kt
lemnik: Potentially not, I've moved them into the `precomputeValues` instead which actually cleans things up quite a bit.

embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/capture/metadata/ClockDrift.kt
```diff
@@ -0,0 +1,19 @@
...
+data class ClockDrift(
+    val networkDriftMillis: Long?,
+    val gnssDriftMillis: Long?,
+) {
+    companion object {
+        /**
+         * Computes the drift between the wall clock and an auxiliary clock, both expressed in milliseconds since the epoch.
+         * Positive return values indicate the auxiliary clock is behind the wall clock; negative values indicate it is ahead.
+         */
+        @JvmStatic
```
bidetofevil: Do we need this? I hope we don't need this lol

bidetofevil (approved):
LGTM. Probably could get rid of the @JvmStatic on the factory method for ClockDrift but not a big deal.

PR #3438, opened 2026-06-01
Improve NDK load time

lemnik:
## Goal
Reduce the size and loading time of the `embrace-android-instrumentation-crash-ndk` native library.

## Results

- arm64 `libembrace-native.so` reduced to 206.7kb (down from )
- symbol count reduced to 6075 (down from 15223)
- `System.loadLibrary` time reduced, macro-benchmark average of ~14.89 (down from ~26.47ms)
  - Samsung A7 Lite Tablet running Android 14
  - Improvement % is reasonably consistent on flagship devices

These are example histograms of the `install-native-crash-signal-handlers` span captured in a macro-benchmark:

old_hist
new_hist

## Testing
Relied largely on existing tests (there is a debug linker version script to retain exported symbols that the connected tests rely on), and macro-benchmarking.

### DIFF
--- a/embrace-android-instrumentation-crash-ndk/src/main/cpp/CMakeLists.txt
+++ b/embrace-android-instrumentation-crash-ndk/src/main/cpp/CMakeLists.txt
@@ -7,6 +7,8 @@ cmake_minimum_required(VERSION 3.4.1)
 
 set (CMAKE_CXX_STANDARD 17)
 
+add_compile_options(-ffunction-sections -fdata-sections)
+
 # Creates and names a library, sets it as either STATIC
 # or SHARED, and provides the relative paths to its source code.
 # You can define multiple libraries, and CMake builds them for you.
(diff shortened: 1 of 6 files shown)


bidetofevil (approved):
This looks good to me but it's been a while since I've reviewed C* code. Are there tests that we can add to verify that things work as expected at runtime, or will the existing ones fail already if this had a weird edge case?

bidetofevil:
FYI, we have a macrobenchmark test that runs the startup for 50 iterations several times and take some sort of median or mean after discarding outliers. Those have been pretty good at assessing load time. Have you found them?

lemnik:
> This looks good to me but it's been a while since I've reviewed C* code. Are there tests that we can add to verify that things work as expected at runtime, or will the existing ones fail already if this had a weird edge case?

The existing tests do some exercising of the native code, but notable don't deeply test things like "crash and restart". I can look at adding something, but it'll need a more advanced test harness that we're currently using.

> FYI, we have a macrobenchmark test that runs the startup for 50 iterations several times and take some sort of median or mean after discarding outliers. Those have been pretty good at assessing load time. Have you found them?

I had not noticed that. I did find that I had to put the NDK loader back onto the main thread to consistently get the NDK specific results into Perfetto (they were mostly being discarded for ending outside of the normal startup).

bidetofevil (approved):
LGTM

PR #3440, opened 2026-06-01
Moshi -> Kotlin Serialization

lemnik:
## Goal
Replace Moshi with Kotlin Serialization / JSON in the SDK (Gradle Plugin excluded for now).

This PR touches a huge number of files, but can reasonably be read commit-by-commit. The order of refactoring was:

- Add Kotlin Serializationn
- Add `@SerialName` and similar annotations, and custom serializer equivalents of our Moshi Adapters
- Replace the serialization code itself with Kotlin Serialization
  - required removing the reflection calls and replacing them with Kotlin Serialization structures
  - some `inline <reified T>` extension utility functions were added to `PlatformSerializer.kt` to avoid reflection and keep the `interface` surface clean
- Remove the Moshi annotations, references and dependency

## Notes

- With the Moshi KSP plugin removed we had no active KSP plugins in the `payload` module. This means that `com.google.devtools.ksp` started producing errors, so it has been removed.

## Testing
Tests are *effectively* unchanged. Tests that deal directly with serialization had to have call-sites updated to Kotlin Serialization friendly calls, such as:

```
return serializer.toJson(result, List::class.java)
```
to
```
return serializer.toJson<List<LegacyExceptionInfo>>(result)
```


bidetofevil (approved):
This looks good overall. A couple of general questions:

- How will regressions to some of the more complex serializers like `EnvelopeResourceSerializer` be surfaces? Will existing tests break loudly if there's a mismatch between the serialization and deserialization?
- How is versioning handled? In the same way as before, i.e. new keys that aren't in old JSONs will deserialize into nulls, and old keys that aren't needed will just be dropped?

lemnik:
> This looks good overall. A couple of general questions:
> 
> * How will regressions to some of the more complex serializers like `EnvelopeResourceSerializer` be surfaces? Will existing tests break loudly if there's a mismatch between the serialization and deserialization?

That is the intention, both the format and the encode/decode logic should match the old adapters exactly.

> * How is versioning handled? In the same way as before, i.e. new keys that aren't in old JSONs will deserialize into nulls, and old keys that aren't needed will just be dropped?

Yes, exactly: old keys are ignored and missing values are decoded as their defaults:
https://github.com/embrace-io/embrace-android-sdk/pull/3440/changes#diff-2412340a6a377dd97b9ab066f3365debe9fdcbf81ae1ccfbf023eefc257e57deR9-R11

PR #3452, opened 2026-06-08
Kotlin Serialization - Binary Codec

lemnik:
## Goal
Add a binary cache with a custom ordered binary cache format to accelerate startup and reduce the number of files required during `Embrace.start`. The file format has a single version tag for compatibility and a unit test to track breaking changes. The encoder and decoder are both built on-top of Kotlin Serialization for ease of maintenance.

This PR also encapsulated the `deviceId` in its own dedicated class as the binary cache allows us to bypass the `SharedPreferences` load on fast-path startups by adding both the `deviceId` and `etag` into the cache file.

## Testing
New unit tests added for the new startup. Also manually tested.


embrace-android-config/src/main/kotlin/io/embrace/android/embracesdk/internal/config/store/RemoteConfigStoreImpl.kt
```diff
@@ -22,26 +29,48 @@ internal class RemoteConfigStoreImpl(
...
 
-    override fun loadResponse(): ConfigHttpResponse? {
-        try {
+    // binary fast-path cache. Not created up-front: its absence is a clean cache miss that falls
+    // back to [configFile]/[etagFile].
+    private val cachedConfigFile = File(storageDir, "cached_config")
+
+    override fun loadResponse(): StoredConfigResponse? = loadFromCache() ?: loadFromJson()
+
+    private fun loadFromCache(): StoredConfigResponse? {
```
bidetofevil, line 38: What happens if the load fails because the data is in an expected format? Should we delete what's persisted so we don't fail again next time, perhaps log an internal error?

embrace-android-config/src/main/kotlin/io/embrace/android/embracesdk/internal/config/store/RemoteConfigStoreImpl.kt
```diff
@@ -22,26 +29,48 @@ internal class RemoteConfigStoreImpl(
...
+                deviceId = cached.deviceId,
+            )
+        } catch (exc: Exception) {
+            null
+        }
+    }
+
+    private fun loadFromJson(): StoredConfigResponse? {
+        return try {
             val cfg = configFile.inputStream().buffered().use {
```
bidetofevil, line 59: Same comment as `loadFromCache()`

bidetofevil (approved):
LGTM. A couple of non-blocking comments about failure to read from the cache and perhaps purging it if the JSON/binary can't be read successfully (e.g. corrupt data)

embrace-android-config/src/main/kotlin/io/embrace/android/embracesdk/internal/config/store/RemoteConfigStoreImpl.kt
lemnik: Deleting the files for deserialization failures definitely makes sense 👍

PR #3471, opened 2026-06-15
Experimental Smoothness Vital

lemnik:
## Goal
Experimental smoothness metric to track important frame-drops that are clearly visible during times when the user would be most aware of them (internally named "focal moments"). These typically start with user interactions and end once the layout "settles".

Frame counts and drops are normalised to 60fps for simplicity. Screen refresh rates are used to adjust the actual frame counts to 60fps, where at a refresh rate of 30hz every frame is counted twice (x2) while at 120hz every frame only counts as 0.5 frames.

## Testing
Unit tests for the instrumentation, and manual testing for tuning the reporting and settling.


embrace-android-instrumentation-vitals/README.md
```diff
@@ -0,0 +1,13 @@
...
+automatically. The module owns the shared OS plumbing — per-frame metrics, touch interaction
+listeners, and a Choreographer vsync source — and feeds it to the smoothness vital.
+
+- **Smoothness** — framerate quality during a user interaction. A *focal moment* opens on a touch and
+  concludes once the interaction goes quiet (the screen settles); the frames rendered during it are
+  correlated with the interaction to gauge whether the user experienced dropped frames. Loosely
+  inspired by the web's INP metric, but the signal of interest is framerate quality during the
+  interaction rather than time-to-paint. See `SMOOTHNESS.md`.
+
+Requires API 24+ (`Window.addOnFrameMetricsAvailableListener`); registers nothing on older devices.
```
bidetofevil, line 13: I think that's extremely fair. We might consider upping our minApi support to 23 or something soon

embrace-android-instrumentation-vitals/src/main/kotlin/io/embrace/android/embracesdk/internal/vitals/FocalInteractionCallbacks.kt
```diff
@@ -0,0 +1,37 @@
...
+
+/**
+ * Receives the raw signals captured by the Vitals instrumentation. Implemented by the smoothness tracker.
+ */
+internal interface FocalInteractionCallbacks {
+
+    /**
+     * A frame was redrawn: [vsyncNanos] is its vsync time, [jankNanos] how far it overran its budget.
+     */
+    fun onFrame(vsyncNanos: Long, jankNanos: Long)
```
bidetofevil: do we actually get nano second resolution or is this to match OTel?

embrace-android-instrumentation-vitals/src/main/kotlin/io/embrace/android/embracesdk/internal/vitals/FocalInteractionCallbacks.kt
```diff
@@ -0,0 +1,37 @@
...
+
+    /**
+     * A new screen began (Activity resumed).
+     */
+    fun onScreenStart()
```
bidetofevil, line 19: Is this mean to be tied to something specific, or just an arbitrary starting point that we are currently pegging to Activity onResume?

embrace-android-instrumentation-vitals/src/main/kotlin/io/embrace/android/embracesdk/internal/vitals/VitalsDataSource.kt
```diff
@@ -0,0 +1,135 @@
...
+    }
+
+    internal fun displayRefreshIntervalNanos(): Long {
+        val refreshRate = try {
+            args.systemService<DisplayManager>(Context.DISPLAY_SERVICE)
+                ?.getDisplay(Display.DEFAULT_DISPLAY)
+                ?.refreshRate
+                ?: NORMALIZED_REFRESH_RATE
+        } catch (e: Throwable) {
+            logger.logError("Failed to read display refresh rate; defaulting to 60Hz", e)
```
bidetofevil: Will this fire multiple times? If so, could be pretty noisy (not sure if this logs an internal error...)

embrace-android-instrumentation-vitals/src/main/kotlin/io/embrace/android/embracesdk/internal/vitals/VitalsFrameMetricsListener.kt
```diff
@@ -0,0 +1,33 @@
...
+        window: Window,
+        frameMetrics: FrameMetrics,
+        dropCountSinceLastInvocation: Int,
+    ) {
+        try {
+            val vsyncNanos = frameMetricsStrategy.vsyncNanos(frameMetrics)
+            val jankNanos = frameMetricsStrategy.jankNanos(frameMetrics)
+            focalCallbacks.onFrame(vsyncNanos, jankNanos)
+        } catch (e: Throwable) {
+            logger.logError("Failed to process frame metrics", e)
```
bidetofevil: Same comment about frequency of this potentialy happening

embrace-android-instrumentation-vitals/src/main/kotlin/io/embrace/android/embracesdk/internal/vitals/VitalsDataSource.kt
```diff
@@ -0,0 +1,135 @@
...
+        handlerThread = thread
+        val handler = Handler(thread.looper)
+
+        // Below API 31 the budget comes from [refreshRate]; track refresh-rate changes to keep it live.
+        if (Build.VERSION.SDK_INT < Build.VERSION_CODES.S) {
+            args.systemService<DisplayManager>(Context.DISPLAY_SERVICE)
+                ?.registerDisplayListener(displayListener, handler)
+        }
+
+        val vitalsScheduler = HandlerVitalsScheduler(handler)
```
bidetofevil: Can the fact that a Handler is used be hidden inside the `HandlerVitalsScheduler` so the instrumentation don't need to care/worry about the threading model?

embrace-android-instrumentation-vitals/src/test/kotlin/io/embrace/android/embracesdk/internal/vitals/FrameMetricsStrategyTest.kt
```diff
@@ -0,0 +1,124 @@
...
+import org.junit.Assert.assertTrue
+import org.junit.Test
+import org.junit.runner.RunWith
+import org.robolectric.util.ReflectionHelpers
+
+@RunWith(AndroidJUnit4::class)
+internal class FrameMetricsStrategyTest {
+
+    private fun setSdkInt(level: Int) {
+        ReflectionHelpers.setStaticField(Build.VERSION::class.java, "SDK_INT", level)
```
bidetofevil: Is this to avoid roboletric? FYI, Android 17 prevents this :sweat_smile:

embrace-android-instrumentation-vitals/src/test/kotlin/io/embrace/android/embracesdk/internal/vitals/FrameMetricsStrategyTest.kt
```diff
@@ -0,0 +1,124 @@
...
+import org.junit.Assert.assertEquals
+import org.junit.Assert.assertTrue
+import org.junit.Test
+import org.junit.runner.RunWith
+import org.robolectric.util.ReflectionHelpers
+
+@RunWith(AndroidJUnit4::class)
+internal class FrameMetricsStrategyTest {
+
+    private fun setSdkInt(level: Int) {
```
bidetofevil: Stylistic nit: we normally put helper methods at the bottom of the class so we get the test/public methods first

embrace-android-instrumentation-vitals/src/test/kotlin/io/embrace/android/embracesdk/internal/vitals/RecordingFocalInteractionCallbacks.kt
```diff
@@ -0,0 +1,17 @@
+package io.embrace.android.embracesdk.internal.vitals
+
+/** Test double that counts the signals delivered to a [FocalInteractionCallbacks]. */
+internal class RecordingFocalInteractionCallbacks : FocalInteractionCallbacks {
```
bidetofevil: To denote its test-only usage, we usually use the naming convention of `Fake<Interface>` for these types of classes.

embrace-android-semconv/src/main/kotlin/io/embrace/android/embracesdk/semconv/EmbSmoothnessAttributes.kt
```diff
@@ -0,0 +1,30 @@
...
+ * DO NOT EDIT - AUTOGENERATED FILE
+ */
+@Suppress("unused")
+object EmbSmoothnessAttributes {
+
+    /**
+     * Number of frames rendered during the focal moment.
+     */
+    @ExperimentalSemconv
+    const val SMOOTHNESS_FRAME_COUNT: String = "smoothness.frame_count"
```
bidetofevil, line 17: We can revisit these names when we get closer to making this public and we know exactly what shape we want this data in. Ideally, we would use names that we can turn in to OTel semconv without needing to change

bidetofevil (commented):
I would add a remote feature flag (default to off) and ship this!

embrace-android-instrumentation-vitals/src/main/kotlin/io/embrace/android/embracesdk/internal/vitals/FocalInteractionCallbacks.kt
lemnik: Frame timings are reported as nanosecond resolution.

embrace-android-config/src/main/kotlin/io/embrace/android/embracesdk/internal/config/behavior/AutoDataCaptureBehaviorImpl.kt
```diff
@@ -56,4 +57,7 @@ class AutoDataCaptureBehaviorImpl(
 
     override fun isNavigationStateCaptureEnabled(): Boolean =
         thresholdCheck.isBehaviorEnabled(remote?.pctNavigationStateCaptureEnabled) ?: NAVIGATION_STATE_CAPTURE_ENABLED_DEFAULT
+
+    override fun isSmoothnessCaptureEnabled(): Boolean =
```
bidetofevil, line 61: Do you want one flag for each POC vital you add or just a generic one for all of these we're going to test?

embrace-android-instrumentation-vitals/src/test/kotlin/io/embrace/android/embracesdk/internal/vitals/FrameMetricsStrategyTest.kt
```diff
@@ -0,0 +1,107 @@
...
+
+    private fun strategy(intervalNanos: Long = 16_000_000L, multiplier: Double = 2.0) =
+        FrameMetricsStrategy.create(intervalNanos, multiplier)
+
+    /**
+     * A mocked [FrameMetrics] that returns the given values for the metrics the strategies read. The
+     * `*_FRAME`/`DEADLINE` constants are compile-time ints, so this works even on SDKs where those
+     * metrics don't exist — the strategy under test simply won't query them there.
+     */
+    private fun frameMetrics(
```
bidetofevil, line 112: Any reason why we don't just create a FakeFrameMetric so we don't have to set up mocks?

embrace-android-instrumentation-vitals/src/test/kotlin/io/embrace/android/embracesdk/internal/vitals/smoothness/FocalMomentTrackerTest.kt
```diff
@@ -0,0 +1,268 @@
...
+        tracker.onInteractionStart()
+        redraw(vsyncNanos = start + 1_000_000L, jankNanos = 1_000_000_000L / 60) // one dropped 60fps frame -> 1.0
+        tracker.onScreenStop()
+
+        assertEquals(1.0, emitted.single().normalizedDroppedFrames, 0.01)
+    }
+
+    // The tracker reads time from SystemClock.uptimeMillis; Robolectric keeps it paused until advance()
+    // moves it, so vsync times are expressed relative to nowNanos().
+    private fun nowNanos(): Long = SystemClock.uptimeMillis() * 1_000_000L
```
bidetofevil: Is it possible to use the clock instance used elsewhere in the SDK so we test this determinstically with a faked instance later on?

embrace-android-payload/src/main/kotlin/io/embrace/android/embracesdk/internal/config/cache/CachedConfiguration.kt
```diff
@@ -5,7 +5,7 @@ import io.embrace.android.embracesdk.internal.serialization.BinaryVersion
 import kotlinx.serialization.Serializable
 
 @Serializable
-@BinaryVersion(-7272683289759519678L)
+@BinaryVersion(-4156899519244063313)
```
bidetofevil, line 8: Do we need to do this manually? I thought the compiler just takes care of this for us? Or are we interpreting this differently so the compiler-generated signatures might match, but semantically it is different if we change this?

bidetofevil (approved):
Added a few non-blocking comments. Happy to ship this as it is give it's behind a feature flag

embrace-android-payload/src/main/kotlin/io/embrace/android/embracesdk/internal/config/cache/CachedConfiguration.kt
lemnik: We can't get the compiler to do this for us without a symbol/annotation processor, there's a unit test that asserts that the value against a rule and fails if they don't match so it's a matter of copy-pasting the result when it does fail.

embrace-android-instrumentation-vitals/src/test/kotlin/io/embrace/android/embracesdk/internal/vitals/FrameMetricsStrategyTest.kt
lemnik: It's a `final` class with only a copy-constructor.

PR #3473, opened 2026-06-16
Fix flakey AEI test

bidetofevil:
So the delivery order validation isn't really validating the right thing - the deilvery order is based on the time a payload is created (i.e. the timestamp on the file name), not the timestamps of the logs inside that payload, which is what we were validating against. Typically, this technically-wrong timestamp is a good enough proxy since the times the payloads are created is in the same order as the times the correponding logs are recorded, assuming there are no network issues changing the order. To truly validate the thing the right sort order is being adherd to in delivery, we can't look at the log timestamp.

But doing that is gonna take some plumbing changes, and I don't think this is the right time to do it. So to solve the flake, which happens in AEI being while the logs recorded with increasing timestamps, they are all written within the same millisecond, so their priorities are the same when it comes to the time. So to break the tie, it uses UUID, which isn't stable (because of TestUuid generator isn't threadsafe), hence we get stuff out of order.  
  
The quickest way to solving the flake is to opt out of this order assertion, which is what I've done :sweat_smile:

### DIFF
--- a/embrace-android-sdk/src/integrationTest/kotlin/io/embrace/android/embracesdk/testcases/features/AeiFeatureTest.kt
+++ b/embrace-android-sdk/src/integrationTest/kotlin/io/embrace/android/embracesdk/testcases/features/AeiFeatureTest.kt
@@ -15,13 +15,13 @@ import io.embrace.android.embracesdk.fakes.TestAeiData
 import io.embrace.android.embracesdk.fakes.config.FakeEnabledFeatureConfig
 import io.embrace.android.embracesdk.fakes.config.FakeInstrumentedConfig
 import io.embrace.android.embracesdk.fakes.setupFakeAeiData
-import io.embrace.android.embracesdk.semconv.EmbAndroidAttributes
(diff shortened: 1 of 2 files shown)


bidetofevil:
* **#3473** <a href="https://app.graphite.com/github/pr/embrace-io/embrace-android-sdk/3473?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/embrace-io/embrace-android-sdk/3473?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

lemnik (approved):
LGTM

bidetofevil:
### Merge activity

* **Jun 16, 2:38 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/embrace-io/embrace-android-sdk/3473).
* **Jun 16, 2:38 PM UTC**: @bidetofevil merged this pull request with [Graphite](https://app.graphite.com/github/pr/embrace-io/embrace-android-sdk/3473).

PR #3474, opened 2026-06-16
Move OkHttpClient init off the main thread

lemnik:
## Goal
Move the initialisation of the `OkHttpClient` off the main thread. We never actually use of the `OkHttpClient` on the main thread, but we have been forcing its initialization on the main thread by directly passing it to its use sites.

This PR makes the `OkHttpClient` references `Lazy<OkHttpClient>` and also changes the `LazyThreadSafetyMode` to `SYNCHRONIZED` for this one instance (where `singleton` is `PUBLISHED`). This defers the initialization to the first use and as a side-effect moves the init to an IO worker thread.

## Testing
Measured in macrobenchmarks and updated the related unit tests.


bidetofevil (approved):
LGTM.  
  
Do we use this client for the config request we send as soon as we startup? If we have to do that, we should just spin up a new instance in the background thread that does that rather than touch this.

PR #3477, opened 2026-06-16
Transition to background-only user session if app inactive

bidetofevil:
Change `SessionOrchestratorImpl` to support creating background-only user sessions if appropriate. This includes making the initial startup user session one where we haven't classified as background-only yet, and instead do that when we foreground or when a fallback timer fires that expresses we've waited long enough (currently 5 seconds after the SDK has initiailized, since that's the only reliable time we can trigger this). This fallback timer will also classify the startup type.

As a result of this the implementation was refactored so the transitions go through common paths. New tests have been added to cover new edge cases, while existing ones are verified via existing testing.


bidetofevil:
* **#3477** <a href="https://app.graphite.com/github/pr/embrace-io/embrace-android-sdk/3477?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/embrace-io/embrace-android-sdk/3477?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

bidetofevil:
Opened another PR to track this for Graphite reasons

PR #3480, opened 2026-06-16
Transition to background-only user session if app inactive

bidetofevil:
Change `SessionOrchestratorImpl` to support creating background-only user sessions if appropriate. This includes making the initial startup user session one where we haven't classified as background-only yet, and instead do that when we foreground or when a fallback timer fires that expresses we've waited long enough (currently 5 seconds after the SDK has initiailized, since that's the only reliable time we can trigger this). This fallback timer will also classify the startup type.

As a result of this the implementation was refactored so the transitions go through common paths. New tests have been added to cover new edge cases, while existing ones are verified via existing testing.

### DIFF
--- a/embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/session/orchestrator/SessionOrchestratorImpl.kt
+++ b/embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/session/orchestrator/SessionOrchestratorImpl.kt
@@ -68,6 +68,7 @@ internal class SessionOrchestratorImpl(
     private val lock = Any()
     private val userSessionListeners = CopyOnWriteArrayList<UserSessionListener>()
 
+    @Volatile
     private var state = appStateTracker.getAppState()
 
     @Volatile
@@ -151,11 +152,13 @@ internal class SessionOrchestratorImpl(
     }
 
     override fun onForeground() {
-        val transitionType = synchronized(lock) {
+        // Hold the as we cancel the inactivity timer and determine the transition type. The lock would be acquired in transitionState()
+        // anyway so we're just handing on for the entire duration instead of dropping it and reacquiring it again.
+        synchronized(lock) {
             inactivityTimerState?.cancel()
             inactivityTimerState = null
             val metadata = currentUserSession()
-            if (metadata?.isBackgroundOnly == true) {
+            val transitionType = if (metadata?.isBackgroundOnly == true) {
                 TransitionType.BACKGROUND_ONLY_SESSION_END
             } else if (metadata?.isInactive(clock) == true) {
                 TransitionType.INACTIVITY_FOREGROUND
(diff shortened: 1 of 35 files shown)


bidetofevil:
* **#3480** <a href="https://app.graphite.com/github/pr/embrace-io/embrace-android-sdk/3480?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/embrace-io/embrace-android-sdk/3480?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/session/orchestrator/SessionOrchestratorImpl.kt
```diff
@@ -142,11 +152,13 @@ internal class SessionOrchestratorImpl(
     }
 
     override fun onForeground() {
-        val transitionType = synchronized(lock) {
+        // Hold the as we cancel the inactivity timer and determine the transition type. The lock would be acquired in transitionState()
+        // anyway so we're just handing on for the entire duration instead of dropping it and reacquiring it again.
```
lemnik:
```suggestion
        // Hold the lock as we cancel the inactivity timer and determine the transition type. The lock will be acquired in transitionState()
        // anyway so we're just holding it for the entire duration instead of dropping it and reacquiring it again.
```

embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/session/orchestrator/SessionOrchestratorImpl.kt
```diff
@@ -251,16 +263,17 @@ internal class SessionOrchestratorImpl(
...
     override fun addUserSessionListener(listener: UserSessionListener) {
-        userSessionListeners.add(listener)
-        val state = userSessionState
-        if (state is UserSessionState.Active) {
-            try {
-                listener.onSessionStateEvent(SessionStateEvent.UserSessionActive(state.metadata.userSessionId))
-            } catch (e: Exception) {
-                logger.trackInternalError(InternalErrorType.UserSessionCallbackFail, e)
+        synchronized(lock) {
+            userSessionListeners.add(listener)
```
lemnik, line 263: Given that `userSessionListeners` is a `CopyOnWriteArrayList` I'm not completely sure what the lock is doing here?

embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/session/orchestrator/SessionOrchestratorImpl.kt
```diff
@@ -76,6 +83,9 @@ internal class SessionOrchestratorImpl(
     @Volatile
     private var maxDurationTimerState: SessionTimerState? = null
 
+    @Volatile
```
lemnik: I only see `backgroundStartupWindowTimerState` being accessed / assigned under lock, in which case `@Volatile` isn't needed.

lemnik (approved):
LGTM with some minor suggestions

embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/session/orchestrator/SessionOrchestratorImpl.kt
bidetofevil: Yeah I think all of these timers are accessed under the lock already after my race condition fixes, so I can see if I can get rid of the other two as well

embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/session/orchestrator/SessionOrchestratorImpl.kt
bidetofevil: Will modify

embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/session/orchestrator/SessionOrchestratorImpl.kt
bidetofevil: It probably doesn't need to be a CopyOnWriteArrayList anymore, but it's the firing of the event listener - it can be fired twice if a new event arrives just after it's added if we don't lock it.

bidetofevil:
### Merge activity

* **Jun 17, 5:05 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/embrace-io/embrace-android-sdk/3480).
* **Jun 17, 5:05 PM UTC**: @bidetofevil merged this pull request with [Graphite](https://app.graphite.com/github/pr/embrace-io/embrace-android-sdk/3480).

PR #3488, opened 2026-06-18
Bump Kotlin Compiler -> 2.4.0

lemnik:
## Goal
Upgrade the Kotlin plugin & compiler to 2.4.0.

This had a knock-on in the Gradle Plugin where we were previously working-around the non-null return requirement in `Provider`. As such this PR also refactors the Gradle Plugin code to adhere to Gradles "empty" `Provider` contract (where `map` & `flatMap` may return `null` but not `Provider` implementations).

## Testing
Relied on the existing full test suite.


embrace-bytecode-instrumentation-tests/src/test/resources/KotlinInnerListener_expected.txt
```diff
@@ -14,8 +14,6 @@ public final class io/embrace/test/fixtures/KotlinNested$KotlinInnerListener imp
   final synthetic Lio/embrace/test/fixtures/KotlinNested; this$0
 
   // access flags 0x1
-  // signature ()V
-  // declaration: void <init>()
```
lemnik, line 18: This is down to a change in how the Kotlin compiler emits `inner class` constructors

bidetofevil (approved):
So much nicer not having to deal with nulls

PR #3490, opened 2026-06-19
Bump the compileSdk = 37

lemnik:
## Goal
Bump the `compileSdk` to `37` so that we're building against Android 17

## Testing
Relied on existing tests


bidetofevil (approved):
LGTM

PR #3493, opened 2026-06-19
NDK: Avoid secondary crashes when copying the stack

lemnik:
## Goal
Not every frame has `MapInfo` reported for it so we need to null-check these frames to avoid a segmentation fault when handling a native crash.

## Testing
Manually tested


bidetofevil (approved):
Well that seems reasonable

PR #3514, opened 2026-06-23
Don't create background only user sessions when background activity is disabled

bidetofevil:
Only create a new background only user session if background activity is enabled. The runtime difference is that the user session number will not increment if we leave the SDK without a user session when one ends and the app is in the background AND background activity is disabled. This is consistent with the idea that we should only have a user session if the user is active OR if the background activity is enabled.  
  
Added tests to cover various BA-off scenarios, and more broadly assert user session numbers and session part numbers when we can

### DIFF
--- a/embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/session/orchestrator/SessionOrchestratorImpl.kt
+++ b/embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/session/orchestrator/SessionOrchestratorImpl.kt
@@ -572,8 +572,8 @@ internal class SessionOrchestratorImpl(
             // User session being ended explicitly - end current one and start a new one based on the expected endAppState.
             transitionType.endsUserSession -> {
                 clearBackgroundStartupWindowTimer()
-                if (endAppState == AppState.BACKGROUND && !configService.backgroundActivityBehavior.isBackgroundActivityCaptureEnabled()
-                ) {
+                if (endAppState == AppState.BACKGROUND &&
+                    !configService.backgroundActivityBehavior.isBackgroundActivityCaptureEnabled()) {
                     terminateActiveUserSession()
                 } else {
                     transitionToNewUserSession(timestamp, endAppState)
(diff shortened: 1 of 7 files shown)


bidetofevil:
* **#3514** <a href="https://app.graphite.com/github/pr/embrace-io/embrace-android-sdk/3514?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/embrace-io/embrace-android-sdk/3514?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/session/orchestrator/SessionOrchestratorImpl.kt
```diff
@@ -572,7 +572,12 @@ internal class SessionOrchestratorImpl(
             // User session being ended explicitly - end current one and start a new one based on the expected endAppState.
             transitionType.endsUserSession -> {
                 clearBackgroundStartupWindowTimer()
-                transitionToNewUserSession(timestamp, endAppState)
+                if (endAppState == AppState.BACKGROUND && !configService.backgroundActivityBehavior.isBackgroundActivityCaptureEnabled()
+                ) {
```
lemnik: nit: formatting

```suggestion
                if (endAppState == AppState.BACKGROUND &&
                    !configService.backgroundActivityBehavior.isBackgroundActivityCaptureEnabled()) {
```

lemnik (approved):
LGTM

embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/session/orchestrator/SessionOrchestratorImpl.kt
bidetofevil: Oh gross. I must've accidentally hit the enter key or something. Also why didn't detekt catch this?

bidetofevil: Wait. Detekt doesn't like this. Hmm. I'll reformat it locally in a way that detekt DOES like it

bidetofevil:
### Merge activity

* **Jun 24, 2:50 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/embrace-io/embrace-android-sdk/3514).
* **Jun 24, 2:50 PM UTC**: @bidetofevil merged this pull request with [Graphite](https://app.graphite.com/github/pr/embrace-io/embrace-android-sdk/3514).

PR #3548, opened 2026-07-03
Refactor the EmbraceDomainCountLimiter for closest-match & remove locking

lemnik:
## Goal

Replace the "last suffix matches" behaviour with a "closest suffix matches" logic when tracking domain log limits.

The PR also replaces the global lock in `EmbraceDomainCountLimiter` with coupled "limit & count" encapsulation (`LimitCounter`).

## Testing
Added unit tests


embrace-android-infra/src/main/kotlin/io/embrace/android/embracesdk/internal/utils/concurrent/LimitCounter.kt
```diff
@@ -0,0 +1,75 @@
+package io.embrace.android.embracesdk.internal.utils.concurrent
+
+import java.util.concurrent.atomic.AtomicLong
+
+/**
+ * Thread-safe integer based limit counter.
+ */
+@JvmInline
+value class LimitCounter private constructor(private val combined: AtomicLong) {
```
bidetofevil, line 10: Need tests for this or will it be effectively tested with `EmbraceDomainCountLimiterTest`?

bidetofevil (approved):
LGTM about tests. I'm surprised there were not unit tests around the domain counter. We must've been relying on its consumer to do the testing?

embrace-android-infra/src/main/kotlin/io/embrace/android/embracesdk/internal/utils/concurrent/LimitCounter.kt
lemnik: It would be somewhat tested, but you're right - a dedicated test would be much better.

lemnik:
> LGTM about tests. I'm surprised there were not unit tests around the domain counter. We must've been relying on its consumer to do the testing?

Yeah. I only see a reference to it in the `HucLiteDataSourceTest`

PR #3568, opened 2026-07-08
Added NOTICE file

fnewberg:
## Goal

Added NOTICE.md file with licenses for SDK dependencies


bidetofevil (approved):
LGTM. Is the gradle task `generateNotice` already added to the project and run automatically if required or is that something we still have to do?

PR #3585, opened 2026-07-14
Allow more remote config over vital behaviour

lemnik:
## Goal
Add remote config for the Smoothness and ScreenLoad vital experiments.

Each remotely configurable threshold and value is now reported as an attribute with each vital span to avoid any ambiguity.

The new `VitalsRemoteConfig` configuration allows us to tune:
- `smoothness_idle_threshold_ms`
- `smoothness_held_idle_threshold_ms`
- `jank_heuristic_multiplier`
- `screen_load_idle_threshold_ms`
- `screen_load_timeout_ms`
- `screen_load_nav_timeout_ms`

These are all tunable directly to avoid hard-coding experiment groups that may need code changes to update and deploy.,

## Testing
Added the new values to the existing config tests.


fractalwrench (commented):
CI has a failure around using symbols only available from API 24, but otherwise this looks good

fractalwrench (approved):
LGTM

PR #3596, opened 2026-07-15
Bump Lint version

fractalwrench:
## Goal

Builds on #3587 by bumping the lint version which requires altering the JVM target for the lint module.


lemnik (approved):
LGTM

PR #3597, opened 2026-07-15
Allow the build info to be exported for async mapping uploads

lemnik:
## Goal
Allow workflows where the `mapping.txt` or other symbols need to be uploaded only after post-build processing. The Gradle plugin can now be configured to export the build-id, app-id + token, and variant as a JSON file that will be usable in a standalone upload process outside of Gradle.

## Testing
New integration test added.

### DIFF
--- a/embrace-gradle-plugin/src/main/java/io/embrace/android/gradle/plugin/tasks/buildinfo/BuildInfoExport.kt
+++ b/embrace-gradle-plugin/src/main/java/io/embrace/android/gradle/plugin/tasks/buildinfo/BuildInfoExport.kt
@@ -1,0 +1,11 @@
+package io.embrace.android.gradle.plugin.tasks.buildinfo
+
+import com.squareup.moshi.JsonClass
+
+@JsonClass(generateAdapter = true)
+data class BuildInfoExport(
+    val buildId: String,
+    val appId: String,
+    val apiToken: String,
+    val variantName: String,
+)
--- a/embrace-gradle-plugin/src/main/java/io/embrace/android/gradle/plugin/tasks/buildinfo/BuildInfoExport.kt
+++ b/embrace-gradle-plugin/src/main/java/io/embrace/android/gradle/plugin/tasks/buildinfo/BuildInfoExport.kt
@@ -6,6 +6,5 @@ import com.squareup.moshi.JsonClass
 data class BuildInfoExport(
     val buildId: String,
     val appId: String,
-    val apiToken: String,
     val variantName: String,
 )
--- a/embrace-gradle-plugin/src/main/java/io/embrace/android/gradle/plugin/tasks/buildinfo/ExportBuildInfoTask.kt
+++ b/embrace-gradle-plugin/src/main/java/io/embrace/android/gradle/plugin/tasks/buildinfo/ExportBuildInfoTask.kt
@@ -1,0 +1,61 @@
+package io.embrace.android.gradle.plugin.tasks.buildinfo
+
+import io.embrace.android.gradle.plugin.model.AndroidCompactedVariantData
+import io.embrace.android.gradle.plugin.tasks.EmbraceTask
+import io.embrace.android.gradle.plugin.util.serialization.MoshiSerializer
+import org.gradle.api.DefaultTask
+import org.gradle.api.file.RegularFileProperty
+import org.gradle.api.model.ObjectFactory
+import org.gradle.api.provider.Property
+import org.gradle.api.tasks.Input
+import org.gradle.api.tasks.OutputFile
+import org.gradle.api.tasks.TaskAction
+import org.gradle.work.DisableCachingByDefault
+import javax.inject.Inject
+
+/**
+ * Writes the parameters that would be used to upload the mapping file and NDK symbols for a
+ * variant into a JSON file, so they can be reused by external tooling.
+ */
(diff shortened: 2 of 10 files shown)


embrace-gradle-plugin/src/main/java/io/embrace/android/gradle/plugin/tasks/buildinfo/BuildInfoExport.kt
```diff
@@ -0,0 +1,11 @@
+package io.embrace.android.gradle.plugin.tasks.buildinfo
+
+import com.squareup.moshi.JsonClass
+
+@JsonClass(generateAdapter = true)
+data class BuildInfoExport(
```
fractalwrench, line 6: This is a subset of `SdkLocalConfig`. It's probably worth discussing between us whether there's any utility in also exporting the embrace-config.json into this file, as it could be useful in debugging (or might be overkill)

embrace-gradle-plugin/src/main/java/io/embrace/android/gradle/plugin/tasks/buildinfo/ExportBuildInfoTaskRegistration.kt
```diff
@@ -0,0 +1,96 @@
...
+        }
+        return exportTask
+    }
+
+    /**
+     * It fetches all available obfuscation tasks. In practice, a single task will be
+     * executed but multiple tasks may be registered and configured. Because of this, we are
+     * forced to return a list of tasks since we are in configuration phase.
+     */
+    private fun fetchObfuscationTasks(
```
fractalwrench: It looks like there's commonality between this and `JvmMappingFileUploadTaskRegistration`

embrace-gradle-plugin/src/main/java/io/embrace/android/gradle/plugin/tasks/buildinfo/ExportBuildInfoTask.kt
```diff
@@ -0,0 +1,61 @@
...
+    fun onRun() {
+        val buildInfo = BuildInfoExport(
+            buildId = buildId.get(),
+            appId = appId.get(),
+            apiToken = apiToken.get(),
+            variantName = variantData.get().name,
+        )
+        outputFile.get().asFile.apply {
+            parentFile.mkdirs()
+            outputStream().use { serializer.toJson(buildInfo, BuildInfoExport::class.java, it) }
```
fractalwrench, line 49: Is this buffered?

lemnik: Yup (in `MoshiSerializer.toJson`)

embrace-gradle-plugin/src/main/java/io/embrace/android/gradle/plugin/tasks/buildinfo/BuildInfoExport.kt
```diff
@@ -0,0 +1,11 @@
...
+    val buildId: String,
+    val appId: String,
+    val apiToken: String,
```
fnewberg: i don't think we should be exporting the token since that is considered a secret. when a customer uses this data to upload the mapping file, they should be expected to show up with the API token and not have it be provided this way

embrace-gradle-plugin/src/main/java/io/embrace/android/gradle/plugin/tasks/buildinfo/ExportBuildInfoTask.kt
```diff
@@ -0,0 +1,61 @@
...
+    private val serializer = MoshiSerializer()
+
+    @get:Input
+    val buildId: Property<String> = objectFactory.property(String::class.java)
+
+    @get:Input
+    val appId: Property<String> = objectFactory.property(String::class.java)
+
+    @get:Input
+    val apiToken: Property<String> = objectFactory.property(String::class.java)
```
fnewberg: remove for reasons described above

embrace-gradle-plugin/src/main/java/io/embrace/android/gradle/plugin/tasks/buildinfo/BuildInfoExport.kt
bidetofevil: If we are going to get the `appToken` from another place other than the `BuildInfo`, we can probably do the same with `appId` as well, so this may end up just being the buildId, in which case it could just be a String rather than a data class

lemnik: Making it a plain string means there's little or no room for future expansion without some kind of format check.

fnewberg (approved):

embrace-gradle-plugin/src/main/java/io/embrace/android/gradle/plugin/tasks/buildinfo/BuildInfoExport.kt
bidetofevil: Keeping a data class seems reasonable. Works for me.

bidetofevil (approved):
LGTM. Never mind about moving `appId`

PR #3603, opened 2026-07-17
Fixed a time mismatch in the screen-load vital

lemnik:
## Goal
Fix a timing bug in the screen load vital where the time of the event should have been captured before posting the events to the vitals handler thread.

This corrects the timing so that events are always compared relative to when they actually happened (or at least when the `main` thread received them) instead of when the vitals thread received the event.

## Testing
Updated the unit tests to cover these cases.


fractalwrench (approved):
LGTM

PR #3604, opened 2026-07-17
Added frame timing data to the smoothness vital

lemnik:
## Goal
Allow the capturing of per-frame timing data in smoothness vitals, gated by a secondary threshold value under the `vitals` remote config block.

The frame time is taken only as the total fame time (vsync - dispatch) in milliseconds. The value is encoded in a buffer as a varint (LEB128) encoding so most frames will only consume a single byte. The buffer is limited to 1k bytes and is allocated exactly once per process as only one focal moment can be active at a time.

## Testing
Manually tested & added unit tests.


fractalwrench (approved):
LGTM

PR #3610, opened 2026-07-21
Minor huclite improvements

lemnik:
## Goal
Reduce the overheads of the `huclite` instrumentation, caching the reflected `Method` references in the `URLStreamHandler` implementations and inlining the `invokeRequestTriggeringMethod` to avoid allocating a lambda for every wrapped function invocation.

## Testing
Relied on existing tests


embrace-android-instrumentation-huc-lite/src/main/kotlin/io/embrace/android/embracesdk/instrumentation/huclite/LazyMethodLookup.kt
```diff
@@ -0,0 +1,48 @@
...
+    @Throws(IOException::class)
+    operator fun getValue(thisRef: Any?, property: KProperty<*>): Method {
+        when (val s = state) {
+            is Method -> return s
+            is MethodLookup -> {
+                try {
+                    val method = s.resolve()
+                    state = method
+                } catch (_: NoSuchMethodException) {
+                    val error = IOException("Failed to instrument ${s.parentClass.simpleName}")
```
fractalwrench, line 27: Is this a behavior change to how reflection was handled currently? I.e. does this change get propagated to customers and show in their dashboards?

lemnik: There is a small change in behaviour here, but it's in the message of the exception & the fact that all of they are now consistently wrapped in `IOException`. Previous behaviour here:

https://github.com/embrace-io/embrace-android-sdk/blob/main/embrace-android-instrumentation-huc-lite/src/main/kotlin/io/embrace/android/embracesdk/instrumentation/huclite/ReflectionUtils.kt#L17

https://github.com/embrace-io/embrace-android-sdk/blob/main/embrace-android-instrumentation-huc-lite/src/main/kotlin/io/embrace/android/embracesdk/instrumentation/huclite/InstrumentedHttpsURLStreamHandler.kt#L55

https://github.com/embrace-io/embrace-android-sdk/blob/main/embrace-android-instrumentation-huc-lite/src/main/kotlin/io/embrace/android/embracesdk/instrumentation/huclite/DelegatingInstrumentedUrlStreamHandler.kt#L49

fractalwrench (approved):

PR #3611, opened 2026-07-21
Cache pattern compilation

fractalwrench:
## Goal

Caches pattern compilation so that regexes aren't recompiled for every network request.

## Testing

Added unit tests.

### DIFF
--- a/embrace-android-config/src/main/kotlin/io/embrace/android/embracesdk/internal/config/behavior/NetworkBehaviorImpl.kt
+++ b/embrace-android-config/src/main/kotlin/io/embrace/android/embracesdk/internal/config/behavior/NetworkBehaviorImpl.kt
@@ -1,11 +1,11 @@
 package io.embrace.android.embracesdk.internal.config.behavior
 
+import io.embrace.android.embracesdk.internal.config.PatternCache
 import io.embrace.android.embracesdk.internal.config.instrumented.schema.InstrumentedConfig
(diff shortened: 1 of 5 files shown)


embrace-android-config/src/main/kotlin/io/embrace/android/embracesdk/internal/config/behavior/NetworkBehaviorImpl.kt
```diff
@@ -65,11 +66,10 @@ class NetworkBehaviorImpl(
...
 
     override fun isUrlEnabled(url: String): Boolean {
-        val patterns = disabledUrlPatterns ?: remote?.disabledUrlPatterns ?: cfg.getIgnoredRequestPatternList()
-        val regexes = patterns.mapNotNull {
-            runCatching { Pattern.compile(it) }.getOrNull()
-        }.toSet()
-        return regexes.none { it.matcher(url).find() }
+        val patterns = disabledUrlPatterns?.toSet()
+            ?: remote?.disabledUrlPatterns
+            ?: cfg.getIgnoredRequestPatternList().toSet()
```
lemnik: The `toSet()` calls here introduce a new overhead for every call of converting the `List` to a `LinkedHashSet`. Would it possibly make sense for the `PatternCache` to key on a `Collection<String>` instead of strictly requiring a `Set`? I feel like moving the de-duplication into `compiledPatterns` would be reasonable, although we might introduce some cache misses as a result.

lemnik (approved):
LGTM

PR #3613, opened 2026-07-21
Close OkHttp remote config response

fractalwrench:
## Goal

Closes the OkHttp remote config response regardless of status code.

## Testing

Added unit tests


bidetofevil (approved):
Interesting. Did not realize this was the case

PR #3618, opened 2026-07-22
Avoid leaking OkHttp requests that don't hit network interceptor

fractalwrench:
## Goal

OkHttp requests that don't hit network interceptor (such as those with a cache response) currently leak memory by `activeCalls` never removing a reference.

Additionally, a request was not recorded as telemetry for this scenario.

## Testing

Added unit tests.

### DIFF
--- a/embrace-android-instrumentation-okhttp/src/main/kotlin/io/embrace/android/embracesdk/internal/instrumentation/okhttp/OkHttpDataSource.kt
+++ b/embrace-android-instrumentation-okhttp/src/main/kotlin/io/embrace/android/embracesdk/internal/instrumentation/okhttp/OkHttpDataSource.kt
@@ -26,7 +26,8 @@ import okio.Buffer
 import okio.GzipSource
 import okio.buffer
 import java.io.IOException
-import java.util.concurrent.ConcurrentHashMap
+import java.util.Collections
+import java.util.WeakHashMap
 
 /**
  * Captures OkHttp requests as telemetry.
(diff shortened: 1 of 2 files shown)


embrace-android-instrumentation-okhttp/src/main/kotlin/io/embrace/android/embracesdk/internal/instrumentation/okhttp/OkHttpDataSource.kt
lemnik: I think that swapping the `ConcurrentHashMap` for a `WeakHashMap` is a reasonable extra line of defence here. The contention on this map should never be especially high, so a simple `synchronizedMap` should also be a reasonable tradeoff.

```suggestion
    private val activeCalls = Collections.synchronizedMap(WeakHashMap<Call, CallData>())
```

embrace-android-instrumentation-okhttp/src/main/kotlin/io/embrace/android/embracesdk/internal/instrumentation/okhttp/OkHttpDataSource.kt
```diff
@@ -88,8 +88,14 @@ internal class OkHttpDataSource(
...
         val request: Request = chain.request()
         return try {
-            // we are not interested in response, just proceed
-            chain.proceed(request)
+            val response = chain.proceed(request)
+            // The network interceptor removes the call from activeCalls when it records the result.
+            // If the entry is still present here, the network interceptor never ran (e.g. a cached
+            // response, a websocket handshake, etc), so needs cleaning up to avoid leaks.
+            if (callData != null && activeCalls.containsKey(chain.call())) {
+                callData.recordSkippedNetworkInterceptorResult(chain.call(), request, response)
```
fractalwrench: Discussed that it's probably optimal not to record a new span at this point, and instead just discard the span without recording.

bidetofevil: Has the span started? If so, we don't want to just fail/end it because we'll have spurious errors as a result.

fractalwrench: @bidetofevil the current behavior is that the span is not terminated and therefore leaks memory. The current change ends the span.

I think it'd be preferable to discard the span entirely as we don't want to record it. LMK if you disagree

bidetofevil: It definitely should be discarded entirely, as if we didn't start it. But the test seems to validate that a request is logged - shouldn't it verify that it isn't?

PR #3619, opened 2026-07-22
Avoid leaking view span references

fractalwrench:
## Goal

Avoids leaking view span references if a view was tracked via the manual API.

## Testing

Added unit tests


lemnik (approved):
LGTM

PR #3620, opened 2026-07-22
Http info modifier

lemnik:
## Goal
Allow captured HTTP request info to be modified before it is recorded as telemetry. This PR only covers HTTP method and URL for now, but the implementation is open to additional changes in future.

All existing HTTP instrumentation now constructs a `MutableHttpRequestInfo` implementation and passes it to any number of registered (and shared) `HttpRequestInfoModifier`s before recording the telemetry. This allows for easy redaction of the URL being recorded.

## Testing
Introduced new integration & unit tests for the new functionality.

### DIFF
--- a/embrace-android-api/src/main/kotlin/io/embrace/android/embracesdk/network/http/HttpRequestInfo.kt
+++ b/embrace-android-api/src/main/kotlin/io/embrace/android/embracesdk/network/http/HttpRequestInfo.kt
@@ -10,7 +10,7 @@ public interface HttpRequestInfo {
     public val url: String
 
     /**
-     * The request's method. Must be one of the following: GET, PUT, POST, DELETE, PATCH.
+     * The request's method (`GET`, `PUT`, `POST`, `DELETE`, `PATCH`, etc.).
      */
     public val httpMethod: String
 }
--- a/embrace-android-api/src/main/kotlin/io/embrace/android/embracesdk/network/http/HttpRequestInfo.kt
+++ b/embrace-android-api/src/main/kotlin/io/embrace/android/embracesdk/network/http/HttpRequestInfo.kt
@@ -1,0 +1,16 @@
+package io.embrace.android.embracesdk.network.http
+
+/**
+ * Information captured about an HTTP request by instrumentation.
+ */
+public interface HttpRequestInfo {
+    /**
+     * The request's URL. Must start with http:// or https://
+     */
+    public val url: String
+
+    /**
+     * The request's method. Must be one of the following: GET, PUT, POST, DELETE, PATCH.
+     */
+    public val httpMethod: String
+}
--- a/embrace-android-instrumentation-network-common/src/main/kotlin/io/embrace/android/embracesdk/internal/instrumentation/network/NetworkRequestDataSourceImpl.kt
+++ b/embrace-android-instrumentation-network-common/src/main/kotlin/io/embrace/android/embracesdk/internal/instrumentation/network/NetworkRequestDataSourceImpl.kt
@@ -7,6 +7,7 @@ import io.embrace.android.embracesdk.internal.arch.limits.NoopLimitStrategy
 import io.embrace.android.embracesdk.internal.arch.schema.EmbType
 import io.embrace.android.embracesdk.internal.arch.schema.ErrorCodeAttribute
 import io.embrace.android.embracesdk.internal.arch.schema.SchemaType
+import io.embrace.android.embracesdk.internal.network.http.MutableHttpRequestInfoImpl
 import io.embrace.android.embracesdk.internal.network.logging.DomainCountLimiter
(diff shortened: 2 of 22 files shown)


embrace-android-api/src/main/kotlin/io/embrace/android/embracesdk/network/http/HttpRequestInfo.kt
```diff
@@ -0,0 +1,16 @@
...
+ * Information captured about an HTTP request by instrumentation.
+ */
+public interface HttpRequestInfo {
+    /**
+     * The request's URL. Must start with http:// or https://
+     */
+    public val url: String
+
+    /**
+     * The request's method. Must be one of the following: GET, PUT, POST, DELETE, PATCH.
```
fractalwrench: Do we want to support other methods or is that all our instrumentation supports right now?

embrace-android-sdk/src/main/kotlin/io/embrace/android/embracesdk/internal/api/delegate/NetworkRequestApiDelegate.kt
```diff
@@ -28,6 +32,18 @@ internal class NetworkRequestApiDelegate(
         }
     }
 
+    override fun addHttpRequestInfoModifier(modifier: HttpRequestInfoModifier) {
```
fractalwrench, line 35: This means requests that happen after SDK init but before this function is invoked could contain PII

embrace-android-api/src/main/kotlin/io/embrace/android/embracesdk/network/http/HttpRequestInfo.kt
lemnik: That doc was copy-pasta from the `EmbraceNetworkRequest`. I've updated the doc to be more relaxed.

embrace-android-sdk/src/main/kotlin/io/embrace/android/embracesdk/internal/api/delegate/NetworkRequestApiDelegate.kt
lemnik: I've added to the KDoc for the public `NetworkRequestApi.addHttpRequestInfoModifier` to point out this limitation. I think it's a reasonable limitation as it pairs with the existence of the `removeHttpRequestInfoModifier` function.

fractalwrench (commented):
Happy in principle but would like @bidetofevil's opinions on whether we need a way to set callbacks before the SDK has initialized

embrace-android-instrumentation-network-common/src/main/kotlin/io/embrace/android/embracesdk/internal/instrumentation/network/NetworkRequestDataSourceImpl.kt
```diff
@@ -76,13 +84,19 @@ class NetworkRequestDataSourceImpl(
     }
 
     override fun startRequest(startData: RequestStartData): String? {
-        if (!configService.networkBehavior.isUrlEnabled(startData.url)) {
+        // Apply any registered modifiers so the reported url/method reflect the consumer's changes.
+        // The underlying HTTP request is not affected.
+        val info = httpRequestInfoModifierChain.apply(MutableHttpRequestInfoImpl(startData.httpMethod, startData.url))
```
bidetofevil, line 89: Should the filtering and limiting be applied to the URL before or after the modificaiton chain? I kind of feel it's the former since the change is applied to what the instrumentation says, but what you are limiting/filtering is the actual domain/path of the original request.

embrace-android-sdk/src/main/kotlin/io/embrace/android/embracesdk/internal/api/delegate/NetworkRequestApiDelegate.kt
bidetofevil: Any harm if we allow these to be added before the SDK starts? We typically force all setup to happen prior to start being called (e.g. otel exporter) partly because that's just how some OTel stuff like resources work, but also to ensure there's consistency and that processors and exporters are applied uniformly.  
  
If there is some cascading effect that initializes a bunch of stuff or asks for things that aren't initialized, can we buffer the calls and add them during startup so we can achieve that?

embrace-android-instrumentation-network-common/src/main/kotlin/io/embrace/android/embracesdk/internal/instrumentation/network/NetworkRequestDataSourceImpl.kt
lemnik: Yeah. I think you're right - will patch that

fractalwrench (approved):
LGTM

PR #3621, opened 2026-07-22
Add remove API for user session listener

fractalwrench:
## Goal

Adds a remove API for the user session listener so that it's possible to deregister.

## Testing

Added unit tests


bidetofevil (approved):
LGTM

PR #3622, opened 2026-07-22
Improve RN bundle hashing performance

fractalwrench:
## Goal

RN bundle IDs are hashed using MD5 to identify the source file. Android's implementation was inefficient as it loaded the entire file into memory & processed it using a small chunk size. I've addressed those problems in this changeset which should speed up RN init & also reduce the chance of OOMs if the bundle is large.

## Testing


hamudi-ahmed (approved):
Nice! LGTM

bidetofevil (approved):
LGTM

PR #3623, opened 2026-07-22
Close payload file when delivering

fractalwrench:
## Goal

Alters the scheduling service so the file is closed after a request has finished, preventing a resource leak.

## Testing

Added unit tests.


bidetofevil (approved):
LGTM

PR #3625, opened 2026-07-22
Set limits on session span event count

fractalwrench:
## Goal

Alters the limits on the session span so the maximum system event count is set at a lower, more reasonable limit, and the system attribute count is bounded.

## Testing

Updated tests.


bidetofevil (commented):
How did you get to this limit? The original one was based on some apps that can set their ceilings to things like breadcrumbsa a lot higher.   
  
Perhaps we can derive this based on the config being applied?

fractalwrench:
@bidetofevil do you have context on what sort of numbers we're talking about for breadcrumbs? I assumed all folks would be <1000 in normal usage, particularly given that a 'system event' is meant to be controlled by us. What are your opinions on the system attribute limits?

bidetofevil:
Some orgs can set breadcrumbs to as high as 10K, I believe.  It's based on the ultimate ceiling, which is effectively that.  
  
I assume this was validated somewhere before, but I guess it just wasn't?

lemnik (approved):
LGTM

bidetofevil (approved):
The current max of the data Fredric showed has 1000 events on spans - that's likely one of the state spans, so on the session span, which the max was designed for, this didn't even come close.  
  
I think I'm ok with this so long a the we don't have to support 10K breadcrumb apps

embrace-android-otel/src/main/kotlin/io/embrace/android/embracesdk/internal/otel/spans/EmbraceSpanFactoryImpl.kt
```diff
@@ -355,8 +355,17 @@ private class EmbraceSpanImpl(
     }
 
     override fun addSystemAttribute(key: String, value: String) {
-        systemAttributes[key] = value
-        spanRepository.notifySpanUpdate()
+        val max = dataValidator.otelLimitsConfig.getMaxSystemAttributeCount()
+        if (systemAttributes.containsKey(key) || systemAttributes.size < max) {
```
bidetofevil, line 422: I wonder if we could do a type-based limit like for spans, where we allow 1000 system events and 10K breadcrumb events, so you can only exceed 2K if you're putting massive amounts of breadcrumbs in, which we allow, but only in the session span?

fractalwrench:
Updated implementation as discussed internally so that the max number of session span events is 1000, and that breadcrumbs have a separate, remotely configurable limit.

bidetofevil (approved):
LGTM

fractalwrench:
Updated implementation further so that it's driven by remote config rather than local config. Functionally this otherwise remains the same.

PR #3626, opened 2026-07-22
Reduce AEI memory usage

fractalwrench:
## Goal

Reduce AEI memory usage by bounding the max trace size it can read (without affecting existing logic that records whether a trace exceeded the limit). Additionally reduces the max AEI count to AOSP's default.

## Testing

### DIFF
--- a/embrace-android-instrumentation-app-exit-info/src/main/kotlin/io/embrace/android/embracesdk/internal/instrumentation/aei/ApplicationExitInfoExt.kt
+++ b/embrace-android-instrumentation-app-exit-info/src/main/kotlin/io/embrace/android/embracesdk/internal/instrumentation/aei/ApplicationExitInfoExt.kt
@@ -107,7 +107,7 @@ private fun InputStream.readBytesCapped(charLimit: Int): ByteArray {
  * and truncated without ever reading the entire source into memory.
  */
 private fun Reader.readTextCapped(charLimit: Int): String {
-    val sb = StringBuilder()
+    val sb = StringBuilder(READ_BUFFER_SIZE)
     val buffer = CharArray(READ_BUFFER_SIZE)
     while (sb.length <= charLimit) {
         val count = read(buffer)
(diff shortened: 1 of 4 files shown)


embrace-android-instrumentation-app-exit-info/src/main/kotlin/io/embrace/android/embracesdk/internal/instrumentation/aei/ApplicationExitInfoExt.kt
```diff
@@ -72,15 +76,49 @@ private fun ApplicationExitInfo.readAeiTrace(
...
+    }
+    return out.toByteArray()
+}
+
+/**
+ * Reads at most one buffer beyond [charLimit] chars, so oversized traces can be detected
+ * and truncated without ever reading the entire source into memory.
+ */
+private fun Reader.readTextCapped(charLimit: Int): String {
+    val sb = StringBuilder()
```
lemnik: nit: maybe we should pre-size the `StringBuilder` as well?

```suggestion
    val sb = StringBuilder(READ_BUFFER_SIZE)
```

embrace-android-instrumentation-app-exit-info/src/main/kotlin/io/embrace/android/embracesdk/internal/instrumentation/aei/ApplicationExitInfoExt.kt
```diff
@@ -72,15 +76,49 @@ private fun ApplicationExitInfo.readAeiTrace(
...
+/**
+ * Reads at most one buffer beyond [charLimit] bytes, so oversized traces can be detected
+ * and truncated without ever reading the entire source into memory.
+ */
+private fun InputStream.readBytesCapped(charLimit: Int): ByteArray {
+    val out = ByteArrayOutputStream(READ_BUFFER_SIZE)
+    val buffer = ByteArray(READ_BUFFER_SIZE)
+    while (out.size() <= charLimit) {
+        val count = read(buffer)
+        if (count == -1) {
```
lemnik, line 97: nit: we can limit to strictly `charLimit` instead of a multiple of `READ_BUFFER_SIZE` pretty easily:

```suggestion
        val remaining = charLimit - out.size()
        val toRead = min(buffer.size, remaining)
        val count = read(buffer, 0, toRead)
        if (count == -1) {
```

fractalwrench: The goal is to exceed charLimit _slightly_ as the current business logic needs to detect when the AEI trace size was exceeded: https://github.com/embrace-io/embrace-android-sdk/blob/8ca9762f09c65b964a26078a24b2d33fa3a93b46/embrace-android-instrumentation-app-exit-info/src/main/kotlin/io/embrace/android/embracesdk/internal/instrumentation/aei/ApplicationExitInfoExt.kt#L62

lemnik (approved):
LGTM

PR #3627, opened 2026-07-22
Release ANR interval references

fractalwrench:
## Goal

Releases ANR interval references by clearing the least valuable sample whenever the max interval limit is exceeded. This means we should only retain 5 intervals with stacktraces rather than 100, and also avoid unnecessary allocations in `getThreadBlockageIntervals()`.

## Testing


bidetofevil (approved):
LGTM. Why keep stuff around that we know we won't use when we can determine that at insertion time. Nice!

PR #3628, opened 2026-07-23
Cleanup temp files in storage directory

fractalwrench:
## Goal

Cleans up temporary files that may be hanging around in the storage directory, along with removing files whose names are malformed.

## Testing

Added unit tests.


lemnik (approved):
LGTM

PR #3629, opened 2026-07-23
Add test for measuring heap allocation

fractalwrench:
## Goal

Adds a test that attempts to give rough measurements of heap allocation for common operations on the Embrace API. This is achieved by repeatedly running each operation and then forcing GC to provide an average measurement.

I've added test cases for creating spans and logs of varying complexity. The span test cases provide similar values to what we've observed in Perfetto profiling.

### DIFF
--- a/embrace-microbenchmark/src/androidTest/kotlin/io/embrace/android/embracesdk/benchmark/HeapAllocationBenchmark.kt
+++ b/embrace-microbenchmark/src/androidTest/kotlin/io/embrace/android/embracesdk/benchmark/HeapAllocationBenchmark.kt
@@ -143,8 +143,8 @@ class HeapAllocationBenchmark {
 
     /**
      * Allocates [firstBatch] then [secondBatch] items produced by [makeItem], snapshotting the used heap
-     * after a forced GC between the two batches. Then returns
-     * the retained bytes per item added in the second batch.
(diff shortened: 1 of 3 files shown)


embrace-microbenchmark/src/androidTest/kotlin/io/embrace/android/embracesdk/benchmark/HeapAllocationBenchmark.kt
```diff
@@ -0,0 +1,215 @@
...
+        )
+    }
+
+    private fun freshAttributes(count: Int): Map<String, String> =
+        (0 until count).associate { "attribute.key.$it" to "attribute-value-$it" }
+
+    /**
+     * Allocates [firstBatch] then [secondBatch] items produced by [makeItem], snapshotting the used heap
+     * after a forced GC between the two batches. Then returns
+     * the retained bytes per item added in the second batch.
```
lemnik: nit:doc formatting 

```suggestion
     * Allocates [firstBatch] then [secondBatch] items produced by [makeItem], snapshotting the used heap
     * after a forced GC between the two batches. Then returns the retained bytes per item added in
     * the second batch.
```

lemnik (approved):
LGTM

PR #3630, opened 2026-07-23
Avoid retaining OtelSpanArgs

fractalwrench:
## Goal

Avoid retaining OtelSpanArgs so that its collections are not retained.

## Testing

Observed ~200 byte reduction in relevant benchmark in `HeapAllocationBenchmark`


lemnik (approved):
LGTM

PR #3631, opened 2026-07-23
Flatten EmbraceSpanData layer

fractalwrench:
## Goal

Flattens the `EmbraceSpanData` layer as it just acted as an intermediate between `SpanData` and `Span` which caused unnecssary alloc.

## Testing

Updated unit tests.


bidetofevil (approved):
I don't even recall why these were different. Perhaps a reason that is no longer needed? Anyway, if everything still works, that's great

PR #3632, opened 2026-07-23
Reduce allocations in log/span export

fractalwrench:
## Goal

Reduces allocations in log/span export by filtering out private spans first, and avoiding another instantiation of attributes for each span/log record.


lemnik (approved):
LGTM

PR #3637, opened 2026-07-24
Remove dead code and allocation churn

fractalwrench:
## Goal

Removes various bits of dead code and allocation churn from the Tracing implementation.


bidetofevil (approved):
`removeSessionPartEvent()` seems like an analogue to `addSessionPartEvent()`, but I guess if we're not calling it, it's not necessary - and we can always add it back when. LGTM

PR #3638, opened 2026-07-24
Optimize span flushing

fractalwrench:
## Goal

Optimizes span flushing. Previously a `Set` was constructed which called `equals()` on every `EmbraceSpanData` object (and its events, links, etc) in the collection. This was chosen to deal with the demands of concurrent calls.

However, it's safe to assume that the number of completed spans will only ever grow while the flush lock is acquired. Therefore we can obtain the size of the completed spans, then poll & add elements to a new array that is returned. This avoids the potentially expensive equality checks.

## Testing

Updated unit tests.


lemnik (approved):
LGTM

PR #3639, opened 2026-07-24
Memoise session link attributes

fractalwrench:
## Goal

Memoises the session ID attributes for added as links to each span so that allocations are marginally reduced.

## Testing

Added unit tests.


lemnik (approved):
LGTM

PR #3640, opened 2026-07-24
Remove leaked request references

fractalwrench:
## Goal

`NetworkRequestDataSource` keeps a reference to active requests that wasn't always closed in a collection, which leaked memory. This changeset alters it so that the reference is always removed after a request completes.

## Testing

Added unit tests.


bidetofevil (approved):
LGTM

PR #3641, opened 2026-07-24
Avoid duplicating Window.Callback registrations

lemnik:
## Goal
Avoid duplicating and leaking `Window.Callback`s in the callback chain.

## Changes
Track whether the `Window.Callback` for Compose Tap, and Vitals instrumentation has already been registered by keeping a weak-reference to the `Window`.

## Testing
Added new unit tests


fractalwrench (approved):
LGTM

PR #3644, opened 2026-07-24
Combine SpanSink and SpanRepository

fractalwrench:
## Goal

Combines `SpanSink` and `SpanRepository` as they are always used together in the same place and can be rationalised further when together.


bidetofevil (commented):
Nice. LGTM

bidetofevil (approved):

PR #3645, opened 2026-07-24
Use log record processor

fractalwrench:
## Goal

Use a `LogRecordProcessor` rather than wrapping `LoggerProvider/Logger` interfaces. This sort of approach should reduce the amount of boilerplate code required by Embrace's SDK (and therefore allocations), along with generally simplifying things.

## Testing

Updated unit tests.


embrace-android-otel/src/main/kotlin/io/embrace/android/embracesdk/internal/otel/logs/EmbraceLogRecordProcessor.kt
```diff
@@ -0,0 +1,53 @@
...
+ * that has [skipMetadataContextKey] set to true. This is used for logs that describe a past session
+ * (e.g. native crashes replayed on the next launch), which carry their own session identity and must
+ * not be stamped with the current session's metadata.
+ */
+class EmbraceLogRecordProcessor(
+    private val uuidSource: UuidSource,
+    private val skipMetadataContextKey: Provider<ContextKey<Boolean>>,
+) : LogRecordProcessor {
+
+    private val metadataProviderRef = AtomicReference<Provider<Map<String, String>>> { emptyMap() }
```
lemnik, line 30: Couldn't this rather be a plain `@Volatile`? It doesn't seem to be used as an atomic?

embrace-android-otel/src/main/kotlin/io/embrace/android/embracesdk/internal/otel/logs/EmbraceLogRecordProcessor.kt
```diff
@@ -0,0 +1,53 @@
...
+ * API, as they all flow through the SDK's logger.
+ *
+ * Metadata enrichment can be suppressed for a specific log record by emitting it with a [Context]
+ * that has [skipMetadataContextKey] set to true. This is used for logs that describe a past session
+ * (e.g. native crashes replayed on the next launch), which carry their own session identity and must
+ * not be stamped with the current session's metadata.
+ */
+class EmbraceLogRecordProcessor(
+    private val uuidSource: UuidSource,
+    private val skipMetadataContextKey: Provider<ContextKey<Boolean>>,
```
bidetofevil, line 27: Can we just hardware a key in there in our semantic conventions that, once detected, will do this skipping? There's a desire to codify this style of async telemetery as something like `app.async`, and we can pioneer that in our impl

embrace-android-otel/src/main/kotlin/io/embrace/android/embracesdk/internal/otel/impl/EmbLogger.kt
```diff
@@ -1,43 +0,0 @@
-package io.embrace.android.embracesdk.internal.otel.impl
-
-import io.embrace.android.embracesdk.internal.otel.logs.EventService
-import io.opentelemetry.kotlin.attributes.AttributesMutator
-import io.opentelemetry.kotlin.context.Context
-import io.opentelemetry.kotlin.logging.Logger
-import io.opentelemetry.kotlin.logging.SeverityNumber
-
-class EmbLogger(
```
bidetofevil, line 9: So we no longer need our own implementation of the `Logger` interface now given the enrichment we do is all via the `EmbraceLogRecordProcessor`?

PR #3653, opened 2026-07-27
Reduce internal span limit

fractalwrench:
## Goal

Reduces the internal span limit as it was set fairly high and this has the potential to cause issues on the P99.9 of sessions.


bidetofevil (commented):
Should reduce the limits of internal span producing instrumentation as well like network status. etc? Feels like we should indicate that we are dropping telemetry/metadta because we are "full" and show something in the dashboard

PR #3655, opened 2026-07-27
Avoid capturing content length in OkHttp requests with no header

fractalwrench:
## Goal

Avoid capturing content length in OkHttp requests that have no explicit Content-Length header. I've created a local config flag `sdk_config.networking.capture_okhttp_response_body_size` that allows opting into this behaviour. This runs a risk of memory exhaustion for the current instrumentation & making this alteration better matches what HttpUrlConnection instrumentation does.

## Testing

Added unit tests.


lemnik (approved):
LGTM

bidetofevil (approved):
I would be really curious how often this happens, but this expensive fallback being off by default seems reasonable. Maybe we can look at the number after we release this to compare how many requests don't have body size compared to before?

PR #3656, opened 2026-07-27
Release previous session span

fractalwrench:
## Goal

Release a reference to the previous session span that caused the entire span object to remain in memory. Instead only information required for a link is kept.


lemnik (approved):
LGTM

PR #3658, opened 2026-07-27
Fix EmbSpan attributes read

fractalwrench:
## Goal

Fix `EmbSpan` attributes so the filter allows reading them back via the OTel API.


bidetofevil (approved):
Yikes. Good find!

PR #3659, opened 2026-07-27
Optimize span implementation

fractalwrench:
## Goal

Optimizes allocations when creating spans by removing one map allocation, and removing `EmbTrace` calls on the hot path that instantiated string keys.


bidetofevil (approved):
LGTM. Do we have some other way to measure in benchmarking or perfetto traces how long each operation takes

PR #3660, opened 2026-07-27
Remove cyclic refs in `WeakHashMap<Window, *>`

lemnik:
## Goal
Replace the `WeakHashMap<Window, *>` in Vitals & Compose instrumentation with a weak hash set since the maps were only used as markers. This way `Window` references cannot be cyclic through these maps as they no longer have a right-hand-side reference.

## Testing
Tested with example app and debugger


bidetofevil (approved):
Nice. LGTM

PR #3662, opened 2026-07-28
Avoid round trip allocation

fractalwrench:
## Goal

Avoids unnecessary allocations when reading events/links from `EmbSpan`.


lemnik (approved):
LGTM

PR #3664, opened 2026-07-28
Impose a vitals-specific span limit

lemnik:
## Goal
Limit the number of vital spans that can be generated. Default of 250 per session part, with a RemoteConfig override.

## Testing
New unit test coverage


fractalwrench (approved):
LGTM

PR #3665, opened 2026-07-28
Release collections after span stopped

fractalwrench:
## Goal

Release unused collections after a span has stopped to free up memory.

## Testing

Added unit test.


embrace-android-otel/src/main/kotlin/io/embrace/android/embracesdk/internal/otel/spans/EmbraceSpanFactoryImpl.kt
```diff
@@ -196,13 +196,26 @@ private class EmbraceSpanImpl(
...
+    /**
+     * Once a span has stopped it has been exported to an independent [Span] payload, so its own
+     * event/link collections are redundant. Release them to allow GC on the collections during the
+     * remainder of the session part.
+     */
+    private fun releaseRetainedData() {
+        systemEvents.clear()
+        customEvents.clear()
+        systemLinks.clear()
+        customLinks.clear()
```
lemnik: All of these are concurrent collections, so `clear()` isn't O(1) on any of them - do we want to consider either fully replacing the collections with empties (to preserve nullability) or accept the null checks and null them all out?

bidetofevil: If `clear()` isn't constant time, blow them away

bidetofevil (approved):
LGTM. Feel free to do the change Jason suggested though - that seems good and easy,

embrace-android-otel/src/main/kotlin/io/embrace/android/embracesdk/internal/otel/spans/EmbraceSpanFactoryImpl.kt
fractalwrench: Updated to null things out instead.

PR #3666, opened 2026-07-28
Flatten SpanService implementation

fractalwrench:
## Goal

Flattens the `SpanService` implementation so that `SpanServiceImpl` is the only relevant class that is invoked.

## Testing


embrace-android-otel/src/main/kotlin/io/embrace/android/embracesdk/internal/otel/spans/SpanServiceImpl.kt
```diff
@@ -3,36 +3,87 @@ package io.embrace.android.embracesdk.internal.otel.spans
...
+     * already observed [otelComponents] as null cannot add to the buffer after the replay has finished.
+     */
+    private val bufferLock = Any()
+
+    private val bufferedCalls = ConcurrentLinkedQueue<BufferedRecordCompletedSpan>()
+
+    /**
+     * Lifetime count of buffered calls.
+     */
+    private val bufferedCallsCount = AtomicInteger(0)
```
lemnik: As far as I can see both of these are always accessed under `bufferLock` - so could be reasonably changed to a single `ArrayDeque` (and `bufferedCalls.size`) or something similar?

embrace-android-otel/src/main/kotlin/io/embrace/android/embracesdk/internal/otel/spans/SpanServiceImpl.kt
```diff
@@ -3,36 +3,87 @@ package io.embrace.android.embracesdk.internal.otel.spans
...
 import io.embrace.android.embracesdk.spans.ErrorCode
 import io.opentelemetry.kotlin.OpenTelemetry
 import io.opentelemetry.kotlin.tracing.Tracer
-import java.util.concurrent.atomic.AtomicBoolean
+import java.util.concurrent.ConcurrentLinkedQueue
+import java.util.concurrent.atomic.AtomicInteger
 
 /**
- * Implementation of the core logic for [SpanService]
+ * Implementation of [SpanService].
```
bidetofevil, line 15: Very nice!

bidetofevil (approved):
Very nice. LGTM functionally - Jason's suggestion seems good too.

embrace-android-otel/src/main/kotlin/io/embrace/android/embracesdk/internal/otel/spans/SpanServiceImpl.kt
fractalwrench: Updated to use `ArrayDeque`

PR #3667, opened 2026-07-28
Reduce scheduling overhead for ScreenLoads

lemnik:
## Goal
Avoid scheduling a timer on every tap for ScreenLoad vitals. The tap now captures the start time, but the timer is scheduled on the navigation start event. The timeout is derived from the tap timestamp (rather than the navigation event).

## Testing
Added unit tests to ensure the ScreenLoad time remains "tap" -> "settle"


fractalwrench (approved):
LGTM

PR #3668, opened 2026-07-28
Flatten tracing api delegate

fractalwrench:
## Goal

Moves the timestamp normalization to `SpanService` from `TracingApiDelegate`.


bidetofevil (approved):
Oops. Missed this when I reviewed first.

PR #3669, opened 2026-07-28
Avoid allocation churn on TelemetryDestination

fractalwrench:
## Goal

Avoid allocation churn on TelemetryDestination by using `SpanEvent` and `ErrorCodeAttribute` rather than `EmbraceSpanEvent` and `ErrorCode`.


bidetofevil (approved):
Nice.

PR #3670, opened 2026-07-28
Reduce boxing in EmbraceSpanImpl

fractalwrench:
## Goal

Avoids boxing some longs in `EmbraceSpanImpl`.

## Testing


bidetofevil (approved):
LGTM. Non-blocking comment about sentinal value that you can take or leave if it matters.

PR #3671, opened 2026-07-28
Remove the use of `serializer<T>()` from init functions

lemnik:
## Goal
Using the `reified` extension function of `PlatformSerializer.toJson` is convenient but carries a hidden reflection cost at runtime which we want to avoid. This PR changes the config related classes to create or reference the serializers directly which avoids the `Reflection.typeOf` calls.

## Testing
Relied on existing tests.


bidetofevil (approved):
LGTM

fractalwrench (approved):

PR #3674, opened 2026-07-29
Reduce references on EmbraceSpanImpl

fractalwrench:
## Goal

Reduce references on `EmbraceSpanImpl` by creating a class that holds dependencies.


bidetofevil (approved):
LGTM

PR #3676, opened 2026-07-29
Pre-warm shared preferences

fractalwrench:
## Goal

Attempts to pre-warm shared preferences at the very start of SDK initialization on a background thread. This may help reduce the amount of time the config service spends waiting on `awaitLockedLoaded` later.


bidetofevil (approved):
Why not. On the poor performing devices, I wonder if this will get processed before the SDK starts up - but there isn't a tonne of downside, unless this could make the call slower because it's not on the main thread?

bidetofevil:
@lemnik do you think throwing the shared prefs init onto a background thread can backfire if non-main-thread access makes it slower and processed later, blocking the call on the main thread?

lemnik:
> @lemnik do you think throwing the shared prefs init onto a background thread can backfire if non-main-thread access makes it slower and processed later, blocking the call on the main thread?

As the code stands I don't the worst case path that this introduces is significantly worse than the current worst case. The current `SharedPrefsImpl`[kicks off background loading on init](https://cs.android.com/android/platform/superproject/+/android-latest-release:frameworks/base/core/java/android/app/SharedPreferencesImpl.java;l=141-153) but I don't think that's always been the case, so sticking with the "background `getSharedPreferences()` -> background load" seems reasonable.

lemnik (approved):

PR #3677, opened 2026-07-29
Reduce synthetic lambdas in sdk initialization

fractalwrench:
## Goal

Attempts to reduces the number of synthetic lambdas in SDK initialization, as well as removing some useless `lazy` initialization as nearly all properties are initialized eagerly.

## Testing


bidetofevil (approved):
Nice. Is there a determistic path for the cascading init on the SDK start path that we should be aware of so we don't accidentally reference something earlier than expected and push the init of various things up earlier than required? Or are we just relying on integration tests to blow up if that happens?

fractalwrench:
Integration tests should blow up. Also generally speaking property declaration order should prevent that.

PR #3678, opened 2026-07-29
Alter per session-part span limits

fractalwrench:
## Goal

This changeset adds limits that enforce a maximum number of spans per session-part. It makes the following changes:

1. Reduces the max internal span count from 5000 to 1500 (confirmed via data analysis that this doesn't affect the P99.9 of sessions)
2. A limit of 1000 for network request spans (data analysis shows that these dominate all other spans)


bidetofevil (approved):
tl;dr, looks good providing we can move the network span limit up to 2K. Further:  
  
Looking at the data, 1500 non-network internal spans is probably a reasonable cap. At all percentiles, total span - network spans is < 1000.   
  
As long as we can mark a session as having reached the cap and we can look at that data, I'm ok with that.  
  
Limiting network spans is a tricker issue to me. Our current limits allow 1000 per domain by default, which would be odd since we're capping the total to that. As well, P99.9 is over 1000 network spans - which means 0.1% of session parts have more than that. With ~120M session parts daily, that's 120K, which is pretty significant to me.  
  
Can be bump this, for now, to 2000, which would cover almost up to P99999, so reducing the violators to ~1200? I think some of the work to kill logging ones for cached responses will reduce the actual usage further, and allow us to move this down to 1000. (We might also need to make this configurable for those who actually have that many...)

fractalwrench:
Bumped the max count of network spans to 2000 as suggested.

bidetofevil (approved):
LGTM

fnewberg:
@fractalwrench can we make the network span limit be remotely configurable? i would much rather be able to adjust it for customers that way rather than trying to find a value that is perfect for all customers

fractalwrench:
Updated changeset so that the limits are remotely configurable.

lemnik (approved):
LGTM

PR #3679, opened 2026-07-29
Provide configurable timeout for network request spans

fractalwrench:
## Goal

Provide a configurable timeout for network request spans. This will be used to timeout network requests that are otherwise left open-ended. A future changeset will hook this up so that the span actually terminates.

## Testing

Added unit tests.


bidetofevil (approved):
Very nice. LGTM.

lemnik (approved):
LGTM

fractalwrench:
Updated config so that timeout can be driven by remote config.

PR #3680, opened 2026-07-29
Shared MapSerializer  in SharedPrefsStore

lemnik:
## Goal
Create a single common `MapSerializer<String, String>` used `SharedPrefsStore. getStringMap()` so that we don't allocate one for every `getStringMap()` call

## Testing
Existing tests

### DIFF
--- a/embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/prefs/SharedPrefsStore.kt
+++ b/embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/prefs/SharedPrefsStore.kt
@@ -42,7 +42,7 @@ internal class SharedPrefsStore(
 
     override fun getStringMap(key: String): Map<String, String>? {
         val mapString = impl.getString(key, null) ?: return null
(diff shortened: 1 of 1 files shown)


embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/prefs/SharedPrefsStore.kt
```diff
@@ -62,4 +62,8 @@ internal class SharedPrefsStore(
             -1
         }
     }
+
+    private companion object {
+        val mapSerializer = MapSerializer(String.serializer(), String.serializer())
```
fractalwrench, line 67: Is this object thread-safe?

fractalwrench (commented):
One question around thread safety of a map serializer, otherwise this looks good

embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/prefs/SharedPrefsStore.kt
lemnik: yes

fractalwrench (approved):
LGTM

PR #3681, opened 2026-07-29
Use correct implicit context storage for opentelemetry-kotlin

fractalwrench:
## Goal

Use thread-local context storage when opentelemetry-kotlin is used in the SDK, so that existing behavior is matched.


bidetofevil (approved):
LGTM for now.  
  
But do we actually want this in the long run? For the migration to using Kotlin SDK impl, this will minimize changes, but for the Embrace SDK, I think we should talk about severing the tie to this (by default) and making folks enable this.  
  
Maybe as a 10.0 thing?

fractalwrench:
Yeah, I think tbd for 10.0. For now maintaining existing behavior feels best for a seamless switch.

PR #3682, opened 2026-07-29
Use LogRecordProcessor rather than wrapping OpenTelemetry's Logger

fractalwrench:
## Goal

Use `LogRecordProcessor` rather than wrapping OpenTelemetry's Logger is preferable as it reduces the amount of book-keeping of collections & simplifies our logging pipeline. This supersedes #3645 as it greatly simplifies how 'metadata' was appended to individual logs, by asking the AEI/native crash instrumentation to append their own session information, and otherwise allowing our processor to add it.

The existing logic for how log records are processed and exported wrt user-supplied processors does not change.

## Testing

Updated existing tests.


embrace-android-sdk/src/integrationTest/resources/user_session_aei_log_active.json
```diff
@@ -32,31 +32,15 @@
         },
         {
           "key": "emb.session_part_id",
-          "value": "__EMBRACE_TEST_SESSION_PART_ID__"
-        },
-        {
-          "key": "emb.state",
```
bidetofevil, line 38: Why are these no longer on the AEI logs? The initial state should still be populated when we can start logging telemetry, I thought?

bidetofevil (dismissed):
Looks great! Logs are simple and we just have to ensure the order of the processors is something we can control - if we can discard the wrapping and enrich in the processor, that can potentially lead to further simplification - `EventService` is already s simpel wrapper over the OTel logging interface so we don't have to eat that dep. Ideally everything could be like that - or go the opposite direction and don't involve OTel at all until the very last moment.

embrace-android-sdk/src/integrationTest/resources/user_session_aei_log_active.json
fractalwrench: I've altered the LogRecordProcessor so that the initial state is still populated (this was problematic for AEI/native crashes only). I now have an approach that should maintain the existing behavior for both.

embrace-android-otel/src/main/kotlin/io/embrace/android/embracesdk/internal/otel/logs/EmbraceLogRecordProcessor.kt
```diff
@@ -0,0 +1,56 @@
...
+    override suspend fun shutdown(): OperationResultCode = OperationResultCode.Success
+
+    /**
+     * Whether the record describes a process that has already died, in which case none of the current
+     * metadata applies to it. A native crash is resurrected by a later process and carries the whole state of
+     * the session that crashed - its session IDs, app state and session properties - so enriching it would
+     * describe the wrong session entirely.
+     */
+    private fun Map<String, Any>.belongsInCurrentProcess(): Boolean =
+        get(EmbType.System.NativeCrash.key) == EmbType.System.NativeCrash.value
```
bidetofevil, line 55: Nice. When we have `app..async` or whatever, we can use that instead of looking for specific types;

bidetofevil (approved):
LGTM

PR #3683, opened 2026-07-29
Add tests for parity between KMP/opentelemetry-java OpenTelemetry implementations

fractalwrench:
## Goal

Add integration tests that check export parity between KMP/opentelemetry-java OpenTelemetry distributions by asserting against golden files.


bidetofevil (approved):
LGTM

PR #3684, opened 2026-07-29
Remove event service

fractalwrench:
## Goal

Removes the EventService as this class now simply forwards to a `Logger`.


bidetofevil (approved):
Haha, I thought we had to keep the interface for dependency reasons, but if not, woohoo! Another relunctant abstraction laid to bed.

PR #3692, opened 2026-07-29
Internal experiment tracking service definition and  implementation

bidetofevil:
Internal service to track experiments and feature flags along with the internal data model used in the SDK.

PR #3690, opened 2026-07-29
Add emb.experiments attribute to semantic conventions

bidetofevil:
Add attribute name to use to carry the experiments records payload

PR #3693, opened 2026-07-29
Public API for Experiments

bidetofevil:
Public API that allows the tracking of experiments and feature flags, Calls before the SDK starts are buffered and the result is applied to all telemetry that the SDK logs.

### DIFF
--- a/embrace-android-api/api/embrace-android-api.api
+++ b/embrace-android-api/api/embrace-android-api.api
@@ -58,18 +58,15 @@ public abstract interface annotation class io/embrace/android/embracesdk/annotat
 public abstract interface annotation class io/embrace/android/embracesdk/annotation/NotTracedActivity : java/lang/annotation/Annotation {
 }
 
-public final class io/embrace/android/embracesdk/experiments/TrackedExperiment {
-	public fun <init> (Ljava/lang/String;JLjava/lang/String;)V
-	public synthetic fun <init> (Ljava/lang/String;JLjava/lang/String;ILkotlin/jvm/internal/DefaultConstructorMarker;)V
-	public final fun getId ()Ljava/lang/String;
-	public final fun getStartTimeMs ()J
-	public final fun getVariant ()Ljava/lang/String;
+public abstract interface class io/embrace/android/embracesdk/experiments/TrackedExperiment {
+	public abstract fun getId ()Ljava/lang/String;
+	public abstract fun getStartTimeMs ()J
+	public abstract fun getVariant ()Ljava/lang/String;
 }
 
-public final class io/embrace/android/embracesdk/experiments/TrackedFeatureFlag {
-	public fun <init> (Ljava/lang/String;J)V
-	public final fun getId ()Ljava/lang/String;
-	public final fun getStartTimeMs ()J
+public abstract interface class io/embrace/android/embracesdk/experiments/TrackedFeatureFlag {
+	public abstract fun getId ()Ljava/lang/String;
+	public abstract fun getStartTimeMs ()J
 }
 
 public abstract interface class io/embrace/android/embracesdk/internal/api/BreadcrumbApi {
@@ -86,10 +83,18 @@ public abstract interface class io/embrace/android/embracesdk/internal/api/Embra
 }
 
 public abstract interface class io/embrace/android/embracesdk/internal/api/ExperimentApi {
+	public abstract fun createExperiment (Ljava/lang/String;JLjava/lang/String;)Lio/embrace/android/embracesdk/experiments/TrackedExperiment;
+	public abstract fun createFeatureFlag (Ljava/lang/String;J)Lio/embrace/android/embracesdk/experiments/TrackedFeatureFlag;
 	public abstract fun trackExperiment ([Lio/embrace/android/embracesdk/experiments/TrackedExperiment;)V
 	public abstract fun trackFeatureFlag ([Lio/embrace/android/embracesdk/experiments/TrackedFeatureFlag;)V
 	public abstract fun untrackExperiment ([Ljava/lang/String;J)V
 	public abstract fun untrackFeatureFlag ([Ljava/lang/String;J)V
+}
+
+public final class io/embrace/android/embracesdk/internal/api/ExperimentApi$DefaultImpls {
+	public static fun createExperiment (Lio/embrace/android/embracesdk/internal/api/ExperimentApi;Ljava/lang/String;JLjava/lang/String;)Lio/embrace/android/embracesdk/experiments/TrackedExperiment;
+	public static synthetic fun createExperiment$default (Lio/embrace/android/embracesdk/internal/api/ExperimentApi;Ljava/lang/String;JLjava/lang/String;ILjava/lang/Object;)Lio/embrace/android/embracesdk/experiments/TrackedExperiment;
+	public static fun createFeatureFlag (Lio/embrace/android/embracesdk/internal/api/ExperimentApi;Ljava/lang/String;J)Lio/embrace/android/embracesdk/experiments/TrackedFeatureFlag;
 }
 
 public abstract interface class io/embrace/android/embracesdk/internal/api/InstrumentationApi {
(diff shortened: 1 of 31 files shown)


bidetofevil:
* **#3836** <a href="https://app.graphite.com/github/pr/embrace-io/embrace-android-sdk/3836?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#3835** <a href="https://app.graphite.com/github/pr/embrace-io/embrace-android-sdk/3835?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#3834** <a href="https://app.graphite.com/github/pr/embrace-io/embrace-android-sdk/3834?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#3833** <a href="https://app.graphite.com/github/pr/embrace-io/embrace-android-sdk/3833?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#3832** <a href="https://app.graphite.com/github/pr/embrace-io/embrace-android-sdk/3832?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#3711** <a href="https://app.graphite.com/github/pr/embrace-io/embrace-android-sdk/3711?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#3710** <a href="https://app.graphite.com/github/pr/embrace-io/embrace-android-sdk/3710?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#3709** <a href="https://app.graphite.com/github/pr/embrace-io/embrace-android-sdk/3709?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#3693** <a href="https://app.graphite.com/github/pr/embrace-io/embrace-android-sdk/3693?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/embrace-io/embrace-android-sdk/3693?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#3692** <a href="https://app.graphite.com/github/pr/embrace-io/embrace-android-sdk/3692?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#3691** <a href="https://app.graphite.com/github/pr/embrace-io/embrace-android-sdk/3691?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#3690** <a href="https://app.graphite.com/github/pr/embrace-io/embrace-android-sdk/3690?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `integration/experiments`


<h2></h2>

embrace-android-api/api/embrace-android-api.api
```diff
@@ -58,6 +58,20 @@ public abstract interface annotation class io/embrace/android/embracesdk/annotat
 public abstract interface annotation class io/embrace/android/embracesdk/annotation/NotTracedActivity : java/lang/annotation/Annotation {
 }
 
+public final class io/embrace/android/embracesdk/experiments/TrackedExperiment {
+	public fun <init> (Ljava/lang/String;JLjava/lang/String;)V
```
fractalwrench: An interface (or avoiding the need for this type entirely) would be preferable. Concrete classes tie us to a specific implementation. Same for `TrackedFeatureFlag`

embrace-android-api/api/embrace-android-api.api
```diff
@@ -71,6 +85,13 @@ public abstract interface class io/embrace/android/embracesdk/internal/api/Embra
 	public abstract fun startView (Ljava/lang/String;)Z
 }
 
+public abstract interface class io/embrace/android/embracesdk/internal/api/ExperimentApi {
+	public abstract fun trackExperiment ([Lio/embrace/android/embracesdk/experiments/TrackedExperiment;)V
+	public abstract fun trackFeatureFlag ([Lio/embrace/android/embracesdk/experiments/TrackedFeatureFlag;)V
```
fractalwrench, line 89: I'd suggest deconstructing the parameters so these just take primitives, as that circumvents the need for an extra type + instantiation, and is still backwards compatible if we need to add extra fields via default params.

embrace-android-api/src/main/kotlin/io/embrace/android/embracesdk/internal/api/ExperimentApi.kt
```diff
@@ -0,0 +1,33 @@
+package io.embrace.android.embracesdk.internal.api
+
+import io.embrace.android.embracesdk.experiments.TrackedExperiment
+import io.embrace.android.embracesdk.experiments.TrackedFeatureFlag
+
+/**
+ * The public API used to track experiment and feature-flag state on a given device for the current app instance. IDs are unique across
+ * experiments and feature flags. The SDK does not persist this state, and users need to call this each time the app starts up.
```
fractalwrench, line 10: Would be good to mention explicitly that this can be called before the SDK is initialized, whenever an experiment starts. Or after.

embrace-android-api/src/main/kotlin/io/embrace/android/embracesdk/internal/api/ExperimentApi.kt
```diff
@@ -0,0 +1,33 @@
...
+    /**
+     * Tracks the given experiment memberships that this app instance has been bucketed into.
+     */
+    public fun trackExperiment(vararg experiments: TrackedExperiment)
+
+    /**
+     * Stops tracking the given experiments on this app instance at the given time. An experiment can only be untracked after it has
+     * first been tracked. Untracking an experiment that has not been tracked has no affect.
+     */
+    public fun untrackExperiment(vararg experimentIds: String, endTimeMs: Long)
```
fractalwrench, line 29: This approach couples different experiment IDs with the same end time, which might not necessarily be true if our SDK is starting late in the lifecycle. Is this what we want?

embrace-android-api/api/embrace-android-api.api
bidetofevil: Sure

embrace-android-api/api/embrace-android-api.api
bidetofevil: We want to do a bulk API, hence having these objects. I'd much rather than having arrays or somehow managing state so that each of the hundreds of calls don't rebuild the blob.

embrace-android-api/src/main/kotlin/io/embrace/android/embracesdk/internal/api/ExperimentApi.kt
bidetofevil: Will add. It's not only that you COULD do that, but the fact that folks SHOULD do that which we should get across as well

embrace-android-api/src/main/kotlin/io/embrace/android/embracesdk/internal/api/ExperimentApi.kt
bidetofevil: So the API is tracking when the device is contacted or finds out that one more more experiments have ended. I would imagine that would come via a payload from the server or something that tells them the state of more than one experiment, and hence it taking effect at the same time. Given that, turning all those experiments off seems to be the correct call.   
  
Also, if there are separate times, then they can also make separate calls if they choose.

embrace-android-api/src/main/kotlin/io/embrace/android/embracesdk/experiments/TrackedFeatureFlag.kt
```diff
@@ -0,0 +1,22 @@
...
+     */
+    public val id: String
+
+    /**
+     * The time at which the flag started applying to the device, in milliseconds since the epoch.
+     */
+    public val startTimeMs: Long
+}
+
+internal class TrackedFeatureFlagImpl(
```
fractalwrench, line 19: Could the implementation detail be separated out from the `embrace-android-api` module, along with the defaults in `ExperimentApi`? This is an approach that's worked fairly well in opentelemetry-kotlin so I'd be keen to maintain it in embrace-android-sdk too unless it's unavoidable

fractalwrench (commented):
LGTM pending comments

fractalwrench:
Merging as-is, pending comments will be addressed in a separate PR.

PR #3697, opened 2026-07-30
Remove clock device-sleep edge-case

lemnik:
## Goal
`NormalizedIntervalClock` may have been falsely reporting drift when the device slept between:
```
        val newTime = baseline + monotonicClock()
        val prev = lastTime.getAndSet(newTime) 
```

Leading a thread to calculate the drift against a stale (and raced) timestamp.

## Design
`SystemClock::elapsedRealtime` cannot go backwards by contract, but the new logic in `NormalizedIntervalClock` should reliably detect if the supplied clock source ever *did* go backwards by more than 60 seconds. This does come at a cost however, since we may now CAS loop several times on any given `now()` call (where `getAndSet` was a single exclusive-exchange on most devices).

## Testing
Fixed the existing tests to assert `internalErrorMessages` instead of `errorMessages` (which were never being logged by the `NormalizedIntervalClock`).

### DIFF
--- a/embrace-android-utils/src/main/kotlin/io/embrace/android/embracesdk/internal/clock/NormalizedIntervalClock.kt
+++ b/embrace-android-utils/src/main/kotlin/io/embrace/android/embracesdk/internal/clock/NormalizedIntervalClock.kt
@@ -5,6 +5,7 @@ import io.embrace.android.embracesdk.internal.logging.InternalErrorType
 import io.embrace.android.embracesdk.internal.logging.InternalLogger
 import java.util.concurrent.atomic.AtomicBoolean
 import java.util.concurrent.atomic.AtomicLong
+import kotlin.math.max
 
 /**
  * A clock which uses [android.os.SystemClock.elapsedRealtime] that is normalized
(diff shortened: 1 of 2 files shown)


embrace-android-utils/src/main/kotlin/io/embrace/android/embracesdk/internal/clock/NormalizedIntervalClock.kt
```diff
@@ -24,23 +24,35 @@ class NormalizedIntervalClock(
...
-        val prev = lastTime.getAndSet(newTime)
-        if (prev > 0L && newTime < prev - driftThresholdMs) {
-            if (hasLoggedDrift.compareAndSet(false, true)) {
-                logger?.trackInternalError(
-                    InternalErrorType.InternalInterfaceFail,
-                    IllegalStateException(
-                        "NormalizedIntervalClock drifted back in time by more than threshold. Delivery is likely out-of-order.",
-                    ),
-                )
+        while (true) {
```
fractalwrench: Could we bound this by N attempts just in case something goes wrong with the contained logic (either now or in future)?

fractalwrench (approved):
LGTM pending one comment

embrace-android-utils/src/main/kotlin/io/embrace/android/embracesdk/internal/clock/NormalizedIntervalClock.kt
lemnik: We could (I'm assuming the method exits with `newTime` and does not attempt to move `lastTime` at that point), and in that case a low limit (something like 5 or 8) seems like a better idea, or a limit relative to the available concurrency (`availableProcessors` or similar).

fractalwrench (approved):

PR #3698, opened 2026-07-30
Additional test cases for comparing OTel distribution behavior

fractalwrench:
## Goal

Builds on #3683 by enhancing the test cases that compare behavior between the compat and KMP OpenTelemetry distributions.


bidetofevil (approved):
LGTM

PR #3699, opened 2026-07-30
Update AGENTS.md

fractalwrench:
## Goal

Update AGENTS.md following suggestions from `claude doctor` to avoid derivable information & reduce the number of tokens loaded into context.


bidetofevil (approved):
Who am I to doubt what a real doctor says?

PR #3700, opened 2026-07-30
Timeout network request spans after configurable duration

fractalwrench:
## Goal

This changeset times out network request spans after a locally configurable duration (defaults to 10m - well over the P99.9 for requests). On reaching a timeout the request will be stopped and marked as failed. This does alter the behavior of the instrumentation but is the only way to fulfil the constraints of accurate traceparent injection _and_ freeing up the span created in the OTel layer.

I've implemented this by altering the span storage layer so that it can timeout any span, by passing a timeout in ms when it is first created. This is tracked and swept at periodic intervals. If any span exceeds this interval, the span will be stopped.

This only applies to network request spans for now but could be applied to others in future to help avoid resource leaks.

## Testing

Added unit tests.


bidetofevil (approved):
LGTM

lemnik (approved):
Should the timeout maybe be non-nullable to avoid boxing?

Otherwise LGTM!

PR #3702, opened 2026-07-30
Fix R8 warning

fractalwrench:
## Goal

`opentelemetry-kotlin` 0.6.0 adds a new runtime dependency on `io.opentelemetry:opentelemetry-extension-trace-propagators`, which has classes that implement `io.opentelemetry.sdk.autoconfigure.spi.ConfigurablePropagatorProvider`.  This is causing R8 warnings when compiling, such as [this one](https://github.com/embrace-io/embrace-android-sdk/actions/runs/30540000440/job/90865197591).

This showed up as a sporadic test flake due to CI caching the mavenLocal snapshot that runs the gradle plugin integration tests. I've fixed this so that those tests always refresh dependencies to ensure they're correct.


lemnik (approved):
LGTM

PR #3703, opened 2026-07-30
Update config controlling opentelemetry-kotlin mode

fractalwrench:
## Goal

Updates how the config controlling opentelemetry-kotlin's mode works. A remote config flag was available but fundamentally the OpenTelemetry layer needs to initialize much sooner than the config service, so both the remote/local flags were disregarded in favour of a default. This means that the flag doesn't work and opentelemetry-java will always be used.

I've chosen to fix this by relying on the local config flag only. An alternative would be to refactor the SDK initialization so that config is always loaded before the OpenTelemetry layer, but this feels suboptimal as it delays customers from creating traces of their application.


bidetofevil (approved):
So.... with this, it means we turn it on for everyone and allow a custom shut off locally only? Well this sure makes roll out faster :sweat_smile:

fractalwrench:
@bidetofevil happy to discuss the alternative of altering SDK init order. I felt that this was the better tradeoff though as that approach could substantially delay the Tracing API + Logging API's usability at startup

bidetofevil:
I actually like this. We should be hella confident about this and really watch when a new app adopts, and if things go bad, we revert and pull that version with a patch. If this fails, it should be loud - and if it's only failing subtlely, slower rollouts won't make it easy for us to detect.

PR #3705, opened 2026-07-30
Optimize EmbraceSpanImpl collection memory usage

fractalwrench:
## Goal

Optimizes the memory usage of collections in `EmbraceSpanImpl` by using non-concurrent implementations behind a lock.


bidetofevil (approved):
LGTM , but@lemnik taking a look before or after merging would be good too

embrace-android-otel/src/main/kotlin/io/embrace/android/embracesdk/internal/otel/spans/EmbraceSpanFactoryImpl.kt
```diff
@@ -125,31 +128,55 @@ private class EmbraceSpanImpl(
...
-    private var customEvents: Queue<EmbraceSpanEvent>? = ConcurrentLinkedQueue()
+    private var customEvents: MutableList<EmbraceSpanEvent>? = null
 
     @Volatile
-    private var systemLinks: Queue<EmbraceLinkData>? = ConcurrentLinkedQueue()
+    private var systemLinks: MutableList<EmbraceLinkData>? = null
 
     @Volatile
-    private var customLinks: Queue<EmbraceLinkData>? = ConcurrentLinkedQueue()
+    private var customLinks: MutableList<EmbraceLinkData>? = null
```
lemnik: Since these are all under lock now, we can remove the `@Volatile`s and leave the fencing to the `synchronized` blocks

lemnik (approved):
LGTM!

PR #3707, opened 2026-07-31
Fix AnyValue assertions

fractalwrench:
## Goal

There's a bug in how opentelemetry-kotlin stringifies AnyValue objects compared to opentelemetry-java, so I've tweaked the test assertions now to fix a test failure that was introduced recently.


bidetofevil (approved):
LGTM

PR #3714, opened 2026-08-03
Report whether the running device appears to use eMMC

lemnik:
## Goal
Attempt to report whether a device appears to have an eMMC module (instead of UFS) as its primary storage.

Detection is a best-effort based on the block device names when they are accessible to the app. When no detection was possible then `null` is reported (and the resource attribute is omitted entirely) rather than report a false-positive. The detection also checks for the existence of "normal" flash storage to differentiate between eMMC as primary storage, and the existence of a TF card for secondary storage.

## Testing
New unit tests for the logic, along with manual testing of the logic across a range of real devices and Android versions.


fractalwrench (commented):
Happy with the detection, I think it'd be good to get consensus on how we want to report it in the payload, whether we should send on 100% of devices or a specific subset, and whether we want to export this over OTLP or not

bidetofevil (approved):
Resource is the right place if we can get it quickly enough. Once my Resource unification PRs get in, it will show up as an OTel resource attribute - so we should probaby think of a good semconv approved named for it (we dont' need that now - but we will need it when I rebase my PRs on this lol)

PR #3715, opened 2026-08-03
Optimize EmbTrace implementation

fractalwrench:
## Goal

Optimizes the EmbTrace implementation to reduce any impact it might have any performance (or correctness) of traces. I've made a few changes here:

- `EmbTrace` _must_ be called with `trace` now, as most of the other variants were unused, and led to codepaths where a trace wasn't closed which could unbalance Perfetto traces for folks consuming our SDK
- Removed inline reified fun which used reflection
- Avoided constructing a string each time the trace was invoked, in the best case


bidetofevil (approved):
LGTM

PR #3716, opened 2026-08-03
Setup Gradle parallel tooling

fractalwrench:
## Goal

Enables parallel tooling in Gradle: https://docs.gradle.org/current/userguide/performance.html#sec:configure_tooling_api_actions_parallelism


lemnik (approved):
LGTM

PR #3717, opened 2026-08-03
Avoid two instances of I/O during SDK init

fractalwrench:
## Goal

Avoids two instances of I/O during SDK init by moving calls to `createNewFile` to the point where files are actually written. This exploits the fact that we now read from a cached version of config, so don't always need to create files at init (which requires unnecessary syscalls).

## Testing

Enhanced existing test file.


lemnik (approved):
Very nice!

PR #3718, opened 2026-08-03
Report the device SOC when it is available

lemnik:
## Goal
Report the SOC model (which is manufacturer specific) when it is available and not a known placeholder value. The SOC in the device can't always reliably be determined by the device make & model.

## Testing
Unit tested, and manually tested on real devices.


fractalwrench (commented):
Similar comments to #3714 - happy with the detection, but think it'd be good to determine how we want to send the data

embrace-android-payload/src/main/kotlin/io/embrace/android/embracesdk/internal/payload/EnvelopeResource.kt
```diff
@@ -65,6 +65,7 @@ import kotlinx.serialization.Serializable
  * @param screenResolution The screen resolution. Previous name: d.sr
  * @param numCores (Android) The number of CPU cores the device has. Previous name: d.nc
  * @param usesEmmcStorage (Android) Whether the device appears to use an eMMC module for its primary storage.
+ * @param deviceSocModel (Android) The model name of the SoC the device is built around.
```
bidetofevil, line 68: Lets expand the acronym in the description?

bidetofevil (approved):
LGTM. Resource seems like the right place for this if it can be obtained without introducing material delays to SDK start

PR #3720, opened 2026-08-03
Optimize service loading

fractalwrench:
## Goal

Allows R8 to optimize SPI by replacing reflexive instantiations with a direct call. Our previous use of service loader and keep rules were blocking this.


lemnik (approved):
LGTM

PR #3721, opened 2026-08-03
Remove session.id as span attribute

fractalwrench:
## Goal

Removes `session.id` which is always duplicated by the value of `emb.user_session_id`.

## Testing


bidetofevil (changes requested):
I think we need the ability to add back the `sesion.id` for SDK users who want to use the Embrace session as their own definition of session and put the ID in attribute name for that in semantic conventions

fractalwrench:
@bidetofevil that could be achieved in a `SpanProcessor`/`LogRecordProcessor` that introspects the attributes. I imagine that folks would mainly care about OTLP-export for that use-case?

bidetofevil:
Ah, so we're saying they can do it themselves via a processors using the our API to get user session? I can buy that.   
  
We'll need to document this in the release notes.

bidetofevil (approved):
LGTM as long as we document how folks can get this back via custom processors

PR #3722, opened 2026-08-03
Alter worker thread used for config requests

fractalwrench:
## Goal

Alters the worker thread that the config HTTP request is enqueued on. The config request is enqueued on the IoRegWorker thread. In the worst case of a request taking ~1 min to timeout, this could block everything on the IoRegWorker (shared prefs warming, native crash handler installation, cached data sending). It feels more reasonable to enqueue it on the HttpRequestWorker instead.

I've also altered the `PayloadResurrectionService` which had a similar problem.

### DIFF
--- a/embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/resurrection/PayloadResurrectionServiceImpl.kt
+++ b/embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/resurrection/PayloadResurrectionServiceImpl.kt
@@ -33,8 +33,6 @@ import io.embrace.android.embracesdk.semconv.EmbSessionAttributes
 import io.opentelemetry.kotlin.semconv.SessionAttributes
 import java.io.InputStream
 import java.util.concurrent.CopyOnWriteArrayList
-import java.util.concurrent.TimeUnit
-import java.util.concurrent.TimeoutException
 import java.util.zip.GZIPInputStream
 import java.util.zip.ZipException
 import kotlin.math.max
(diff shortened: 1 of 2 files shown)


embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/resurrection/PayloadResurrectionServiceImpl.kt
```diff
@@ -290,18 +288,15 @@ internal class PayloadResurrectionServiceImpl(
             else -> null
         }
 
-        // Synchronously provide the payload to the IntakeService, blocking on the returned Future
+        // Hand the payload to the IntakeService without blocking. The write is queued on the data
+        // persistence worker while this thread moves on to the next payload; the envelope is already
+        // fully in memory, so we assume the cached copy is reasonably safe to delete.
         if (resurrectedPayload != null) {
```
bidetofevil, line 294: This blocking was happening because of race conditions. Does changing the thread this is running on eliminate them? (I don't think so but let me know if I'm mistaken :sweat_smile:)  
  
IIRC, the deletes of the tombstones might happen before the intakes are processed and persisted, so we can lose data as a result. Doing this async will make can lead to deliveries being out of order. As well, this could lead to the resurrection being declared done without the delivery queue being fully populated, which means sessions could be delivered out of order.  
  
This is the original PR that made this change: https://github.com/embrace-io/embrace-android-sdk/pull/3159

fractalwrench: This is helpful context. I think I'll close this PR given that the main win was around a long-lived HTTP request.

PR #3723, opened 2026-08-03
Load config as first operation

fractalwrench:
## Goal

Alters the SDK so that config is loaded from disk as the very first operation. #3703 was limited by the fact the OpenTelemetry SDK was created before config was loaded, whereas this feels like the more robust approach (independent of whatever decision we take on whether it makes sense to control distribution via a remote config).


bidetofevil (approved):
This makes sense. Is there a possibility that a corrupt persisted config could lead to the SDK being "stuck" and never enablible or disablible remotely?

fractalwrench:
The change in order shouldn't affect that. The `isSdkDisabled()` check still happens at the same place in `InitializedModuleGraph`, after the config HTTP request has been kicked off.

PR #3728, opened 2026-08-04
Fix build by loosening a test assertion

fractalwrench:
## Goal

Fix build by loosening a test assertion that was checking for a specific snapshot version.


lemnik (approved):
LGTM

PR #3729, opened 2026-08-04
Alter config worker

fractalwrench:
## Goal

Alters the worker thread that the config HTTP request is enqueued on. The config request is enqueued on the IoRegWorker thread. In the worst case of a request taking ~1 min to timeout, this could block everything on the IoRegWorker (shared prefs warming, native crash handler installation, cached data sending). It feels more reasonable to enqueue it on the HttpRequestWorker instead.

Broken out from #3722


lemnik (approved):
LGTM

PR #3731, opened 2026-08-04
Offload system service call in PayloadSourceModule to background

fractalwrench:
## Goal

`PayloadSourceModule` was making a system service call that is possible to move to the background within `EmbraceMetadataService`.


bidetofevil (approved):
LGTM

PR #3732, opened 2026-08-04
Avoid system service calls in AEI/thermal state instrumentation

fractalwrench:
## Goal

Avoids system service calls in AEI/thermal state instrumentation by offloading work to the background thread.


bidetofevil (approved):
LGTM

PR #3733, opened 2026-08-04
Move system service call to background

fractalwrench:
## Goal

Refactors the `NetworkConnectivityService` so that the `ConnectivityManager` is obtained in the background. If the service isn't populated the SDK assumes the network is available.


bidetofevil (approved):
LGTM

PR #3734, opened 2026-08-04
Log flush optimization

fractalwrench:
## Goal

Optimizes the flushing of batched logs by using a similar approach to #3638 (for spans). `Log` is a data class so the current implementation calls `equals()` on every object in its collection.


bidetofevil (approved):
LGTM

PR #3735, opened 2026-08-04
Fix emb-sdk-init span

fractalwrench:
## Goal

The emb-sdk-init span missed the first few bits of SDK initialization after several refactors extracted code. This changeset sets the start time at the very start of SDK init so that the time is accurate.


bidetofevil (approved):
Nice. This should change the `emb-embrace-init` childspan too, eh?

PR #3737, opened 2026-08-04
Add instrumented test that checks for strictmode violations

fractalwrench:
## Goal

Adds an instrumented test that initializes the SDK and checks for strictmode violations, failing the test case if any violations that aren't in an allowlist are detected. This should hopefully allow us to catch any regressions that adds strictmode violations to SDK init.


bidetofevil (approved):
LGTM

PR #3738, opened 2026-08-04
Harden config layer

fractalwrench:
## Goal

Hardens the remote config layer by catching exceptions that currently failed SDK init, and deletes malformed files.

## Testing

Added unit test coverage.


bidetofevil (approved):
LGTM

PR #3739, opened 2026-08-05
Resolve device id lazily

fractalwrench:
## Goal

Resolve device ID lazily rather than eagerly in `ConfigServiceImpl` to avoid hitting disk for the value until absolutely critical


bidetofevil (approved):
LGTM

PR #3740, opened 2026-08-05
Address some strictmode violations

fractalwrench:
## Goal

`CoreModuleImpl` eagerly initializes SharedPreferences at SDK init. This refactors it to make access lazy for all of SDK init, in the happy path where the binary config cache exists already.


bidetofevil (approved):
LGTM

PR #3742, opened 2026-08-05
Improve system service lookup

fractalwrench:
## Goal

Delays calls to `getSystemService` so that they either happen in the background or later on during SDK init (which will allow further optimizations later on).


bidetofevil (approved):
LGTM

PR #3743, opened 2026-08-05
Avoid device ID shared prefs read

fractalwrench:
## Goal

Fixes an eager init of `deviceId` that was triggering `SharedPreferences` in the SDK init warm path.


bidetofevil (approved):
LGTM

PR #3744, opened 2026-08-05
Avoid reading shared prefs in Otel only mode

fractalwrench:
## Goal

Avoids reading `SharedPreferences` when the SDK is configured only to export via OTLP. Currently we check device ID when the threshold is set at the default of 100f - this changeset optimizes that I/O away.


bidetofevil (approved):
LGTM

PR #3745, opened 2026-08-05
Remove cancelled tasks from queue

fractalwrench:
## Goal

Removes cancelled tasks from the queue so that they can be GC'd as soon as they are cancelled. Currently we rely on the worker draining the queue to remove tasks, which is bad if each queue in the task could potentially contain a payload.


bidetofevil (approved):
LGTM

PR #3746, opened 2026-08-05
Make periodic cache interval configurable via remote config

fractalwrench:
## Goal

Alters the periodic cache interval so that we can tweak it via remote config. This is intended as a short-term control that can be used while wider data persistence improvements take place.


bidetofevil (approved):
Can we not publicize this, i.e. don't document it? :sweat_smile:

PR #3747, opened 2026-08-05
Refresh payload in progress checks

fractalwrench:
## Goal

The scheduling service does not check whether payloads exist on disk. Check each time something is scheduled and remove deleted payloads from the in-progress cache, so they can't block delivery.


bidetofevil (approved):
Nice. I assume no race conditions get introduced?

fractalwrench:
I don't think so. This is just tracking the locations of files on disk so removing from those collections should have the effect of simply skipping the deleted files.

PR #3748, opened 2026-08-05
Initialize instrumentation async

fractalwrench:
## Goal

Adds the ability to initialize instrumentation asynchronously with a `DataSource` and converts `AeiDataSource` to use this as an initial test. Fundamentally we are already processing AEI records asynchronously, but we load the instrumentation at SDK init time on the main thread.

If this approach seems good, I'll aim to move a few more non-critical `DataSource` implementations to the background.

## Testing


bidetofevil (approved):
LGTM

PR #3749, opened 2026-08-05
Fix synchronization of start calls

fractalwrench:
## Goal

Properly synchronizes on `Embrace.start()`, along with `stop()` and `disable()`. Previously we were relying on `ModuleInitBootstrapper.init()` which did not exclude every potential double-call.


bidetofevil (approved):
LGTM

PR #3750, opened 2026-08-06
Rename lifecycle tracking classes

fractalwrench:
## Goal

Renames lifecycle tracking classes to better reflect they are tracking the foreground/background process state.


bidetofevil (approved):
LGTM

PR #3751, opened 2026-08-06
Alter OTLP export to avoid runBlocking

fractalwrench:
## Goal

Fixes the OTLP export pathway so that it doesn't call `runBlocking`.

## Testing


bidetofevil (approved):
Very nice way of managing composite exporters. Feels like a good thing to push upstream eventually too.

PR #3752, opened 2026-08-06
Separate androidx process lifecycle tracking

fractalwrench:
## Goal

Separates out androidx's process lifecycle tracking implementation with our SDK's business logic on how to handle changes in the process state.


bidetofevil (approved):
LGTM. After this, we shouldn't have to deal with the AndroidX lifecycle at all? Is there future plans to create a single overarching lifecycle so we have one place to handle this, activity, and even navigation/window lifecycle events too via the EventBus?

fractalwrench:
>After this, we shouldn't have to deal with the AndroidX lifecycle at all

Yes, once we've battle-tested this implementation it would be feasible to remove that dependency completely.

PR #3754, opened 2026-08-06
Make app startup counter lazy init

fractalwrench:
## Goal

Alters the app startup counter to use lazy init so that it isn't resolved until later in SDK initialization. This allows `SharedPreferences` a chance to prewarm.


bidetofevil (approved):
LGTM. The intention of the eager init was so startup crashes will be less likely to kill the app before we had a chance to increment. But the trade-off isn't worth it if everything is lazy now, so this looks good.

PR #3755, opened 2026-08-06
Make instrumentation initialize async

fractalwrench:
## Goal

Updates more instrumentation to initialize async, where this is feasible.


bidetofevil (approved):
LGTM. Any concerns about the thermal and power save instrumentations losing the initial state that the app is initialized under if the delta is significant?

fractalwrench:
IMO losing initial state if the delta is very high (seconds) is an acceptable trade-off.

PR #3756, opened 2026-08-06
Implement process lifecycle tracking

fractalwrench:
## Goal

Implements process lifecycle tracking based on ProcessLifecycleOwner, without the dependency.

## Testing

Added unit tests.


bidetofevil (approved):
Looks great!

PR #3757, opened 2026-08-06
Initialize network connectivity checks async

fractalwrench:
## Goal

Initializes network connectivity checks on a background thread rather than main during SDK init.


bidetofevil (approved):
LGTM for what we are trying to do here. But we may want to further refactor this later on - `register()` for the service is already posting the registration to a background worker, so adding those listeners in the main thread won't make them get the latest status any faster. 

The original idea -- to get the current status first and then replay it to any added listeners -- was partially undone by the registration of the service being async, so we might as well do these in order all in the background so we get the benefit of the determistic init order AND not suffer the perf hit.  
  
I can take care of that after this is merged unless you want to?

fractalwrench:
That would be great!

PR #3759, opened 2026-08-06
Push back shared prefs first read

fractalwrench:
## Goal

Pushes back the first read of SharedPreferences by making some properties lazy.


bidetofevil (approved):
LGTM

PR #3761, opened 2026-08-06
Provide config to control process lifecycle tracking

fractalwrench:
## Goal

Provides a config that controls which process lifecycle tracking implementation should be used.

## Testing

Added unit tests.


bidetofevil (approved):
LGTM.  
  
Is this more like a safety valve or do we anticipate some differences in production, like the removal of certain race conditions, the order of activity callback invocations as we transition from foreground/background, etc.?

fractalwrench:
By default our own implementation is off. However, I'd like to flip that after we've validated the implementation and verified with a few customers. There are some subtle differences in how lifecycle is handled that are a consequence of using our own implementation, but IMO the benefit is worth it.

PR #3762, opened 2026-08-06
Avoid setting jailbroken in SharedPreferences

fractalwrench:
## Goal

Avoid setting jailbroken in SharedPreferences.


bidetofevil (approved):
LGTM

PR #3763, opened 2026-08-06
Avoid storing screen resolution in shared prefs

fractalwrench:
## Goal

Avoid storing screen resolution in shared prefs as it's probably faster to make the API call than retrieve from disk.


bidetofevil (approved):
Interesting implication tbf. I wonder if this applies to anything else?

PR #3764, opened 2026-08-06
Optimize how ordinals are incremented

fractalwrench:
## Goal

Optimizes how ordinals are incremented by moving everything under one `edit`.


bidetofevil (approved):
Nice. LGTM

PR #3765, opened 2026-08-07
Batch shared preferences writes

fractalwrench:
## Goal

Every `apply()` call to `SharedPreferences` [enqueues a disk write](https://cs.android.com/android/platform/superproject/+/android-latest-release:frameworks/base/core/java/android/app/SharedPreferencesImpl.java;l=484?q=sharedpreferencesimpl) on a background thread. We can reduce the number of disk writes by batching edits of SharedPreferences within our SDK.

### DIFF
--- a/embrace-android-infra/src/main/kotlin/io/embrace/android/embracesdk/internal/store/KeyValueStore.kt
+++ b/embrace-android-infra/src/main/kotlin/io/embrace/android/embracesdk/internal/store/KeyValueStore.kt
@@ -39,4 +39,11 @@ interface KeyValueStore {
      * Performs a batch edit of values in the key-value store.
      */
     fun edit(action: KeyValueStoreEditor.() -> Unit)
+
+    /**
+     * Coalesces every [edit] performed on the calling thread inside [action] into a single commit,
+     * which is issued when the outermost [batch] returns. As with [edit], the commit happens even if
+     * [action] throws.
+     */
+    fun batch(action: () -> Unit) = action()
 }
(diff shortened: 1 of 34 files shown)


embrace-android-infra/src/main/kotlin/io/embrace/android/embracesdk/internal/store/KeyValueStore.kt
```diff
@@ -39,4 +39,11 @@ interface KeyValueStore {
      * Performs a batch edit of values in the key-value store.
      */
     fun edit(action: KeyValueStoreEditor.() -> Unit)
+
+    /**
+     * Coalesces every [edit] performed on the calling thread inside [action] into a single commit,
+     * which is issued when the outermost [batch] returns. As with [edit], the commit happens even if
+     * [action] throws.
+     */
+    fun batch(action: () -> Unit) = action()
```
bidetofevil, line 51: It almost feels like this should be the default - or at least change the naming of `edit` that each call will be done on its own along with the perf penalty that is paid - so the perf semantics are clear and it only gets used sparingly when you are only editing one thing in the store?

bidetofevil (approved):
LGTM. One non-blocking naming suggestion to push ppl to use the batch impl and discourage and callout the use of the existing `edit` as only appropriate if you are adding/changing one key.

embrace-android-infra/src/main/kotlin/io/embrace/android/embracesdk/internal/store/KeyValueStore.kt
fractalwrench: Renamed per feedback

PR #3766, opened 2026-08-07
Create a module for persisting sessions

fractalwrench:
## Goal

Creates a module that will be used for persisting sessions to disk.


bidetofevil (approved):
NICE

PR #3797, opened 2026-08-13
Implement SessionPartDirectory

fractalwrench:
## Goal

Implements `SessionPartDirectory`, which will be responsible for holding the metadata that allows session part directories to be constructed and manipulated.

## Testing

Added unit tests.


bidetofevil (approved):
Here we go :-)

PR #3802, opened 2026-08-13
Implement persistence layer config

fractalwrench:
## Goal

Implements a config that controls how the persistence layer should behave (i.e. whether we should enable multi-file writes in parallel).

## Testing

Added unit tests.


embrace-android-payload/src/main/kotlin/io/embrace/android/embracesdk/internal/config/remote/MultiFilePersistenceRemoteConfig.kt
```diff
@@ -0,0 +1,14 @@
+package io.embrace.android.embracesdk.internal.config.remote
+
+import kotlinx.serialization.SerialName
+import kotlinx.serialization.Serializable
+
+@Serializable
+data class MultiFilePersistenceRemoteConfig(
```
bidetofevil: Fredric now prefers config keys in the main remote config object if it's just going to be that, rather than nesting it in another object. Are we expecting more keys here to justify the wrapping object?

bidetofevil (approved):
LGTM - no blocking comment about whether we should have the key just be in the RemoteConfig rather than nested.

embrace-android-payload/src/main/kotlin/io/embrace/android/embracesdk/internal/config/remote/MultiFilePersistenceRemoteConfig.kt
fractalwrench: Hadn't realised there was a change in preference. I can update it to be a top-level config for now

PR #3803, opened 2026-08-13
Implement manifest file writer

fractalwrench:
## Goal

Implements a class that writes the manifest file for a session part.

## Testing

Added unit tests.


bidetofevil (approved):
LGTM. We'll have to rework this quite a bit when the resource attributes aren't going to be known in advance and we store them all as map entries 😅

PR #3804, opened 2026-08-13
Implement initial session reconstruction service

fractalwrench:
## Goal

Implements the session reconstruction service, initially just deserializing the manifest file into an `Envelope<SessionPartPayload>`.

## Testing

Added unit tests.


bidetofevil (approved):
LGTM

PR #3809, opened 2026-08-14
Encode session part directory version in filename

fractalwrench:
## Goal

Encodes a version for the session part directory in its filename. This will be used to indicate changes in the format that are notable enough to require modifications to our persistence/deserialization logic.

## Testing

Added unit tests.


bidetofevil (approved):
LGTM

bidetofevil (approved):
LGTM

PR #3812, opened 2026-08-14
Reconstruct metadata in session part payloads

fractalwrench:
## Goal

Alters the `SessionReconstructionService` so that it deserializes metadata and adds it to session part payloads.

## Testing

Added unit tests.


bidetofevil (approved):
LGTM

PR #3813, opened 2026-08-14
Create session span file writer

fractalwrench:
## Goal

Creates a class that writes a session span file.

## Testing

Added unit tests.

### DIFF
--- a/embrace-android-session-persistence/src/main/proto/embrace/session/persistence/session_span.proto
+++ b/embrace-android-session-persistence/src/main/proto/embrace/session/persistence/session_span.proto
@@ -1,0 +1,12 @@
+syntax = "proto3";
+
+package embrace.session.persistence;
+
+import "embrace/session/persistence/span.proto";
+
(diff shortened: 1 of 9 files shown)


embrace-android-session-persistence/src/main/proto/embrace/session/persistence/session_span.proto
```diff
@@ -0,0 +1,12 @@
+syntax = "proto3";
+
+package embrace.session.persistence;
+
+import "embrace/session/persistence/span.proto";
+
+option java_package = "io.embrace.android.embracesdk.internal.session.persistence";
+
+message SessionSpan {
```
bidetofevil, line 9: Should this be SessionPartSpan to be consistent?

bidetofevil (approved):
LGTM. Non-blocking comment about using the name "SessionPartSpan" rather than just "SessionSpan" here and everywhere else for consistency.

PR #3814, opened 2026-08-14
Read session span

fractalwrench:
## Goal

Reads the session span in the `SessionReconstructionService`.

## Testing

Added unit tests.

### DIFF
--- a/embrace-android-session-persistence/src/main/kotlin/io/embrace/android/embracesdk/internal/session/persistence/SessionReconstructionService.kt
+++ b/embrace-android-session-persistence/src/main/kotlin/io/embrace/android/embracesdk/internal/session/persistence/SessionReconstructionService.kt
@@ -56,15 +56,32 @@ class SessionReconstructionService(
             EnvelopeMetadataProto::format_version,
         )?.toPayload() ?: return null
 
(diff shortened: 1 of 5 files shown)


embrace-android-session-persistence/src/main/kotlin/io/embrace/android/embracesdk/internal/session/persistence/SessionReconstructionService.kt
```diff
@@ -56,15 +56,32 @@ class SessionReconstructionService(
...
+        val sessionSpan = readPartFile(
+            partDir,
+            SESSION_SPAN_FILE_NAME,
+            SessionSpan.ADAPTER,
+            SessionSpan::format_version,
+        ) ?: return null
+
+        val span = sessionSpan.span
+        if (span == null) {
+            trackFailure(IOException("Session span file has no span: ${partDir.path}"))
```
bidetofevil, line 68: Do we care about the actual path in the error message in production? This will make precise aggregation harder if we want to narrow down this type of failure at this location. It's fine if you think it's useful though.

bidetofevil (approved):
LGTM. Non-blocking comment about path details in the internal error exception.

PR #3815, opened 2026-08-14
Write span snapshots

fractalwrench:
## Goal

Implement a class that is capable of writing span snapshots to a file.

## Testing

Added unit tests.


bidetofevil (approved):
LGTM

PR #3826, opened 2026-08-17
Address review feedback

fractalwrench:
## Goal

Renames the `SessionSpan` protobuf object to `SessionPartSpan` and removes a path from an error message.


lemnik (approved):
LGTM

PR #3842, opened 2026-08-18
Support reading span snapshots

fractalwrench:
## Goal

Support reading span snapshots in the `SessionReconstructionService`.

## Testing

Added unit tests.


lemnik (approved):
LGTM

PR #3844, opened 2026-08-18
Implement limit on file size when reconstructing session

fractalwrench:
## Goal

Adds an overall limit on file size when reconstructing sessions. This is meant as a backstop rather than the only limit, and the data limits for individual items within the session will be applied in a future changeset.

## Testing

Added unit tests.


lemnik (approved):
LGTM

PR #3845, opened 2026-08-19
[Integration] Experiments feature

fractalwrench:
## Goal

Integration branch for the experiments feature


embrace-android-sdk/src/main/kotlin/io/embrace/android/embracesdk/internal/api/delegate/ExperimentApiDelegate.kt
```diff
@@ -0,0 +1,135 @@
...
+        if (!sdkCallChecker.started.get()) {
+            val admitted = admitEntries(ids) ?: return
+            pendingEvents.add(PendingEvent.Untrack(action, kind, admitted, endTimeMs))
+        } else {
+            untrackNow(action, kind, ids, endTimeMs)
+        }
+    }
+
+    private fun trackNow(action: String, data: List<TrackedData>) {
+        if (sdkCallChecker.check(action)) {
```
fnewberg, line 95: a thread that reads `started == false` and is descheduled can enqueue after `flushPendingCalls()` has already drained. `flushPendingCalls()` runs once, so the entry is never delivered

embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/capture/experiment/ExperimentTrackingServiceImpl.kt
```diff
@@ -0,0 +1,135 @@
...
+        var updated = false
+        synchronized(lock) {
+            data.forEach { entry ->
+                if (trackRecord(entry.toRecord())) {
+                    updated = true
+                }
+            }
+        }
+        if (updated) {
+            publishRecords()
```
fnewberg: since `publishRecords` happens outside the lock, it is possible that  two racing `track` calls can read snapshots in one order and write the attribute in the other, leaving the session-part span holding the older
version.

embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/capture/experiment/ExperimentRecord.kt
```diff
@@ -0,0 +1,21 @@
...
+ */
+internal data class ExperimentRecord(
+    val kind: ExperimentKind,
+    val id: String,
+    val variant: String?,
+    val startTimeMs: Long,
+    val endTimeMs: Long?,
+) {
+
+    fun serialize(): String =
```
fnewberg, line 14: in the [apple SDK PR](https://github.com/embrace-io/embrace-apple-sdk/pull/734/changes#diff-dcbe66391722fdc9d9ef69e3d6dd01d90ecbeab7a2bcce8cdfc81f4e111c284c), we had a similar implementation to this and changed it so that the object holds the serialized value, with escaping and serialization only happening on creation or mutation of a record, with `endTimeMs` being the only thing that can mutate. 

this prevents potentially large numbers of calls when you have many records and have to re-serialize _all_ of them every time you  add a new record

embrace-android-sdk/src/main/kotlin/io/embrace/android/embracesdk/EmbraceImpl.kt
```diff
@@ -125,6 +129,7 @@ internal class EmbraceImpl(
                         // not fully initialized, but the SDK shouldn't catastrophically throw after this point,
                         // so we allow external calls.
                         sdkCallChecker.started.set(true)
+                        experimentApiDelegate.flushPendingCalls()
```
fnewberg: i'll leave it up to you to decide on where and how to address this, but i think we need to rethink how this is done currently. this is the current call stack from here

```
  Embrace.start(context)
  └─ EmbraceImpl.start()                                          EmbraceImpl.kt:132
     └─ ExperimentApiDelegate.flushPendingCalls()                 ExperimentApiDelegate.kt:52-62
        │
        └─ while (true) { pendingEvents.poll() ?: break }         :56        ← N iterations
           └─ ExperimentApiDelegate.trackNow(action, data)        :58 → :82-86
              ├─ sdkCallChecker.check(action)                     :83        ← N public-API events
              └─ ExperimentTrackingServiceImpl.track(data)        ExperimentTrackingServiceImpl.kt:28-40
                 │
                 ├─ synchronized(lock) { data.forEach { trackRecord(it.toRecord()) } }   :30-36
                 │  └─ records[key] = record                                             :78
                 │     cacheValid = false                                                :79   ← invalidates
                 │
                 └─ if (updated) publishRecords()                                        :38
                    └─ publishRecords()                                                  :93-97
                       ├─ getRecords()                                                   :56-66
                       │  └─ if (!cacheValid)                                            :57   ← always true
                       │     └─ records.values.joinToString(";") { it.serialize() }       :61   ← k records
                       │        └─ ExperimentRecord.serialize()                  ExperimentRecord.kt:14-16
                       │           ├─ id.escape()                                        :18-20  → 3 × replace()
                       │           └─ variant?.escape()                                  :18-20  → 3 × replace()
                       │
                       └─ telemetryDestination.addSessionPartAttribute(EMB_EXPERIMENTS, value)  :95
                          └─ TelemetryDestinationImpl.addSessionPartAttribute()   TelemetryDestinationImpl.kt:189-193
                             └─ EmbraceSpanFactoryImpl.addSystemAttribute()       EmbraceSpanFactoryImpl.kt:432-440
```

`ExperimentTrackingServiceImpl` has a serialization cache (`cacheValid` / `cachedRecords`, `:25-26`), but every mutation invalidates it at `:79`. Inside the flush, each buffered event mutates and then immediately publishes, so the cache never survives to be read twice. Every publish re-serializes the entire accumulated record set from scratch.

The k-th buffered event publishes with k records in the map, so total serialize() calls across the flush is Σ(k=1..N) k = N(N+1)/2:

| Buffered single-entry calls   | serialize() calls during flush |
|-------------------------------|--------------------------------|
| 100                           | 5,050                          |
| 1,000                         | 500,500                        |
| 5,000 (PENDING_ENTRY_LIMIT)   | 12,502,500                     |
  
This results in a lot of redundant attribute writes through `addSystemAttribute` → `notifySpanUpdate()`, and `sdkCallChecker.check()` calls. We should aim to just publish once after we've drained the queue.

fnewberg: `flushPendingCalls()` is the first statement after `started.set(true)`, inside the window the
comment two lines above explicitly designates as must-not-throw ("the SDK shouldn't catastrophically
throw after this point"). It's also the first thing on that path ever to resolve
`experimentTrackingService` — `EmbraceImplFieldDelegate` returns `null` until `started` is true, so
nothing here has been exercised before this call.

It's also the only statement in the block driven by customer input: it replays
customer-supplied strings in customer-chosen quantity, up to `PENDING_ENTRY_LIMIT` (5,000) entries.
Everything following in the lines `:133-139` operates on the SDK's own state.

Consider moving the flush below `bootstrapper.triggerPayloadSend()` at `:137`, or give it its own inner
try/catch, so a bad replay can't skip the rest of SDK init.

embrace-android-api/src/main/kotlin/io/embrace/android/embracesdk/experiments/TrackedExperiment.kt
```diff
@@ -0,0 +1,23 @@
+package io.embrace.android.embracesdk.experiments
+
+/**
+ * An experiment this app instance has been bucketed into.
+ */
+public interface TrackedExperiment {
```
fnewberg, line 6: [claude flagged this. not sure how realistic of an issue this is]

`TrackedExperiment` is a public interface, so nothing stops somebody from implementing it, and
`experiments.map { it.toData() }` in `ExperimentApiDelegate` is evaluated as the argument expression
 before `sdkCallChecker.check()` and with no guard:

- a Java caller passing `Arrays.asList(exp, null)` throws NPE out of `Embrace.trackExperiments` on
  the caller's thread;
- `override val id get() = throw IllegalStateException()` propagates into the app.

`LogsApiDelegate.kt:159` and `:239` already use `runCatching` for exactly this, so we should consider using that here as well. `TrackedFeatureFlag` has the same shape.

If we make both final classes with public `@JvmOverloads` constructors, that removes the hazard,
gives `equals`/`hashCode`, fixes the Java-ergonomics gap (callers currently write 
`trackExperiment("id", null, null)` with boxed `Long`s), and deletes six permanent `DefaultImpls`
ABI entries. Failing that, `runCatching` around the mapping, or seal the interfaces so only
`createExperiment`/`createFeatureFlag` can produce them.

embrace-android-api/src/main/kotlin/io/embrace/android/embracesdk/internal/api/ExperimentApi.kt
```diff
@@ -0,0 +1,75 @@
...
+     * Creates a [TrackedExperiment] with an optional [variant] and a timestamp [startedAt] in milliseconds from epoch observed from
+     * the client-side that denotes when the experiment variant was first applied.
+     */
+    public fun createExperiment(id: String, variant: String? = null, startedAt: Long? = null): TrackedExperiment
+
+    /**
+     * Tracks a single experiment membership that this app instance has been bucketed into. A null [startedAt] means the time at which
+     * this call is made will be used.
+     */
+    public fun trackExperiment(id: String, variant: String? = null, startedAt: Long? = null): Unit =
```
fnewberg, line 23: i think most of our other APIs, e.g. `addUserSessionProperty`,  return a boolean to indicate success or failure. do we want to stick with that convention or are we moving away from that?

embrace-android-api/src/main/kotlin/io/embrace/android/embracesdk/experiments/TrackedFeatureFlag.kt
```diff
@@ -0,0 +1,18 @@
+package io.embrace.android.embracesdk.experiments
+
+/**
+ * A single feature flag that has been enabled on this app instance.
+ */
+public interface TrackedFeatureFlag {
```
fnewberg, line 6: No `variant`. Apple's `TrackedFeatureFlag` has one and we should align with that

embrace-android-sdk/src/main/kotlin/io/embrace/android/embracesdk/internal/api/delegate/ExperimentApiDelegate.kt
```diff
@@ -0,0 +1,141 @@
...
+        }
+        val admitted = entries.take(remaining)
+        bufferedEntryCount.addAndGet(admitted.size)
+        return admitted
+    }
+
+    private fun TrackedExperiment.toData(): TrackedData =
+        TrackedData.Experiment(
+            id = id,
+            startTimeMs = startedAt ?: now(),
```
fnewberg, line 128: `now()` is sampled per entry rather than once per call to `Embrace.trackExperiments(...)`. it is called here and `:123` while mapping the list, so entries in one bulk call that omit a start time get different timestamps.

Apple samples `let now = Date()` once per call to its equivalent of `Embrace.trackExperiments(...)`, so the same customer code produces different data per platform.

embrace-android-api/src/main/kotlin/io/embrace/android/embracesdk/internal/api/ExperimentApi.kt
```diff
@@ -0,0 +1,75 @@
...
+
+import io.embrace.android.embracesdk.experiments.TrackedExperiment
+import io.embrace.android.embracesdk.experiments.TrackedFeatureFlag
+
+/**
+ * The public API used to track experiment and feature-flag state on a given device for the current app instance. IDs are unique within
+ * each kind, so an experiment and a feature flag may share an ID. The SDK does not persist this state, and users need to call this each
+ * time the app starts up.
+ */
+public interface ExperimentApi {
```
fnewberg, line 11: should this be marked with `@InternalApi` to match other interfaces in `internal.api`?

embrace-android-api/src/main/kotlin/io/embrace/android/embracesdk/internal/api/ExperimentApi.kt
```diff
@@ -0,0 +1,75 @@
+package io.embrace.android.embracesdk.internal.api
+
+import io.embrace.android.embracesdk.experiments.TrackedExperiment
+import io.embrace.android.embracesdk.experiments.TrackedFeatureFlag
+
+/**
+ * The public API used to track experiment and feature-flag state on a given device for the current app instance. IDs are unique within
+ * each kind, so an experiment and a feature flag may share an ID. The SDK does not persist this state, and users need to call this each
+ * time the app starts up.
+ */
```
fnewberg, line 10: is this the appropriate location to document drop behavior, e.g.

- the limits (500 records, 128-char id, 128-char variant by default);
- that over-length values are **dropped, not truncated** — worth stating explicitly because the rest
  of this SDK truncates, so the safe assumption is wrong here;
- that an over-length _variant_ drops the whole record, not just the variant;
- that re-tracking an ID after untracking it is permanently ignored;
- that untrack retains the record and its slot against the cap rather than freeing it.

embrace-android-config/src/main/kotlin/io/embrace/android/embracesdk/internal/config/behavior/ExperimentBehaviorImpl.kt
```diff
@@ -0,0 +1,33 @@
+package io.embrace.android.embracesdk.internal.config.behavior
+
+import io.embrace.android.embracesdk.internal.config.remote.RemoteConfig
+import kotlin.math.min
+
+class ExperimentBehaviorImpl(
+    private val remote: RemoteConfig?,
+) : ExperimentBehavior {
+
+    override fun getMaxExperimentCount(): Int = min(
```
fnewberg, line 10: should we protect against invalid values, e.g. -1?

embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/capture/experiment/ExperimentTrackingServiceImpl.kt
```diff
@@ -0,0 +1,135 @@
...
+        is TrackedData.FeatureFlag -> ExperimentRecord(
+            kind = ExperimentKind.FEATURE_FLAG,
+            id = id.stripWhitespace(),
+            variant = null,
+            startTimeMs = startTimeMs,
+            endTimeMs = null,
+        )
+    }
+
+    private fun isValid(record: ExperimentRecord): Boolean {
```
fnewberg, line 155: should `startTimeMs` and `endTimeMs` be validated? the Apple SDK rejects non-representable dates before they reach the encoder (`Date.swift:33-53`), so data from the two platforms can potentially disagree.

embrace-android-sdk/src/main/kotlin/io/embrace/android/embracesdk/internal/api/delegate/ExperimentApiDelegate.kt
```diff
@@ -0,0 +1,141 @@
...
+
+    // Use the system clock if the SDK hasn't been initialized and the SDK clock is unavailable.
+    private fun now(): Long = clock?.now() ?: System.currentTimeMillis()
+
+    /**
+     * Return the entries to be allowed given the cap. Any entries that will put the total over the cap will be dropped.
+     */
+    private fun <T> admitEntries(entries: List<T>): List<T>? {
+        if (entries.isEmpty()) {
+            return entries
```
fnewberg: nitpick: since this returns the empty list unchanged rather than signaling "nothing to do", this will get added to the list of pending events

embrace-android-otel/src/test/kotlin/io/embrace/android/embracesdk/internal/otel/spans/EmbraceSpanImplTest.kt
```diff
@@ -494,6 +495,18 @@ internal class EmbraceSpanImplTest {
...
     }
 
+    @Test
+    fun `reserved experiment key attribute is not settable on internal and public spans via the addAttribute method`() {
+        listOf(embraceSpan, createInternalEmbraceSdkSpan()).forEach { span ->
+            with(span) {
+                assertTrue(start())
+                assertFalse(addAttribute(key = EmbCommonAttributes.EMB_EXPERIMENTS, value = "spoof"))
+                assertNull(attributes()[EmbCommonAttributes.EMB_EXPERIMENTS])
+                assertNull(embraceSpan.snapshot()?.attributes?.findAttributeValue(EmbCommonAttributes.EMB_EXPERIMENTS))
```
fnewberg: The assertion doesn't use the loop variable. Inside `forEach { span -> with(span) { … } }` this line
reads `embraceSpan.snapshot()`, so the internal-span iteration re-checks the public span and the
internal case is never actually asserted. `:503-504` are correct; only this line is wrong.

embrace-android-sdk/src/test/kotlin/io/embrace/android/embracesdk/internal/api/delegate/ExperimentApiDelegateTest.kt
```diff
@@ -0,0 +1,312 @@
...
+        val flushedIds = fakeExperimentTrackingService.trackedData.map { it.id }
+        assertEquals(PENDING_ENTRY_LIMIT, flushedIds.size)
+        assertTrue(flushedIds.contains("exp-0"))
+        assertTrue(flushedIds.contains("exp-kept"))
+        assertFalse(flushedIds.contains("exp-dropped"))
+        assertFalse(flushedIds.contains("exp-after-full"))
+    }
+
+    @Test
+    fun `replay after SDK start passes every buffered entry to the store, which enforces the configured cap`() {
```
fnewberg: The test name claims something nothing does. `…which enforces the configured cap` passes
`RemoteConfig(experimentMaxCount = 2)`, but the delegate never reads `ConfigService` — its cap is the compile-time `MAX_EXPERIMENT_COUNT_LIMIT` — and `FakeExperimentTrackingService` enforces no cap at all. Nothing in this test can fail if the configured cap stops being honored.

fnewberg (changes requested):

embrace-android-sdk/src/main/kotlin/io/embrace/android/embracesdk/internal/api/delegate/ExperimentApiDelegate.kt
bidetofevil: `flushPendingCalls()` will only be invoked if `started` is set to true, and it can only ever get back to false if the SDK is disabled, which means the telemetry shouldn't be recorded anyway. 

SDK start in practice is considered a one-way street, and you can't stop and the instrumentation is not expected to survive a stop, i.e. the SDK isn't restartable once it's stopped.

embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/capture/experiment/ExperimentTrackingServiceImpl.kt
bidetofevil: The `emb.experiments` snapshot takes whatever is in memory and write it to the attribute, and it is not mean to be read as the single source of true. That job belongs to the data in memory, which is synchronized, and its modification is not subject to races. 

To put it another way, the snapshot is always accurate to the point of when it was taken. The concern that what is on the session span is stale is true only in the sense that the in-memory data is updated but the snapshot hasn't, which is not something we guarantee to be atomic. 

The small window between when the data in memory is updated to when the background thread runs to update the session part span is an accepted trade off we make to not further delay the calling thread any time we log telemetry.

embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/capture/experiment/ExperimentRecord.kt
bidetofevil: This method is only called when we serialize ALL the in-memory data onto the session part span. 

For a bulk call, that happens once (i.e. `publishRecords()`). For a single-experiment syntactic sugar call, we have to make sure that when it exits, we queued a write of the current value to the session part span, so we can't get away we calling it every time. But we explicit discourage that pattern and point folks to the bulk API, which would not have this problem when multiple experiments are being added/updated. And if they use it in a loop when all the experiment data to apply are known, they shouldn't or they will incur that penalty.

embrace-android-sdk/src/main/kotlin/io/embrace/android/embracesdk/EmbraceImpl.kt
bidetofevil: Ugh. Yeah this should be using the bulk API. I'll fix this. It slipped through the rounds of refactoring.

bidetofevil: We need to flush the pending calls early so the telemetry recorded after that is properly tagged. It is not considered an optional part of the init - if the experiments application is not applied, we consider the entire SDK  startup a failure and not try to run with an unknown error.

Also, if we throw here, along with every subsequent call, it is caught at line 140, so no harm no foul, even if the unexpected happens, in the worst case.

embrace-android-config/src/main/kotlin/io/embrace/android/embracesdk/internal/config/behavior/ExperimentBehaviorImpl.kt
bidetofevil: These are coming from our remote service - we can validate the data before using it, but I think in general, we expect data coming from our side to be validated already, so this is not done.

If we want to do this kind of protection, we should probably do it at a higher level - when we ingest and deserialize the object. We drop values that don't conform to our expectations (e.g. non-Ints are dropped), so we can validate negative values too. But if we do it here, we should probably do it for every remote config

fnewberg: > But if we do it here, we should probably do it for every remote config

yep. let's discuss during the next guild meeting

embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/capture/experiment/ExperimentTrackingServiceImpl.kt
fnewberg: fair enough 🙂

embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/capture/experiment/ExperimentRecord.kt
fnewberg: > This method is only called when we serialize ALL the in-memory data onto the session part span.

right, but that happens a _lot_ with the current implementation. take a look at what the apple SDK does and see if it's worth making the same change here that we made from the original implementation

embrace-android-sdk/src/main/kotlin/io/embrace/android/embracesdk/EmbraceImpl.kt
fnewberg: > We need to flush the pending calls early so the telemetry recorded after that is properly tagged. It is not considered an optional part of the init - if the experiments application is not applied, we consider the entire SDK startup a failure and not try to run with an unknown error.

probably quicker to discuss this on a call

embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/capture/experiment/ExperimentTrackingServiceImpl.kt
bidetofevil: We can. We don't normally validate customer-provided timestamps beyond trying to detect a nano-milli mismatch when we write to the OTel and fail there, but since this is going into a blob, we can do it here too.

embrace-android-sdk/src/main/kotlin/io/embrace/android/embracesdk/internal/api/delegate/ExperimentApiDelegate.kt
bidetofevil: We can return null here to by pass that.

embrace-android-otel/src/test/kotlin/io/embrace/android/embracesdk/internal/otel/spans/EmbraceSpanImplTest.kt
bidetofevil: This is to ensure that the change somehow didn't make onto the session span, not validating the span in the loop. I can add a comment to make it more obvious

embrace-android-sdk/src/test/kotlin/io/embrace/android/embracesdk/internal/api/delegate/ExperimentApiDelegateTest.kt
bidetofevil: Ah, I didn't remove/refactor the test after the limits requirements/impl changed. I'll remove this test if we already have one that tests the limits to the buffer.

embrace-android-sdk/src/main/kotlin/io/embrace/android/embracesdk/EmbraceImpl.kt
bidetofevil: Actually, while I can bulkify this and coalesce all the similar calls into a single bulk call, the intent was just to replay the calls the user has made. So if they made 10 non-bulk calls in a loop, they'll get 10 record rewrites - but if they use the API correctly, it should just be one call.

If truly want to do one write at SDK startup time, I can create an internal "don't commit" version of the track and untrack methods and just not expose this publicly. That'll involve a bit more complexity and really reduce the hit to the bare minimum.

Under most usage scenarios, you won't really want to untrack an experiment, but I can see multiple teams track different experiments on their own. Rewriting this attribute a handful of times, one for each team's bulk call, won't typically take that long, but doing this will make it go faster at the expense of a more complex service.

Is this an optimization you think we need?

embrace-android-otel/src/test/kotlin/io/embrace/android/embracesdk/internal/otel/spans/EmbraceSpanImplTest.kt
bidetofevil: Ignore my last comment if you'd read it - I read the test wrong. Yeah, I'll fix this

embrace-android-sdk/src/test/kotlin/io/embrace/android/embracesdk/internal/api/delegate/ExperimentApiDelegateTest.kt
bidetofevil: This doesn't do anything now. We are now only buffering calls that add entries to the system, and the test case above this already verifies the total number of records that an be buffered. I'll just delete this.

embrace-android-sdk/src/main/kotlin/io/embrace/android/embracesdk/EmbraceImpl.kt
fnewberg: > Is this an optimization you think we need?

yes

bidetofevil: Per discussion, I've added a PR (#3928) to do as much combining as possible BEFORE SDK startup without introducing the additional complexity on the service that I was concerned about.

This ensures all tracking calls made before the SDK starts are combined and replayed in the same order, which is the important workflow. Untracking calls are combined as much as possible in a best-effort way, but they can trigger additional blob construction if used before the SDK starts.

Experiment tracking calls made after the SDK starts will not have the benefit of the coalescing as we don't want to introduce the kind of edit/commit transaction concepts into this API like SharedPreferences. Post SDK startup, we won't expect tight loops of consecutive blob-mutating API calls

bidetofevil: We didn't discuss this yesterday. Are you good with my assertion? I think failing the entire SDK when a cross-cutting component like experiments fails to load is the right thing to do. This isn't instrumentation, which could fail individually -- this affects metadata, so we should fail loudly and consistently -- without crashing the app, of course.

fnewberg: i have a few concerns about this

- would we be aware that this was happening? in customers' eyes i think this would come across as missing data
- if i had a choice as a dev, i'd rather take session parts without experiment info rather than not having any session parts at all. throwing out all data just because something went wrong with experiment config seems wrong. if we took this approach, we should include some indicator that this failure happened

bidetofevil (approved):
Looks good to me, the maker of these changes lol

embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/capture/experiment/ExperimentRecord.kt
bidetofevil: Per discussion, calls pre-sdk start (which is where the bulk of the experiments will be applied) has been combined as much as possible, which under typical usage, will result in one session-part-span-update/serialization. I'll consider this resolved.

embrace-android-sdk/src/main/kotlin/io/embrace/android/embracesdk/EmbraceImpl.kt
bidetofevil: Per Slack discussion, I'll identify all the modular pieces of SDK startup that can fail without affecting core functionality (i.e. recording and delivering of telemetry) and allow them to fail without taking down the subsequent steps in SDK init, including this. 

Note: these steps already shouldn't throw given the internals, but we'll add another level of safety so we can better catalogue unexpected and non-fatal failures in the future.

embrace-android-config/src/main/kotlin/io/embrace/android/embracesdk/internal/config/behavior/ExperimentBehaviorImpl.kt
bidetofevil: Did not discuss in the meeting but I've raised this in Slack. I don't consider this blocking since this is the status quo, so we can pursue that async.

embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/capture/experiment/ExperimentTrackingServiceImpl.kt
bidetofevil: Same as the remote-config semantic validation: will discuss this issue that exists beyond this feature and handle it async from this.

fnewberg (approved):

PR #3853, opened 2026-08-20
Create components for writing session part directory

fractalwrench:
## Goal

Adds components that will be used to write the session part directory.

## Testing

Added unit tests.

### DIFF
--- a/embrace-android-telemetry-persistence/src/main/kotlin/io/embrace/android/embracesdk/internal/delivery/storage/StoredEntryIndex.kt
+++ b/embrace-android-telemetry-persistence/src/main/kotlin/io/embrace/android/embracesdk/internal/delivery/storage/StoredEntryIndex.kt
@@ -1,0 +1,110 @@
+package io.embrace.android.embracesdk.internal.delivery.storage
+
+import io.embrace.android.embracesdk.internal.clock.Clock
+import io.embrace.android.embracesdk.internal.logging.InternalErrorType
+import io.embrace.android.embracesdk.internal.logging.InternalLogger
+import io.embrace.android.embracesdk.internal.utils.threadSafeToList
+import java.io.File
+import java.io.FileNotFoundException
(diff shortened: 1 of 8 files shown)


embrace-android-telemetry-persistence/src/main/kotlin/io/embrace/android/embracesdk/internal/delivery/storage/StoredEntryIndex.kt
```diff
@@ -0,0 +1,110 @@
...
+        }
+
+        newEntry ?: return false
+
+        // remove entries by count
+        val count = entries.size
+        if (count < storageLimit) {
+            return false
+        }
+        val input = entries.plus(newEntry)
```
lemnik: nit: using the `plus` operator by name here seems a bit arbitrary?

```suggestion
        val input = entries + newEntry
```

embrace-android-telemetry-persistence/src/main/kotlin/io/embrace/android/embracesdk/internal/delivery/storage/StoredEntryIndex.kt
```diff
@@ -0,0 +1,110 @@
...
+        val removalCount = input.size - storageLimit
+        if (removalCount < 0) {
+            return false
+        }
+        val removals = input.sortedWith(layout.removalComparator).take(removalCount)
+        removals.forEach(::delete)
```
lemnik: Would be potentially be worth having a `removalCount == 1` fast path as well here? It seems like it could be a common case, and `minOf` would avoid the full overhead of a sort.

It might also be worth considering making `input` mutable? It's a private copy and an in-place sort is going to be just that little bit more efficient.

bidetofevil (approved):
LGTM. If you saw a comment on this, it's because I had a brain fart lol

PR #3854, opened 2026-08-20
Write session part directory in session transitions

fractalwrench:
## Goal

Writes the session part directory when session transitions occur.

## Testing

Added unit tests.


bidetofevil (approved):
LGTM

PR #3855, opened 2026-08-20
Create session part writer

fractalwrench:
## Goal

Extracts code added in #3854 to `SessionPartWriter`, which will be used to encapsulate all operations that write telemetry for the current session part.

## Testing

Added unit tests


bidetofevil (approved):
LGTM

PR #3856, opened 2026-08-20
Write metadata file in session part directory

fractalwrench:
## Goal

Writes the metadata file in the current session part directory. This is written when the session part is created and when the user information changes.

The threading model is important as it currently enforces correctness - all session persistence happens on a dedicated background thread. If the session part ends before a enqueued write has happened, the enqueued write will still persist information to the correct directory.

## Testing

Added unit tests.

### DIFF
--- a/embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/session/orchestrator/SessionPartWriter.kt
+++ b/embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/session/orchestrator/SessionPartWriter.kt
@@ -9,4 +9,9 @@ interface SessionPartWriter {
      * A new session part has started. Creates the directory that holds its telemetry.
      */
     fun onSessionPartStarted(timestamp: Long, userSessionId: String, sessionPartId: String)
+
+    /**
+     * User information has changed and should be persisted.
+     */
+    fun onUserInfoChanged()
 }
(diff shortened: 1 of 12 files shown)


embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/session/orchestrator/SessionPartWriterImpl.kt
```diff
@@ -2,50 +2,78 @@ package io.embrace.android.embracesdk.internal.session.orchestrator
...
     /**
-     * The directory that telemetry is currently written to. Only ever touched on [worker], so that
-     * a write always targets the session part it was queued for rather than a later one.
+     * The writers for the session part that telemetry is currently written to. Each targets one
+     * fixed directory, so a write that was queued for a session part cannot land in a later one.
      */
-    @Volatile
-    private var activeDirectory: SessionPartDirectory? = null
-
+    private val current = AtomicReference<PartWriters?>(null)
```
lemnik: This looks like a `@Volatile` again? Do we maybe need the atomic side of it later?

lemnik (approved):

embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/session/orchestrator/SessionPartWriter.kt
```diff
@@ -9,4 +9,9 @@ interface SessionPartWriter {
      * A new session part has started. Creates the directory that holds its telemetry.
      */
     fun onSessionPartStarted(timestamp: Long, userSessionId: String, sessionPartId: String)
+
+    /**
+     * User information has changed and should be persisted.
+     */
+    fun onUserInfoChanged()
```
bidetofevil, line 16: I'm wondering if it makes sense to make this more generic, i.e. something in the metadata object has changed and needs updating? This feels too specific and tied to what we store there right now - but I can see this expanding in the future.

bidetofevil (approved):
LGTM. One non-blocking comment about the method name on the interface

embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/session/orchestrator/SessionPartWriter.kt
fractalwrench: Addressed here: https://github.com/embrace-io/embrace-android-sdk/pull/3887

PR #3859, opened 2026-08-21
Write resource session part file

fractalwrench:
## Goal

Writes the resource session part file when a new session part starts.

## Testing

Added unit tests.


bidetofevil (approved):
LGTM

PR #3862, opened 2026-08-21
Add a callback for completed spans

fractalwrench:
## Goal

Adds a callback for completed spans that will be used to append them to a file in the session part directory.

## Testing

Added unit tests.


bidetofevil (approved):
LGTM

PR #3863, opened 2026-08-24
Memoize the `SessionIdsSnapshot` returned from `SessionIdsProviderImpl.getActiveSessionIds`

lemnik:
## Goal
Avoid allocating a new `SessionIdsSnapshot` every time `SessionIdsProviderImpl.getActiveSessionIds` is called. These values don't change very frequently and `SessionIdsSnapshot` is immutable, so shared instances reduce our memory consumption.

## Testing
New unit tests added


fractalwrench (approved):
LGTM

PR #3878, opened 2026-08-25
Write session part span

fractalwrench:
## Goal

Writes the session part span when the session part is initially created. I have not altered the implementation to write on the session part end or at regular intervals during the session - that will come in future changesets.

## Testing

Added unit tests.


embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/session/orchestrator/SessionPartWriterImpl.kt
```diff
@@ -84,6 +88,18 @@ class SessionPartWriterImpl(
         }
     }
 
+    // TODO: the session span is not yet written on a regular interval while the part is active.
+
+    /**
+     * Writes the session span as it stands right now.
+     */
+    private fun queueSessionSpanWrite(writers: PartWriters) {
```
bidetofevil, line 96: Do we want to do any debouncing here?

bidetofevil (approved):
LGTM. Non blocking comment about debouncing session write queuing - not sure if it'll be taken care of later. Ignore if it is.

embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/session/orchestrator/SessionPartWriterImpl.kt
fractalwrench: Yes. This will be addressed in future changesets

PR #3879, opened 2026-08-25
Write the session span on its end

fractalwrench:
## Goal

When the session part ends, write the session part to disk.

## Testing

Added unit tests.


embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/session/orchestrator/SessionPartWriterImpl.kt
```diff
@@ -63,6 +64,17 @@ class SessionPartWriterImpl(
         queueSessionSpanWrite(writers)
     }
 
+    override fun onSessionPartEnded(sessionPartId: String) {
```
bidetofevil, line 67: Do we want to prevent future writes to this session span if there's a race condition for this to be updated right after its end had been triggered?

bidetofevil (approved):
LGTM. Non-blocking comment about preventing races by blocking future write requests after the session part ends

embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/session/orchestrator/SessionPartWriterImpl.kt
fractalwrench: Yes. I'll address this in a separate changeset but we should clear out the `writers` reference at this point.

PR #3880, opened 2026-08-25
Periodically write session part span

fractalwrench:
## Goal

Hooks up the session part span so that it will be written to disk at periodic intervals. Currently this uses the same periodic caching mechanism as the monolithic payload caching.

## Testing

Added unit tests.

### DIFF
--- a/embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/session/orchestrator/SessionOrchestratorImpl.kt
+++ b/embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/session/orchestrator/SessionOrchestratorImpl.kt
@@ -442,6 +442,7 @@ internal class SessionOrchestratorImpl(
                                             )
                                         }
                                         updatePeriodicCacheAttrs()
+                                        sessionPartWriter?.onPeriodicWrite()
                                         payloadFactory.snapshotPayload(state, timestamp, zygote)
                                     }
                                 }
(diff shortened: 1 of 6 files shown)


embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/session/orchestrator/SessionOrchestratorImpl.kt
```diff
@@ -442,6 +442,7 @@ internal class SessionOrchestratorImpl(
                                             )
                                         }
                                         updatePeriodicCacheAttrs()
+                                        sessionPartWriter?.onPeriodicWrite()
```
bidetofevil, line 445: This will eventually replace line 446 to snapshot the current session part? During this period that we do both, is there a need to make sure these are done atomically so there's no race condition that could make the payloads differ?

bidetofevil (approved):
LGTM. Non-blocking question about the new and old ways of persistence and whether they need to be synchronized if we want to ensure the payloads from the old and new methods always match?

embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/session/orchestrator/SessionOrchestratorImpl.kt
fractalwrench: IMO it's not necessary to ensure the old/new methods match 1:1, other than for the purposes of test assertions (where we can control for that anyway)

PR #3881, opened 2026-08-25
Coalesce session part file writes

fractalwrench:
## Goal

Coalesce writes to the session part directory so that there is never more than one enqueued for each session part file, thus preventing unnecessary backpressure.

## Testing

Added unit tests.

### DIFF
--- a/embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/session/orchestrator/SessionPartWriterImpl.kt
+++ b/embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/session/orchestrator/SessionPartWriterImpl.kt
@@ -17,10 +17,13 @@ import io.embrace.android.embracesdk.internal.spans.CurrentSessionPartSpan
 import io.embrace.android.embracesdk.internal.utils.UuidSource
 import io.embrace.android.embracesdk.internal.worker.BackgroundWorker
 import java.io.File
+import java.util.concurrent.Future
+import java.util.concurrent.atomic.AtomicReference
 
 /**
  * Writes session part telemetry to disk, if the multi-file persistence layer is enabled.
- * All filesystem work is queued on a single-threaded [worker].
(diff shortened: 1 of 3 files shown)


embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/session/orchestrator/SessionPartWriterImpl.kt
```diff
@@ -137,5 +139,21 @@ class SessionPartWriterImpl(
...
+    /**
+     * Queues the writes for one file in a session part. Each write persists the whole file so
+     * we should cancel and remove any enqueued writes as they will be superseded by later ones.
+     * A write that has already started is left to finish.
+     */
+    private class CoalescingWriteQueue(private val worker: BackgroundWorker) {
+        private val pending = AtomicReference<Future<*>?>(null)
+
+        fun submit(runnable: Runnable) {
+            pending.getAndSet(worker.submit(runnable))?.cancel(false)
```
bidetofevil, line 156: Do we need to remove already enqueued writes or should be not queue new ones if there is already one enqueued? I'm assuming any pending writes will just take what the current state is, so replacing the job or just dropping the new one should do the same thing. Or is there a race condition hidden in the implementation nuances?

bidetofevil (approved):
LGTM. Non-blocking comment about how we do the coalescing and whether we can just drop new ones if there is already a pending one

embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/session/orchestrator/SessionPartWriterImpl.kt
fractalwrench: IMO we need to remove enqueued writes for a few reasons:

1. State is generally taken at capture time, not write time, because the SDK could have moved onto another session by the time we write
2. We want to prevent backpressure and holding unnecessary object references in runnables

PR #3882, opened 2026-08-25
Write session files synchronously on crash

fractalwrench:
## Goal

If a JVM crash happens the process is about to terminate, so any session files should be written synchronously on disk.

## Testing

Added unit and integration tests.

### DIFF
--- a/embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/session/orchestrator/SessionPartWriterImpl.kt
+++ b/embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/session/orchestrator/SessionPartWriterImpl.kt
@@ -6,6 +6,7 @@ import io.embrace.android.embracesdk.internal.envelope.metadata.EnvelopeMetadata
 import io.embrace.android.embracesdk.internal.envelope.resource.EnvelopeResourceSource
 import io.embrace.android.embracesdk.internal.envelope.session.SESSION_ENVELOPE_TYPE
 import io.embrace.android.embracesdk.internal.envelope.session.SESSION_ENVELOPE_VERSION
+import io.embrace.android.embracesdk.internal.logging.InternalErrorType
 import io.embrace.android.embracesdk.internal.logging.InternalLogger
 import io.embrace.android.embracesdk.internal.otel.spans.EmbraceSdkSpan
 import io.embrace.android.embracesdk.internal.session.persistence.SessionManifestWriter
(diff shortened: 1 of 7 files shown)


embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/session/orchestrator/SessionPartWriterImpl.kt
```diff
@@ -115,12 +130,35 @@ class SessionPartWriterImpl(
...
+                action()
+            } catch (exc: Throwable) {
+                logger.trackInternalError(errorType, exc)
+            }
+        } else {
+            submit(Runnable(action))
+        }
+    }
+
+    private fun enabled(): Boolean {
```
bidetofevil, line 159: Since you're checking whether we're in a crash, it's not clear what this is enabling now. It feels like the implementation is overload with the enabling of the multi-file writer and whether or not we are handling a crash at the moment?

fractalwrench: Attempted to address this on #3907

embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/session/orchestrator/SessionPartWriterImpl.kt
```diff
@@ -115,12 +130,35 @@ class SessionPartWriterImpl(
...
         }
     }
 
-    private fun enabled(): Boolean = configService.persistenceBehavior.isMultiFilePersistenceEnabled()
+    /**
+     * Queues [action] with [submit], or runs it on the calling thread if the process is crashing
+     * and the [worker] has already been sealed by [onCrash]. Telemetry gathered inside [action] can
+     * throw, so a failure is tracked here rather than escaping onto a crashing thread.
+     */
+    private fun execute(
```
lemnik, line 143: nit: given the structure of this function - it looks like a good candidate for `inline` to avoid some of the lambda allocations

embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/session/orchestrator/SessionPartWriterImpl.kt
bidetofevil: Cool. Looks good.

bidetofevil (approved):
LGTM after my comment was addressed

PR #3883, opened 2026-08-25
Invoke callback when envelope resource mutates

fractalwrench:
## Goal

Invoke a callback when the `EnvelopeResource` mutates. This will be helpful in determining when the session part files should be rewritten to disk.

## Testing

Added unit tests.


bidetofevil (changes requested):
Envelope `resource` should be immutable after SDK start, but it's not obvious through this interface that that is true. Can we have some protection around this or make it otherwise obvious that this isn't the same as other change listeners that will always be applied?

fractalwrench:
`EnvelopeResource` is _not_ currently immutable after SDK start (e.g. RN/Unity specific fields are set after SDK init, amongst others). Therefore we do require a listener that rewrites when the information updates to avoid data loss.

Higher up the stack, #3902 alters the proto files so that all the immutable data is stored in one file, and the mutable data stays in another.

bidetofevil (approved):
OK. Lets improve this after everything is said and done.

PR #3884, opened 2026-08-25
Rewrite manifest on state change

fractalwrench:
## Goal

Overwrites the session part manifest file whenever one of the properties in `EnvelopeResource` change.

## Testing

Added unit tests.


bidetofevil (commented):
I'll review this after the question about the implication of immutable resources has on the interfae

bidetofevil (approved):
LGTM. This codifies the departure between the Embrace envelope resource and the OTel resource - when I "rebase" my changes on this, it means the OTel resource just gets a snapshot of the Embrace envelope resource at SDK start time.

I hope we can make this truly immutable in the future, and have all the mutable stuff (which is only the JS bundle ID?) go into metadata.

PR #3888, opened 2026-08-26
Alter span change listener

fractalwrench:
## Goal

Alter the listener that is invoked when a span changes to include a reference to the changed span (required for filtering out the session span), and additionally allow registering multiple listeners.

## Testing

Added unit tests.


bidetofevil (approved):
LGTM

PR #3889, opened 2026-08-26
Invoke span change listener outside of monitor

fractalwrench:
## Goal

Invokes the listener for when a span changes outside of a monitor lock.


bidetofevil (approved):
LGTM

PR #3890, opened 2026-08-26
Handle span mutation callbacks

fractalwrench:
## Goal

Improve where span mutation callbacks are invoked to avoid unnecessary calls before the span is started, and tighten when attributes/status can be set on a span. 

## Testing

Added unit tests.

### DIFF
--- a/embrace-android-otel/src/main/kotlin/io/embrace/android/embracesdk/internal/otel/spans/EmbraceSpanFactoryImpl.kt
+++ b/embrace-android-otel/src/main/kotlin/io/embrace/android/embracesdk/internal/otel/spans/EmbraceSpanFactoryImpl.kt
@@ -121,15 +121,17 @@ private class EmbraceSpanImpl(
     override var status: StatusData
         get() = currentStatus
         set(value) {
-            if (setSpanStatus(value)) {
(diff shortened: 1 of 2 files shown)


embrace-android-otel/src/main/kotlin/io/embrace/android/embracesdk/internal/otel/spans/EmbraceSpanFactoryImpl.kt
```diff
@@ -121,15 +121,17 @@ private class EmbraceSpanImpl(
     override var status: StatusData
         get() = currentStatus
         set(value) {
-            if (setSpanStatus(value)) {
+            if (setSpanStatus(value) && spanStarted()) {
```
bidetofevil, line 124: Shouldn't we call `spanStarted()` first since a span that hasn't started can't have a status?

fractalwrench: That's handled within `setSpanStatus` at L131

bidetofevil (approved):
LGTM

PR #3891, opened 2026-08-26
Add tests for span change listener

fractalwrench:
## Goal

Adds tests that assert the behavior of the span change listener.


bidetofevil (approved):
LGTM

PR #3892, opened 2026-08-26
Add LeakDetector

lemnik:
## Goal
Implement a GC aware `LeakDetector` that can keeps track of references to objects with known constrained lifecycles (such as `Activity`, `Fragment`, `View`, etc.).

Each object must be registered (`trackOpen`) as soon as possible after its creation, and marked as closed (`trackClosed`) once its standard lifecycle ends (`Activity.onDestroy`, etc.). The `LeakDetector` then keeps track of how many GC cycles the reference remains valid for and stops reporting the reference if it is garbage collected.

## Testing
Added new unit tests for the `LeakDetector` itself.

### DIFF
--- a/embrace-android-instrumentation-leaks/src/main/kotlin/io/embrace/android/embracesdk/instrumentation/leaks/LeakDetector.kt
+++ b/embrace-android-instrumentation-leaks/src/main/kotlin/io/embrace/android/embracesdk/instrumentation/leaks/LeakDetector.kt
@@ -1,0 +1,345 @@
+package io.embrace.android.embracesdk.instrumentation.leaks
+
+import android.os.Build
+import android.os.Debug
+import io.embrace.android.embracesdk.internal.clock.Clock
+import java.lang.ref.PhantomReference
+import java.lang.ref.Reference
+import java.lang.ref.ReferenceQueue
+import java.lang.ref.WeakReference
+import java.util.WeakHashMap
+import java.util.concurrent.atomic.AtomicBoolean
+import kotlin.concurrent.thread
+
+/**
+ * Detects objects that outlive the end of their own lifecycle, and tracks how many GC cycles they go on to survive.
+ *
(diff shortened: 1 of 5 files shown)


embrace-android-instrumentation-leaks/src/main/kotlin/io/embrace/android/embracesdk/instrumentation/leaks/LeakDetector.kt
```diff
@@ -0,0 +1,345 @@
...
+            return null
+        }
+
+        ref.clear()
+
+        val referent = ref.target.get() ?: return null
+
+        if (ref.suspected) {
+            val suspect = ConfirmedSuspect(
+                referent,
```
fractalwrench, line 179: I may have read this wrong as there's a lot of moving parts, but it looks like this code gets a strong reference to an `Activity` and holds it on a `HashSet`, and could therefore leak it

fractalwrench: Is there any sort of testing we could do that would detect these sorts of regressions in the `LeakDetector`?

lemnik: `ConfirmedSuspect` is itself a `WeakReference`, so `referent` here is only weakly referenced (though the `suspect` itself is held strongly in the `HashSet`. If `referent` is GC'd we remote `suspect` from the `HashSet` once we get it from the `ReferenceQueue` and [dispatch](https://github.com/embrace-io/embrace-android-sdk/pull/3892/changes#diff-827ac561c07388608e454e1404f538299c23f39839854832cc857a7d8f8419a2R282) it. 🤔

Adding some regression tests is an excellent idea though 👍

fractalwrench (approved):
LGTM

PR #3895, opened 2026-08-26
Write span snapshots

fractalwrench:
## Goal

Initial changeset for writing span snapshots. This hooks up all the necessary infrastructure & writes span snapshots at the start and end of a session part only. Changes within the session will be left for a future date.

## Testing

Added unit tests.

### DIFF
--- a/embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/session/orchestrator/SessionPartWriterImpl.kt
+++ b/embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/session/orchestrator/SessionPartWriterImpl.kt
@@ -9,11 +9,13 @@ import io.embrace.android.embracesdk.internal.envelope.session.SESSION_ENVELOPE_
 import io.embrace.android.embracesdk.internal.logging.InternalErrorType
 import io.embrace.android.embracesdk.internal.logging.InternalLogger
 import io.embrace.android.embracesdk.internal.otel.spans.EmbraceSdkSpan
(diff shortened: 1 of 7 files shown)


embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/session/orchestrator/SessionPartWriterImpl.kt
```diff
@@ -164,6 +169,24 @@ class SessionPartWriterImpl(
         }
     }
 
+    /**
+     * Writes in-flight spans to a snapshot file.
+     */
+    private fun queueSpanSnapshotsWrite(writers: PartWriters) {
+        // TODO: future: don't call this every time the listener is invoked as it's expensive to
+        // obtain _all_ the spans for every change. Currently this write only happens at session
```
bidetofevil, line 177: Won't this leave stale span snapshots if there's a native crash or unexpected process termination in the middle of a session part?

fractalwrench: Yes. Given this isn't enabled yet, I'd like to deliver that in a future changeset.

bidetofevil: Sounds good

bidetofevil (approved):
LGTM. Comment will be addressed in the future.

PR #3898, opened 2026-08-27
Remote config gate activity leak detection

lemnik:
## Goal

Gate the activity leak detection functionality behind a new `RemoteConfig` threshold: `"pct_activity_leak_detection_enabled"`.

This is a standard 0-100% threshold, with a default behaviour of "off".

## Testing
New unit tests


fractalwrench (approved):
LGTM

PR #3899, opened 2026-08-27
Treat missing files as empty spans

fractalwrench:
## Goal

Treats missing files as empty spans instead of discarding the entire payload. This also avoids an unnecessary write when starting a session part and no spans are present.

## Testing

Added unit tests.


bidetofevil (approved):
LGTM

PR #3901, opened 2026-08-27
Optimize threshold check

fractalwrench:
## Goal

Optimizes the BehaviorThreshold check by putting the calculation behind `lazy` rather than re-calculating every time we access something in the `ConfigService`.


bidetofevil (approved):
LGTM

PR #3902, opened 2026-08-27
Split mutable properties from resource

fractalwrench:
## Goal

Splits mutable properties in `EnvelopeResource` from the manifest file so that it's only written once. The mutable values are now written to the metadata file.

## Testing


bidetofevil (commented):
Will review once the issue of the "mutable resource" is resolved

bidetofevil (approved):
LGTM.

PR #3903, opened 2026-08-27
Optimize writing of span snapshots

fractalwrench:
## Goal

Optimizes the writing of span snapshots by serializing on the worker thread, not the calling thread.


bidetofevil (approved):
LGTM

PR #3905, opened 2026-08-27
Identify specific leak references

lemnik:
## Goal
Allow more reasonable re-duplication & aggregation work by identifying each suspected leak reference. This adds the `identityHashCode` to each of the leak suspects reported (as a hex string) so that 2 different suspected leaks of the same class can be distinguished. 

## Testing
New unit test added.


fractalwrench (approved):
LGTM

PR #3906, opened 2026-08-28
Regression tests for the LeakDetector

lemnik:
## Goal
Add a set of regression tests to `LeakDetector` (using `System.gc()`) to ensure that the `LeakDetector` itself never accidentally becomes a source of memory leaks.


fractalwrench (approved):
LGTM

PR #3907, opened 2026-08-28
Alter write seal logic

fractalwrench:
## Goal

Alter write seal logic, addressing feedback raised on #3882.


bidetofevil (approved):
LGTM

PR #3909, opened 2026-08-28
Add initial test for multi-file persistence

fractalwrench:
## Goal

Adds an initial integration test that asserts multi-file persistence delivers sessions. Future changesets will enhance the actual assertions.


bidetofevil (approved):
Nice. LGTM

PR #3910, opened 2026-08-28
Fragment & Fragment View leak detection

lemnik:
## Goal
Detect leaked `Fragment`s and fragment `View`s when fragments are used.

This PR adds `Fragment` support directly to the `embrace-android-instrumentation-leaks` module using a `compileOnly` dependency on `androix.fragment` and a `runCatching` to handle cases where apps don't use `Fragment`s (a no-op callback implementation is used in this case).

`Fragment.getView()` can return `null` when we get the `onFragmentViewDestroyed` callback so we track the `Fragment`/`View` association with a double sided `WeakHashMap` (encapsulated as `FragmentViewMap`) since `View`s may strong-reference their `Fragment`s via tags.

## Testing
Manual testing and new unit tests


fractalwrench (approved):
LGTM

PR #3920, opened 2026-09-01
Alter JDK used in CI

fractalwrench:
## Goal

#3911 failed as the setup-java action has removed support for the 'adopt' distribution.


bidetofevil (approved):
LGTM. Hopefully nothing funky happens as a result!

PR #3921, opened 2026-09-01
Implement CoalescingWriteQueue

fractalwrench:
## Goal

Implements `CoalescingWriteQueue`. This attempts to add debouncing to write operations so that a burst of changes does not result in a burst of writes to disk, by delaying the task by a configurable number of milliseconds.

If a write is already in-progress, it is left to execute and no attempt to cancel is made. If a write has been superseded and is not yet executing, it is removed from the queue and cancelled.

Finally, this also adds a mechanism to run the task synchronously on the calling thread. This will be beneficial for cases such as JVM crashes where the process is terminating.

This is not fully hooked up yet and will be updated to do so in a later PR.

## Testing

Added tests.

### DIFF
--- a/embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/session/orchestrator/CoalescingWriteQueue.kt
+++ b/embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/session/orchestrator/CoalescingWriteQueue.kt
@@ -1,0 +1,84 @@
+package io.embrace.android.embracesdk.internal.session.orchestrator
+
+import io.embrace.android.embracesdk.internal.worker.BackgroundWorker
+import java.util.concurrent.Future
+import java.util.concurrent.FutureTask
+import java.util.concurrent.TimeUnit
+import java.util.concurrent.TimeoutException
+import java.util.concurrent.atomic.AtomicLong
+import java.util.concurrent.atomic.AtomicReference
+
+/**
+ * Debounces the writes for one file in a session part. Each write persists the whole file, so a
+ * write that is superseded before it runs can be dropped. The intent behind this class
+ * is that a burst of changes typically only cost one write, rather than one per change.
+ *
+ * A write is held as a [FutureTask] armed with a scheduled trigger that runs it
+ * [delayMs] later. Cancelling that trigger does not cancel the [FutureTask] it wraps. This
+ * lets [shutdown] force the task to run early rather than queueing the work twice.
+ */
+internal class CoalescingWriteQueue(
+    private val worker: BackgroundWorker,
+    private val delayMs: Long = DEFAULT_DELAY_MS,
+) {
+    private val pending = AtomicReference<PendingWrite?>(null)
+
+    /**
+     * Orders writes so that concurrent calls to [submit] can be compared by age. This handles interleaving
+     * during submission.
+     */
+    private val sequence = AtomicLong()
+
+    /**
+     * Arms [runnable] to run [delayMs] from now, disarming whatever was armed before it.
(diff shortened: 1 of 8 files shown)


embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/session/orchestrator/CoalescingWriteQueue.kt
```diff
@@ -0,0 +1,84 @@
...
+     */
+    fun submit(runnable: Runnable) {
+        val seq = sequence.incrementAndGet()
+        val task = FutureTask(runnable, Unit)
+        val trigger = runCatching {
+            worker.schedule<Unit>(task, delayMs, TimeUnit.MILLISECONDS)
+        }.getOrNull()
+        val write = PendingWrite(seq, task, trigger)
+
+        while (true) {
```
bidetofevil, line 44: So the algo here is basically:

- If there's a `pending` `write` that is newer than this task, discard the future task that will run, i.e. this old and superseded so go away
- Otherwise, if `pending` has not changed since it was obtained earlier, set pending to this `write` request
- And if `pending` has changed, go through this process again. Every task should either be explicitly discarded because a newer one is pending, or set as the next pending to run, with the old one canceled.

I don't think I can spot a deadlock. And the first condition ensures that older jobs are cleaned up if a new one gets in, while the second condition ensures older queue'd writes are superseded by new ones. 

So it seems like as long as the `runnable` for all tasks are equivalent or at least does a superset of what older ones do, then I think we're good.

Should we state that last part explicitly? This isn't so much coalescing than debouncing - perhaps a name that indicates that is more appropriate to telegraph that, i.e. older runnables are explicitly discarded when new ones are preferred. So if you submit heterogenous runnables, this will not do what you want it to do.

And given the runnables should be doing the same thing, would it make sense to give this queue a singular runnable (or a runnable factory) to explicitly enforce that? Then this just requires a trigger and everything is managed internally.

bidetofevil (approved):
Code looks good.

One non-blocking comment about the nature of this thing, particularly about how the runnable is set and whether or not we should prevent different runnables from being queued that could supersede but not replace what older runnables are doing. 

This is effectively a debouncer and doesn't really combine runnables, so it would be nice if that was more clear from the name and the API (e.g. don't let the runnables do different things if you're going to discard older ones just because newer ones exist)

embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/session/orchestrator/CoalescingWriteQueue.kt
```diff
@@ -0,0 +1,84 @@
...
+/**
+ * Debounces the writes for one file in a session part. Each write persists the whole file, so a
+ * write that is superseded before it runs can be dropped. The intent behind this class
+ * is that a burst of changes typically only cost one write, rather than one per change.
+ *
+ * A write is held as a [FutureTask] armed with a scheduled trigger that runs it
+ * [delayMs] later. Cancelling that trigger does not cancel the [FutureTask] it wraps. This
+ * lets [shutdown] force the task to run early rather than queueing the work twice.
+ */
+internal class CoalescingWriteQueue(
```
bidetofevil, line 20: The functionality to run this synchronously that the PR mentioned: is that work that is still pending, or am I not seeing how it can be done?

fractalwrench: `shutdown()` should achieve this

PR #3922, opened 2026-09-01
Enhance multi file persistence test cases

fractalwrench:
## Goal

Enhances the `MultiFilePersistenceParityTest` test cases to provide some reasonable coverage that persistence matches the results of the legacy persistence approach.

Several test cases are failing as the implementation is not yet complete. These will be addressed in future changesets.


bidetofevil (approved):
LGTM

PR #3923, opened 2026-09-01
Add module for telemetry limits

fractalwrench:
## Goal

Adds a module that will be used to share telemetry limits between where we write data and where we read it, so that the enforced limits are the same. Currently this module is empty.


bidetofevil (approved):
LGTM if there isn't another lower-level module we can define these in

PR #3928, opened 2026-09-01
Combine buffered experiment API calls when SDK starts

bidetofevil:
Combine buffered experimental API calls into as few calls into the service as possible without add edit/commit transaction semantics to the internal service. Each call will generate the `emb.experiments` blob from scratch, which which should minimize.

### DIFF
--- a/embrace-android-sdk/src/main/kotlin/io/embrace/android/embracesdk/internal/api/delegate/ExperimentApiDelegate.kt
+++ b/embrace-android-sdk/src/main/kotlin/io/embrace/android/embracesdk/internal/api/delegate/ExperimentApiDelegate.kt
@@ -49,15 +49,60 @@ internal class ExperimentApiDelegate(
         untrack("untrack_feature_flag", ExperimentKind.FEATURE_FLAG, ids, endedAt ?: now())
     }
 
+    /**
+     * Flush buffered experiment API calls by combine tracking and untracking into as few
+     * calls down to the service as possible. The ordering of tracking calls will be preserved,
+     * first writer still wins, but all untracking calls will be replayed after the tracking
+     * calls in the order they were called. Later untracking calls using the same end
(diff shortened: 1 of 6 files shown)


embrace-android-sdk/src/main/kotlin/io/embrace/android/embracesdk/internal/api/delegate/ExperimentApiDelegate.kt
```diff
@@ -49,16 +49,61 @@ internal class ExperimentApiDelegate(
         untrack("untrack_feature_flag", ExperimentKind.FEATURE_FLAG, ids, endedAt ?: now())
     }
 
+    /**
+     * Flush buffered experiment API calls by combine tracking and untracking into as few
```
fnewberg:
```suggestion
     * Flush buffered experiment API calls by combining tracking and untracking into as few
```

embrace-android-sdk/src/main/kotlin/io/embrace/android/embracesdk/internal/api/delegate/ExperimentApiDelegate.kt
```diff
@@ -49,16 +49,61 @@ internal class ExperimentApiDelegate(
...
+                            } else {
+                                flagsToUntrack[event.endTimeMs] = existing + event.ids
+                            }
+                        }
+                    }
+                }
             }
         }
+        experimentTrackingService?.track(toTrack)
+        expToUntrack.entries.forEach {
```
fnewberg, line 103: add a comment explaining that we don't want to incur the additional complexity of combining untracks and that they are unlikely to occur at volume before startup

fractalwrench (approved):
LGTM pending review comments + CI

bidetofevil:
The test compile break is wild. My usual test command didn't rerun what was modified - I might need to not rely on `testRelease` to run all the tests that are necessary.

PR #3933, opened 2026-09-02
Add leak detection instrumentation as default

lemnik:
## Goal
Add `embrace-android-instrumentation-leaks` as a default dependency in the `embrace-android-sdk` module, so that memory leak detection is a default (remotely configured) instrumentation.


bidetofevil (approved):
LGTM. This means it's on by default as well?

lemnik:
> LGTM. This means it's on by default as well?

Nope, it's [off by default](https://github.com/embrace-io/embrace-android-sdk/pull/3898/changes#diff-2b78ba825bef9e660864dc0539772db235377ed90f62227861c272f55a91b8f3R23) with remote config gate, but should be included by default so that we can turn it on

PR #3935, opened 2026-09-02
Move bulk commit logic into service

bidetofevil:
Move logic to combine various experiment API calls into the internal service and make the delegate only keep track of the API calls which can be replayed in order when the SDK initializes.

---


fractalwrench (approved):
LGTM

PR #3937, opened 2026-09-03
Resolve test failures in persistence implementation

fractalwrench:
## Goal

Resolves a couple of test failures in the persistence implementation by specifying the `processIdentifier` attribute correctly and altering how empty span collections are handled.


bidetofevil (approved):
LGTM

PR #3938, opened 2026-09-03
Resolve test failures in persistence implementation

fractalwrench:
## Goal

Resolves a couple of test failures in the persistence implementation by retaining data in the span object until all writes have completed. This is only enabled when this feature is, and in practice the objects should not be held for long.


bidetofevil (approved):
LGTM. The setter for this can be removed once we're fully productionized on this, i.e. we will always retain after stop?

PR #3941, opened 2026-09-03
Loosen persistence test assertion

fractalwrench:
## Goal

Loosens an assertion in the persistence test that assumed attributes in the session span should be persisted in the same order for both persistence mechanisms.


bidetofevil (approved):
LGTM. Sorting before diffing is probably a good idea in most places anyway

PR #3942, opened 2026-09-03
Add config to enable URL fragment redaction in the WebViews

lemnik:
## Goal
Introduce config (for `embrace-config.json`) to control the capturing / redaction of URL fragments captured by the `WebView` instrumentation.

## Testing
Added unit test to check the expected default value, and naming (which will become a contract with the Gradle plugin in a later PR).


fractalwrench (approved):
LGTM. Do we have plans for a remote config too?

PR #3944, opened 2026-09-03
Gradle plugin enum config support

lemnik:
## Goal
Implement support for config functions that return `enum` constants to the Gradle plugin, alongside the existing support for `string`, `bool`, `int`, `map`, etc.

## Testing
Unit tests to cover the new bytecode modification.


fractalwrench (approved):
LGTM

PR #3945, opened 2026-09-03
Carry over spans in new persistence mechanism

fractalwrench:
## Goal

If a span finishes and a session part isn't available yet on startup, buffer the call as per the legacy mechanism. This allows traces such as emb-sdk-init to be included in the payload.


bidetofevil (approved):
LGTM

PR #3948, opened 2026-09-04
Bound the span file size when reading/writing

fractalwrench:
## Goal

Bound the file size of span/span snapshots when reading/writing to a reasonable backstop value.

### DIFF
--- a/embrace-android-session-persistence/src/main/kotlin/io/embrace/android/embracesdk/internal/session/persistence/CompletedSpansReader.kt
+++ b/embrace-android-session-persistence/src/main/kotlin/io/embrace/android/embracesdk/internal/session/persistence/CompletedSpansReader.kt
@@ -16,10 +16,11 @@ private const val SPANS_TAG = 1
  * not reported: every record written in full before it is returned. A record that is all
  * present but does not decode is corruption and throws.
  */
-internal fun readCompletedSpans(source: BufferedSource): List<SpanProto> {
+internal fun readCompletedSpans(source: BufferedSource, maxBytes: Long = MAX_PART_FILE_BYTES): List<SpanProto> {
     val spans = mutableListOf<SpanProto>()
     val reader = ProtoReader(source)
     reader.beginMessage()
+    var remaining = maxBytes
 
     while (true) {
         val record = try {
@@ -32,6 +33,10 @@ internal fun readCompletedSpans(source: BufferedSource): List<SpanProto> {
                 }
             }
         } catch (exc: EOFException) {
+            return spans
+        }
+        remaining -= record.size
+        if (remaining < 0) {
             return spans
         }
         spans.add(SpanProto.ADAPTER.decode(record))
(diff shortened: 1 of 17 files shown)


embrace-android-session-persistence/src/main/kotlin/io/embrace/android/embracesdk/internal/session/persistence/CompletedSpansReader.kt
```diff
@@ -16,10 +16,11 @@ private const val SPANS_TAG = 1
  * not reported: every record written in full before it is returned. A record that is all
  * present but does not decode is corruption and throws.
  */
-internal fun readCompletedSpans(source: BufferedSource): List<SpanProto> {
+internal fun readCompletedSpans(source: BufferedSource, maxBytes: Long = MAX_PART_FILE_BYTES): List<SpanProto> {
```
bidetofevil, line 19: Is blob we persist always compressed? 3MB should be OK if that's true, but may not if we are storing anything on disk, even temporarily, that is uncompressed.

Or is this all just protobuf now so we're already pretty efficient, and 3MB is an unreasonably large amount already so we don't have to worry?

embrace-android-session-persistence/src/main/kotlin/io/embrace/android/embracesdk/internal/session/persistence/SessionPartFiles.kt
```diff
@@ -25,14 +25,17 @@ internal const val SPAN_SNAPSHOTS_FILE_NAME = "span_snapshots.pb"
...
  * The temporary file is always cleaned up, so a failed write leaves the directory as it was found.
  * Throws [IOException] if the file could not be written; callers are responsible for reporting that
  * as an internal error of the appropriate type.
  */
-internal fun writeAtomically(partDir: File, fileName: String, encode: (OutputStream) -> Unit) {
+internal fun writeAtomically(partDir: File, fileName: String, maxBytes: Long, encode: (OutputStream) -> Unit) {
     val tmpFile = File.createTempFile(fileName, ".tmp", partDir)
     try {
-        tmpFile.outputStream().buffered().use(encode)
+        LimitedOutputStream(tmpFile.outputStream().buffered(), maxBytes).use(encode)
```
bidetofevil, line 38: We should probably record an internal error if this limit is hit - perhaps capping it at once per process?

bidetofevil (approved):
LGTM. A non-blocking comment and suggestion.

embrace-android-session-persistence/src/main/kotlin/io/embrace/android/embracesdk/internal/session/persistence/CompletedSpansReader.kt
fractalwrench: We don't compress data on the write path with the new mechanism. This is a tradeoff discussed in the internal doc as it increases the write operation time by >2x, and the overall footprint of the files is assumed to be small. JSON payloads that are reconstructed _are_ gzip compressed.

embrace-android-session-persistence/src/main/kotlin/io/embrace/android/embracesdk/internal/session/persistence/SessionPartFiles.kt
fractalwrench: This is implemented in the changeset for this callsite

PR #3949, opened 2026-09-04
Thread through crash shutdown path

fractalwrench:
## Goal

Threads through the necessary behavior for handling the crash shutdown path. The `CoalescingWriteQueue` now simply submits the task to the worker immediately and the caller is responsible for waiting with a timeout.

Additionally I spotted a race where writes could happen to a session part that was ending so I tightened the logic to prevent that.


bidetofevil (approved):
Very nice!

PR #3951, opened 2026-09-04
Allow disabling legacy periodic caching path

fractalwrench:
## Goal

Allows disabling the legacy periodic caching path so that instead of running in parallel, the new mechanism replaces the old mechanism.

## Testing

Updated existing tests.


bidetofevil (approved):
Sweet. I wonder how many more simplifications we can do if we don't have to worry about random tasks trying to snapshot and update, and that all payload modification paths are event-based and queued up in an orderly fashion.

PR #3952, opened 2026-09-04
Read session part on IO thread

fractalwrench:
## Goal

Refactors code to make it more explicit that session parts should be read on the IO thread rather than the session-persistence thread.


bidetofevil (approved):
LGTM. 

I wonder if we can start using annotations to markup which workers/threads we are expecting certain code blocks or functions to run in so that we can proactively ensure this doesn't get messed up accidentally? Obviously not for this PR, but as a future thing to pick up.

PR #3955, opened 2026-09-04
Avoid and forbide use of non-performant means of obtaining serializers

bidetofevil:
Reference Kotlin Serialization serializers directly instead of relying on helper methods that could trigger classloading in apps where those classe aren't preloaded by AOT. Band this via a detekt rule. Also share serializer instance if possible to avoid race conditions that might create contention in sensitive areas.

---


fractalwrench (approved):
LGTM pending CI

PR #3956, opened 2026-09-04
Record APK optimization attributes

bidetofevil:
Record as SDK init span attributes indications that AOT might have optimized the app APK to improve performance. This is useful for tracking during app startup but it is likely useful as a proxy for performance drags for particular app instances over all.

---


embrace-android-infra/src/main/kotlin/io/embrace/android/embracesdk/internal/utils/ByteArrayExtensions.kt
```diff
@@ -0,0 +1,39 @@
...
+}
+
+/**
+ * Whether the bytes starting at [offset] equal [contentToFind]
+ */
+fun ByteArray.regionMatches(offset: Int, contentToFind: ByteArray): Boolean {
+    if (offset < 0 || offset + contentToFind.size > size) {
+        return false
+    }
+    var j = 0
```
lemnik, line 31: Tiramisu has a vectorised `Arrays.equals` implementation

```suggestion
    if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.TIRAMISU) {
        return Arrays.equals(this, offset, contentToFind.size, contentToFind, 0, contentToFind.size)
    }
    
    var j = 0
```

fractalwrench (commented):
I prefer using the Android framework's APIs to calculate the CPU ABI as it looks like this method requires I/O to read the APK file? That doesn't seem ideal to force onto the SDK init path

embrace-android-instrumentation-startup-trace/src/main/kotlin/io/embrace/android/embracesdk/internal/instrumentation/startup/ArtOptimizationState.kt
```diff
@@ -0,0 +1,87 @@
...
+        private fun getIsa(primaryAbi: String): String? = when (primaryAbi) {
+            "arm64-v8a" -> "arm64"
+            "armeabi-v7a" -> "arm"
+            "x86_64" -> "x86_64"
+            "x86" -> "x86"
+            else -> null
+        }
+
+        private const val ART_COMPILER_FILTER_KEY: String = "compiler-filter"
+        private const val NUL_CHAR = "\u0000"
```
lemnik, line 84: Given that this is always matched as a cstring, consider maybe writing it out as a byte-array here would be a better option since it would definitely avoid the concat and byte-encoding steps? A comment would keep it readable.

embrace-android-utils/src/main/kotlin/io/embrace/android/embracesdk/internal/SystemInfo.kt
```diff
@@ -40,6 +40,12 @@ data class SystemInfo(
      * Name of the model of the device.
      */
     val deviceModel: String = getDeviceModel(),
+
+    /**
+     * The primary supported ABI, expected to be the one the current app instance is running under. [Build.UNKNOWN] if
+     * the platform does not report one or the first reported one isn't in [SUPPORTED_ABIS].
+     */
+    val primaryAbi: String = getPrimaryAbi(),
```
lemnik, line 48: Do we know how this will behave on 64bit devices running 32bit OS builds?

embrace-android-infra/src/main/kotlin/io/embrace/android/embracesdk/internal/utils/ByteArrayExtensions.kt
bidetofevil: Nice. I literally would not have thought that this

embrace-android-instrumentation-startup-trace/src/main/kotlin/io/embrace/android/embracesdk/internal/instrumentation/startup/ArtOptimizationState.kt
bidetofevil: Sure. I was definitely prioritizing readability here but a comment will work just as well.

embrace-android-utils/src/main/kotlin/io/embrace/android/embracesdk/internal/SystemInfo.kt
bidetofevil: Same as before tbh. The old impl has a is32bit flag but it was never checked.  I can take a look at this after, just to button this up?

embrace-android-infra/src/main/kotlin/io/embrace/android/embracesdk/internal/utils/ByteArrayExtensions.kt
bidetofevil: Hmm. Android not a dependency at this layer. I'll have to plumb this through the caller. 

How much of a savings are we going to get if we don't need to do this by hand?

bidetofevil: Looks like even the non-vectorized implementations shoudl be faster, so I'll just use that always and rip out the code that checks for android version lol

bidetofevil:
> I prefer using the Android framework's APIs to calculate the CPU ABI as it looks like this method requires I/O to read the APK file? That doesn't seem ideal to force onto the SDK init path

This doesn't change how the ABIs are being read -- it's still going through `BUILD.SUPPORT_ABIS` as it did before. It's just where it's being read is in a different place.

embrace-android-infra/src/main/kotlin/io/embrace/android/embracesdk/internal/utils/ByteArrayExtensions.kt
bidetofevil: See the comments. Unless minSdk is 35, vectorized version get stripped out by R8 and replaced with a different implementation because of a performance issue. See: https://r8.googlesource.com/r8/+/a442f74aa82e5992cd709b5654ecd31ef4071247

The loop is actually faster sometimes by mircrobenchmark testing, but using the Kotlin method to compare effectively calls the 2-parameter version of the `Arrays.equal()` which is pretty much fine given what we are using it for.

embrace-android-infra/src/main/kotlin/io/embrace/android/embracesdk/internal/utils/FileExtensions.kt
```diff
@@ -0,0 +1,15 @@
...
+
+/**
+ * Return the contents of the file as a [ByteArray] if it's smaller than [maxBytes] bytes. Otherwise, return the first [maxBytes].
+ * This will re-throw any exceptions or errors encountered while reading the file.
+ */
+fun File.readHead(maxBytes: Int): ByteArray = RandomAccessFile(this, "r").use { file ->
+    val size = minOf(file.length(), maxBytes.coerceAtLeast(0).toLong()).toInt()
+    val bytes = ByteArray(size)
+    file.readFully(bytes)
+    bytes
```
lemnik: This looks like we could reasonably be using a sequential `InputStream` instead of a `RandomAccessFile`:
```suggestion
fun File.readHead(maxBytes: Int): ByteArray = DataInputStream(inputStream()).use { iStream ->
    val size = minOf(length(), maxBytes.coerceAtLeast(0).toLong()).toInt()
    val bytes = ByteArray(size)
    iStream.readFully(bytes)
    bytes
```

bidetofevil: Sure - let me have a look at this

embrace-android-sdk/src/main/kotlin/io/embrace/android/embracesdk/internal/injection/SdkInitActions.kt
```diff
@@ -282,6 +283,12 @@ internal fun ModuleGraph.markSdkInitComplete(sdkInitDurationsProvider: () -> Map
                         packageInfo = instrumentationModule.instrumentationArgs.packageInfo,
                         nowMs = initModule.clock.now(),
                         prefsFileSizeProvider = { defaultPrefsFile(coreModule.context)?.length() },
+                        artOptimizationProvider = {
```
bidetofevil, line 286: The `attributesProvider()` runs on the background thread when the SDK init span is recorded, so everything done here is on the background and doesn't block the main thread.

embrace-android-instrumentation-startup-trace/src/main/kotlin/io/embrace/android/embracesdk/internal/instrumentation/startup/ArtOptimizationState.kt
```diff
@@ -0,0 +1,97 @@
...
+                    } else {
+                        null
+                    }
+                }.getOrNull()
+                ArtOptimizationState(
+                    artCompilerFilter = filter,
+                    hasAppImage = art.isFile,
+                )
+            } catch (_: Throwable) {
+                null
```
fractalwrench, line 51: We should probably log an internal error?

fractalwrench (approved):
LGTM pending one comment

embrace-android-instrumentation-startup-trace/src/main/kotlin/io/embrace/android/embracesdk/internal/instrumentation/startup/ArtOptimizationState.kt
bidetofevil: I thought about that, but i don't know if this will just get really noisy because a particular OEM or type of app has some weirdness such that this data can't be obtained.

Like with the other attributes where the reading of the data isn't exactly bound to a supported API, I'll do a sweep of production data to see there are systemic reason why certain devices can't capture certain attributes?

Another thing we can do is finally do that type split in the internal errors so we can have different severities - then we can log this as a warning and not worry about it polluting the entire internal error population?

PR #3962, opened 2026-09-07
Alter write debounce delay

fractalwrench:
## Goal

Alters the delays used for debouncing writes so that they are more likely to avoid bursts of writes, without overly increasing the time for data to get persisted. Additionally altered queueing so that the first write is always written. 

## Testing

Added unit tests.


lemnik (approved):
LGTM

PR #3963, opened 2026-09-07
Persist span snapshot on span changes

fractalwrench:
## Goal

Persists span snapshot on span changes rather than just at the session part start/end.

## Testing

Added unit tests


lemnik (approved):
LGTM

PR #3965, opened 2026-09-07
Avoid creating periodic-cache worker

fractalwrench:
## Goal

Avoids creating a periodic-cache worker when multi-file persistence is enabled, which saves the instantiation of one unnecessary thread.

## Testing

Updated unit tests.


lemnik (approved):
LGTM

PR #3967, opened 2026-09-07
Extract span limits

fractalwrench:
## Goal

Extracts the limits for span data capture so that in future it can be enforced when reading spans too.


lemnik (approved):
LGTM

PR #3969, opened 2026-09-07
Only log internal error once

fractalwrench:
## Goal

If a session part directory does not exist the SDK should only log an error once, rather than propagate multiple different types of error message. We likely don't want to attempt recovery from this situation as it indicates I/O problems (e.g. no disk space) or that the telemetry got pruned.

## Testing


lemnik (approved):
LGTM

PR #3970, opened 2026-09-07
Avoid redundant session envelope creation

fractalwrench:
## Goal

Avoids creating a redundant session envelope when multi-file persistence is enabled. This was previously done because the envelope construction has side effects, but this changeset extracts out those side effects to avoid the expensive cost of constructing an entire payload.

## Testing

Added unit tests.


lemnik (approved):
LGTM

PR #3971, opened 2026-09-07
Gracefully handle corrupt span records

fractalwrench:
## Goal

Gracefully handle corrupt span records by attempting to continue if one is invalid, rather than dropping _all_ spans.


lemnik (approved):
LGTM

PR #3972, opened 2026-09-07
Increase debounce time

fractalwrench:
## Goal

Increases the time for debouncing operations so that file writes happen less frequently by default. Note that the first write, session part end, and crashes all flush these writes immediately.


lemnik (approved):
LGTM

PR #3973, opened 2026-09-07
Don't notify on crash

fractalwrench:
## Goal

Don't attempt to read a completed session part when a crash has happened. Leave it for the next process launch as it's very unlikely to complete.


lemnik (approved):
LGTM

PR #3974, opened 2026-09-07
WebView leak detection

lemnik:
## Goal
Add leak detection of `WebView` instances to the `embrace-android-instrumentation-leaks`.

## Testing
Added new unit tests.


fractalwrench (approved):
LGTM

PR #3977, opened 2026-09-08
Enforce limit on large persisted span

fractalwrench:
## Goal

Enforces a reasonable upper limit when reading spans so that we don't have an unbounded allocation if a single span is very large.


bidetofevil (approved):
LGTM.

Do we need to declare this at write time too so we won't get into a situation where a span record is bigger than what we can read back?

fractalwrench:
Yes. That is enforced higher up the stack.

PR #3978, opened 2026-09-08
Persist session span with other spans

fractalwrench:
## Goal

Instead of storing the session part span in a separate file, persist it with the span snapshots/completed spans file.

This requires dropping the `emb.heartbeat_unix_time_nanos` property (or at least altering how it's recorded) and therefore requires some additional thought/discussion before merging this.


bidetofevil (approved):
LGTM

PR #3979, opened 2026-09-08
Enforce overall span count limit when reading data

fractalwrench:
## Goal

Enforces an overall limit on the number of spans when deserializing persisted data.

## Testing

Added unit tests.

### DIFF
--- a/embrace-android-session-persistence/src/main/kotlin/io/embrace/android/embracesdk/internal/session/persistence/SessionReconstructionService.kt
+++ b/embrace-android-session-persistence/src/main/kotlin/io/embrace/android/embracesdk/internal/session/persistence/SessionReconstructionService.kt
@@ -68,21 +68,14 @@ class SessionReconstructionService(
         }
         val metadata = metadataProto.toPayload()
 
-        val sessionSpan = readPartFile(
-            partDir,
-            SESSION_SPAN_FILE_NAME,
-            SessionPartSpan.ADAPTER,
-            SessionPartSpan::format_version,
-        ) ?: return null
+        val span = readSessionSpan(partDir) ?: return null
 
-        val span = sessionSpan.span
-        if (span == null) {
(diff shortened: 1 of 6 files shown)


embrace-android-session-persistence/src/main/kotlin/io/embrace/android/embracesdk/internal/session/persistence/SessionReconstructionService.kt
```diff
@@ -112,14 +105,29 @@ class SessionReconstructionService(
...
+        val sessionSpan = readPartFile(
+            partDir,
+            SESSION_SPAN_FILE_NAME,
+            SessionPartSpan.ADAPTER,
+            SessionPartSpan::format_version,
+        ) ?: return null
+
+        val span = sessionSpan.span
+        if (span == null) {
+            trackFailure(IOException("Session span file has no span"))
```
bidetofevil, line 118:
```suggestion
            trackFailure(IOException("Session part span file has no spans"))
```

embrace-android-session-persistence/src/main/kotlin/io/embrace/android/embracesdk/internal/session/persistence/SessionReconstructionService.kt
```diff
@@ -241,7 +255,21 @@ class SessionReconstructionService(
...
+
+    private class SpanBudget {
+
+        var remaining: Int = MAX_PERSISTED_SPANS
+            private set
+
+        var exceeded: Boolean = false
+            private set
+
+        fun spend(spans: Int, truncated: Boolean) {
```
bidetofevil, line 268: Is a negative `remaining` valid or should be coerce it up to 0 if it ever gets there?

bidetofevil (approved):
LGTM. A couple of non-blocking suggestions

PR #3981, opened 2026-09-08
Enforce overall limit on json files

fractalwrench:
## Goal

Drops any JSON payload file that has exceeded a backstop limit, to prevent memory exhaustion. This limit is set high and is unlikely to be met in practice.

## Testing

Added unit tests.

### DIFF
--- a/embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/resurrection/PayloadResurrectionServiceImpl.kt
+++ b/embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/resurrection/PayloadResurrectionServiceImpl.kt
@@ -31,6 +31,7 @@ import io.embrace.android.embracesdk.internal.session.getUserSessionProperties
 import io.embrace.android.embracesdk.internal.utils.EmbTrace
 import io.embrace.android.embracesdk.internal.utils.Provider
 import io.embrace.android.embracesdk.semconv.EmbSessionAttributes
+import java.io.IOException
 import java.io.InputStream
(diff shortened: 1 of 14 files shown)


embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/resurrection/PayloadResurrectionServiceImpl.kt
```diff
@@ -451,14 +454,22 @@ internal class PayloadResurrectionServiceImpl(
...
+     */
+    private fun StoredTelemetryMetadata.loadDecompressedPayload(): InputStream? {
+        if (cacheStorageService.payloadSizeBytes(this) > maxPayloadBytes) {
+            throw IOException(OVERSIZED_PAYLOAD_MSG)
+        }
+        return cacheStorageService.loadPayloadAsStream(this)?.let {
             try {
                 GZIPInputStream(it)
             } catch (_: ZipException) {
                 null
```
bidetofevil, line 469: No need to log if there's an issue with reading the zipped file?

bidetofevil (approved):
LGTM. One non-blocking comment.

embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/resurrection/PayloadResurrectionServiceImpl.kt
fractalwrench: IIRC we had this and it ended up being pretty noisy due to partially written files when the device ran out of disk space

PR #3982, opened 2026-09-08
Merge manifest and metadata files

fractalwrench:
## Goal

Instead of storing the manifest and metadata in separate files, persist them both together. This saves one file block and potential complexity of another file, for the cost that sometimes the metadata will change and the whole file will need writing again.


bidetofevil (approved):
LGTM. I think this is a good simplification given the overhead of rewriting the immutable resource isn't great, and metadata is also not something that changes a lot at this moment.

PR #3984, opened 2026-09-08
Write scenarios for persistence benchmarking

fractalwrench:
## Goal

Writes scenario code that will be used for benchmarking the different persistence mechanisms.


embrace-benchmark-common/src/main/kotlin/io/embrace/android/embracesdk/benchmark/scenario/ScenarioSpec.kt
```diff
@@ -0,0 +1,40 @@
...
+    val id: String,
+    val description: String,
+    val action: ScenarioScope.() -> Unit,
+)
+
+/**
+ * What a scenario is given to work with: the SDK under test, and control over the passage of time
+ * and the app's foreground state.
+ */
+interface ScenarioScope {
```
bidetofevil, line 18: Nice. Is this something we can generalize?

bidetofevil (approved):
LGTM. 

I like the idea of abstracting the definition of a testing scenario. Perhaps we can extract this concept so we have less boilerplate when writing macrobenchmark tests?

PR #3985, opened 2026-09-08
Avoid reporting unnecessary internal error

fractalwrench:
## Goal

Avoids reporting an unnecessary internal error when a span snapshot is superseded by a completed span.


bidetofevil (approved):
LGTM

PR #3986, opened 2026-09-08
Remove unnecessary lock

fractalwrench:
## Goal

Removes unnecessary locks as the writes now all happen on a single-threaded executor.


bidetofevil (approved):
My favourite way of handling concurrency is getting rid of the concurrency. YAY!

PR #3987, opened 2026-09-08
Drain spans on load

fractalwrench:
## Goal

Drain spans when reading them back from persisted data so that they're only held in memory once, rather than twice per object layer.

### DIFF
--- a/embrace-android-session-persistence/src/main/kotlin/io/embrace/android/embracesdk/internal/session/persistence/DecodedSpans.kt
+++ b/embrace-android-session-persistence/src/main/kotlin/io/embrace/android/embracesdk/internal/session/persistence/DecodedSpans.kt
@@ -1,10 +1,28 @@
 package io.embrace.android.embracesdk.internal.session.persistence
 
+import io.embrace.android.embracesdk.internal.payload.Span
+
 /**
  * The spans read back from a completed spans file, including the first exception (if any) found when decoding
  */
 internal class DecodedSpans(
-    val spans: List<SpanProto>,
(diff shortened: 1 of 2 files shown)


embrace-android-session-persistence/src/main/kotlin/io/embrace/android/embracesdk/internal/session/persistence/DecodedSpans.kt
```diff
@@ -1,10 +1,28 @@
 package io.embrace.android.embracesdk.internal.session.persistence
 
+import io.embrace.android.embracesdk.internal.payload.Span
+
 /**
  * The spans read back from a completed spans file, including the first exception (if any) found when decoding
  */
 internal class DecodedSpans(
-    val spans: List<SpanProto>,
+    private val decoded: MutableList<SpanProto>,
```
bidetofevil, line 9: I had a bunch of comments about `decoded` being modified during the drain, but then I read further and realized that `decoded` is never modified after it's passed into this class 😅 

Do you think we need to formalize/guarantee that invariant so that the iteration logic can make that assumption too? Or at least if it was added to, it doesn't affect the drain, e.g. get the first index to remove and iterate down from it to 0, leaving any new additions (which should never happen) still in the list

bidetofevil (approved):
LGTM - one suggestion which you can feel free to leave if you think is overkill (or add on to if you think it's not sufficient) about locking the iteration at the start of the drain so additions don't screw up the order (though we'll still be vulnerable to removals...)

PR #3988, opened 2026-09-08
Delete session part directory on successful store

fractalwrench:
## Goal

Delete the session part directory when the `IntakeService` has successfully stored the envelope, to avoid potential data loss.

## Testing

Added unit tests.


bidetofevil (approved):
LGTM

PR #3989, opened 2026-09-08
Avoid permanently delaying debounced task

fractalwrench:
## Goal

Avoids permanently delaying debounced task by specifying a delay in the write queue, so that continuous bursts of updates don't stop a task from executing.

## Testing

Added unit tests.


bidetofevil (approved):
Ah nice. Take that, backpressure!

PR #3990, opened 2026-09-08
Optimize write for completed spans

fractalwrench:
## Goal

Improves performance for writing completed spans by keeping the file open for appends & tracking the file size ourselves.

## Testing

Added unit tests.


embrace-android-session-persistence/src/main/kotlin/io/embrace/android/embracesdk/internal/session/persistence/CompletedSpansWriter.kt
```diff
@@ -35,33 +38,87 @@ class CompletedSpansWriter(
         try {
             writeImpl(spans)
         } catch (exc: Throwable) {
+            discardSpanFile()
             trackFailure(exc)
```
bidetofevil, line 42: Should the failure tracking for the write operation go before the discard (i.e. swap lines 41 and 42), which also could fail if the SpanFile couldn't be closed? Seems like knowing what what caused the discard to fail is more important than knowing the discard failed, if we can only choose one or if we were reading these in order.

bidetofevil (approved):
LGTM. One non-blocking, pedantic comment that you can take or leave 😅

bidetofevil (approved):
LGTM. I'll leave in my pedantic suggestion but feel free to take or leave

PR #3994, opened 2026-09-10
Add macrobenchmark test scaffold

fractalwrench:
## Goal

Adds a module for macrobenchmark that can be used for assessing performance of the persistence layer (and other features, in time).


lemnik (approved):
LGTM

PR #3996, opened 2026-09-10
Create microbenchmark for writing session

fractalwrench:
## Goal

Creates a microbenchmark for encoding a representative session payload from scratch.


lemnik (approved):
LGTM

PR #4000, opened 2026-09-10
Measure size on disk of session

fractalwrench:
## Goal

Measures the disk size taken by the session with representative tests, and serves as a guard against that size increasing in future without alerting us.


lemnik (approved):
LGTM

PR #4001, opened 2026-09-10
Widen serialization/deserialization benchmarks

fractalwrench:
## Goal

Widens the serialization/deserialization benchmarks so that they assess performance against small, medium, and large spans.


lemnik (approved):
LGTM

PR #4004, opened 2026-09-11
Make SDK init attributes more robust and improve testing

bidetofevil:
Make sure each SDK init attribute separates error and unsupported states properly, with errors always recording the attribute with an error value that is outside the range of valid values for that attribute as well as log an internal error. Unless otherwise stated, a missing attribute will be due to that particular device not supporting the logging of that attribute. Some attributes, like low memory state, will not add the attribute unless it's in an anomalous state, and to disambiguate that from a device that doesn't support that attribute will require additional information not available at instrumentation time, but will be at analysis time (e.g. filter out certain Android versions or OEMs that don't support a particular thing).


embrace-android-instrumentation-startup-trace/src/main/kotlin/io/embrace/android/embracesdk/internal/instrumentation/startup/ArtOptimizationState.kt
```diff
@@ -10,56 +14,69 @@ import java.io.File
...
+                    // We read the chunk of the header the filter is expected to be in and scan the bytes for the value.
+                    findArtCompilerFilter(odex.readHead(HEADER_SCAN_BYTES))
+                } else {
+                    // Not finding the file is not an error, so simply return that fact.
+                    ART_COMPILER_FILTER_NOT_FOUND
+                }
+            }.getOrElse { throwable ->
+                logger.trackAttributeError(ART_COMPILER_FILTER, throwable)
+                STRING_ERROR
             }
```
lemnik: isn't this just a `try {} catch(throwable: Throwable) {}` at this point? I'm not sure `runCatching` adds much.

embrace-android-instrumentation-startup-trace/src/main/kotlin/io/embrace/android/embracesdk/internal/instrumentation/startup/SdkInitEnvironmentAttributes.kt
```diff
@@ -25,98 +33,239 @@ import kotlin.math.roundToLong
...
+        throwable = IllegalStateException("Attribute failed to record: $key", cause),
+    )
+}
+
+/**
+ * Runs [block] to contribute the attributes derived from one source, keeping a failure in it from costing us the
+ * attributes derived from every other source. Each attribute inside the block is expected to handle its own computation
+ * failure, but this provides a wrapper so internal errors can be logged for the failure of the whole block.
+ */
+internal inline fun MutableMap<String, String>.putAttributes(
```
lemnik: nit: maybe this could be named more clearly?
```suggestion
internal inline fun MutableMap<String, String>.safePutAttributes(
```

fractalwrench (approved):
LGTM

embrace-android-instrumentation-startup-trace/src/main/kotlin/io/embrace/android/embracesdk/internal/instrumentation/startup/ArtOptimizationState.kt
bidetofevil: You're right - it was doing something before such that it had an `onSuccess()` but the refactor render this an awkward idiom to use. I'll change this.

embrace-android-instrumentation-startup-trace/src/main/kotlin/io/embrace/android/embracesdk/internal/instrumentation/startup/SdkInitEnvironmentAttributes.kt
bidetofevil: Will do

PR #4007, opened 2026-09-11
Add script to copy perfetto trace

fractalwrench:
## Goal

Adds a script that runs a macrobenchmark test and copies the perfetto trace to a predictable location on disk.


bidetofevil (approved):
LGTM. How does this overlap with the analysis skills and tooling Ive been working on?

fractalwrench:
@bidetofevil I think that once these changes are in, we can update the skill so that it uses these constituent tools. For measuring SDK operation times in particular I think this is going to yield some better ways of analyzing data

PR #4008, opened 2026-09-11
Alter benchmark to cover persistence layer

fractalwrench:
## Goal

Alters the macrobenchmark so that it covers persisting sessions and the script pulls the created Perfetto trace.


bidetofevil (approved):
LGTM

PR #4009, opened 2026-09-11
Add a module for perfetto analysis

fractalwrench:
## Goal

Adds the skeleton of a module that will eventually be used for local perfetto analysis.


bidetofevil (approved):
LGTM. Again, how does this align with my WIP analysis work?

PR #4011, opened 2026-09-11
Track dirty span snapshots

fractalwrench:
## Goal

Refactors the `SessionPartWriter` so that dirty span snapshots (i.e. ones that have changed) are tracked. In future this will allow us to serialize just those parts in the payload.


embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/session/orchestrator/SessionPartWriter.kt
```diff
@@ -29,10 +30,10 @@ interface SessionPartWriter {
     fun onSpanCompleted(spans: List<Span>)
 
     /**
-     * A span the session part is recording has changed, so the span snapshots on disk are stale.
-     * This includes the session span itself, which is snapshotted until the part ends.
+     * [span] has changed, so the snapshot of it on disk is stale. This includes the session span
+     * itself, which is snapshotted until the part ends.
      */
-    fun onSpanSnapshotChanged()
+    fun onSpanSnapshotChanged(span: EmbraceSdkSpan)
```
bidetofevil, line 36: This means each change will have to be processed individually, eh? Is there an opportunity for buffering at some point to at least deal with the mega backpressure cases or is that overkill for now?

embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/session/orchestrator/SpanSnapshotTracker.kt
```diff
@@ -0,0 +1,65 @@
...
+import java.util.Collections
+import java.util.concurrent.ConcurrentHashMap
+
+/**
+ * Tracks the spans that are recording, and those among them whose latest state has yet to reach
+ * disk.
+ */
+class SpanSnapshotTracker {
+
+    private val inFlight: MutableSet<EmbraceSdkSpan> = Collections.newSetFromMap(ConcurrentHashMap())
```
bidetofevil, line 14: Any requirements to maintain order in any of these collections? Or is the underlying set's iteration order already by insertion so we're good there? And do we have to worry about multiple updates of the same span delaying the write by changing the iteration order because of reinsertion?

bidetofevil (approved):
LGTM. A few non-blocking questions

embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/session/orchestrator/SessionPartWriter.kt
fractalwrench: We do some debouncing that prevents too many writes at once which helps handle backpressure

embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/session/orchestrator/SpanSnapshotTracker.kt
fractalwrench: I don't think there's any requirement for maintaining order here

PR #4013, opened 2026-09-11
Begin reading Perfetto file

fractalwrench:
## Goal

Begins reading the Perfetto trace file generated via macrobenchmark. I've slimmed down the size of the trace file substantially by only recording atrace sections, which should be sufficient for the needs of reporting operations tracked via `android.os.Trace`. I've additionally validated that the code can check whether a trace file is actually a Perfetto trace.

## Testing

Added unit tests.


bidetofevil (approved):
LGTM. With perfetto access, we can build much more granular benchmarking and profiling in regular macrobenchmark tests. This is awesome!

PR #4014, opened 2026-09-11
Add integration test for crashed session

fractalwrench:
## Goal

Adds an integration test that asserts with multi-file persistence crashes don't lose data.


bidetofevil (approved):
LGTM

PR #4020, opened 2026-09-14
Implement `SpanCollectionReader`

fractalwrench:
## Goal

Adds `SpanCollectionReader` which will be used for reading both completed spans and span snapshots. Initially prototyped in #3998

## Testing

Added unit tests.


bidetofevil (approved):
LGTM

PR #4021, opened 2026-09-14
Interpret traces from perfetto

fractalwrench:
## Goal

Adds interpretation for traces from perfetto. This maps the Protobuf objects onto a model layer that sorts the events and organises them by thread.

## Testing

Added unit tests.


bidetofevil (approved):
LGTM

PR #4023, opened 2026-09-14
Allow generating stats from perfetto via cli

fractalwrench:
## Goal

Allows generating stats from perfetto traces via the shell script. Stats can be outputted in either JSON or markdown. I performed some sanity checking against ui.perfetto.dev to validate that the slices match what should be expected.

## Testing

Added unit tests & performed sanity checking.


bidetofevil (approved):
LGTM

PR #4024, opened 2026-09-14
Calculate wall clock time for operations

fractalwrench:
## Goal

Builds on #4023 by allowing the calculation of % wall-time spent on an operation for a given thread.


bidetofevil (approved):
LGTM

PR #4025, opened 2026-09-14
Produce html report to read perfetto json

fractalwrench:
## Goal

Produce a HTML report that is capable of reading and interactively displaying the information generated in #4023


bidetofevil (approved):
LGTM

PR #4026, opened 2026-09-14
Add append option to span snapshot writer

fractalwrench:
## Goal

Alters the span snapshot writing to handle appends, without hooking up the change to production code. Initially prototyped in #3998

## Testing

Added unit tests.


bidetofevil (approved):
LGTM

PR #4027, opened 2026-09-14
Alter trace section names

fractalwrench:
## Goal

Alters the names used for tracing in the persistence layer so that it's simpler to distinguish multi-file (mf) from single-file (sf).


bidetofevil (approved):
LGTM

PR #4028, opened 2026-09-14
Add counters for file persistence I/O

fractalwrench:
## Goal

Adds counters for file persistence I/O so that it will be possible to measure the I/O throughput of both solutions.


embrace-android-infra/src/main/kotlin/io/embrace/android/embracesdk/internal/utils/CountingOutputStream.kt
```diff
@@ -0,0 +1,26 @@
...
+/**
+ * Counts the bytes written through it
+ */
+class CountingOutputStream(private val delegate: OutputStream) : OutputStream() {
+
+    var written: Long = 0
+        private set
+
+    override fun write(b: Int) {
+        written++
```
bidetofevil: Do we want to first write to the delegate, then do the bookkeeping, in case the delegate throws?

bidetofevil (approved):
LGTM. Comment about the order of delegating to the wrapped stream and when we increment the bytes written counter. Feels like we should confirm the write was successful first before adding to the counter?

If this is handled in some other way or what we have is preferred, feel free to merge away.

PR #4029, opened 2026-09-14
Make output an argument

fractalwrench:
## Goal

Make `output` a required argument when invoking the `analyse_trace.sh` script.


bidetofevil (approved):
LGTM

PR #4030, opened 2026-09-14
Setup skeletons of scripts

fractalwrench:
## Goal

Setup boilerplate for scripts that will be used to aggregate and compare benchmark runs.


bidetofevil (approved):
LGTM

PR #4031, opened 2026-09-14
Basic EventBus implementation

lemnik:
## Goal
A basic `EventBus` implementation to be shared between the instrumentation modules. The `EventBus` is available to instrumentation directly through the `InstrumentationArgs` objects.

`EventBus` handles event class hierarchies, allowing common super-types between modules to reduce the coupling between related modules.

```kotlin
sealed interface NavigationEvent {
  class ActivityStarted(val name: String): NavigationEvent
  class ActivityStopped(val name: String): NavigationEvent
```

can be observed with just:

```kotlin
eventBus.addHandler { event: NavigationEvent ->
```

## PR Stack Info

A future PR introduces state (sticky / replay) events, and a class hierarchy cache to somewhat simplify the dispatch code.

## Testing
Added basic unit tests covering the `EventBus`.


embrace-android-infra/src/main/kotlin/io/embrace/android/embracesdk/internal/utils/event/EventBus.kt
```diff
@@ -0,0 +1,142 @@
...
+        synchronized(this) {
+            handlers?.let(action)
+            dispatch.clear()
+        }
+    }
+
+    /**
+     * Resolves and caches the handlers for a concrete event class, on first emit of that class and again after any registration change.
+     */
+    @Synchronized
```
fractalwrench, line 88: nit: my preference would be to synchronize on an explicit lock that's a private property

PR #4032, opened 2026-09-14
EventBus based tap instrumentation

lemnik:
## Goal
Refactors the `taps` and `compose-tap` instrumentation modules to emit and report events via the `EventBus`.

## Testing
Updated unit tests, and tested manually


embrace-android-instrumentation-api/src/main/kotlin/io/embrace/android/embracesdk/internal/arch/ui/TapSignal.kt
```diff
@@ -0,0 +1,10 @@
+package io.embrace.android.embracesdk.internal.arch.ui
+
+/**
+ * A tap on a named UI element, emitted by whichever UI framework integration observed it.
+ */
+data class TapSignal(
```
fractalwrench, line 6: I initially thought we could think about storing the event bus and its events in a separate module from instrumentation-api, but I guess we can see how it grows first. It'd be good to place them in their own package so it's slightly simpler to move if we choose to do that

PR #4033, opened 2026-09-14
Introduce State events

lemnik:
## Goal
Add support for sticky / stateful events to the `EventBus`.

These events go through a completely distinct channel to normal events (`emitState` vs `emit`, `addStateHandler ` vs `addHandler`, etc.). Evert stateful event type is keyed against a `StateKey` which side-steps the complexities and questions that come from polymorphic compatible event dispatch (which `NavigationEvent` do we replay?).

## Testing
New unit tests & manual testing


embrace-android-infra/src/main/kotlin/io/embrace/android/embracesdk/internal/utils/event/EventBus.kt
```diff
@@ -3,140 +3,234 @@ package io.embrace.android.embracesdk.internal.utils.event
...
-        mutate(handlersFor(type)) { it.add(handler) }
+        mutate(handlersFor(registered, type)) { it.add(handler) }
     }
 
     inline fun <reified E : Any> addHandler(handler: EventHandler<E>) {
         addHandler(E::class.java, handler)
     }
 
+    /**
+     * Registers [handler] against [key], replaying the value retained against it before returning. Registering a handler already
```
fractalwrench, line 65: The doc could be more explicit that if the value is null, nothing is returned.

PR #4083, opened 2026-09-15
Avoid grouping by thread

fractalwrench:
## Goal

Avoid grouping by thread name and ID as this was splitting operations we want to combine (such as span snapshot changes) and is additionally complicated when comparing cold start traces where the main thread may not be set.


embrace-perfetto-analysis/src/main/kotlin/io/embrace/android/embracesdk/internal/perfetto/OperationStats.kt
```diff
@@ -3,13 +3,11 @@ package io.embrace.android.embracesdk.internal.perfetto
 import kotlinx.serialization.Serializable
 
 /**
- * What one section cost on one thread, across every occurrence recorded there. A section that ran on
- * two threads is two of these.
+ * What one section cost across every occurrence the trace recorded, wherever it ran. A section that
```
bidetofevil, line 6: Hmm. A cross thread operation may be harder to nail down if it comprises multiple different run contexts that span threads. What operations require these and does this amalgamation change what we can say about such an operation?

fractalwrench: I found it more useful grouping operations by their trace name, rather than by thread, although I can see there'd be benefit on displaying stats on what thread an operation ran on also. I'd suggest we do that as a follow up or add-on?

bidetofevil: Follow up is fine. It's a bit of an edge case anyway, and if are testing these numbers rigourously, it'll just mean more noise, which we can compensate with more data - or better metadata. Lets see how widespread this is before doing anything,

PR #4085, opened 2026-09-15
Reorganise packages in embrace-perfetto-analysis

fractalwrench:
## Goal

Reorganise packages in embrace-perfetto-analysis module to avoid too many files in the root.


bidetofevil (approved):
LGTM

PR #4086, opened 2026-09-15
Create model layer for holding iteration data

fractalwrench:
## Goal

Creates a model layer with classes that hold data on how iterations (i.e. multiple runs of a macrobenchmark) should be represented in memory. I've additionally setup some HTML output for this, which takes up a lot of the diff. The next changeset will wire everything up.


bidetofevil (approved):
State stuff LGTM. The rendering code.... I only skimmed through 😅

PR #4087, opened 2026-09-15
Hookup span snapshots

fractalwrench:
## Goal

Hooks up span snapshots so that they are written to an append-only file from `SessionPartWriter`.

## Testing

Added unit tests.


bidetofevil (approved):
LGTM

PR #4088, opened 2026-09-15
Calculate aggregate stats for multiple macrobenchmark runs

fractalwrench:
## Goal

Calculates aggregate stats for multiple macrobenchmark runs and produces a HTML report displaying the findings.


bidetofevil (approved):
LGTM

PR #4089, opened 2026-09-15
Create skeleton for comparing macrobenchmarks

fractalwrench:
## Goal

Creates the skeleton of the interface that will be used to compare macrobenchmark runs against each other for comparative perf analysis of features.


bidetofevil (approved):
LGTM

PR #4090, opened 2026-09-15
Compare performance between two macrobenchmark directories

fractalwrench:
## Goal

Compares performance between two macrobenchmark directories so that operations can have their performance meaningfully compared. The following changeset will bring a HTML report.


bidetofevil (approved):
LGTM

PR #4091, opened 2026-09-15
Handle session reader backpressure

fractalwrench:
## Goal

Handles backpressure within `SessionPartReader` by sequentially burning through the queue rather than attempting to read _everything_ at once, which would cause memory spikes if more than one session part directory was present. This also prevents potential data loss as it guarantees that `IntakeService` has persisted the payload rather than assuming the model will be safely persisted later.

## Testing

Added unit tests.


bidetofevil (approved):
Ah nice. LGTM.

PR #4092, opened 2026-09-15
Display results for performance comparison

fractalwrench:
## Goal

Display results as HTML when comparing performance between two macrobenchmark runs.


bidetofevil (approved):
LGTM

PR #4093, opened 2026-09-15
Avoid retaining span references when not required

fractalwrench:
## Goal

Avoid retaining a reference to span collection objects when multi-file persistence is not enabled, so that GC can tidy up unused memory.


bidetofevil (approved):
LGTM

PR #4094, opened 2026-09-15
Delete new session part directories when functionality disabled

fractalwrench:
## Goal

Deletes the new session part directories when the functionality is disabled.

## Testing

Added unit tests.


bidetofevil (approved):
LGTM

PR #4096, opened 2026-09-16
Allow feature flags to have variants

bidetofevil:
Allow feature flags to be tracked with an optional variant, just like an experiment.


fractalwrench (commented):
What is the difference between a feature flag and an experiment with this change? The interfaces are functionally equivalent

bidetofevil:
> What is the difference between a feature flag and an experiment with this change? The interfaces are functionally equivalent

The difference is that it is surfaced to customers as two different things so they can have two different UX on the dashboard, among other things. Under the hood, they are essentially the same thing, and given the new requirements, even their feature sets aligned, so part of this change is to minimize the actual internal differences while keeping the surface differences.

fractalwrench (approved):

PR #4098, opened 2026-09-16
Alter size limit

fractalwrench:
## Goal

Alters the size limit when writing to take into account that protobufs are uncompressed, and the 3Mb upload limit refers to compressed gzip payloads.


bidetofevil (approved):
LGTM

PR #4099, opened 2026-09-16
Handle imbalanced session part start/end

fractalwrench:
## Goal

Report an internal error when a session part start/end call is imbalanced, and attempt to repair the imbalance.

## Testing

Updated unit tests. A few test cases were deleted that following this changeset became duplicates of other test cases in the same file.


embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/session/orchestrator/SessionPartWriterImpl.kt
```diff
@@ -96,6 +96,11 @@ class SessionPartWriterImpl(
             directoryStore.create(writers.directory)
 
             synchronized(bufferLock) {
+                // a part should always end before the next one starts.
+                current?.let { orphan ->
```
lemnik, line 101: nit: is it maybe worth nulling `current` here so there's no chance this gets stuck (if somehow `finish` failed)?

lemnik (approved):
LGTM - With a very optional suggestion

PR #4100, opened 2026-09-16
Use empty list if spans not readable

fractalwrench:
## Goal

Supply an empty list instead of `null` if one of the spans/span snapshot files aren't readable for whatever reason, rather than dropping the entire session part.


lemnik (approved):
LGTM!

bidetofevil (approved):
LGTM

PR #4107, opened 2026-09-17
Clear specific handed over spans

fractalwrench:
## Goal

Clear specific handed over spans rather than the whole collection.


bidetofevil (approved):
LGTM

PR #4108, opened 2026-09-17
Alter how temp files are named

fractalwrench:
## Goal

Atomic writing is achieved by writing to a temporary file and then renaming the file. Creating a temporary file results in a call to `SecureRandom` and an extra call to `createNewFile`. Using a fixed name in the current directory avoids that cost and should always be deleted regardless.


bidetofevil (approved):
LGTM

PR #4109, opened 2026-09-17
Increase max span size

fractalwrench:
## Goal

Increases the max span size limit to 1Mb to account for a theoretical very large session part span with lots of attributes, breadcrumbs, and experiments.


bidetofevil (approved):
LGTM. Should we have a special internal error if we hit the size max for a span that is a session span, just so we know that it's happening? Or should we wait until we hit SOME oversized ones before differentiating?

fractalwrench:
I think waiting for _some_ oversized spans to be hit sounds good. I don't expect it to be hit frequently given our current guardrails

PR #4111, opened 2026-09-17
Improve error handling for span file reading

fractalwrench:
## Goal

Improves error handling for span file reading by reporting corrupted files via internal errors, and returning any spans decoded up to that point rather than dropping everything.

## Testing

Added unit tests.


embrace-android-session-persistence/src/main/kotlin/io/embrace/android/embracesdk/internal/session/persistence/SpanCollectionReader.kt
```diff
@@ -31,6 +38,9 @@ internal class SpanCollectionReader(
         reader.nextTag().takeIf { it != -1 }
     } catch (exc: EOFException) {
         null
```
bidetofevil, line 40: Could this not be a sign of corruption as well?

bidetofevil (approved):
LGTM. One non-blocking comment about whether EOFException are considered corruption as well

embrace-android-session-persistence/src/main/kotlin/io/embrace/android/embracesdk/internal/session/persistence/SpanCollectionReader.kt
fractalwrench: Potentially - it's more likely to be when a process died mid-write, which this reader attempts to recover from.

PR #4112, opened 2026-09-17
Remove special-casing for the session span

fractalwrench:
## Goal

Removes special-casing for the session span. It is now persisted as part of the span snapshots/completed spans so doesn't require special handling.


bidetofevil:
Is there a mechanism now to ensure that all the session part's associated spans (e.g. state spans) are delivered and the session part span updated (with span links or metadata) before it is locked and flushed? Or is that handled in a different part of the lifecycle now, so some thing else is doing the coordination and this part is just about the persistence and the building of the payload?

PR #4113, opened 2026-09-17
Use same format for snapshots/spans

fractalwrench:
## Goal

Use the same protobuf message for span snapshots/completed spans to reduce code duplication.


bidetofevil (approved):
Nice. LGTM

PR #4114, opened 2026-09-17
Combine completed span/span snapshot readers

fractalwrench:
## Goal

Combines shared code between the completed span/span snapshot readers and extract it to `SpanRecordDecoder`. This ensures that behavior around limits and corruption are handled consistently.


bidetofevil (approved):
LGTM

PR #4115, opened 2026-09-17
Implement span queue

fractalwrench:
## Goal

Implements an object that holds a collection of span objects in a queue, that are eventually drained. On drain (and if a large number of objects are added) the queue is compacted by removing duplicates (i.e. objects with the same span ID).

This is the first in a series of changes intended to simplify the logic for writing span snapshots/completed spans, and ensure that each append uses a reasonable delay before writing as perf testing showed the current state of affairs negatively impacted I/O.

## Testing

Added unit tests.


embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/session/orchestrator/SpanQueue.kt
```diff
@@ -0,0 +1,73 @@
...
+
+/**
+ * Buffers telemetry in order before it is written to disk.
+ *
+ * Telemetry is added and compacted to remove duplicates if [COMPACT_THRESHOLD] is exceeded.
+ * [drain] removes from the queue after performing a [compact].
+ *
+ * [identityOf] supplies the key a record is deduplicated by (the span ID).
+ */
+internal class SpanQueue<T>(
```
bidetofevil: There's nothing particularly span-y about this since `T` can be anything, as long as the data follows the superceding/compaction rules. Is there a reason why we don't name this more generically so it can be used to do a similar type of draining for other items we want to process in a queue that can be similarly compacted? I'm thinking something like debouncing state changes, for instance.

bidetofevil (approved):
Looks real good. A non-blocking comment about naming and whether we can make this more generic (by name at least) given that I can see use cases outside of spans

PR #4116, opened 2026-09-17
Differentiate system state values with clashing non-system ones

bidetofevil:
Keep track of the type of a state value by using the attribute `emb.state.value_type` on session part state spans and transition events. On logs, the key for a particular state's value type should be `<state-value-key-name>.value_type`.

We do this by checking whether the value class implements `TypedStateValue` and returning true on the method `isSystemType()`. It is effectively a fancy boolean that will allow for future extension and avoid a sealed class with one implementation.


fractalwrench (commented):
I think it'd be good to break this into two PRs: one that creates the typed state interfaces + infra, and one that makes the navigation instrumentation etc use it. Would also be cool to get some more backstory behind the motivation for this change, and get CI passing

bidetofevil:
Broke this down. I realized I put nothing in this PR's description about navigation lol. I'll add more context to the next PR.

lemnik (approved):
LGTM

PR #4119, opened 2026-09-18
Add TelemetryWriteScheduler

fractalwrench:
## Goal

Adds `TelemetryWriteScheduler` which provides both a more performant and simpler way of managing span writes to disk.

## Testing

Added unit tests.


embrace-android-core/src/main/kotlin/io/embrace/android/embracesdk/internal/session/orchestrator/TelemetryWriteScheduler.kt
```diff
@@ -0,0 +1,115 @@
...
+ * tasks that are not already in-progress.
+ *
+ * Note that telemetry is always written on the worker - the caller is responsible for waiting for writes to finish
+ * in scenarios such as a crash.
+ */
+internal class TelemetryWriteScheduler<T>(
+    private val worker: BackgroundWorker,
+    private val delayMs: Long,
+    private val queue: TelemetryQueue<T>,
+    private val guard: (Runnable) -> Runnable,
```
bidetofevil, line 25: How will this be used in the future? It looks like a wrapper that can do callback-y things when a write task is provisioned. Are we trying to do external bookkeeping or notification of some sort?

bidetofevil (approved):
LGTM. Just a question about what the planned use for `guard` will be going forward

bidetofevil:
Never mind - the use of guard is answered in the next PR lol

PR #4121, opened 2026-09-18
Simplify metadata writes

fractalwrench:
## Goal

Simplifies metadata writes by driving them through `TelemetryWriteScheduler`.


bidetofevil (approved):
This is sick :-) LGTM

PR #4122, opened 2026-09-18
Simplify snapshot writes

fractalwrench:
## Goal

Simplifies span snapshot writes by using `TelemetryWriteScheduler`.


bidetofevil (approved):
LGTM

PR #4123, opened 2026-09-18
Remove dead code

fractalwrench:
## Goal

Removes dead code from the session persistence.


bidetofevil (approved):
Wow. We hardly knew ya!

PR #4124, opened 2026-09-18
Alter resurrection order

fractalwrench:
## Goal

Alter resurrection order so that persisted session parts are processed before the payload resurrection happens.

## Testing

Added integration test case.


bidetofevil (approved):
LGTM. What a subtle change that could break the whole thing. So existing test cases, even integration ones, didn't cover this? The issue looks structure unless we are queuing on two different threads here.

fractalwrench:
Existing tests run against the legacy layer - new tests will be added further up the stack

PR #4125, opened 2026-09-18
Avoid deep equality comparisons

fractalwrench:
## Goal

Avoid deep equality comparisons when removing items from a queue, which are expensive to perform.


bidetofevil (approved):
LGTM

PR #4126, opened 2026-09-18
Avoid writing if the queue is empty

fractalwrench:
## Goal

Avoids writing if the queue is empty, which prevents some unnecessary serialization of metadata.


lemnik (approved):
LGTM

bidetofevil (approved):
LGTM. One lingering question that might just be me being slow right now -- is there a possibility that there's a race when flush drops a write task, and just before it returns, the queue becomes non-empty, and somehow there's no flush scheduled until there's another incoming to trigger it?

fractalwrench:
I don't see where that case would happen

PR #4127, opened 2026-09-18
Apply limit consistently to closed span files

fractalwrench:
## Goal

`discardFile()` reset the record count to 0 and closed the file. The next span write opened the file again, which would allow the spans in the file to escape the 4000 per session part limit.

I've renamed `discardFile()` to `close()` and retained the `recordCount` for the duration of the session part, and added unit tests to catch this case.


lemnik (approved):
LGTM

bidetofevil (approved):
LGTM

PR #4128, opened 2026-09-18
Anchor the network capture expiry time correctly

lemnik:
## Goal
Anchor the `NetworkCaptureRuleRemoteConfig.expiresIn` to the delivery time of the config data. This is the closest reasonable match to the server behaviour.

## Changes

- `expiresIn` is expressed in seconds, not milliseconds (doc comment fixes)
- rule validity is now checked with the anchor time:
```kotlin
now < configDeliveredAt + TimeUnit.SECONDS.toMillis(rule.expiresIn)
```

This PR pulls in a fairly substantial number of files, as the `Clock` needs to be pulled through the `PersistentConfig` so that the delivery timestamp can be stored persistently (alongside `deviceId`), which leads to a substantial dragnet of fakes & tests.

## Testing
New unit tests introduced


fractalwrench (approved):
LGTM, assuming we have discussed and reached consensus with iOS/backend folks that this is the desired behavior

PR #4129, opened 2026-09-18
Use typed state values to differentiate special no-screen navigation states

bidetofevil:
Leverage typed state values to do this

---


bidetofevil:
We are using this to differentiate special state values that don't map to an actual screen (e.g. like the app is backgrounded) so that there would be no clashes when we write the string form in the attribute. We shouldn't prevent folks from having a destination call "Initializing" or "Backgrounded", and storing a special unlikely-to-clash token instead makes the data less intelligible on its own and opaque. Ditto with encoding it somehow.

lemnik (approved):
LGTM

